PR #19, opened 2011-09-28
trie - reducing memory use

timwintle:
This adds space for static 1-charater strings which are used for transitions in the trie implementation - which avoids multiple allocations for the same string.

Over a large (dense) set of 8-character strings, I observed memory use reduced from 388M to 283M after this change.

### DIFF
--- a/Bio/trie.c
+++ b/Bio/trie.c
@@ -71,31 +71,6 @@ static void free_suffix(char *s) {
     }
 }
 
-/**
- * Old method without using lookup table
- */
-/*
-static char* duplicate(const char* s) {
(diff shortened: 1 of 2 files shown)


peterjc:
We've decided to declare ``Bio.trie`` as obsolete with the view of deprecating it and then removing it from Biopython, see #1658 and http://biopython.org/wiki/Deprecation_policy - at this point that choice can easily be reverted. Or, since code should have all been dual licensed, it would be possible to fork the ``Bio.trie`` code and release it as a standalone 3-clause BSD licensed module on PyPI.

In light of this, I am going to close this issue.

PR #37, opened 2012-05-04
Fixing Entity copy method to preserve child ordering.

asford:
Pull request to fix issue #3349.


peterjc:
@asford Would you agree to dual licence your Biopython contributions under the *Biopython License Agreement* and the *BSD 3-Clause License* as per #898 please?

PR #73, opened 2012-09-25
-w option in RAxML requires absolute path

Jingping:
-w      FULL (!) path to the directory into which RAxML shall write its output files

```
          DEFAULT: current directory
```


peterjc:
@Jingping would you agree to dual license your Biopython contributions as discussed on #898 so we can switch to the standard 3-clause BSD license? (Apologies if you already have done so, and I have overlooked this)

PR #158, opened 2013-02-12
motif IC method fixed for non-uniform base distribution

sefakilic:
The ic() method in Motif class computes information content as

\sum_b [2 + f(b) * lg(f(b))], where b is base. This formula computes information content for only one position in the motif. The current implementation assumes equal frequency of bases. I modified it to handle nonuniform base frequencies. For each position in the motif, the information content is computed as

\sum_b [-bg(b) * lg(bg(b)) + f(b) * lg(f(b))],

where bg(b) is the background frequency of base b, f(b) is the frequency of base b in that position of the motif.

### DIFF
--- a/Bio/Motif/_Motif.py
+++ b/Bio/Motif/_Motif.py
@@ -137,8 +137,14 @@ class Motif(object):
         """
         res=0
         pwm=self.pwm()
+        # compute background entropy for one column
+        bg_entropy=0
+        for a in self.alphabet.letters:
+            if background[a]!=0:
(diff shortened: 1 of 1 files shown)


peterjc:
It looks like this PR has gotten a bit stale. After doing the Biopython 1.66 release I removed the deprecated ``Bio.Motif`` module. This work would need to be revised to go into ``Bio.motifs`` instead.

sefakilic:
Bio.motifs Bio.motifs.matrix.`mean` methods seems take into account the non-uniform background distribution, so this PR is obsolete.

PR #196, opened 2013-07-02
Fixing a bug in SeqIO.PdbIO in parsing pdb-atom sequence

matsu3shiro:
Fixing a problem that SeqIO.PdbIO outputs erroneous amino acid sequence
if the ATOM record has more than one missing segments.

SeqIO.PdbIO includes a bug in parsing amino acid sequences
 from PDB ATOM records (pdb-atom).  
It outputs erroneous amino acid sequence
 if the ATOM record has more than one missing segments.  

The original code trys to insert 'X's for each gap position, 
 but it adds the C-terminal sequence from the gap
 each time it fills a gap in addition to the 'X's.  
In contrast, the insertion of C-terminal sequence is only required 
 at the final gap position.  
Thus, this code outputs an amino acid sequence
 in which C-terminal sequence is erroneously repeated
 if the coordinate file includes more than one gap positions.  

I fixed this bug by moving lines 173-174 out of the FOR loop
 starting from line 166.


peterjc:
@matsu3shiro Could you comment on issue #898 if you agree to dual license your Biopython contributions under both the "Biopython License Agreement" and the "BSD 3-Clause License". Many thanks, Peter.

peterjc:
Hello again @matsu3shiro - sorry to bother you, but we would still like your agreement to dual license your contribution 5303fbdbeb4f1f0a0fb865a02c975c27a92a399d to Biopython  under both the "Biopython License Agreement" and the "BSD 3-Clause License". Many thanks, Peter.

PR #234, opened 2013-09-05
Improved parsing of GN in the SwissProt parser.

joshainglis:
Small change to parsing of GN in the SwissProt parser.
Now creates a dictionary for the gene name tokens.


peterjc:
It looks like between us we forgot about this pull request, sorry about that.

Are you interested in rebasing this to the current master branch (and thinking about what to do with the old ``GN`` test cases as discussed above)?

joshainglis:
> Superseded by #4162 .
> 
> @joshainglis  Please let us know if you would like to be included in the list of contributors.

Well this is a blast from the past!
Yeah sure, that would be nice, thanks.

PR #241, opened 2013-09-28
Add checks for unicode byte order marks

evolvedmicrobe:


peterjc:
@evolvedmicrobe would you agree to dual license your contributions to Biopython to date (including this and #239) as per #898?

evolvedmicrobe:
@peterjc absolutely!

PR #247, opened 2013-10-17
fixed typo @ Bio.SeqIO.FastaIO.FastaWriter

stanti:


peterjc:
@stanti it may not really be necessary for this little fix, but would you agree to dual licence your Biopython contributions under the *Biopython License Agreement* and the *BSD 3-Clause License* as per #898 please?

peterjc:
Sorry, found where you agreed here: https://github.com/biopython/biopython/issues/898#issuecomment-236926537

PR #292, opened 2014-02-19
Treeconstructor methods

jwillis0720:
Adding methods to tree constructor to work with phylip packages for neighbor joining that suffers from speed. Also converting to PIP8 standards.


peterjc:
Sorry, it looks like we forgot about this pull request. Rebasing looks a bit tedious, so copying the proposed new ``convert_distance_matrix `` method looks easier.

@lijax, @JoaoRodrigues - what do you think?

etal:
Just copying the `convert_distance_matrix` method sounds good to me. I'm not sure why it's associated with the `DistanceCalculator` class instead of `_DistanceMatrix` -- it doesn't use `self`, but it does require a `_DistanceMatrix` instance. Maybe it would be best to copy it into the `_DistanceMatrix` class definition instead.

An example of Phylip's input format is given in the [documentation for the neighbor program](http://evolution.genetics.washington.edu/phylip/doc/neighbor.html). That page says several Phylip programs take that input format, and they are [listed here](http://evolution.genetics.washington.edu/phylip/progs.data.dist.html).

Since this format is a fairly general and human-readable table (though columns are not necessarily aligned prettily in this implementation), it could even be used as the `__str__` method -- or `format`, `format_phylip`, or `to_phylip` to reserve the option for another human-readable string representation. (But note that the original use case for this was for huge distance matrices.)

Potential enhancement: It looks like the format allows multiple spaces between values in the table, so aligned columns would be OK: `ljust` the taxon names in the first column to the width of the longest name, and use a float format of `%.6f` or so for the numeric values.

peterjc:
@jwillis0720 Are you happy to dual licence your contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

@etal Can I leave this with you please?

jwillis0720:
Yes thank you

etal:
OK, I think I can take care of this next week.

peterjc:
@jwillis0720 would you agree to dual license your Biopython contributions as discussed on #898 so we can switch to the standard 3-clause BSD license?

PR #352, opened 2014-07-25
Interface analysis

aparente:
When trying to call Interface I get the error:

<ipython-input-2-aa768c5df9d3> in <module>()
----> 1 from Bio.PDB import Interface

/Users/Angelica/anaconda/lib/python2.7/site-packages/Bio/PDB/Interface.py in <module>()
      8 from Bio.PDB.Entity import Entity
      9 from Bio.Data import IUPACData
---> 10 from Bio.SCOP.Raf import to_one_letter_code
     11 from Bio.PDB.PDBParser import PDBParser
     12 from Bio.PDB.NACCESS import NACCESS

ImportError: cannot import name to_one_letter_code

After some inspection of Raf.py I noticed that there is no object to_one_letter_code but one_letter_code


Bio/PDB/Interface.py
```diff
@@ -0,0 +1,226 @@
...
+        super_imposer.set_atoms(ref_atoms, alt_atoms)
+        
+        print "RMS = %0.2f" % super_imposer.rms
+        
+        return super_imposer.rms
+        
+    def _get_neighbors_id(self, chain):
+        "Creates a dictionary of tuples from the neighbors dictionary"
+        
+        self.neighbors_id={}
```
mschmo: There's a few places where you don't include spaces between your assignment. Try to update these to stay consistent.

Bio/PDB/Interface.py
```diff
@@ -0,0 +1,226 @@
...
+        "Defined the fraction of native contacts between 2 interfaces"
+        
+        #Creates neighbors dictionary for each interface residue
+        if not self.neighbors:
+            self.set_neighbors()
+        if not mobile.neighbors:
+            mobile.set_neighbors()
+        
+        #Calculation will be done only one one chain thanks to symetry
+        for c in self.get_chains():
```
mschmo: You may want to include a check to make sure there are keys, or `chain` will never get assigned a value and you'll end up with a `NameError` below.
Also, instead of executing a loop here you could probably do something like `chain = self.child_dict.keys()[0]`

Bio/PDB/Interface.py
```diff
@@ -0,0 +1,226 @@
...
+    def fcc(self, mobile):
+        "Defined the fraction of native contacts between 2 interfaces"
+        
+        #Creates neighbors dictionary for each interface residue
+        if not self.neighbors:
+            self.set_neighbors()
+        if not mobile.neighbors:
+            mobile.set_neighbors()
+        
+        #Calculation will be done only one one chain thanks to symetry
```
mschmo: to one

Bio/PDB/Interface.py
peterjc: If it helps, we have a proposed pre-commit hock on issue #493, which would automate checking these little style things.

mschmo: That's a great idea. I don't like nitpicking, so let a computer do that.

JoaoRodrigues: Hi,

As it's said right in the beginning of this pull request, this is an
experimental branch that was never merged into the master.

Thanks for the comments though, if I'll get round to look into this code,
I'll take them into account.

Matthew Schmoyer <notifications@github.com> escreveu no dia qua, 25/03/2015
às 16:57:

peterjc:
Can we close this old PR? @JoaoRodrigues?

PR #360, opened 2014-09-10
Change Newick tree parsing to handle numeric names

bsmith89:
Numeric taxa names in Newick trees are parsed as branch confidence.
For instance a newick file

```
(((0:0.1,1:0.1)0.99:0.1,2:0.1)0.98:0.0);
```

would be parsed as a tree with three, unnamed terminal nodes with confidence values of 0, 1, and 2.

I believe this is a bug, since ~~branches to~~ (edit 9/12) terminal nodes shouldn't have confidence values.
I've edited `Bio/Phylo/NewickIO.py` so that it behaves as I would expect, and written a unit test.

### DIFF
--- a/Bio/Phylo/NewickIO.py
+++ b/Bio/Phylo/NewickIO.py
@@ -217,9 +217,10 @@ class Parser(object):
     def process_clade(self, clade):
         """Final processing of a parsed clade. Removes the node's parent and
         returns it."""
-        if (clade.name and not (self.values_are_confidence or
-                                self.comments_are_confidence)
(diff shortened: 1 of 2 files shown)


peterjc:
@bsmith89 would you agree to dual license your Biopython contributions as discussed on #898 so we can switch to the standard 3-clause BSD license?

(Apologies if you already have done so, and I have overlooked this)

peterjc:
RE: https://github.com/biopython/biopython/issues/898#issuecomment-527530728 - Thank you Byron!

PR #366, opened 2014-10-05
Added ability to change the seqfeature source

ctSkennerton:
Added a new kwarg to BioSQL.Loader._load_seqfeature_basic that
allows the source to be changed dynamically based on the value
of the source qualifier found in the annotations of the SeqRecord

Background:
GFF files set aside column 2 for the source of the feature that
is set to the value of the program/algorithm that generated that
feature. Parsing a gff file using BCBio.GFF returns a SeqRecord
object that contains the source as a qualifier key-value pair.
Normally these key-value pairs are loaded into the
seqfeature_qualifier_value table, but the BioSQL wiki suggests that the
source should instead go into the term table.

The current code in biopython sets the value of source for a
seqfeature to be the same value regardless of whether the SeqRecord
object contains the source qualifier. This change adds in code that
will search for a source qualifier in the SeqRecord, set the value to
the source of the seqfeature and remove the key from the SeqRecord
qualifiers dict so that it is not mistakenly added to the
seqfeature_qualifier_value table

### DIFF
--- a/BioSQL/Loader.py
+++ b/BioSQL/Loader.py
@@ -751,12 +751,26 @@ class DatabaseLoader:
     def _load_seqfeature(self, feature, feature_rank, bioentry_id):
         """Load a biopython SeqFeature into the database (PRIVATE).
         """
-        seqfeature_id = self._load_seqfeature_basic(feature.type, feature_rank,
-                                                    bioentry_id)
+        # records loaded from a gff file using BCBio.GFF will contain the value
+        # of the 2nd column of the gff as a feature qualifier. The BioSQL wiki
+        # suggests that the source should not go in with the other feature 
+        # mappings but instead be put in the term table
+        # (http://www.biosql.org/wiki/Annotation_Mapping)
+        try:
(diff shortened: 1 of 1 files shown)


peterjc:
Thanks to Brian Osborne (@bosborne) for testing this and comparisons with BioPerl:
http://lists.open-bio.org/pipermail/biopython-dev/2015-June/020971.html
http://lists.open-bio.org/pipermail/biopython-dev/2015-June/020973.html

peterjc:
Rebased and applied as https://github.com/biopython/biopython/commit/b709f7e05c674698e791c9d877871fea8375c1f1 and preceding commit (there was a whitespace change on the master preventing an automatic merge).

Thank you Conner https://github.com/biopython/biopython/commit/884707e040404dc9b93704cc43f2f9170b09a1e8 - and thank you Brian for testing this.

Bonus points if anyone wants to write a unit test for this in ``Tests/common_BioSQL.py``.

bosborne:
Peter,

I’ll write the test(s), good practice!

BIO

PR #390, opened 2014-10-24
Changed majority_consensus, strict_consensus and get_support to

xzy3:
accept an iterable of trees rather then a list. This allows
consensus trees to be generated for much larger sets of trees by
streaming trees from the disk.

### DIFF
--- a/Bio/Phylo/Consensus.py
+++ b/Bio/Phylo/Consensus.py
@@ -12,6 +12,7 @@ adam consensus.
 from __future__ import division
 
 import random
+import itertools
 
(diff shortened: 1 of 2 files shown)


peterjc:
@xzy3 would you agree to dual license your Biopython contributions as discussed on #898 so we can switch to the standard 3-clause BSD license? (Apologies if you already have done so, and I have overlooked this)

xzy3:
@peterjc Sure. I hereby give my explicit permission to switch to the BSD 3-clause license.

PR #431, opened 2014-11-16
Blast NCBIXML.py: new feature: coverage calculation

herbstk:
I would like to provide a new feature for the NCBIXML parser. The resulting object now has two additional properties (`.alignment.qcovs` and `.hsp.qcovhsp`) which are calculated directly from the data extracted from a XML file as they are not included in this format. I think it makes sense to provide this additional information as it makes sophisticated filtering easier. The code is based on the [NCBI Blast C++ Source code](http://www.ncbi.nlm.nih.gov/IEB/ToolBox/CPP_DOC/lxr/source/src/) and I tested it by emulating blast's tabular output and which I then compared to the output of a blast_formatter run.

### DIFF
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -18,6 +18,42 @@ from functools import reduce
 
 __docformat__ = "restructuredtext en"
 
+# some helper functions for coverage logic
+def length(range):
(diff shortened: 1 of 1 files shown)


peterjc:
I'm going to close this old pull request, but thank you for the ideas and feedback.

PR #437, opened 2014-12-02
landscape.io basic configuration

tiagoantao:
The microscopic file needed to configure landscape.io

### DIFF
--- a/.landscape.yaml
+++ b/.landscape.yaml
@@ -1,0 +1,2 @@
+doc-warnings: yes
+strictness: veryhigh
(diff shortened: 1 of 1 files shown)


tiagoantao:
@peterjc  Should we add this or just close it?

peterjc:
The changes look good - I tried cherry-picking them but was getting a bad object message. Is something out of sync?

Anyway - go ahead and apply the changes as a clean commit to the current master, or with a rebase.

(I'd prefer you not to do a simple merge because the year long branch will just confuse things when viewing the history)

PR #438, opened 2014-12-03
Proposal of new Biopython module: bold

carlosp420:
As I mentioned in an email to the dev list some time ago, I have been working on module to perform calls to the BOLD database via their API. The BOLD database contains more than 1 million public DNA barcode sequences (part of the COI gene). One of the most interesting services is the possibility of sending the barcode sequence and retrieving the taxon identification and more metadata from the BOLD servers.

I just migrated the code to Biopython from a temporal Github repository. You can see the documentation here https://bold.readthedocs.org/en/latest/usage.html that covers all the API methods provided by BOLD.

This module includes unittests for 99% coverage. The tests and docstrings have been tested in Python 2.6, 2.7, 3.3, 3.4 and pypy.

I completed all the work that I could think of, hence the pull request. I am open to feedback on this.


ericmjl:
@carlosp420, this is something that just came in handy as I was embarking on the final stages of this paper I'm working on! Thank you for building this module, I give it an up-vote for inclusion in biopython. After all, biopython has the ability to query multiple databases, so the BOLD database should be one as well.

carlosp420:
I am glad that it was helpful!

peterjc:
It is always a good sign when a second person finds your code directly useful - thanks for commenting @ericmjl :)

I see we never did bring this proposal to the main discussion mailing list, better late than never:
http://lists.open-bio.org/pipermail/biopython-dev/2015-May/020932.html

carlosp420:
yes, I had forgotten about this!

PR #446, opened 2014-12-12
setup.py now uses setuptools if available

uweschmitt:
I tested my modifications building source, binary egg and wheel distribution files. Local install and develop install work too.

I added this change in order to support creation of binary wheels, especially on Windows. The windows installer is not "virtualenv aware" which restricts its use.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,12 +27,20 @@ import sys
 import os
 import shutil
 
-from distutils.core import setup
-from distutils.core import Command
-from distutils.command.install import install
-from distutils.command.build_py import build_py
-from distutils.command.build_ext import build_ext
-from distutils.extension import Extension
+try:
+    from setuptools import setup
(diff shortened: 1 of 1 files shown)


peterjc:
I'm proposing we merge this now that Biopython 1.65 is out, and see if there are any unexpected side-effects: http://lists.open-bio.org/pipermail/biopython-dev/2015-January/020816.html

peterjc:
OK, somewhat later, but now that Biopython 1.67 is out I have cherry-picked this as 489d200406c3e6e9e4f14e68ac6f48bd4095b7c6

Apologies for forgetting about this pull request. I was reminded of it as we'll have to use setuptools if we are going to provide wheels on PyPI: e317c0178074d10a9a5b8e5c50ddf81e06f58c3e

Thank you!

PR #457, opened 2015-01-07
Fix codeml tree parsing

### LINKED ISSUE #453: Regexp for matching trees in PAML (codeml) output
Hi all,
I've noticed that the parser doesn't always capture trees from codeml output. I can see in the code (here: https://github.com/biopython/biopython/blob/master/Bio/Phylo/PAML/_parse_codeml.py#L174) that the regex for deciding if the line contains a tree is "\\(\\(+" which I take to be 'left parenthesis followed by one or more left parentheses'. An example tree from one of my result files looks like this:

(GLN_GLV_YALI0E29865G: 8.494319, GLN_GLV_SAKL0A04796G: 1.449617, (((AFL078W: 4.846475, GLN_GLV_KLLA0E19449G: 3.372082): 1.075125, (GLN_GLV_ZYRO0G08558G: 2.931415, ((SMIK_6_402: 0.320244, (SBAY_16_110: 0.368757, (SKUD_16_82: 0.015513, PORT_16_86: 0.008132): 0.279255): 0.132022): 2.861413, (KNAG0D05130: 6.335563, KAFR0K00370: 2.459791): 0.491734): 0.868678): 1.003015): 0.720969, (KWAL56_23998: 2.000445, GLN_GLV_KLTH0H04510G: 1.613190): 2.588057): 0.560065);

i. e. starts with only one parenthesis and naturally the regex doesn't match. Am I missing something here or should the regex actually match lines which start with just one parenthesis in order to always correctly capture trees? If so, is there any lines which don't contain trees but start with a single parenthesis?

Best,
Greg

brandoninvergo:
Fix #453 

I opted to err on the side of caution with regards to the regexp. I didn't trust putting `.*` in the middle of it so I was more explicit about what should be found between the opening and closing parentheses, taking into account the rather, erm, broad definition of the Newick format used in codeml output.


peterjc:
You'll need to tweak the new test for Python 2.6, quoting the TravisCI results:

```
ERROR: Test finding trees in the results, in response to bug #453, where
  File "/home/travis/build/biopython/biopython/Tests/test_PAML_codeml.py", line 519, in testTreeParseVersatility
    self.assertIn("NSsites", results)
AttributeError: 'ModTest' object has no attribute 'assertIn'
```

Rather than ``self.assertIn(a, b)`` we're using the pattern ``self.assertTrue(a in b)`` instead.

brandoninvergo:
Ah right, I forgot that assertIn is new.  I pushed a fix.

brandoninvergo:
Still failing, I see.  I'll check back in on it tomorrow.

peterjc:
Same basic problem, ``.assertIsNot(...)`` is also new in Python 2.7 onwards. Just use ``.assertTrue(...)`` or similar.

brandoninvergo:
Oops, missed that one.  Trying again...

peterjc:
Green light from TravisCI :)

Rebased and applied to master, see https://github.com/biopython/biopython/commit/4ba8d192e26e7f65e626685546107d8f9454ae95 and parent commits.

Thanks!

PR #469, opened 2015-02-01
fast rmsd based structure superimposer

xulesc:
Added a QCP method based structure super imposer which is twice as fast as the SVD based structure super imposer. Comparitive perormance benchmakr can be seen here https://github.com/xulesc/algos/blob/master/sq.bench.png generated using https://github.com/xulesc/algos/blob/master/svd_qcp_bench.py.


xulesc:
added patch for python 3 (tested with python 3.4)

JoaoRodrigues:
Hi,

Just had a quick look at the code. Wouldn't you benefit from using numpy
sum/mean/sqrt functions in the inner functions of the superimposer? Maybe
even take advantage of the vectorization of the arrays?

Cheers and good work

2015-02-02 11:04 GMT+01:00 Anuj Sharma <notifications@github.com>:

xulesc:
hi,

thanks for the comments. the main computation heavy part of the algo is implemented in C. i changed the _inner_product to use numpy sum. the other inner function _rms is not actually used in the super imposition its only for non-aligned rmsd calculation so does not really impact the performance. if there is something specific you noticed that ive missed let me know and i will fix it.

cheers

xulesc:
general query, would you need anything else from me in order to complete integration of the pull request? please let me know if there is something else i can do, thanks!

peterjc:
Given the C/NumPy code, maybe @mdehoon might have some useful comments?

Or @etal perhaps? [The original authors are no longer active on the Biopython mailing list]

JoaoRodrigues:
I think it would be nice to fully disclose the modifications to the
original C code. Are they just the Python bindings?

xulesc:
the modification is pretty much addition of python bindings, some return statements and removal of other functions that are not used by this module. i am not sure what you mean 'fully disclose' link to the original code is included in the comments on the c file.

JoaoRodrigues:
This is what I meant, what were the modifications. If they are minor
changes it's ok, but it's worth, I think, to include exactly what changes
were made to the original. Just so people that read the source know what
changed and if the algorithm produces the same results exactly.

2015-02-10 11:17 GMT+01:00 Anuj Sharma <notifications@github.com>:

peterjc:
@JoaoRodrigues the only C code is the new file ``Bio/QCPSuperimposer/qcprotmodule.c``, the GitHub tab "Files Changed" on this pull request shows the overall changes from all the commits on the pull request.

As I understand it, the code does not change the existing ``Bio.SVDSuperimposer`` but adds ``Bio.QCPSuperimposer`` as an alternative.

xulesc:
@peterjc  that is correct. the qcprotmodule.c is based on an implementation from another source. 
@JoaoRodrigues i can add to the comments an entry saying what the changes are as i described in my comment above to you if that works.

JoaoRodrigues:
@peterjc I understand, but since the C code is taken from another source and modified, it would be better to include explicitly the modifications in the header, just like a commit message for example. If I was to use this code, I'd want to know exactly what changed between the original source and this, lest I make wrong assumptions.

@xulesc That would be great, thanks.

peterjc:
@JoaoRodrigues sorry, I understand your question now: How does ``qcprotmodule.c`` differ from the 3-clause BSD licensed ``qcprot.c`` from http://theobald.brandeis.edu/qcp/ ?

JoaoRodrigues:
@peterjc Exactly. @xulesc already replied and he's willing to add these to the header.

xulesc:
updated

mdehoon:
Nice work! I have just a few comments.

First, I would suggest to change sum(diff * diff) by dot(diff, diff) as sum(diff * diff) makes a new array diff * diff first and then sums it, while dot(diff, diff) avoids making a new array.

For py_FastCalcRMSDAndRotation, can we get a more informative docstring? It's hard to understand from "fast rotation & rmsd" what this function does. In particular I am wondering if there is some way to use numpy instead.

Finally I think that this module should be under Bio.PDB, if it's relevant to Bio.PDB only.

xulesc:
@mdehoon 
>> First, I would suggest to change sum(diff * diff) by dot(diff, diff) as sum(diff * diff) makes a new array diff * diff first and then sums it, while dot(diff, diff) avoids making a new array.
ill fix this and upload a patch

>> For py_FastCalcRMSDAndRotation, can we get a more informative docstring? It's hard to understand from "fast rotation & rmsd" what this function does. In particular I am wondering if there is some way to use numpy instead.
i can add a docstring to explain the function. i am not a numpy pro but i was not able to make it faster with py functions in my intial attempts. primarily because a lot of the operations are not matrix like (sort of expected since its a replacement for doing SVD which is matrix ops.)

>> Finally I think that this module should be under Bio.PDB, if it's relevant to Bio.PDB only.
i am not sure i agree with it being under Bio.PDB because the module basically superimposes structures not necessarily protein structures but i am not very well versed with the BioPython layout so if that seems more reasonable i can move it.

mdehoon:
I have a vague recollection that there were plans to make a Bio.structures module.
Anyway in the past we have made too many top-level modules (like SVDSuperimposer), so while Bio.PDB may not be ideal, I think it's better than making yet another top-level module.

Otherwise I think this pull request is good to go.

JoaoRodrigues:
The Bio.structures module is on hold... a lot of development needed there,
or at least a lot of thought. Yet, I finished my Ph.D in December so I've
been actually had some time to think properly about it and read a bit about
it, I'll post some ideas soon.

As for the placement, I think Bio.PDB is a good place. This will only be
useful for the structures anyway so why keep it out?

Thanks for going over this @mdehoon

Cheers



2015-02-12 1:32 GMT+01:00 mdehoon <notifications@github.com>:

xulesc:
ok i have moved the module into Bio.PDB, i hope this was the intent of the suggestion.

@mdehoon  thanks you for your comments
@JoaoRodrigues i have been working in on structure related topic in my PhD and would be happy to contribute to such a module if and when you draw up your plans

cheers

JoaoRodrigues:
Bump: @peterjc. This could be merged if nothing is failing with Travis?

peterjc:
@mdehoon are you happy on the C side?

(Given the number of commits, it might be clearer to squash this down to one as part of applying this to the master?)

mdehoon:
I'm a bit lost with the number of commits here. I wrote two comments above, but they may already have been taken care of in the later commits. If so, feel free to ignore them.

xulesc:
hi,
@mdehoon ill change the // entry. as for the py3 and init method. as it is written currently initqcprot is not going to be called for py3 the function PyInit_qcprotmodule will be used. this should work ok i think?

xulesc:
i had addressed your earlier comments. i thought peters comment about squashing the commits into 1 made sense so thats what i was doing.

xulesc:
ok. i have updated the c comments to follow ansi. the commits have been squashed to 1.

mdehoon:
Looks good to me. Thanks!

peterjc:
Thanks all - applied as https://github.com/biopython/biopython/commit/eac50059fd327ede9431676dd1aa8f008f9fa726

xulesc:
thanks all

PR #472, opened 2015-02-10
Write feature qualifiers alphabetically by key.

glebkuznetsov:
The motivation for this change is that if you are iterating on versions of, say, a Genbank, then it's helpful for the feature qualifiers to be written in a deterministic order so you can more easily track changes.

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -346,7 +346,8 @@ class _InsdcWriter(SequentialSequenceWriter):
                + self._wrap_location(location) + "\n"
         self.handle.write(line)
         # Now the qualifiers...
-        for key, values in feature.qualifiers.items():
+        for key in sorted(feature.qualifiers.keys()):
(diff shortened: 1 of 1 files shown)


peterjc:
Alphabetical sorted is probably better than whatever order the dictionary used (as now).

Are you aware of any NCBI guidance or apparent usage pattern here?

glebkuznetsov:
A quick search doesn't reveal anything for me. I don't feel strongly about the commit either way.  It's just useful on my end. Offering it up in case you judge it useful.

peterjc:
Thanks - I'm currently travelling but plan to merge this and acknowledge you as contributing to the next release. Probably in a day or two?

peterjc:
It is a bit late to ask now, but would you agree to dual license your Biopython contributions as discussed on #898 please? Thank you.

glebkuznetsov:
Yes I agree.

peterjc:
Great, thanks!

Issue #406, opened 2014-11-03
PEP8 cleanup (without functional code changes)


peterjc:
Tracking issue for applying small PEP8 changes like white space which do NOT alter the code behaviour, and discussion of dividing labour to avoid stepping on each other's toes. i.e. You can comment here if you are planning to work on a batch of modules.

Ideally this will allow us to routinely use a git pre-commit hook like https://github.com/cbrueffer/pep8-git-hook as outlined on the mailing list thread, http://mailman.open-bio.org/pipermail/biopython-dev/2014-October/020741.html

CC @carlosp420 and @twrightsman

peterjc:
Having been using the commit hook on (some of) my own machines, there are still plenty of (harmless) white space issues in the code base which do not follow PEP8, e.g. spaces around mathematical operators. I've been fixing this as an when I happen to touch a file for another reason.

carlosp420:
I included these exceptions in my pre-commit hook:

E261 	at least two spaces before inline comment
E225 	missing whitespace around operator
E226  	missing whitespace around arithmetic operator
E227 	missing whitespace around bitwise or shift operator
E228 	missing whitespace around modulo operator

as suggested here
http://mailman.open-bio.org/pipermail/biopython-dev/2014-October/020748.html

Maybe those are the spaces that you keep finding?


Dr. Carlos Peña
Laboratory of Genetics
Department of Biology
University of Turku
20014 Turku
FINLAND

* Associate Editor: Revista peruana de Biología
http://is.gd/TwbW

* The Nymphalidae Systematics Group
http://nymphalidae.utu.fi/db.php

peterjc:
Yes, and in general I found following these makes the code a little bit clearer - although for some of the more complex mathematical formulas it can be useful to omit the spaces around multiplication, to emphasise visually that happens before any addition/subtraction.

peterjc:
I'm going to mark this as closed, and focus next on adopting the pre-commit git hook, new issue #493

PR #473, opened 2015-02-11
Implement rest of ABIF spec

erasche:
This PR implements the rest of the ABIF spec, nearly every key listed in their documentation.

My users were clamouring for channel data from each individual channel for viz. purposes, and I figured "well, while I'm there..." and went ahead and added in the rest of the keys.

- all keys from all instrument models are searched for in the ABIF data structure
- labels are made available for anyone wishing to e.g. import the data structure and annotate tracks for their datasets.


peterjc:
I'm going to ask @bow for comments. Also, given their input on https://github.com/biopython/biopython/commit/b20f3641a5eaae5df1d25de5ece8b7d8db441e3e I'm also curious to hear from Mike Cariaso (@cariaso) and/or David Bulger (@DavidBulger).

On a general point, it will need some tests for the new fields - perhaps requiring some more sample files? This is particularly important given this is a binary file format (which can make writing code valid under both Python 2 and 3 a little harder).

erasche:
Sure, I'll expand the test cases. It shouldn't necessitate any new files, the existing ones have many of these fields alreadt.

bow:
Thanks for adding this, @erasche! Indeed, many of the keys were implemented in the version we have now.

I'm at work now so I can't take a full look at it yet, but I agree with @peterjc that more tests would be useful.

I'll write code-specific comments in the `diff` directly.

Bio/SeqIO/AbiIO.py
```diff
@@ -33,23 +33,250 @@
...
+        'APrN1': 'Analysis Protocol settings name',
+        'APrV1': 'Analysis Protocol settings version',
+        'APrX1': 'Analysis Protocol XML string',
+        'CMNT1': 'Sample Comment',
+        'CTID1': 'Container Identifier, a.k.a. plate barcode',
+        'CTNM1': 'Container name, usually identical to CTID, but not necessarily so',
+        'CTTL1': 'Comment Title',
+        'CpEP1': 'Capillary type electrophoresis. 1 for a capillary based machine. 0 for a slab gel based machine.',
+        'DATA1': 'Channel 1 raw data',
+        'DATA1': 'Channel 1 raw data',
```
bow: These look like duplicate key-value pairs ;)...

DavidBulger: I agree with Bow - they look like duplicates. On another note, we could
have additional useful tags available if we have the program check which
ABI version is being used and then add the tags specific for that version.
This would not take up much running time and would allow for tags like POS,
which are required to generate plots of the sequence and evaluate
heterozygous mutations. Let me know if you need more test files as I have
hundreds of files with various mutations and quality scores available.
Unfortunately, I do not have test files from all the different ABI
versions, which would be useful if a couple lines of code are added to
check the version and read the file accordingly.

All the Best,
David

On Thu, Feb 12, 2015 at 7:39 AM, Wibowo Arindrarto <notifications@github.com

erasche: @bow nice catch, thanks!

@DavidBulger I didn't see the POS tag in the spec, that sounds quite useful. I have a couple dozen test files, but they're pretty assuredly all from the same instrument, so I'd be very interested in having some more test data. 
I had avoided implementing any type of instrument detection for the very reason of not having any test data to validate my answers. Additionally I was concerned that testing for the presence/abscence of some tags as an instrument detection method would be error-prone. For example, see `abi_prism_3100/3100-Avant` which has a few different keys from the `general` set, but all of the keys there are optional.

> "Additional useful tags available...if we...check which ABI version is being used and then add tags specific for that version"

Currently the `abif_raw` data structure comes back with any found keys from any instrument, see [L427](https://github.com/biopython/biopython/pull/473/files#diff-9bba5ad30827093582367314de9a01ebR427), I just merge all instrument keys available as there aren't instrument-specific differences in key definitions. Or did you mean something else by that? 

Thanks for the feedback, all!

DavidBulger: I was referring to checking the instrument ID and then adding all the keys
that have different keys on different instruments (
http://www6.appliedbiosystems.com/support/software_community/ABIF_File_Format.pdf).
However, this may require revision of the SeqRecord file. As for the POS
tag, I was generically referring to the peak position data, which has
different tags for different instruments. By combining the peak position
key with the DATA keys, one can create X-Y plots of the sequence results
and perform various statistics on the sequencing data (relative peak
heights, area under the curve, etc).

Thanks,
David

erasche:
I've added more test cases following in the style of the existing test cases that were there (happy to expand upon them or refactor them if y'all would like). 

- Keys added to each of the individual test files (the individual files already in use in the test cases actually had different sets of keys)
- The single large test case that already tested some of the `abif_raw` data now has supplied data for the rest of the keys.

cariaso:
looks like an improvement to me. no reservations. full speed ahead.

On Thu, Feb 12, 2015 at 6:20 AM, Peter Cock <notifications@github.com>
wrote:

> I'm going to ask @bow <https://github.com/bow> for comments. Also, given
> their input on b20f364
> <https://github.com/biopython/biopython/commit/b20f3641a5eaae5df1d25de5ece8b7d8db441e3e>
> I'm also curious to hear from Mike Cariaso (@cariaso
> <https://github.com/cariaso>) and/or David Bulger (@DavidBulger
> <https://github.com/DavidBulger>).
>
> On a general point, it will need some tests for the new fields - perhaps
> requiring some more sample files? This is particularly important given this
> is a binary file format (which can make writing code valid under both
> Python 2 and 3 a little harder).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/biopython/biopython/pull/473#issuecomment-74055451>.
>



-- 
--
Mike Cariaso
http://www.cariaso.com

Bio/SeqIO/AbiIO.py
erasche: (Maybe I'm still not clear on your comment, if so, my apologies in advance).

Instrument ID seems to be pretty reliably specified in `MODL1` so that can be extracted.

However, as I was building the `_INSTRUMENT_SPECIFIC_TAGS` dictionary, I didn't notice a single case where multiple (different) keys on different instruments mapped to the same data. All of the keys in the `general` set I checked the definitions of, to ensure they were the same for each instrument, before merging them into the general section.

>  peak position data, which has different tags for different instruments

Searching through the official spec for all instances of the word peak, I'm really not seeing anything on instruments other than the 3500. If that information is tagged differently in ABIF files from different instruments, it's not in the spec. If you have examples I'd be happy to add restructure the code to accomodate that.

The data structure in my commit was largely generated from a `pdftotext` rendering of the PDF file, followed by some transforms and then manually merging the keys from each instruments if they had identical definitions.

DavidBulger: This looks great! I apologize for my previous comments, I had not realized
that this was specifically what you were doing (PLOC1 is the tag for my
instrument, which you have already added) :) This is a feature I have been
hoping someone would implement. Thank you!

erasche: Ah, great, happy to help!

Bio/SeqIO/AbiIO.py
```diff
@@ -203,7 +422,11 @@ def _abi_parse_header(header, handle):
         # only parse desired dirs
         key = _bytes_to_string(dir_entry[0])
         key += str(dir_entry[1])
-        if key in (list(_EXTRACT) + _SPCTAGS):
+
+        global_tag_listing = []
+        for x in _INSTRUMENT_SPECIFIC_TAGS:
+            global_tag_listing += _INSTRUMENT_SPECIFIC_TAGS[x].keys()
+        if key in global_tag_listing:
```
bow: Another comment :) ~

I see that this is done inside the `while` loop (i.e. the `global_tag_listing` list being repeatedly rebuilt). Would be an idea to create if before the `while` loop?

Also, you can shorten the code a bit by working on the dictionary `values()` directly. Something like:

```
global_tag_listing = []
for t in _INSTRUMENT_SPECIFIC_TAGS.values():
    global_tag_listing += t.keys()
```

bow:
@erasche thanks again for the pull request and also everyone for the inputs + discussion. I added another small comment on the code, but other than that, I think this is good to go.

Bio/SeqIO/AbiIO.py
erasche: Oh jeez, can't believe I didn't catch that. That's what I get for trying to modify everything in place and not really improve upon the code.

I've moved it to the top just after `_INST...` initialization. Thanks, the `.values()` is a good suggestion.

erasche:
I've updated the PR with fixes addressing the comments. 
Thanks for the feed back everyone!

peterjc:
There was an initial TravisCI failure from a network issue, I set it to re-run...

erasche:
@peterjc looks like the build passed.

peterjc:
Rebased and merged to master, see commits leading up to https://github.com/biopython/biopython

Thank you @erasche and thank you everyone else all for your reviews and comments.

peterjc:
According to our buildbot it looks like we may need a tweak to keep Jython happy... the annoying ``Method code too large!`` has been triggered on ``test_SeqIO_AbiIO.py``, likely due to building the large dictionary ``_INSTRUMENT_SPECIFIC_TAGS`` in on go.

See http://testing.open-bio.org/biopython/builders/Linux%2064%20-%20Jython%202.7%20-%20Online/builds/47/steps/shell/logs/stdio

Usually this can be resolved by building the dictionary incrementally, i.e.

```
_INSTRUMENT_SPECIFIC_TAGS = {}
_INSTRUMENT_SPECIFIC_TAGS['general'] = ...
_INSTRUMENT_SPECIFIC_TAGS['abi_prism_3100/3100-Avant'] = ...
```

I'll tackle this in the coming week once back in the office (unless anyone else wants to get to it first?)

erasche:
Happy to make those changes monday morning.

Seems like data I'd normally move to a .yaml data structure in one of my
own projects, but I've never been comfortable enough with pkg_resources to
start doing that here (unless someone thinks that'd be preferable to
hardcoded dict).
21. feb. 2015 18.22 skrev "Peter Cock" <notifications@github.com>:

peterjc:
Also, spotted via https://landscape.io/github/biopython/biopython/53/diff there are some more duplicate keys in the ``_INSTRUMENT_SPECIFIC_TAGS`` dictionary: ``DATA2``, ``DATA3`` and ``DATA4``.

erasche:
@peterjc okay, refactored/fixed, should be OK to test now.

peterjc:
Rebased as https://github.com/biopython/biopython/commit/f65afe975f21e5588ffc95d035987981e67e3c66 and https://github.com/biopython/biopython/commit/89d167c069d295f96a415ce30ed65b3a0f257d4a - see also https://github.com/biopython/biopython/commit/b979d6b089a75649b4812527ac69083fa6f019af (and issue #406).

I've tested this locally with Jython 2.6, the buildbot should do so again tonight. Thanks!

peterjc:
The buildbot only had Jython 2.7b1, which fails. My local machine had Jython 2.7b2 and worked.

Jython 2.7 Beta 4 was recently released (February 2015), so I'll update the buildslave to use that...

erasche:
Ah...joy. Sorry for all of the trouble regarding this file.

peterjc:
Not your fault - Jython on this buildslave was overdue for an update anyway... I might give up and run the Jython builds on my Mac instead, or another Linux box. Hmm.

erasche:
Okay, need some implementation advice y'all for my next PR regarding ABI files.

Now that this PR has been merged I find myself needing access to some of the keys I failed to include earlier (specifically `DATA9-12`, which [apparently](https://code.google.com/p/seqtrace/source/browse/trunk/seqtrace/core/sequencetrace.py#1087) contains the processed data I need to plot peaks). Since I've yet to determine what manner of magic is required in order to obtain the "processed" data (`DATA9-12`) from the "raw data" (`DATA1-4`), I need to add in code to allow access to keys that may not be [universally applicable](https://github.com/biopython/biopython/pull/473/files#diff-9bba5ad30827093582367314de9a01ebR168) (Currently those are tried and always fail as, e.g. `DATA105-199` will never be found in the file).

As I see it, the best option would be to simply expand the `__global_tag_listing` to include the extra ~200 keys. I really don't like that option as the dict is already pretty massive, adding 200 more keys is a pretty ugly solution, so I figured I'd comment here in case anyone has an opinion on this.

peterjc:
There's always the dictionary we had already added of all the raw data - can you just use that?

erasche:
Should I read that as suggesting the removal of [L451](https://github.com/biopython/biopython/blob/72441d3fcfa77b20df08a37034ab444c119e99a2/Bio/SeqIO/AbiIO.py#L451)? That would likely fix my issue.

DavidBulger:
Yes, in my previous fork of the AbiIO I had to change the data input from
DATA1-4 to DATA9-12 as well for my SeqEval program to work. The original
AbiIO file really mostly used the sequence information instead of the
actual data. A simple change from DATA1-4 to DATA9-12 should fix the
problem.

Here is the AbiIO from my previous fork with those changes:
https://github.com/DavidBulger/BiopythonAbiTracer/blob/master/Bio/SeqIO/AbiIO.py

All the Best,
David

erasche:
Summary of options as I see them:

- replace 1-4 with 9-12 (fixes our issues, but not really extensible in case there's )
- remove line 412, stop checking if the tag is "known" to the global tag dict before adding it to the data structure to process
  - if so, should we remove the huge dict? It's probably nice to have it for programmers, but at that point there would be cases where descriptions wouldn't be available for specific tags (e.g. DATA105..DATA109)

cariaso:
this might be overkill to build, but is a lazy load on first use appropriate?

DavidBulger:
What if you just expand the instrument specific tags for 'DATA9-12' into
four separate keys 'DATA9', 'DATA10', 'DATA11', and 'DATA12' with their own
separate descriptions?

erasche:
@cariaso not sure how one might accomplish that here, are there lazy load examples that I could work from?

@DavidBulger my only concern is that while I'm working on it now, that someone else may want other keys later; I figured I'd try and "future proof" this a bit so no one has to revisit it later for adding *yet more* keys based on instrument specifics.

DavidBulger:
I think those extra DATA files are there for the raw and processed data for
additional dyes (up to a couple hundred extras). For all practical
purposes, there should not be more than 1-10 extra dyes for Sanger
sequencing. What if you just expand the instrument specific tags for
'DATA9-12', 'DATA105-115', and 'DATA205-215'? That should be fairly
future-proof without being too terribly bulky.

DavidBulger:
I can only find information for machines with up to 5 dyes (
http://tools.lifetechnologies.com/content/sfs/posters/cms_088713.pdf). So I
could be wrong about the extra data files holding extra dye information.
Also, when the model changes, this file would need to be revisited again to
add more keys for the new model. Nevertheless, having access to all the
tags would be more thorough. Good luck and thanks again for helping with
this!

PR #501, opened 2015-04-01
Fix SearchIO Blast textual parser to handle files with long bitscores.

### LINKED ISSUE #500: SearchIO: Error if blast text contains long bitscores
I've encountered an error while parsing a blast textual file that contains long bitscore. This is an example of text that makes the parse go wrong:
<pre>
Query= ID

Length=191482
                                                                      Score        E &lt;----
Sequences producing significant alignments:                          (Bits)     Value

  ID                                                             ...  1.683e+05  0.0  

...

Query  78     TTGATAGCGAAGAAACCGATGCGCCTCCGGCACCGTTCATCTCCTCCTCGCTCCAGATGG  137
              ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
Sbjct  91269  TTGATAGCGAAGAAACCGATGCGCCTCCGGCACCGTTCATCTCCTCCTCGCTCCAGATGG  91210
</pre>

The problem happens beacuse the bitscore is longer than usual and so more spaces than 5 are present between "Score" and "E" (where the little arrow in the text is pointing).

So line 214 [here](https://github.com/biopython/biopython/blob/6e001a420c2265c7f4b3547f5180eb99d96f8428/Bio/Blast/NCBIStandalone.py#L212-L218) should better be changed to use a regex.

hcraT:
This pull request contain the change that resolve #500

### DIFF
--- a/Bio/Blast/NCBIStandalone.py
+++ b/Bio/Blast/NCBIStandalone.py
@@ -40,6 +40,7 @@ from Bio.Application import _escape_filename
 
 __docformat__ = "restructuredtext en"
 
+_score_e_re = re.compile(r'Score +E')
 
 class LowQualityBlastError(Exception):
     """Error caused by running a low quality sequence through BLAST.
@@ -76,10 +77,6 @@ class _Scanner(object):
     Methods:
      - feed     Feed data into the scanner.
     """
-    def __init__(self):
-        """__init__(self)"""
-        self._score_e_re = re.compile(r'Score +E')
-
     def feed(self, handle, consumer):
         """S.feed(handle, consumer)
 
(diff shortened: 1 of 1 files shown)


peterjc:
@bow - do you have time to review and test this?

My only minor suggestion would be that we should do ``re.compile(r'Score +E'))`` once and use this consistently here and elsewhere in the file? It should be a fraction faster but more importantly would ensure we use the same search pattern consistently.

bow:
@hcraT This looks good :), and I agree with @peterjc 's suggestion. The regex looks more suitable as a 'private' module-level variable instead of a 'private' instance variable since it will only be compiled only once then.

Thanks for the merge request by the way!

hcraT:
I updated the pull request and moved the regex to the module level as suggested.

@bow I like to contribute back (even though this is a very minor contribution). :)

peterjc:
Squashed down into one commit: https://github.com/biopython/biopython/commit/4ca5eee4abddcfd8a870f8146ac00b8558b087b4

Thank you!

PR #502, opened 2015-04-01
Small fixes and improved parsing for models in the MMCIF parser, inserti...

biomadeira:
...on codes and line_counter.

I was having a problem in parsing some multiple model mmCIF structures. Particularly when the length of the models were variable. I ended up making a few additional changes for consistency mostly. 

Kind regards,
Fábio

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -21,6 +21,8 @@ class MMCIF2Dict(dict):
             tokens = self._tokenize(handle)
             token = next(tokens)
             self[token[0:5]] = token[5:]
+            i = 0
+            n = 0
             for token in tokens:
                 if token == "loop_":
                     loop_flag = True
@@ -46,6 +48,8 @@ class MMCIF2Dict(dict):
                 else:
                     self[key] = token
                     key = None
+
+    # Private methods
 
     def _tokenize(self, handle):
         for line in handle:
--- a/Bio/PDB/StructureBuilder.py
+++ b/Bio/PDB/StructureBuilder.py
@@ -156,9 +156,8 @@ class StructureBuilder(object):
                     disordered_residue.disordered_add(new_residue)
                     self.residue = disordered_residue
                     return
-        residue = Residue(res_id, resname, self.segid)
-        self.chain.add(residue)
-        self.residue = residue
+        self.residue = Residue(res_id, resname, self.segid)
+        self.chain.add(self.residue)
 
     def init_atom(self, name, coord, b_factor, occupancy, altloc, fullname,
                   serial_number=None, element=None):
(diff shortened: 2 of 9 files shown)


JoaoRodrigues:
Looks good,  is there a multi model unit test in the suite? Otherwise we could add a small one.

biomadeira:
@JoaoRodrigues I extended testModel() in test_MMCIF.py with just another parsing step using one of structures which was failing. Both the structures pass now so I would assume I didn't break anything else in the process... Just for sanity I ran the other PDB tests and it seems to be fine. All fine in Travis as well.
Thanks

Bio/PDB/MMCIF2Dict.py
```diff
@@ -47,6 +49,8 @@ def __init__(self, filename):
                     self[key] = token
                     key = None
 
+    # Private methods
```
peterjc: Methods with a single leading underscore are by Python convention private, so the comment is redundant (but harmless)

Bio/PDB/StructureBuilder.py
```diff
@@ -196,15 +195,15 @@ def init_atom(self, name, coord, b_factor, occupancy, altloc, fullname,
                                   % (duplicate_fullname, fullname,
                                      self.line_counter),
                                   PDBConstructionWarning)
-        atom = self.atom = Atom(name, coord, b_factor, occupancy, altloc,
-                                fullname, serial_number, element)
+        self.atom = Atom(name, coord, b_factor, occupancy, altloc,
+                         fullname, serial_number, element)
```
peterjc: Is the good reason for adding a new public attribute ``.atom`` here?

biomadeira: @peterjc It doesn't make much difference, as we are initiating self.atom anyway in that 
`atom = self.atom = Atom()`
But I think I changed it by looking at the style used in init_chain() and init_residue() where the self.chain/residue were used...

peterjc: Sorry, yes - it was already exposed as ``.atom``. My mistake.

The old version just used the local variable ``atom`` as an alias, which would be very slightly faster than ``self.atom``.

peterjc:
@JoaoRodrigues are you happy to handle this PR?

JoaoRodrigues:
Sounds good to me, I'm testing it now on a few cases here.

Em qui, 2 de abr de 2015 às 11:03, Peter Cock <notifications@github.com>
escreveu:

JoaoRodrigues:
Good to go on my side.

peterjc:
Thank you both - merged to master. See also https://github.com/biopython/biopython/commit/b31c789c903ed157a0e0ff545b424500a9d15591

biomadeira:
No problem and thanks to you both ;)

PR #350, opened 2014-07-10
Add MAF support to AlignIO

adamnovak:
This would close #5, which has been open for a couple of years.

I've taken the MAF support that @polyatail wrote and merged it against the current mainline Biopython. All the relevant unit tests pass.

Is there anything blocking the merge of this code? I really need MAF support, and I'm sure others would like to have it too.

PR #503, opened 2015-04-03
Updated the MafIO.py from 2012

blaiseli:
I tried to make the MAF parser more internally consistent with respect to start end end coordinates and checked that the alignments returned by `MafIndex.search` and `MafIndex.get_spliced` were correct.

The code seems functional, I used it for a project.
I didn't check whether it complied with some Biopython standards or something, though.


blaiseli:
I just saw that another pull request existed from july 2014 proposing the integration of the MAF parser to Biopython. I had done some searches before starting to work on my own fork, and the only thing I found was the old 2012 version from polyatail, not the re-worked version from Adam Nowak. Is MAF support available or not in the main Biopython distribution as of now (april 2015) ?

adamnovak:
As far as I know, #350 or an alternative PR still has not been merged, and MAF IO support is still not in mainline biopython.

I'm not sure what's going on with this PR on pypy3 with sqlite, but something is broken.

Issue #510, opened 2015-04-08
GenBank parser runs into infinite loop


kblin:
While parsing input files that for some reason end while in the Features table, the GenBank code designed to skip empty lines triggers an infinite loop.

Pull request with a patch coming up.

peterjc:
Closed with merge of #511, thanks.

PR #511, opened 2015-04-08
GenBank: Avoid infinite loop while parsing

kblin:
While parsing input files that for some reason end while in the Features
table, the GenBank code designed to skip empty lines triggered an
infinite loop.

This patch fixes the infinite loop by breaking out of the "consume empty
lines" loop when readline() returns '' (readline()'s way of
saying "end of file") while still supporting the original "consume empty
lines" use case where readline() will return '\n'.

Please note that the provided test case causes the unpatched code to get
stuck in an infinite loop without the provided patch.
This fixes issue #510.


peterjc:
Thank you

PR #519, opened 2015-04-16
Fix for improper 

jecorn:
I attempted to file a bug report for this on the redmine tracker, but the tracker appears to be down.

This is a fix for SeqFeature CompoundLocation.extract to properly deal with features on the reverse strand. The old behavior was to revcomp sequences but still join them in the original order, leading to nonsense. Reverse strand sequences are now revcomped *and* reversed in order, so that join yields the correct sequence.

For example,
old behavior: [ATGC][GCAT] -> [GCAT][ATGC]
new behavior: [ATGC][GCAT] -> [ATGC][GCAT]

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -1227,7 +1227,10 @@ class CompoundLocation(object):
     def extract(self, parent_sequence):
         """Extract feature sequence from the supplied parent sequence."""
         # This copes with mixed strand features & all on reverse:
-        parts = [loc.extract(parent_sequence) for loc in self.parts]
+        if self.strand == -1:
+                parts = [loc.extract(parent_sequence) for loc in reversed(self.parts)]
+        else:
+                parts = [loc.extract(parent_sequence) for loc in self.parts]
         # We use addition rather than a join to avoid alphabet issues:
         f_seq = parts[0]
         for part in parts[1:]:


peterjc:
Yes, our RedMine is currently dead (someone is working on it), but we'd already moved to GitHub issues for new issues anyway.

I think you've misunderstood a deliberate design choice with the ``CompoundLocation``, but I'll try to take a look at this later. If you could add a self contained example here that would be great. Thanks.

jecorn:
For sequences on the reverse strand, I'm not sure why it would be good to reverse complement (FeatureLocation.extract) but then assemble in the forward direction (CompoundLocation.extract). I can see just assembling forward and allowing the user to reverse complement later, or reverse complementing and reversing order (my pull request).

I have a genbank-format test case file from NCBI gene whose CDSes are reassembled backwards. But how I can attach that here?

jecorn:
Now I know what's going on. I have a whole bunch of genbank-format files from NCBI, but it seems they have mis-formatted location ranges! The individual location ranges SpanA < SpanB, but each range is in reverse order within the feature so that LocX > LocY. See below for the example.

So it seem the problem is with malformed input instead of Biopython itself. But would it be worthwhile to add some kind of check for this in Biopython?

     CDS             join(complement(71114..71192),complement(38200..38304),
                     complement(31122..31320),complement(29064..29202),
                     complement(28266..28354),complement(26938..27046),
                     complement(24933..25063),complement(24409..24463),
                     complement(23352..23432),complement(23161..23251),
                     complement(18652..18734),complement(16248..16357),
                     complement(14333..14489),complement(13094..13238),
                     complement(12342..12472),complement(11175..11309),
                     complement(10288..10389),complement(9989..10094),
                     complement(9049..9141),complement(8897..8964),
                     complement(8437..8537),complement(7511..7664),
                     complement(7028..7095),complement(6438..6546),
                     complement(6267..6344),complement(4906..5006),
                     complement(3549..3648),complement(2938..3057),
                     complement(2691..2762),complement(2457..2547),
                     complement(1905..2011))

peterjc:
The idea of the ``CompoundLocation`` is to hold all the fragments (exons) in biological order. The first exon in the list should contain the start codon of a CDS feature. This covers nasty cases like mixed strand features too.

Use gist.github.com for text files, or an NCBI URL ought to be fine. If you have found a record with broken annotation (this does happen sometimes) then we may have to contact the NCBI curators....

peterjc:
P.S. The TravisCI results show your change breaking ``test_SeqIO_features.py`` (which I would have expected).

jecorn:
Yes indeed. This is clearly a case of bad inputs vs bad biopython (if only I'd notice that the input was systematically broken) and the diff shouldn't be merged.  Any way for me to remove the pull request?

peterjc:
I can click on "Close". Or you can delete the branch (``git push jecorn :feature-order-fix``).

But I'd prefer we keep this discussion going as a tracking point for dealing with these apparently mis-annotated entries in the NCBI (they need to be fixed).

peterjc:
@jecorn did you contact the NCBI?

Could you list some of the accessions you found with this problem? Or should I just close this issue?

jecorn:
Hi Peter
Turns out my student generated the genbank files by exporting from a
program (Geneious) that itself slurps from NCBI. But Geneious unfortunately
mangles the format. So problem is with another program rather than official
files or bio python.

Sent from my phone. Sorry for any typos.

@jecorn <https://github.com/jecorn> did you contact the NCBI?

Could you list some of the accessions you found with this problem? Or
should I just close this issue?

—
Reply to this email directly or view it on GitHub
<https://github.com/biopython/biopython/pull/519#issuecomment-101270124>.

peterjc:
Thanks for the update - I'll leave you to raise a bug report with Geneious then. Thank you for your feedback & time.

PR #520, opened 2015-04-18
Miscellaneous Bio.PDB fixes as suggested by landscape.io

JoaoRodrigues:
Mostly PEP8 corrections (mixed indentations, exception not specified, unused variables). Also added checks to .DSSP and .NACCESS to notify the user if there is any error during the calculations and raise the exception if this error prevents output from being created.


JoaoRodrigues:
Spotted a typo. Will re-issue in a minute.

Issue #481, opened 2015-03-02
mmCIF parser and insertion codes


ingvarlagerstedt:
I get an exception when attempting to parse an mmCIF file with insertion codes, the corresponding PDB file parses fine.

When using auth_seq_id, then pdbx_PDB_ins_code should be used for the insertion code. When using label_seq_id, there is no need to use pdbx_PDB_ins_code.

I used 3v11 as an example, which throws an exception on atom 5662, the first atom in the second residue 17 in chain D.

The easy fix is to just ignore auth_seq_id.  The slightly harder is to handle pdbx_PDB_ins_code.

My example program:

```
#!/usr/bin/python 
import sys
from traceback import print_exc
from Bio.PDB import MMCIFParser
from Bio.PDB import PDBParser

def proc(eId):
  structure_id = eId.upper()
  filename = eId+".cif"
  try:
    p=MMCIFParser()
    structure=p.get_structure(structure_id, filename)
  except:
    print "Exception"
    print_exc()

  filename = eId+".pdb"
  try:
    pdb = PDBParser()
    structure=pdb.get_structure(structure_id, filename)
  except:
    print "Exception"
    print_exc()

def main(argv):

  for e in ['3v11',]:
    proc(e)

if __name__ == "__main__":
    main(sys.argv[1:])
```

gave the following output:

```
Exception
Traceback (most recent call last):
  File "mmCIF.py", line 12, in proc
    structure=p.get_structure(structure_id, filename)
  File "/Library/Python/2.7/site-packages/biopython-1.65-py2.7-macosx-10.8-intel.egg/Bio/PDB/MMCIFParser.py", line 25, in get_structure
    self._build_structure(structure_id)
  File "/Library/Python/2.7/site-packages/biopython-1.65-py2.7-macosx-10.8-intel.egg/Bio/PDB/MMCIFParser.py", line 131, in _build_structure
    name, element=element)
  File "/Library/Python/2.7/site-packages/biopython-1.65-py2.7-macosx-10.8-intel.egg/Bio/PDB/StructureBuilder.py", line 234, in init_atom
    residue.add(atom)
  File "/Library/Python/2.7/site-packages/biopython-1.65-py2.7-macosx-10.8-intel.egg/Bio/PDB/Residue.py", line 82, in add
    "Atom %s defined twice in residue %s" % (atom_id, self))
PDBConstructionException: Atom P defined twice in residue <Residue C het=  resseq=17 icode= >
[...]

PR #523, opened 2015-04-20
Further fixes which should also correct problems introduced in my previo...

biomadeira:
...us PR.

I have further fixed the parser, this time for good I believe. My previous PR https://github.com/biopython/biopython/pull/502 fixed things generally but unfortunately had a bug which although correctly parsing insertion codes would still raise duplicated residues' error because of an incorrect if statement that only looked for sequence identifier. (The bit below)

``` python
if current_residue_id != resseq:
    current_residue_id = resseq
    int_resseq = int(resseq)
    structure_builder.init_residue(resname, hetatm_flag, int_resseq, icode)
```

Now it checks whether there is a different insertion code and initiates a new residue if that is the case. 

``` python
if current_residue_id != resseq:
    current_residue_id = resseq
    current_icode = None
    if current_icode != icode:
        current_icode = icode
    structure_builder.init_residue(resname, hetatm_flag, resseq, icode)
else:
    if current_icode != icode:
        current_icode = icode
        structure_builder.init_residue(resname, hetatm_flag, resseq, icode)
```

This problem was the reason I quite believed there could be actual problems in the mmCIF files as addressed in this issue https://github.com/biopython/biopython/issues/481.

A strong reason to believe the parser works as it should this time is because besides the code passing the tests, I have actually loaded every mmCIF in current state of the PDB without problems whatsoever. 

Kind regards,
Fábio

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -64,7 +64,15 @@ class MMCIF2Dict(dict):
                     token += line
                 yield token
             else:
-                tokens = shlex.split(line)
+                try:
+                    tokens = shlex.split(line)
+                except ValueError:
+                    # error "No closing quotation"
+                    line = line.replace("'",'"')
+                    # if odd - add a closing " to that line
+                    if not line.count('"') % 2 == 0:
+                        line = '{}"'.format(line)
+                    tokens = shlex.split(line)
                 for token in tokens:
                     yield token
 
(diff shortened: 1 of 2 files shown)


peterjc:
Sorry this seems to have been forgotten about - Is this old pull request still relevant? If so, it would need updating...

biomadeira:
@peterjc This has been long ago... I am sure people would have noticed if there were major problems with the mmCIF parser.

peterjc:
OK, thanks.

nanand2:
This is still an issue

peterjc:
@nanand2 it sounds like you have a test case where this breaks (or something similar goes wrong). Please open a new issue with full details.

PR #533, opened 2015-04-29
Added functions for kinematics

SuperRaccoon:
Not sure how useful this is. This is for a small class assignment in which we must submit a pull request to an open source project on Github.

### DIFF
--- a/rateFunctions.py
+++ b/rateFunctions.py
@@ -1,0 +1,26 @@
+#use the appropriate function of your choice
+#k is always positive
+#t is always positive as well
+#A0 is the initial amount
+
+
+#the functions all return the expected concentration/amount of A at time t
+
+def firstOrder(k, A0, t): 
+	return (A0*(2.71828182845904523536028747135266249775724709369^(-1*t*k)))
+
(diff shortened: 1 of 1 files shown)


rateFunctions.py
```diff
@@ -0,0 +1,26 @@
...
+def zeroOrder(k,A0,t):
+	return ((-1*k*t)+A0)
+
+
+
+
+#this function gives reaction rate k for an Arrenheius reaction
+#give T in KELVINS, and Ea in JOULES
+def Arrenheius(A,Ea,T):
+	return (A*2.71828182845904523536028747135266249775724709369^(-1*Ea/(8.314*T)))
```
mschmo: This magic number should be stored in a variable.

etal: This is Euler's *e*, available in `math.e` -- and since it's only used for exponentials here, this expression can be:

    A * math.exp(-Ea / (8.314 * T))

mschmo: Sweet, didn't even notice that. Can do same for `firstOrder()`.

Issue #536, opened 2015-05-05
mmCIF parser file handler


sunhwan:
mmCIF parser currently only accept filename as an argument. The MMCIF2Dict.py should be edited as below to be consistent with other parsers.

Current:

```
from Bio._py3k import input as _input

import shlex

class MMCIF2Dict(dict):

    def __init__(self, filename):
        with open(filename) as handle:
            loop_flag = False
```

Edit: 

```
from Bio.File import as_handle
from Bio._py3k import input as _input

import shlex

class MMCIF2Dict(dict):

    def __init__(self, file):
        with as_handle(file) as handle:
            loop_flag = False
```

peterjc:
That looks good, except ``file`` is a reserved word in Python so better not used as the argument name. I would leave it as ``filename``, but add a docstring to explain this can be a handle.

(The whole module would really benefit from some proper docstring documentation)

Could you do this as a pull request please?

peterjc:
The pull request #537 looks good, other than needing a docstring. Thank you.

If you wanted to go further, it would be great to add unit tests for ``MMCIF2Dict`` in ``test_MMCIF.py`` (using a handle and using a filename).

speleo3:
Looks like this ticket can be closed (should have been closed with the merge of #537).

peterjc:
Thanks!

PR #537, opened 2015-05-05
Edited MMCIF2Dict to use either filename or file handler

sunhwan:
MMCIF2Dict is edited to use either filename or file handler, which is consistent behavior of other PDB parsers.

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -14,8 +14,14 @@ import shlex
 
 
 class MMCIF2Dict(dict):
+    """Parse a mmCIF file and return a dictionary."""
 
     def __init__(self, filename):
+        """Parse a mmCIF file and return a dictionary.
+
+        Arguments:
+         - file - name of the PDB file OR an open filehandle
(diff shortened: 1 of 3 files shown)


peterjc:
That looks good for #536, but could you also add a docstring please?

sunhwan:
I'm new to this project. Is adding "test_<module>.py" in the same directory is the convention here? I don't see any other unit test in the same directory. So, let me know. Thanks.

peterjc:
There's a chapter on our test framework in the tutorial, but yes ``run_tests.py`` looks for ``test_*.py`` files. http://biopython.org/DIST/docs/tutorial/Tutorial.html

In this case, as I commented on #536, you should expand the existing ``test_MMCIF.py`` file (rather than adding a new ``test_*.py`` file).

sunhwan:
Thanks, docstring and test is added as well.

peterjc:
Applied to master as commits leading up to https://github.com/biopython/biopython/commit/187c44937ebf1eb553d7e1f95104e8ee42dd6dda

Thank you :+1:

PR #539, opened 2015-05-06
Parse more with Emboss.PrimerSearch.read()

bsmith89:
Adds three new properties to Emboss.PrimerSearch.Amplifier objects which
are filled in upon reading a PrimerSearch output file:
-  seq_id: a string
-  seq_description: the first line after the ID
-  binding_sites: a dictionary of primer sequence mapped to a tuple of hit
   location and number of mismatches

The location element is the output from PrimerSearch minus 1, to be zero
indexed.

Appropriate tests of the new functionality have been added to the
PrimersearchParseTest test-case in Test/test_EmbossPrimer.py


peterjc:
@widdowquinn Is this something you would be able to review/comment on Leighton?

bsmith89:
Any chance of this being merged?

peterjc:
The code and tests look fine, I was just hoping to delegate reviewing it to @widdowquinn or someone ;)

Are there any concerns here with different versions of primer3 output - or is that all stable?

bsmith89:
I wish I could answer that question, but a quick browse of the primer3 and emboss documentation don't tell me much.  I'll leave it to people more informed.

peterjc:
Cross reference #1752

peterjc:
Looking at the current master branch, nothing like this was ever merged - so there is some value in resurrecting this branch/PR but it wouldn't be trivial to do. I am sadly therefore going to close this, apologies that this wasn't addressed more timely.

PR #545, opened 2015-05-21
Change a couple py2 prints to py3 in Tutorial

anderspitman:


peterjc:
Well spotted, but we're trying to use print statements/functions which will work on both Python 2 and 3 unmodified. This means keeping things very simple...

peterjc:
Belatedly cherry-picked and included as https://github.com/biopython/biopython/commit/392a4ee3e13deeb268fe10cbecdd75ab07307100 - see also https://github.com/biopython/biopython/commit/341b69ddeb337a907710a5ab413ad68a13c8f258 and https://github.com/biopython/biopython/commit/3b53514616309f0c318591931d8f248d8e6209b2

This will be part of Biopython 1.67, thank you.

PR #546, opened 2015-05-22
Fix codon alignment using a codon_table

maclandrol:
The user codon_table  is not used when performing codon alignment based on a protein multialignment and nucleotides sequence. I was trying to make a codon alignment using mitochondrial proteins and I got an error resulting of TGA being considered as a stop codon instead of Tryptophan, although the genetic code was provided.


peterjc:
Thanks @maclandrol - this looks sensible. @zruan can you review this please?

P.S. It looks like the build function's docstring is out of date too - it doesn't mention the codon table argument at all.

zruan:
Thank you @maclandrol and @peterjc ! The changes look good to me.
I noticed that there are some typo and missing args, such as codon_table and frameshift. I give take another look at it and have an update shortly.

peterjc:
Great - thanks for looking at this @zruan, and a big thank you to @maclandrol for your first pull request to Biopython :)

PR #552, opened 2015-05-29
bug fix to handle NoneType when writing to newick with branch_length_only=True

ajm:
Writing a newick file with branch_length_only=True can crash with a NoneType exception (I have noticed this when using trees read from PhyloXML files).

The problem is probably with the PhyloXML reader, but as other options (e.g. when writing with support and branch lengths) have the same "or 0.0" to guard against this, maybe this is by design.

### DIFF
--- a/Bio/Phylo/NewickIO.py
+++ b/Bio/Phylo/NewickIO.py
@@ -319,7 +319,7 @@ class Writer(object):
         elif branch_length_only:
             # write only branchlengths, ignore support
             def make_info_string(clade, terminal=False):
-                return (':' + format_branch_length % clade.branch_length) + _get_comment(clade)
+                return (':' + format_branch_length % (clade.branch_length or 0.0)) + _get_comment(clade)
 
         else:
             # write support and branchlengths (e.g. .con tree of mrbayes)
--- a/Tests/test_Phylo.py
+++ b/Tests/test_Phylo.py
@@ -143,6 +143,19 @@ class IOTests(unittest.TestCase):
         os.remove(tmp_filename)
         self.assertEqual(13, count)
 
+    def test_convert_phyloxml_to_newick_branch_length_only(self):
+        """Write phyloxml with bootstrap values to newick format using branch_length_only=True"""
+        trees = Phylo.parse(EX_APAF, "phyloxml")
+        tmp_filename = tempfile.mktemp()
+
+        try :
+            Phylo.write(trees, tmp_filename, "newick", branch_length_only=True)
+            os.remove(tmp_filename)
+            self.assertTrue(True)
+
+        except TypeError:
+            self.assertTrue(False)
+
     def test_int_labels(self):
         """Read newick formatted tree with numeric labels."""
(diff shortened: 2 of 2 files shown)


peterjc:
I'll leave this to @etal to review, but could you also add a test to which would fail prior to your fix? Probably ``test_Phylo.py`` would be the best place...

Thanks!

ajm:
I added a test. Fortunately there was already a suitable PhyloXML example that caused the exception to be thrown.

peterjc:
The TravisCI test failure looked like an unrelated timeout; I asked TravisCI to rerun that test...

etal:
Thanks for catching and testing this. For the fix, there should be a
distinction between a branch length of `0.0` and no branch length specified
(`None`). I think the Newick file should simply not print anything for the
branch length in the latter case. The Newick parser should be able to
handle round-tripping that.

>

ajm:
assertRaises() would not catch the bug because it is to check for the _correct_ raising of an exception. Looking it up, it seems that the correct way to do this is with self.fail() in the except block, so I changed it to that.

peterjc:
Whoops - I read that backwards.

In fact there are already several examples of ``self.assertTrue(False, "some message")`` in our code which ought to be ``self.fail("some message")`` - do you want to fix them on another pull request, or should I do it?

ajm:
I will make the changes for the other tests (I only found 6 instances) and do another pull req.

@etal I am not sure if there is really any correct way to do this. I would imagine that some programs would complain if they were provided with a tree where there were only branch lengths specified on a subset of nodes. I would think setting it to 0.0 is a safer option that will avoid breaking a lot of fragile code.

peterjc:
@etal Was there anything stopping us merging this?

@ajm Are you happy to dual licence your contributions under both the Biopython Licence Agreement AND the 3-Clause BSD license? See #898.

And are you happy to be thanked by name in the ``NEWS`` and ``CONTRIB`` files?

ajm:
@peterjc Yes, the dual licence is fine and by all means include my name. Thanks.

etal:
If we're comfortable with assigning a branch length of 0.0 to the root node when writing Newick files with branch lengths, then this patch looks ready to merge.

peterjc:
Is it easy to special case the root node as (if I understand correctly) it doesn't seem to make sense to give that a branch length?

etal:
Another code path already fills in empty branch length values with 0.0, so this should be fine.

To handle root, since the serialization function is recursive, it might be easier to trim off the 0.0 at the end.

peterjc:
@etal do you want to merge this as is, or further refine it for the root length?

etal:
Seems fine as is.

peterjc:
OK, i'll merge this during the week if you don't beat me to it @etal ;)

peterjc:
We forgot about this one didn't we? Happily rebasing this to the current master was very easy (just black reformatting). I couldn't update the PR (the GitHub permissions settings probably predate that functionality being available by default), so I applied this at the command line instead.

PR #557, opened 2015-06-03
Created indexer for seqxml files

xzy3:


peterjc:
@xzy3 My apologies for not looking at your pull request properly until now.

Your approach using the expat parser and the ``.CurrentByteIndex`` attribute seems much more robust than how we're currently doing the UniProt XML indexing using regular expressions. I like this :)

It looks like currently the code does not properly record the raw record length in bytes, which is probably why you ended up creating the ``_FileChain`` class. I don't think we need this.

Would you be willing to dual licence your contribution under both the current *Biopython License Agreement* and the *3-Clause BSD License*? See #898

xzy3:
It's been well over a year since I did this so I'll try and remember what was going on. The _FileChain is because expat will not parse partial xml files. So _FileChain adds the minimal amount of preamble to get expat to parse the fragment and not complain.

I am willing to license the way you say.

peterjc:
Thank you also regarding the licensing.

I plan to explore your approach using expat, and hopefully generalise it to use on the UniProt XML indexing as well.

xzy3:
whoops this was supposed to be a separate pull request.

PR #564, opened 2015-06-19
Added ability to return GC content as a ratio.

chamkank:
While GC content is usually a percentage, many individuals use it strictly as a ratio. Adding the ability to return a ratio instead of a percentage will provide more options for the user and eliminate the need for extra conversions on the user's part.


yhoogstrate:
`gc` has to be convert to float anyway, so `gc = float(sum(seq.count(x) for x in ['G', 'C', 'g', 'c', 'S', 's']))` would do the trick as well

peterjc:
Closed with #3540.

PR #565, opened 2015-06-22
Fix motif weblogo for Python3, closes #475.

### LINKED ISSUE #475: Py3 compatibility problem of Bio.motifs.Motif.weblogo
Biopython version:

    python3 -c 'import Bio; print(Bio.__version__)'
    1.65

Error when calling ```m.weblogo("motif.png")```:
    
    TypeError: POST data should be bytes or an iterable of bytes. It cannot be of type str.

Relative code: https://github.com/biopython/biopython/blob/master/Bio/Motif/_Motif.py#L663

Thank you guys.

anntzer:
Includes test.

### DIFF
--- a/Bio/motifs/__init__.py
+++ b/Bio/motifs/__init__.py
@@ -479,16 +479,13 @@ class Motif(object):
                   'color3': '',
                   'color4': '',
                   }
-        for k, v in kwds.items():
-            if isinstance(values[k], bool):
-                if not v:
-                    v = ""
-            values[k] = str(v)
 
-        data = urlencode(values)
+        values.update(
+            dict((k, "" if v is False else str(v)) for k, v in kwds.items()))
+        data = urlencode(values).encode("utf-8")
         req = Request(url, data)
         response = urlopen(req)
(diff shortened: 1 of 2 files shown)


peterjc:
I don't immediately see why, but TravisCI says this breaks under Python 2.6 which we still support.

anntzer:
The errors look weird and seem unrelated to the patch.  Any chance you can run the test suite locally?

peterjc:
@barwil Can you look at this please? Note there is something strange on Python 2.6 according to the TravisCI tests (which I did re-run and the failure looks the same).

anntzer:
Fixed now.

peterjc:
Thanks for fixing the Python 2.6 incompatibility. This looks good except I realise we ought to move the new test so that it respects the ``run_tests.py --offline`` switch (see the chapter in our tutorial). I'll do that now...

peterjc:
See https://github.com/biopython/biopython/commit/98ad63d20def54cfb5bec636066964bb446d3f1d and preceding commits. Thanks!

Issue #570, opened 2015-07-03
Complement(join) sequences extracted wrong


kehey:
While using the extract method from CompoundLocation.extract(), there is no check for the strand of the sequence. As a result, all sequences on the reverse strand are concatenated in the wrong order, with the first exon being placed last.

Simply check for strand and do parts.reverse() before pasting the exons together.

See pull request https://github.com/biopython/biopython/pull/571

peterjc:
The ordering is deliberate - the sub-parts should be in biological order. Can you give an example? I suspect the problem is in how your feature locations are being created.

kehey:
I see. The features were created using the genbank parser on the Arabidopsis genbank file of plant ensembl version 27. So I should check whether the genbank file orders them correctly instead, got it, thanks! But that means that the parser does not reorder them if the genbank has them reversed. So i could code a check for that to be pulled into biopython.

peterjc:
The GenBank/EMBL parser should reorder the sub-locations to match what is expected by the ``CompoundLocation`` object model.

If it helps, ``join(complement(A..B),complement(C..D))`` is the same as ``complement(join(C..D,A..B))`` although the former is no longer widely used in the INSDC tables.

What's the URL of this particular Arabidopsis GenBank file?

kehey:
Genbank file: ftp://ftp.ensemblgenomes.org/pub/release-27/plants/genbank/arabidopsis_thaliana/Arabidopsis_thaliana.TAIR10.27.dat.gz
The version of Bio is 1.65

Minimal example that exhibits the issue(using the file locally):

```python
from Bio import SeqIO
from Bio.Seq import Seq
from Bio.SeqRecord import SeqRecord
from Bio.Alphabet import generic_protein, generic_dna
from bcsplaza.plazautils import mp_public2Bayer
from Bio import SeqFeature

fileName = "Arabidopsis_thaliana.TAIR10.27.dat.gz"

bioGenbankRecord = SeqIO.parse(fileName, "genbank")
 for seq_record in bioGenbankRecord :
    for feature in seq_record.features:
        if feature.type == 'CDS':
            if feature.qualifiers['gene'][0] == "AT1G63670":
                test_feature = feature
                test_seq_record = seq_record
```
Looking at test_feature, you indeed see the parts reordered compared to the file:

```
test_feature
```
[...]

PR #571, opened 2015-07-03
Update SeqFeature.py

kehey:
Fix compoundLocation extraction for reverse strands


peterjc:
This break the current deliberate functionality (see unit tests covering mixed strand features etc), let's discuss this on issue #570

Issue #556, opened 2015-06-03
Blastxml parser ValueError: need more than 1 value to unpack


breviata:
I'm running into a problem with the SearchIO xml blast parser. I'm analyzing thousands of files with 50 blast results per file. The script works on about 10k sequences and then I get an error on this one file. If I pass the error, I get about 1000 similar cases of the script failing. (CC @peterjc @bow ) 

**ERROR**

```Traceback (most recent call last):
  File "/home/cstairs/SOFTWARE/MitoPipe_Apr2015_faster2.py", line 464, in <module>
    for blast_record in blast_records:
  File "/opt/perun/python_packages/lib/python2.7/site-packages/Bio/SearchIO/__init__.py", line 316, in parse
    for qresult in generator:
  File "/opt/perun/python_packages/lib/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py", line 198, in __iter__
    for qresult in self._parse_qresult():
  File "/opt/perun/python_packages/lib/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py", line 294, in _parse_qresult
    query_id):
  File "/opt/perun/python_packages/lib/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py", line 395, in _parse_hit
    [a.split(' ', 1) for a in full_id_desc.split(' >')]]
ValueError: need more than 1 value to unpack```

**BIOPYTHON MODULE LINE THAT IS FAILING**

```id_descs = [(x.strip(), y.strip()) for x, y in
            		[a.split(' ', 1) for a in full_id_desc.split(' >')]]```

**LINK TO BLAST OUTPUT FILE**

[OUTPUT file] (https://drive.google.com/file/d/0B8CzGBmKCzfgS2F5eWZKYXJyVDg/view)
Offending sequence: TR7258zzzc0_g2_i1_4

**Comment by peterjc** [comment]
Thanks @breviata - the sample XML file should be really useful here.

@bow - are you able to take a look at this?

**Comment by breviata** [comment]
Hi Peter, There's a URL in that post, since it was too big. [Here again] (https://drive.google.com/file/d/0B8CzGBmKCzfgS2F5eWZKYXJyVDg/view) Let me know if permissions are not working.

**Comment by peterjc** [comment]
Good news, I could download the file ``q48651.xml``, and reproduce the exception.

I could parse it using the older ``NCBIXML`` module:

```
from Bio.Blast import NCBIXML
[...]

PR #575, opened 2015-07-08
Fix for issue #556.

bow:
Instead of relying on raw string splits, we now use regex splits to have the parser more robust. We should be able to handle cases where descriptions are not present, provided each ID starts with the '>' character.


peterjc:
Looks good, thanks.

Issue #525, opened 2015-04-20
HmmerIO text parser breaks on sequence hits without individual domain hits


tryptochan:
I am using the latest Hmmer 3.1b2 with a custom library. There is no description for proteins in the library.

The **hmmer3-text** parser fails on hits with "[No individual domains that satisfy reporting thresholds (although complete target did)]", rasing the ValueError "Expected Hit with query description `qdesc`, found `hqdesc` instead".

It seems only for these hits, like 2eq4_A in the example file, the `hqdesc` is assigned to the query description, while for other hits, it is the placeholder "<unknown description>".

The codes in the `_create_hits` function seem to assign the description. But I have no idea of why this is done like that.
```
if not hit:
   hit.query_description = qdesc
```

[A example file] (https://gist.github.com/tryptochan/4e987981b7c4dd11ac6e)

tryptochan:
I was printing out `qdesc` and `hqdesc` before the assertion statement.

I am a little bit confused now because the descriptions are actually both "unknown description" for other good hits. But for 2eq4, it is "unknown description" versus "Protein-arginine kinase ...". I thought the query description should be parsed and populated to hits and agreement is checked.

bow:
Hi @tryptochan ,

Apologies for the delay, but I finally have time to fully work on this now. The boolean check for `hit` is done because the `query_description` property can be set in two ways: from the hit object itself or from all the HSPs object contained within.

The default is to use the HSPs contained within, since all HSPs contain description about their hits and queries. When this is the case and `Hit` has at least one `HSP` in it, `hit.query_description` actually fetches the HSP's `query_description`.

It is possible, for example in your case, to have hits without HSPs. And in this case, the parser sets the `query_description` manually.

Now, the error occurs because the object that contains the `Hit`s, the `QueryResult` object, expects all of its hits to have the same `query_description` value. The part that raises the exception is a validation step that was added to ensure that querying one hit's `query_description` would be the same as querying any other. 

The bug itself, though, is apparently because the parser does not set the correct query and hit descriptions at the underlying HSP (well, the fragments that the HSP contain) objects.

I've made a fix and will push a pull request soon :).

tryptochan:
Thanks!

jeanrjc:
Hi,

I don't know whether I'd rather open a new issue or not.

[...]

PR #577, opened 2015-07-08
Fix for issue #525

bow:
Also with additional tests to ensure the bug does not happen again.

Unit tests are passing fine :).


peterjc:
Cherry picked as https://github.com/biopython/biopython/commit/e26325113769a1bc6b2db061b1b67d94ab5f87ac (also expanded on the git commit comment, in case we ever lose the GitHub issue information). Thanks!

PR #578, opened 2015-07-09
GenBank: Improve error message from EMBL parser

kblin:
When there is content in an EMBL file after SQ or CO lines that is not // or whitespace, the
parser throws an AssertionError. Unfortunately, the error message is less than helpful.
Improve the error message.

This fixes issue #431


peterjc:
Cherry-picked as https://github.com/biopython/biopython/commit/6fc6ad695c04a656e4492a9f00f35d7a6c441509 thanks!

Issue #459, opened 2015-01-09
DSSP support for mmCIF


mathuin:
The PDB folks are now releasing certain proteins as mmCIF instead of PDB.  mkdssp, the binary generated from DSSP version 2.2.1, supports both mmCIF and PDB for input files.  Bio/PDB/DSSP.py only supports PDB files at this time, and it should support both.

mathuin:
Here's what happens if I try doing it by hand.

```
>>> from Bio.PDB import MMCIFParser
>>> m = MMCIFParser()
>>> s = m.get_structure('4U4R', './4U4R.cif')
>>> model = s[0]
>>> from Bio.PDB import DSSP
>>> d = DSSP(model, './4U4R.cif')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pgd/.virtualenvs/pgd-staging/lib/python2.7/site-packages/Bio/PDB/DSSP.py", line 315, in __init__
    raise PDBException("Structure/DSSP mismatch at %s" % res)
Bio.PDB.PDBExceptions.PDBException: Structure/DSSP mismatch at <Residue SER het=  resseq=2 icode= >
```
Running "mkdssp -i 4U4R.cif -o 4U4R.dssp" generates reasonable output.

JoaoRodrigues:
Hi,

I just tested with a smaller structure ([1CTF](http://www.rcsb.org/pdb/explore/explore.do?structureId=1ctf)) and it works pretty well. A protein/DNA complex also passes the test. So my guess is that there is something odd with this particular structure that causes the problem. I'm trying to reproduce the particular error but parsing this takes quite a long time.. I'll get back to this.

JoaoRodrigues:
Turned out this was an uncaught problem in the MMCIF parser. When creating residues, it was looking only for a change in the residue numbering, which might not always be the case: consecutive chains with the same residue number. PR #580 is fixing this and also providing a nicer error message when trying to write multi-char chain ids using PDBIO.

mathuin:
As this issue was resolved almost a year ago, it's okay to close. :-)

PR #580, opened 2015-07-09
Miscellaneous fixes and improvements to mmCIF and PDB parsers

JoaoRodrigues:
Solves issue #459 and the consequent error it should have given: writing a structure object with multi-char chain identifiers with PDBIO. 

Also added a fast mmCIF parser that extracts only coordinate (and anisotropic b-factor) data and drops the shlex dependency. Practical result: 10x speed improvement.


JoaoRodrigues:
@peterjc @lennax We should revisit the entire structure of the MMCIF parser, particularly the dependency on shlex. It takes ca. 5 minutes to parse a structure because of the shlex.split method calls. Maybe adapting the code used by [PDBx](http://mmcif.wwpdb.org/docs/sw-examples/python/html/index.html) (including their CifWriter) would be something to look into in the future.

peterjc:
I presume merging #835 means we can close this older version.

Issue #584, opened 2015-07-17
SeqFeature.extract() fails when feature has negative offsets


kblin:
On a GenBank file with negative numbers in the feature location line, Biopython will throw a warning like: `BiopythonParserWarning: Couldn't parse feature location: '-2..159'` and then continue. However, this will cause the downstream processing of the featuers to fail:
```
  File "Bio/SeqFeature.py", line 354, in extract
    return self.location.extract(parent_sequence)
AttributeError: 'NoneType' object has no attribute 'extract'
```

kblin:
So the thing is that the genbank parser doesn't create a FeatureLocation object for negative offsets. Is this the intended behaviour? That is: is the correct fix to add an "if not self.location is None" check to SeqFeature.extract(), or do we instead fix the parser to always create a FeatureLocation?

kblin:
It looks like we used to (incorrectly) parse this until bc733da09, but IMHO that patch is doing the right thing.
The code explicitly seems to give a "None" feature location on parsing issues, so I guess the correct way to handle this is to fix SeqFeature.extract()? What would be a good return on a "None" location extract()? As we seem to be unwilling to raise a LocationParserError in the parser, raising some sort of error in the extract doesn't feel right.

peterjc:
Being strict, we could fail immediately during parsing on the invalid location, but that breaks a lot of use cases. Therefore we raise a ``BiopythonParserWarning`` allowing the user to process most of the file. Using ``None`` for the ``SeqFeature``'s ``.location`` seems reasonable to me here.

Rather than the ``AttributeError``, perhaps calling ``.extract(...)`` when the ``.location is None``, I suggest we should raise a ``ValueError`` with a more helpful message (i.e. this feature has no location)?

The only other behaviours I can think of is returning an empty sequence, but that seems risky to me, or having ``.extract(...)`` return ``None``?

kblin:
Ok, I think there's two things we can do. I like the idea of the ValueError in extract. That way the error message isn't confusing.
And talking with a co-worker during lunch we had this idea of a "strict mode" and a "relaxed mode" during parsing, where "relaxed mode" would be what we do right now, and "strict mode" raising a ParserError everywhere we currently issue a warning.

peterjc:
We have this strict/permissive mode in the PDB parser, but I think it is overly complex. To me the more elegant option is use the ``warnings`` library to treat ``BiopythonParserWarning`` as an error if you want to be strict.

[...]

Issue #531, opened 2015-04-28
doctest do not seem to work


ghuls:
doctest do not seem to work.

If you change the output of a doctest (so the output if invalid) and run the doctest command later, it still succeeds:
```
python run_tests.py doctest
```

peterjc:
I *think* this is something subtle about local files versus the installed files. If you reinstall Biopython after making your changes, does it then work?

ghuls:
Thanks, after installing it works.
Is it documented somewhere?

http://biopython.org/DIST/docs/tutorial/Tutorial.html#htoc261

peterjc:
It ought to work without installation. This issue is now the documentation for this bug.

benfulton:
Should doctest run without building (python setup.py build) ?

If so, the doctests that rely on Bio._aligners must be discarded as they cannot work.

If not, I think this issue can be closed. After building but without installing:

$ python run_tests.py doctest
Ran 77 tests in 3.950 seconds

peterjc:
Building without installing then testing ought to work:

```
python setup.py build
python setup.py test
```

Or variations like this:

```
python setup.py build
cd Tests
python run_tests.py doctest
```

```
python setup.py build
cd Tests
python run_tests.py --offline doctest
```

peterjc:
The issue here is sometimes (and I never worked out when), rather than testing the newly built code in the ``build/`` folder (which is what ``run_tests.py`` adds to the path), it uses the (typically older) code from an installed copy of Biopython.

Does this work reliably for you?

0. Remove any installed Biopython
1. Install a really old Biopython
2. Checkout the latest Biopython
3. python setup.py build
4. cd Tests
5. python run_tests.py doctest

benfulton:
This seems to work. I used a Jetstream Ubuntu 18.04 image and installed BioPython 1.62 using Python 2.7.15rc1 and 3.6.5.

77 doctests ran successfully in both cases using the latest BioPython. I then modified a Bio.Align doctest so it would fail, and rebuilt. The doctests successfully, well, failed :)

The only thing that occurs to me is a namespace issue, either in the BioPython code or something caused by older versions of Python themselves. Either way, I think this issue should be closed.

PR #585, opened 2015-07-17
Issue 584

kblin:
After discussing how to properly handle this on issue #584, fix the issue by raising a ValueError, and also explicitly check for the BiopythonParserWarning being raised and the subsequent feature having a "None" FeatureLocation.

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -359,13 +359,13 @@ class SeqFeature(object):
         >>> f.extract(seq)
         Traceback (most recent call last):
            ...
-        ValueError: FeatureLocation is None. Check the sequence file for a valid location.
+        ValueError: The feature's .location is None. Check the sequence file for a valid location.
 
         Note - currently only sub-features of type "join" are supported.
         """
         if self.location is None:
-            raise ValueError("FeatureLocation is None. Check the sequence "
-                             "file for a valid location.")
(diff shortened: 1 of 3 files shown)


kblin:
Durn, hang on, apparently I can't use assertIsNone in Python 2.6.

Bio/SeqFeature.py
```diff
@@ -349,8 +349,23 @@ def extract(self, parent_sequence):
...
+        >>> f = SeqFeature(None, type="domain")
+        >>> f.extract(seq)
+        Traceback (most recent call last):
+           ...
+        ValueError: FeatureLocation is None. Check the sequence file for a valid location.
+
         Note - currently only sub-features of type "join" are supported.
         """
+        if self.location is None:
+            raise ValueError("FeatureLocation is None. Check the sequence "
```
peterjc: ``FeatureLocation`` is a little misleading (as could also be a ``FeatureLocation``), maybe ``The feature's .location is None.``?

kblin:
Ok, sure. Patch coming up.

kblin:
And I just learned that doctests need a `python setup.py build` in order to pick up changes. Sorry for all the fuss.

peterjc:
Yeah, the doctest thing is issue #531 - that one puzzles me.

peterjc:
Thanks Kai

Issue #493, opened 2015-03-25
Enforce basic PEP8 style with a pre-commit git hook


peterjc:
See discussion on #406 and on the mailing list (e.g. http://mailman.open-bio.org/pipermail/biopython-dev/2014-October/020741.html ) where @cbrueffer proposed adopting this pre-commit git hook: https://github.com/cbrueffer/pep8-git-hook

I've been using this on Mac and Linux and it works very nicely.

Before adopting this we need to test it under Windows, where we need to know if it works, safely has no effect, or causes problems.

peterjc:
I've since learnt that we can't enforce a git hook within the repository itself (which surprised me as things like ``.gitignore`` work that way). All we can do is document this (e.g. some projects bundle their pre-commit hook as a script in the repository and include a make command to install it).

peterjc:
Comment from @MarkusPiotrowski on the mailing list:

> Regarding Windows: I'm using Git for Windows and flake8 as pre-commit
> hook which can easily be setup by flake8 itself:
>  ``$ flake8 --install-hook=git``

http://lists.open-bio.org/pipermail/biopython-dev/2016-July/021468.html

This is very simple to install, although likely to be quite verbose and complain about many things already in Biopython (like our legacy mixed case module names).

peterjc:
Thinking about this again, I think we should do something like this:

1. Create new file called ``Scripts/git_pre_commit_hook.py`` or similar (it could be written in bash?) within our repository
2. Provide instructions to set this up as a your pre-commit hook, ``cd biopython`` and run ``ln -s Scripts/git_pre_commit_hook.py .git/hooks/pre-commit``

The new script would ideally cover not just the command line tool ``pep8``, but also ``flake8``, ``pydocstyle`` (for PEP257 docstrings) and eventually ``rst-lint`` as well - covering the same core set of checks as our TravisCI/Tox setup. Indeed, depending on how we resolve #883, the TravisCI/Tox setup might be implemented by calling the same script but with an option to check all the files not just the changed files?

peterjc:
Work in progress here which runs flake8 etc and rst-lint (planned for the TravisCI/Tox setup soon):

https://github.com/peterjc/biopython/blob/pre_commit_linting/Scripts/git_pre_commit_hook.py

peterjc:
[...]

PR #598, opened 2015-08-20
Add the Phenotype module

mgalardini:
The Phenotype module could ideally be used to handle growth curves and other similar experiments.
For the moment it features a parser/writer for Phenotype Microarray data files.

Includes tests...


mgalardini:
Just noticed that the changes included a file in AlignIO, now it should be ok.

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+    
+    A PlateRecord stores all the wells of a particular Phenotype
+    Microarray plate, along with metadata (if any). The single wells can be
+    accessed calling their id as an index or iterating on the PlateRecord:
+    
+    >>> from Bio import Phenotype
+    >>> plate = Phenotype.read("plate.csv", "pm-csv")
+    >>> well = plate['A05']
+    >>> for well in plate:
+    ...    print("%s"%well.id)
```
peterjc: PEP8 spacing would be: ``print("%s" % well.id)`` and four spaces relative to the line above.

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+            signal = float(signal)
+        except ValueError:
+            raise ValueError('Signal should be a number')
+    
+        self._signals[time] = signal
+    
+    def __getitem__(self, time):
+        """Returns a subset of signals or a single signal.
+        """
+        if isinstance( time, slice ):
```
peterjc: PEP8 spacing would be ``if isinstance(time, slice):``

See also #493 for automating this checking.

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+                
+            if time.stop is None:
+                stop = max(self.get_times())
+            else:
+                stop = time.stop
+            
+            time = np.arange(start, stop, time.step)
+            return list(self._interpolate(time))
+            
+        elif isinstance( time, int ) or isinstance( time, float ):
```
peterjc: My eye was caught by the non-PEP8 spacing, but you also I would suggest the more concise ``elif isinstance(time, (int, float)):`` here.

However, you might want to allow ``long`` here too (Python 2 only) in which case we have ```from Bio._py3k import _is_int_or_long``` and we could add a similar version for int, long or float?

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+    
+    >>> well.fit('gompertz')
+    >>> print(well.slope, well.model)
+    (127.94630059171354, 'gompertz')
+    
+    If no function can be fitted, the parameters are left as None, except for
+    the max, min, average_height and area. 
+    """
+    
+    def __init__(self, wellid, plate=PlateRecord(None), signals={}):
```
peterjc: Never use a mutable object as a default argument (see eg https://pythonconquerstheuniverse.wordpress.com/category/python-gotchas/ for why), use ``None`` instead and then perhaps:

```
def __init__(self, wellid, plate=None, signals={}):
    # your code here
    if plate is None:
        self.plate = PlateRecord(None)
    else:
        self.plate = None
    if signals is None:
        self._signals = {}
    else:
        self._signals = signals
```

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+            
+        raise ValueError('Invalid index')
+        
+    def __iter__(self):
+        for time in sorted(self._signals.keys()):
+            yield time, self._signals[time]
+    
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            if self._signals.keys() != other._signals.keys():
```
peterjc: This looks like it would only work with Python 2 dictionaries, under Python 3 you would get iterators which would not compare as equal even if the list of keys was. This tells me we need more tests.

Probably you need  ``if list(self._signals) != list(other._signals):`` (since iterating over a dict yields the keys)

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+        if not isinstance( well, WellRecord ) :
+            raise TypeError('Expecting a WellRecord object')
+                            
+        signals = {}
+        
+        times = set(self._signals.keys()).union(set(well._signals.keys()))
+        for t in sorted(times):
+            signals[t] = self[t] + well[t]
+                            
+        neww = WellRecord(self.id, signals = signals)
```
peterjc: PEP8 spacing would be ``neww = WellRecord(self.id, signals=signals)`` I think.

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+            return
+        
+        self.area = get_area(self.get_signals(), self.get_times())
+        
+        if function is None:
+            sigmoid = avail_func
+        else:
+            sigmoid = (function, )
+        
+        mod = __import__ ('Bio.Phenotype.pm_fitting')
```
peterjc: This worries me - in part given subtle Python 2/3 changes in imports. Would a dictionary of functions work better?

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+        data = json.load(handle)
+    except ValueError:
+        raise ValueError('Could not parse JSON file')
+
+    # We can have one single plate or several
+    # we need to discriminate
+    try:
+        data.keys()
+        data = [data]
+    except:pass
```
peterjc: PEP8 layout, move the ``pass`` to the next line. Also what exception are you expecting here? Is there a clearer way to tell one plate from multiple plates from the json?

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+            plateID = pobj[_csvData][_plate]
+        except TypeError:
+            raise TypeError('Malformed JSON input')
+        except KeyError:
+            raise KeyError('Could not retrieve plate id')
+            
+        # Parse also non-standard plate IDs
+        if not plateID.startswith(_platesPrefix):
+            warnings.warn('Non-standard plate ID found (%s)'%plateID,
+                           RuntimeWarning)
```
peterjc: PEP8 spacing around the ``%`` operator. Also I suggest using our more specific warning via ``from Bio import BiopythonParserWarning``

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+                idx[i] = x.strip()
+                
+        elif data:
+            if plate is None:
+                continue
+            
+            # Workaround for bad-formatted files
+            try:
+                float(line[0])
+            except:
```
peterjc: Avoid ``except:``, this should probably be ``except ValueError:`` right?

Bio/Phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1101 @@
...
+        plate.qualifiers = qualifiers
+        yield plate
+
+def _toOPM(plate):
+    """Helper function to transform a PlateRecord object into a dictionary
+    """
+    d={}
+    
+    for k, v in plate.qualifiers.items():
+        d[k] = v
```
peterjc: Probably you can do that in one go, guessing  ``d = dict(plate.qualifiers.items())`` would work?

Bio/Phenotype/pm_fitting.py
```diff
@@ -0,0 +1,124 @@
+# Copyright 2014 by Marco Galardini.  All rights reserved.
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
+"""This module provides functions to perform sigmoid functions fitting to
```
peterjc: Can you give this docstring a one line module summary? See PEP257 https://www.python.org/dev/peps/pep-0257/

peterjc:
Would you be happy switching ``Bio.Phenomics`` to ``Bio.phenomics`` to better follow PEP257 module naming conventions?

mgalardini:
Hi Peter, I've made all the changes you kindly suggested; I had to struggle a bit to fix a couple of syntax errors in python2.6, but now they are fine.

I'll prepare some documentation as soon as I can.

Best,
Marco

Bio/phenotype/PhenMicro.py
```diff
@@ -0,0 +1,1116 @@
...
+        for k, v in wells.items():
+            plate[k] = WellRecord(k, plate, v)
+        plate.qualifiers = qualifiers
+        yield plate
+
+
+def _toOPM(plate):
+    """Helper function to transform a PlateRecord object into a dictionary
+    """
+    d = {}
```
peterjc: You can drop this line now.

mgalardini:
Thanks for the suggestions, they are in now; I'll prepare the documentation during this week, thanks again for the help!

mgalardini:
Hi,

I've added a chapter to the documentation for the Bio.phenotype module. Unfortunately, in order to get my branch up to date to the current form of the tutorial (several chapters instead of one file) I think that I should have made a merge and not a rebase. Is this gonna be a problem? I think I can probably fix it if needed.

Marco

mgalardini:
Hi,

just wondering if there are any news on this Pull Request; I'm not trying to push it, just curious to know whether it requires more work or not.

Thank you very much.

peterjc:
I must have missed the previous GitHub notification before Christmas. Sorry.

You're right the git history has become quite confused - GitHub thinks 270 files have changed which is likely not a good sign, and makes reviewing your work a lot harder.

As long as you are comfortable with git rebase etc, then you could then force push the updated branch, and this pull request would be updated to match. As always, do take a suitable a backup first - just in case I made a copy of your current branch here: https://github.com/peterjc/biopython/tree/mgalardini_phenomics_2016_01_18

mgalardini:
Thank you, I'll try to fix the situation later today!

mgalardini:
See PR #755

peterjc:
Closing in favour of the cleaned-up version on #755, thanks!

PR #610, opened 2015-09-10
Support clustalo tests on Windows

benfulton:
Tested clustalo on windows. When the exe is unavailable, getoutput raises FileNotFoundError. Catch and swallow that so MissingExternalDependencyError is raised. All tests pass when clustalo.exe is available.


peterjc:
See also https://github.com/biopython/biopython/commit/4d41c8b571cc47dcf40cccf93d489a2655aed9f8 - we can't use ``FileNotFoundError`` under Python 2.

We could use ``WindowsError``, but this merged into its parent class ``OSError`` anyway with Python 3.3 onwards.

PR #613, opened 2015-09-14
Parse structured comment

bosborne:
All tests pass, including that print/compare test.


peterjc:
I've asked on the mailing list in case anyone wanted to comment: http://lists.open-bio.org/pipermail/biopython-dev/2015-October/021120.html

Note the last few commits were comment clarifications rather than adding a URL to the documentation. I'll probably squash this into one commit anyway as part of merging it & dealing with the merge conflict.

Once that's done it would be great to update https://github.com/biopython/biopython/blob/master/Bio/SeqIO/InsdcIO.py to write structured comments...

peterjc:
Applied to master as a squashed commit, thanks @bosborne :)

Issue #605, opened 2015-09-05
Bug in - print_that() as 'map' in Restriction


glinshiz:
If sequence contains restriction sites only for one enzyme I receive error "KeyError:(0,) in line 338 in fine PrintFormat.py

peterjc:
Can you turn that into a short self contained example? That would be very helpful for reproducing the problem - and perhaps even turning it into a unit test.

glinshiz:
Here is the example 
It works in the first case, but don't work in the second case 

from Bio import Restriction
from Bio.Seq import Seq

def main():
        rb = Restriction.RestrictionBatch(["NdeI","EcoRI","BamHI","BsmBI"])

        #In the first case the sequence contains two BamHI restriction sites and one NdeI. Both BamHI and NdeI in the RestrictionBatch list.
        my_seq = Seq('CCAGTCTATAATTCG' + Restriction.BamHI.site+'GCGGCATCATACTCGA' + Restriction.BamHI.site +'ATATCGCGTGATGATA' + Restriction.NdeI.site +'CGTAGTAATTACGCATG')
        Analong = Restriction.Analysis(rb, my_seq)
        Analong.print_as('map')
        Analong.print_that()

        #In the second case the sequence contains two BamHI restriction sites and one EcoRV. Only  BamHI  in the RestrictionBatch list.
        my_seq = Seq('CCAGTCTATAATTCG' + Restriction.BamHI.site+'GCGGCATCATACTCGA' + Restriction.BamHI.site +'ATATCGCGTGATGATA' + Restriction.EcoRV.site +'CGTAGTAATTACGCATG')
        Analong = Restriction.Analysis(rb, my_seq)

        Analong.print_as('map')
        Analong.print_that()

if __name__ == "__main__":
        main()

peterjc:
Confirmed:

```

                17 BamHI
                |                                           
                |                     39 BamHI
                |                     |                     
CCAGTCTATAATTCGGGATCCGCGGCATCATACTCGAGGATCCATATCGCGTGATGATAC
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
GGTCAGATATTAAGCCCTAGGCGCCGTAGTATGAGCTCCTAGGTATAGCGCACTACTATG
1                                                         60

 62 NdeI
 |                                                          
ATATGCGTAGTAATTACGCATG
||||||||||||||||||||||
TATACGCATCATTAATGCGTAC
61                          82


   Enzymes which do not cut the sequence.

[...]

PR #614, opened 2015-09-14
Bug in - print_that() as map in Restriction #605

johnbradley:
I made a small change to fix print_as('map')/print_that() from issue 605.
The root cause was an assumption that there would be markers on every row of annotated seq output.
I created three rather large unit tests.
Simplified example of original bug:

    from Bio import Restriction
    from Bio.Seq import Seq

    rb = Restriction.RestrictionBatch(["NdeI","EcoRI","BamHI","BsmBI"])
    my_seq = Seq('CCAGTCTATAATTCG' +     Restriction.BamHI.site+'GCGGCATCATACTCGAATATCGCGTGATGATACGTAGTAATTACGCATG')
    Analong = Restriction.Analysis(rb, my_seq)
    Analong.print_as('map')
    Analong.print_that()
 
Thanks

### DIFF
--- a/Bio/Restriction/PrintFormat.py
+++ b/Bio/Restriction/PrintFormat.py
@@ -317,14 +317,14 @@ class PrintFormat(object):
         for x in range(60, length, 60):
             counter = x - 60
             l = []
+            cutloc[counter] = l
+            remaining = []
             for key in mapping:
                 if key <= x:
                     l.append(key)
                 else:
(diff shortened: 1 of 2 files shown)


peterjc:
Cherry-picked as https://github.com/biopython/biopython/commit/7b7d79fde5438878fcb5480169220cd0ff76cff5 thank you.

PR #617, opened 2015-09-19
Update chapter_kegg.tex

kozo2:
```python
from Bio import KEGG
request = KEGG.get("ec:5.4.2.2")
```

did NOT work


peterjc:
Thank you. Would you like to be added to the ``NEWS`` and ``CONTIB`` files?

kozo2:
Thanks Peter.
Yes. It is much appreciated if you would do so.

PR #618, opened 2015-09-22
add how to save ec:5.4.2.2.txt

kozo2:
I think it would be better to add this data download process.


kozo2:
I found a miss in Windows.
I will re-send this pull request.

PR #619, opened 2015-09-22
Windows does NOT accept colon(:) for file name.

kozo2:
I think it would be better to rename the enzyme file (and add how to download the file).


peterjc:
I agree, commit merged - thanks!

PR #620, opened 2015-09-23
Support MSAProbs testing on Windows

benfulton:
All tests pass. One test checks for a segmentation fault return code which is probably different on every platform.


peterjc:
Applied as https://github.com/biopython/biopython/commit/dc22c706799f8860eb7eac7500ca1d07b5f936a2 thank you.

Issue #625, opened 2015-09-24
Update Bio.Sequencing.Applications wrapper for samtools 1.x


peterjc:
There have been enough command line changes from samtools 0.1.19 to the 1.x releases that ``test_samtools_tool.py`` now fails as reported by @tillea testing with samtools 1.2

http://lists.open-bio.org/pipermail/biopython/2015-September/015753.html

Filing this as an issue on GitHub to avoid us forgetting about this. CC @saketkc who's offered to work on this.

peterjc:
Looks good under samtools 1.2.1 but I found a regression in samtools' develop branch while testing this: https://github.com/samtools/samtools/issues/475

PR #628, opened 2015-10-06
Fix tests for samtools 1.x support

saketkc:
See #625 
I might not be able to add full support  for all new `samtools-1.x` options. But this fix is backward compatible. Tested with `samtools 0.1.20 and 1.2`


peterjc:
Applied as https://github.com/biopython/biopython/commit/427872dd687324da02e6feac06b5c623fa0734df thanks!

Issue #603, opened 2015-09-01
pairwise2 global(local)xx bug and memory leak


sergpolly:
I was happily using `pairwise2.align.globalxx(s1,s2)` to align nearly identical sequences up until I faced a couple of ~4000 long proteins - `globalxx` stalls there and keeps leaking memory (30 mins of runtime and >100GB of RAM is not a limit, I've never waited any longer). Minimal example:

```
from Bio import pairwise2
seq1 = ''.join(['A',]+['I' for x in range(4000)])
seq2 = ''.join(['V',]+['I' for x in range(4000)])
aln = pairwise2.align.globalxx(seq1,seq2)
```

getting the matches-score only (which is all I need actually) as `pairwise.align.globalxx(seq1,seq2,score_only=True)`, make things better - `globalxx` does yield an answer (seems to be correct), but for those made up 4001 proteins it takes ~5 mins. Smaller sequences (~40,~400) do align well and in a reasonable time (hundreds of milliseconds scale), though.

importing cpairwise2 does not raise any exceptions, so I'm guessing it's C implementation running.
My `sys.version` output is `2.7.6 (default, Jun 22 2015, 17:58:13) [GCC 4.8.2]`
`Bio.__version__` is 1.65
OS `Ubuntu 14.04.3 LTS`

peterjc:
If you are using the C code, this will work:

```
>>> from Bio import pairwise2
>>> from Bio import cpairwise2
>>> pairwise2._make_score_matrix_fast is cpairwise2._make_score_matrix_fast
True
```

Your example also behaves badly on Mac OS X. Deliberately running the Pure Python version:

```
$ cd ~/repository/biopython
$ python
Python 2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from Bio import pairwise2
>>> seq1 = ''.join(['A',]+['I' for x in range(4000)])
>>> seq2 = ''.join(['V',]+['I' for x in range(4000)])
>>> aln = pairwise2.align.globalxx(seq1,seq2)
^C
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "Bio/pairwise2.py", line 317, in __call__
    return _align(**keywds)
  File "Bio/pairwise2.py", line 338, in _align
    score_only)
[...]

PR #631, opened 2015-10-07
pairwise2 global(local)xx bug and memory leak #603

johnbradley:
Changes to reduce memory foot print of pairwise2.align.globalxx.


peterjc:
@MarkusPiotrowski is this work still relevant now that we've merged #782 which updated the C and Python code?

PR #634, opened 2015-10-11
switch to Unittests for test_seq.py

carlosp420:
I [read that contributions for the test suite](http://mailman.open-bio.org/pipermail/biopython-dev/2015-October/021064.html) are needed.

In this patch, I replaced the tests in ``Test/test_seq.py`` for unittests. I also noticed that the test coverage for the module Bio.Seq was at 61%. So I added tests to get higher coverage, currently 96%.

Some of the test cases for exceptions in the code required using assert statements like this:

```python
with self.assertRaises(ValueError):
   do_something_to_trigger_exception
```

Those cases were failing in Python2.6 due to its old unittest library. That's why I am importing the library ``unittest2`` and installing it in ``.travis.yml``. The usage of ``unittest2`` can be removed after Biopython drops support for Python 2.6.

Sorry for the long patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ before_install:
   - "export MSQL=$PG"
   - "if [[ $PY3 == 'yes' ]]; then export MSQL=no; fi"
   - "if [[ $MSQL == 'yes' ]]; then pip install mysql-python; fi"
+#Lib unittest2 is needed for some assertRaises tests under Python 2.6
+  - "if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2; fi"
 
 install:
   - "/usr/bin/yes | python setup.py install"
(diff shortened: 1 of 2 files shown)


peterjc:
This looks good, thank you!

I would like to be able to use ``with self.assertWarns(BiopythonWarning):`` to check warnings are issued rather than ``with warnings.catch_warnings(..)`` which just silences them, but that was only included in Python 3.2 onward. Perhaps we should insist on ``unittest2`` for Python 2 support?

carlosp420:
yes, it is good to have ``self.assertWarns``. I had to change them ``warnings.catch_warnings`` due to the old ``unittest`` library. Maybe ``unittest2`` can be required for Python versions earlier than 3.4?

peterjc:
See http://lists.open-bio.org/pipermail/biopython-dev/2015-October/021086.html

PR #635, opened 2015-10-12
some small fixes to Bio.Seq

carlosp420:
so the code looks prettier.


peterjc:
We currently have a de-facto coding style which does string formatting the old fashioned way familiar to C programmers. This probably reflects the fact the early Biopython code was written before the ``.format()`` method was added to Python.

I wonder if adopting the ``.format()`` method deserves wider discussion?

peterjc:
Applied, thanks.

Switching to ``.format(...)`` seems sensible even if we don't make it a Biopython coding style convention yet.

PR #638, opened 2015-10-15
increased test coverage for Bio.SeqRecord and other fixes

carlosp420:
such as typos and usage of ``format`` method with the ``!r`` conversion in ``Bio.SeqRecord``.
Coverage increased from 54% to 90%.


peterjc:
Unless anyone objects, I'm OK with requiring ``unittest2`` like this under Python 2.6 to get all the tests to run: http://lists.open-bio.org/pipermail/biopython-dev/2015-October/021090.html

Thanks Carlos.

PR #639, opened 2015-10-15
Use container-based testing in Travis-CI

carlosp420:
The new testing infrastructure in Travis-CI is container-based, faster and
probably better.

It has to be activated by using: ``sudo: false``

See [more info](http://docs.travis-ci.com/user/workers/container-based-infrastructure/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 #
 #Note when testing Python 3, the 'python' command will invoke Python 3
 #and similarly for PyPy too.
-
+sudo: false
 language: python
 python:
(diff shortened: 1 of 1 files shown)


carlosp420:
oh well, by having ``sudo: false`` Biopython cannot install packages such as clustalw and similar, which are unlikely to be white-listed by Travis-CI (see [here](http://docs.travis-ci.com/user/apt/)).

close()

peterjc:
Yes, I'd looked at this but felt on balance having more of the optional binary tools installed was worthwhile.

PR #640, opened 2015-10-18
Increased test coverage for Bio.LogisticRegression

carlosp420:
from 88% to 94%.


peterjc:
Thanks!

Note with a minor refactor we could keep compatibility with the Python 2.6 unittest: https://github.com/biopython/biopython/commit/68e5de5b0593937b1ae6a80ff424fbce7920846e

PR #644, opened 2015-10-24
Bio.Phylo._utils: usage of "if variable" and "if not variable"

carlosp420:
PEP8 recommends against using the following statements:
```python
if variable == True:

if variable is True:
```

PEP8 recommends using instead ``if variable`` and ``if not variable`` when testing
for boolean values in variables.
Also removed redundant parentheses.

I guess this PR needs approval from @etal

### DIFF
--- a/Bio/Phylo/_utils.py
+++ b/Bio/Phylo/_utils.py
@@ -419,14 +419,14 @@ def draw(tree, label_func=str, do_show=True, show_confidence=True,
         Graphical formatting of the lines representing clades in the plot can be
         customized by altering this function.
         """
-        if (use_linecollection is False and orientation == 'horizontal'):
+        if not use_linecollection and orientation == 'horizontal':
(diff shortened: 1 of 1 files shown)


etal:
Looks good to me :+1: 

FYI: An issue that's not the case here, but could arise elsewhere, is if the possible values for a keyword argument are `True`, `False`, or `None`, where `None` and `False` are handled differently, then this automatic substitution wouldn't be valid. (I haven't seen these pull requests run into any cases like this in Biopython yet, so I guess you probably knew this and were watching for it already.)

peterjc:
Thanks both - merged.

And yes, this kind of change does need to be looked at manually in case of usage like your example where the variable isn't simply used as a yes/no boolean.

carlosp420:
thanks, luckily I have so far found easy and obvious cases where only True and False were the only options for such variables.

PR #645, opened 2015-10-24
Update README to suggest using pip install

Chris7:
Updated installation instructions to use pip, and provide another set of instructions for Windows users.

### DIFF
--- a/README
+++ b/README
@@ -30,15 +30,25 @@ http://dx.doi.org/10.1093/bioinformatics/btp163 pmid:19304878
 For the impatient
 =================
 
-To build and install Biopython, download and unzip the source code, go to this
-directory at the command line, and type::
+The simplest way to install Biopython is using the pip installer, which by
+default is included in Python versions 2.7.9, 3.4 and later. Biopython may
+be intalled via pip by the following command:
 
-    python setup.py build
-    python setup.py test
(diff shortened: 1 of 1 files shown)


peterjc:
See this thread http://lists.open-bio.org/pipermail/biopython-dev/2015-October/021105.html - not clear yet if we have a consensus, but also the wiki http://biopython.org/wiki/Download and main installation document https://github.com/biopython/biopython/blob/master/Doc/install/Installation.tex will need updating

Chris7:
While we're on readmes, is there any reason you haven't used any markup language to make it easier to navigate?

Chris7:
Linewidths fixed.

peterjc:
The ``README`` is marked up with RST, see the ``README.rst`` symlink in the repository for GitHub's benefit to trigger the formatted view. This symlink file is deliberately not included in the source releases via ``MANIFEST.in``.

peterjc:
Closing - As of  whttps://github.com/biopython/biopython/commit/8a16a61109f40ed353cac3970a9965e3e7505916e do recommend using pip, but as yet do not talk about specific Windows Python packages like Anaconda. You could raise that on the biopython(dev) mailing list to see if this is being widely used?

PR #647, opened 2015-10-26
Bio.Phylo.PAML: replaced "if len(variable) > 0" by "if variable"

carlosp420:
PEP8 recommends the following:

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are
false.

Yes: if not seq:
     if seq:

No:  if len(seq):
     if not len(seq):
```

The variables in question are either lists or dictionaries.

I guess this PR needs approval from @brandoninvergo.


peterjc:
Looks sensible -  any comments @brandoninvergo or are you happy for these changes to be applied?

peterjc:
Rebased and applied, thank you.

PR #649, opened 2015-10-27
Bio.AlignIO: replaced "if len(variable) > 0" by "if variable"

carlosp420:
PEP8 recommends the following:
    
```
    For sequences, (strings, lists, tuples), use the fact that empty sequences are
    false.
    
    Yes: if not seq:
         if seq:
    
    No:  if len(seq):
         if not len(seq):
```


peterjc:
Thanks, merged.

PR #650, opened 2015-10-27
Fix broken or redirected links

dadoskawina:
Help pages at NCBI were moved. I have corrected broken or redirected links in Bio/Entrez/__init__.py


peterjc:
Thank you, pull request merged - this will be in Biopython 1.67, which is when we would next plan to update http://biopython.org/DIST/docs/api/Bio.Entrez-module.html

PR #659, opened 2015-10-30
Bio/motifs/matrix.py: require C-accelerated PWM code under CPython

mruffalo:
Instead of attempting to import the `_pwm` module in every call to
`PositionSpecificScoringMatrix.calculate`, import it once at the beginning
of the class body. If the module can't be loaded and the code is running
under CPython, raise an ImportError. Otherwise, conditionally define a
private `_calculate` method, using the `_pwm` module if available and a
Python implementation if not.

### DIFF
--- a/Bio/motifs/matrix.py
+++ b/Bio/motifs/matrix.py
@@ -7,6 +7,7 @@ and position-specific scoring matrices.
 """
 
 import math
+import platform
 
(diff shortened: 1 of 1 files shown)


peterjc:
This would partly solve #657 - could you review this please @mdehoon ?

Issue #658, opened 2015-10-30
PyPy3 test failure with _sqlite3.InterfaceError


peterjc:
Recently I've noticed this kind of error from PyPy3 under TravisCI, e.g.

https://travis-ci.org/biopython/biopython/jobs/87728288
https://travis-ci.org/biopython/biopython/jobs/87731411
https://travis-ci.org/biopython/biopython/jobs/88086877
https://travis-ci.org/biopython/biopython/jobs/88347907

```
======================================================================
ERROR: test_SeqIO_loading (common_BioSQL.AutoSeqIOTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/common_BioSQL.py", line 889, in test_SeqIO_loading
    self.check('fasta', 'Fasta/wisteria.nu')
  File "/home/travis/build/biopython/biopython/Tests/common_BioSQL.py", line 858, in check
    compare_record(record, db_rec)
  File "/home/travis/build/biopython/biopython/Tests/seq_tests_common.py", line 275, in compare_record
    compare_sequence(old.seq, new.seq)
  File "/home/travis/build/biopython/biopython/Tests/seq_tests_common.py", line 243, in compare_sequence
    assert expected == str(new[i:j:step])
  File "/home/travis/build/biopython/biopython/BioSQL/BioSeq.py", line 102, in __getitem__
    self.start + j)
  File "/home/travis/build/biopython/biopython/BioSQL/BioSeqDatabase.py", line 499, in get_subseq_as_string
    (start + 1, length, seqid))[0])
  File "/home/travis/build/biopython/biopython/BioSQL/BioSeqDatabase.py", line 473, in execute_one
    self.execute(sql, args or ())
  File "/home/travis/build/biopython/biopython/BioSQL/BioSeqDatabase.py", line 483, in execute
    self.dbutils.execute(self.cursor, sql, args)
  File "/home/travis/build/biopython/biopython/BioSQL/DBUtils.py", line 52, in execute
    cursor.execute(sql.replace("%s", "?"), args or ())
  File "/opt/python/pypy3-2.4.0/lib_pypy/_sqlite3.py", line 915, in wrapper
    return func(self, *args, **kwargs)
  File "/opt/python/pypy3-2.4.0/lib_pypy/_sqlite3.py", line 1084, in execute
    return self.__execute(False, sql, [params])
[...]

PR #660, opened 2015-10-30
adding example for KEGG.Enzyme.read()

arosenfeld:
Small example for KEGG.Enzyme.read() added in PR https://github.com/biopython/biopython/pull/653

### DIFF
--- a/Doc/Tutorial/chapter_kegg.tex
+++ b/Doc/Tutorial/chapter_kegg.tex
@@ -19,6 +19,17 @@ Parsing a KEGG record is as simple as using any other file format parser in Biop
 '5.4.2.2'
 \end{verbatim}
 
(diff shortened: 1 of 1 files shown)


peterjc:
Don't worry about the TravisCI pypy3 failure - see #658

We've got a special trick in ``test_Tutorial.py`` for actually testing the snippets in the Tutorial as doctests, which is described in a chapter of the tutorial itself. Ideally we can do that with your example - but in any case, I'd like to merge this when I'm back in the office after the weekend. Thanks!

peterjc:
I'm sorry this got forgotten about for so long. 

Commit https://github.com/biopython/biopython/commit/2c85d112c443d693b3ecdd275e3528313436d278 turned your examples into live doctests via ``test_Tutorial.py``.

Thank you!

PR #661, opened 2015-10-31
Bio.Seq: implementing translation of gapped sequences

carlosp420:
This PR implements the translation of gapped sequences.

* A gap character ``gap=None`` argument has been added to the ``translate`` method and ``_translate`` function.
* The code will try to infer the gap character from the given alphabet. If both alphabet and gap character are missing, it will raise a TranslationError.
* If the given gap character is different from the one in the alphabet, it will raise a ValueError.
* The translation of a seq object with gaps and stop codons will return this alphabet: ``"HasStopCodon(Gapped(ExtendedIUPACProtein(), '~'), '*')"``.

I tried to cover all the cases I could think of as unittests. I also added some docstrings to cover the gapped translation.


peterjc:
Don't worry about the pypy3 failure on TravisCI, see #658

Bio/Seq.py
```diff
@@ -1000,9 +1014,25 @@ def translate(self, table="Standard", stop_symbol="*", to_stop=False,
...
+                raise ValueError("Gap {0!r} does not match {1!r} from alphabet".format(
+                    gap, self.alphabet.gap_char))
+
+        protein = _translate_str(str(self), codon_table, stop_symbol, to_stop,
+                                 cds, gap=gap)
+        if gap and \
+                gap in protein and \
+                stop_symbol in protein:
+            alphabet = Alphabet.HasStopCodon(Alphabet.Gapped(codon_table.protein_alphabet,
+                                                             gap_char=gap))
```
peterjc: You're not using the ``stop_symbol`` in ``HasStopCodon``, so that would default to ``*``.

You've also split the if statement up into more lines than needed with ``\`` characters.

I'd probably do this as two separate if statements,

```
if gap and gap in protein:
    alphabet = Alphabet.Gapped(codon_table.protein_alphabet, gap)
else:
    alphabet = codon_table.protein_alphabet

if stop_symbol in protein:
    alphabet = Alphabet.HasStopCodon(alphabet, stop_symbol)
```

Or the other way round depending which order seems best for the alphabet class wrapping horribleness?

peterjc:
We should raise this on the mailing list to check if anyone thinks it would be a problem for any code expecting a translation exception on a gap codon to suddenly gets a gapped translation back (although this would only happen if their sequence alphabet had been setup using a gap character - otherwise they need to set the gap character explicitly in the translate call).

At very least, this would be a behaviour change both stressing in the ``NEWS`` file and the release announcement.

carlosp420:
Thanks for feedback Peter. 

It seems best to return ``HasStopCodon(Gapped('-'), '*')`` for translation of
gapped seqs with stop codons as this is the result of adding two sequences with
gaps and stop codons:
```
seq1.alphabet = HasStopCodon
seq2.alphabet = Gapped

seq3 = seq1 + seq2

seq3.alphabet = HasStopCodon(Gapped(ExtendedIUPACProtein(), '-'), '*')
```

I also added test cases for translation of gapped sequences and stop codons 
and passing a ``stop_symbol``. This is now handled in the returned sequence and
returned alphabet.

peterjc:
Current behaviour with no gap specified:

```
>>> from Bio.Seq import Seq
>>> Seq("ACT---TAA").translate()
  File "Bio/Seq.py", line 1004, in translate
    stop_symbol, to_stop, cds)
  File "Bio/Seq.py", line 2064, in _translate_str
    "Codon '{0}' is invalid".format(codon))
Bio.Data.CodonTable.TranslationError: Codon '---' is invalid
```

Current behaviour with gapped alphabet:

```
>>> from Bio.Seq import Seq
>>> from Bio.Alphabet import generic_dna, Gapped
>>> Seq("ACT---TAA", Gapped(generic_dna)).translate()
  File "Bio/Seq.py", line 1004, in translate
    stop_symbol, to_stop, cds)
  File "Bio/Seq.py", line 2064, in _translate_str
    "Codon '{0}' is invalid".format(codon))
Bio.Data.CodonTable.TranslationError: Codon '---' is invalid
```

The proposed behaviour adds a gap argument to the translate method, but will look at the alphabet by default:

```
>>> from Bio.Seq import Seq
>>> Seq("ACT---TAA").translate()
  File "Bio/Seq.py", line 1027, in translate
    cds, gap=gap)
  File "Bio/Seq.py", line 2097, in _translate_str
    "Codon '{0}' is invalid".format(codon))
Bio.Data.CodonTable.TranslationError: Codon '---' is invalid
```

So far the same, but with the proposed changes you can now do this:

```
>>> Seq("ACT---TAA").translate(gap="-")
Seq('T-*', HasStopCodon(Gapped(ExtendedIUPACProtein(), '-'), '*'))
```

The following would be a change in behaviour - the gap codon is inferred from the declared alphabet:

```
>>> from Bio.Seq import Seq
>>> from Bio.Alphabet import generic_dna, Gapped
>>> Seq("ACT---TAA", Gapped(generic_dna)).translate()
Seq('T-*', HasStopCodon(Gapped(ExtendedIUPACProtein(), '-'), '*'))
```

Or,

```
>>> Seq("ACT~~~TAA", Gapped(generic_dna, "~")).translate()
Seq('T~*', HasStopCodon(Gapped(ExtendedIUPACProtein(), '~'), '*'))
```


As long as it is consistent, you can use the new argument too:

```
>>> Seq("ACT---TAA", Gapped(generic_dna)).translate(gap="-")
Seq('T-*', HasStopCodon(Gapped(ExtendedIUPACProtein(), '-'), '*'))
```

However, like the existing ``.ungap`` method, a mismatch gives an error, e.g.

```
>>> Seq("ACT---TAA", Gapped(generic_dna)).translate(gap="~")
  File "Bio/Seq.py", line 1024, in translate
    gap, self.alphabet.gap_char))
ValueError: Gap '~' does not match '-' from alphabet
>>> Seq("ACT---TAA", Gapped(generic_dna)).ungap(gap="~")
  File "Bio/Seq.py", line 1112, in ungap
    gap, self.alphabet.gap_char))
ValueError: Gap '~' does not match '-' from alphabet
>>> Seq("ACT---TAA", Gapped(generic_dna)).ungap()
Seq('ACTTAA', DNAAlphabet())
```

peterjc:
Mailing list thread:
- http://lists.open-bio.org/pipermail/biopython-dev/2015-November/021138.html
- http://lists.open-bio.org/pipermail/biopython-dev/2015-November/021139.html
- ...

carlosp420:
yep, that's better. Fixed.

peterjc:
Committed to master as squashed commit https://github.com/biopython/biopython/commit/72c16670a15911c28f7555ba32f1eade56362587 - thank you Carlos, apologies for the delay :)

bgruening:
Great work @carlosp420 and @peterjc!

carlosp420:
:wink:

PR #665, opened 2015-11-05
when parsing paml/codeml config files, split options at first '=' 

cswarth:
`Bio.Phylo.PAML.codeml.read_ctl_file` cannot parse cofig options that have embeded '='

Parsing a codeml option file like this fails,

    seqfile = /purifying/noindel/pop=1000/sample_aln_fixed.fa

because there are '=' characters in the value.
Fix is to only consider the first '=' as separating name from value.

### DIFF
--- a/Bio/Phylo/PAML/codeml.py
+++ b/Bio/Phylo/PAML/codeml.py
@@ -108,7 +108,7 @@ class Codeml(Paml):
                         if "=" not in uncommented:
                             raise AttributeError(
                                 "Malformed line in control file:\n%r" % line)
-                        (option, value) = uncommented.split("=")
+                        (option, value) = uncommented.split("=",1)
                         option = option.strip()
                         value = value.strip()
                         if option == "seqfile":


cswarth:
Huh?  Why is CI failing somewhere in `test_SeqIO_loading (common_BioSQL.AutoSeqIOTests)` when this cannot possibly be affected by the single line change made to `Bio/Phylo/PAML/codeml.py`?

Bio/Phylo/PAML/codeml.py
```diff
@@ -108,7 +108,7 @@ def read_ctl_file(self, ctl_file):
                         if "=" not in uncommented:
                             raise AttributeError(
                                 "Malformed line in control file:\n%r" % line)
-                        (option, value) = uncommented.split("=")
+                        (option, value) = uncommented.split("=",1)
```
peterjc: Minor: There should be a space after the comma following PEP8 style guidelines.

peterjc:
This looks sensible to me, any comments @brandoninvergo ?

It would be nice to add a unit test for this, but that's not essential.

brandoninvergo:
Yes, that looks fine to me too.

peterjc:
@cswarth Are you happy to be thanked by name in the ``NEWS`` and ``CONTRIB`` files? I'm asking because you've not opted to make your name public on your GitHub profile.

cswarth:
@peterjc sure, thanks for the acknowledgement. That was an probably the result of default paranoia on my part - I'll expose the name now.

peterjc:
Great. If you were not aware, your name and email is present in the git commit log anyway - it just isn't quite so obvious as being on the GitHub profile.

PR #668, opened 2015-11-13
various Nexus parsing fixes

cswarth:
- remove '/' from the list of prohibited characters in identifiers when reading nexus files.
  I was unable to access the original nexus spec, but this change was made under the robustness
  principle - be liberal in what you accept, conservative in what you produce.
- allow ',' in nexus comments on internal nodes.  Perviously ',' was
  allowed in the comments on the whole tree (most encapsulating node)
  but through a quirk in the parser ',' could not appear in the
  comments of internal nodes.
- add Nexus test cases to check if comment parsing is working as expected.

I ran into problems parsing Nexus format trees produced from sequences whose name contained a date, e.g. 'patient1_1000_116|1M|XXX|XXX|2011/11/10'

Nexus files produced by Beast and treeannotator often have ',' in the comments on each node.  The Nexus parser was not handling that case.   
I added a couple of test cases to demonstrate the enhanced comment parsing.

### DIFF
--- a/Bio/Nexus/Nexus.py
+++ b/Bio/Nexus/Nexus.py
@@ -36,7 +36,7 @@ SPECIAL_COMMANDS = ['charstatelabels', 'charlabels', 'taxlabels', 'taxset',
                     'charset', 'charpartition', 'taxpartition', 'matrix',
                     'tree', 'utree', 'translate', 'codonposset', 'title']
 KNOWN_NEXUS_BLOCKS = ['trees', 'data', 'characters', 'taxa', 'sets', 'codons']
-PUNCTUATION = '()[]{}/\,;:=*\'"`+-<>'
+PUNCTUATION = '()[]{}\,;:=*\'"`+-<>'
 MRBAYESSAFE = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_'
 WHITESPACE = ' \t\n'
 # SPECIALCOMMENTS = ['!','&','%','/','\\','@'] # original list of special comments
--- a/Bio/Nexus/Trees.py
+++ b/Bio/Nexus/Trees.py
@@ -100,14 +100,20 @@ class Tree(Nodes.Chain):
             subtrees = []
             plevel = 0
             prev = 1
+            incomment = False
             for p in range(1, closing):
-                if tree[p] == '(':
+                if not incomment and tree[p] == '(':
                     plevel += 1
-                elif tree[p] == ')':
+                elif not incomment and tree[p] == ')':
                     plevel -= 1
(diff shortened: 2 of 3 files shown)


peterjc:
The NEXUS paper Meddison et al 1997 PDF appears to be free to download now (or maybe our institute has a subscription now?): http://sysbio.oxfordjournals.org/content/46/4/590 - this is essentially scans of the original, with OCR added to the PDF, so we can't 100% reliably search the text.

Page 596,

    Tokens, Words, Punctuation, and Whitespace
    Words, punctuation, and whitespace together form a NEXUS file. Tokens are
    the words and punctuation of a NEXUS command. Words are separated by
    either whitespace (blank spaces, tabs, etc.) or punctuation. The Appendix
    contains formal definitions of these terms

Page 620,

    Punctuation.—Any of the following text characters are considered punctuation at
    some times: ( ) [ ] { } / \ , ; : = * ' " " + - < > The following punctuation marks have
    special properties: [ ] do not break a word; + and - are allowed as state symbols,
    but none of the rest are allowed; - is considered punctuation except were it is the
    minus sign in a negative number.

The NEXUS format is explicitly defined in terms of tokens delimited by this punctuation character list - you cannot safely remove the ``/`` from the punctuation list.

In principle, given the Nexus format is well defined, we should NOT quietly accept invalid files. Where it is useful to do so, and unambiguous to handle, then parsing with a warning would be a reasonable compromise (see ``BiopythonParserWarning`` class).

Can you share some of the problem input files (e.g. via gist.github.com) which should help determine if there is a problem in the parser (e.g. how it handles quoted identifiers), or if the other tools are producing invalid NEXUS output (quite possible given this looks like a corner case). Thanks.

cswarth:
Thanks for the thoughtful feedback, Peter.  I'll send updates as soon as I can.

peterjc:
@etal Do you know enough about Nexus to review this? Does anyone else from the current developers team?

etal:
This looks good to me, and I like the unit tests. Since Beast is widely used, I agree Biopython ought to accept its output. In general I'd encourage the use of NeXML or phyloXML over Nexus to provide more rigor to the handling of phylogenetic data where that's a priority.

I suspect the original 1997 spec did not anticipate much syntax within the comments. It's reasonable to exempt comments from restrictions that apply to the rest of the format, which this pull request nicely implements, and so it's not really necessary to warn about punctuation characters within comments unless there's a risk of ambiguity in the interpretation.

peterjc:
Cherry-picked as 9377867393b5c1e24430cb8db0ea22a9da8420f0 with NEWS entry and acknowledgement added in a84392184a19c220253c7c133b6abc4abb9dbb77

Thank you and apologies for letting this pull request wait so long.

PR #671, opened 2015-11-16
update test cases for PAML v4.8, and don't crash on long taxa names

cswarth:
- Update yn00 tests with an output file from version 4.8.

- Update regexp in `_parse_yn00.py` to enable long taxa names.  Long taxa names (> 30 characters) can run over data results leading to parsing errors.  Current parser can still be fooled, but by a smaller, more pathological set of taxa names.  Length alone is no longer enough to break it.

- Add tests for long names.  This involved some minor modifications to the existing test harness to support testing with multiple alignment files. No longer write name of alignment file into control file from python; it should already be there.  When scanning for results to parse, match result files by pattern rather than assuming everything in the Results directory is from the same control file.


peterjc:
@brandoninvergo can you review at this pull request please?

brandoninvergo:
Wow, I'm extremely sorry that I missed this. Github notifications are sometimes marked as spam for me, so I didn't see it...and I guess I haven't logged in in months, either.

Anyway, everything in the code looks fine to me.  Thanks, Chris, for fixing this!

peterjc:
Applied as squashed commit https://github.com/biopython/biopython/commit/70c0de5fdfd75aa10b7f0d8f85ce7103579701a4 - acknowledgement in https://github.com/biopython/biopython/commit/10037bbd478cde4368f7da1f6cd08dc9cb558fcc - thank you and apologies we let this contribution sit so long.

PR #673, opened 2015-11-21
AlignIO: improve StockholmIO performance by eliminating array search

wwood:
Hi,

I noticed that reading stockholm format files becomes quite slow when there are large numbers of sequences involved. The slowness gets worse quickly, super-linearly, so cutting it down to something more manageable,

before
```
$ time hmmalign DNGNGWU00016_mingle_output_good_seqs.hmm <(head -n50000 a.fa) | seqmagick convert --input-format stockholm - out2.fa 

real	0m9.803s
user	0m9.972s
sys	0m0.140s
```
after (output was the same)
```
$ time hmmalign DNGNGWU00016_mingle_output_good_seqs.hmm <(head -n50000 a.fa) | seqmagick convert --input-format stockholm - out2.fa 

real	0m2.423s
user	0m2.504s
sys	0m0.132s
```
The specific issue was a repeated search of the `ids` array, so I replaced this with a hash. Only really want an OrderedSet not an OrderedDict, but figured OrderedDict does the job and all the mess incurred is contained the `__next__` function.

Thanks
ben

### DIFF
--- a/Bio/AlignIO/StockholmIO.py
+++ b/Bio/AlignIO/StockholmIO.py
@@ -136,7 +136,7 @@ from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
 from Bio.Align import MultipleSeqAlignment
 from .Interfaces import AlignmentIterator, SequentialAlignmentWriter
-
+from collections import OrderedDict
 
 class StockholmWriter(SequentialAlignmentWriter):
     """Stockholm/PFAM alignment writer."""
@@ -337,7 +337,7 @@ class StockholmIterator(AlignmentIterator):
         # if present it agrees with our parsing.
 
         seqs = {}
-        ids = []
+        ids = OrderedDict() # Really only need an OrderedSet, but python lacks this
         gs = {}
         gr = {}
         gf = {}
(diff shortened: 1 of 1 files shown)


wwood:
Gah, so according to travis it fails on python2.6 due to `OrderedDict` not being a thing there. Does the reader have any suggestions on the best way to implement this? Would a new class in `StockholmIO` that implements an ordered set without relying on `OrderedDict` be a good plan? I'm not very familiar with the biopython code base so might be missing something obvious.

peterjc:
Use this instead,

```
from Bio._py3k import OrderedDict
```

This is our internal equivalent of the ``six`` library, in this case it will use the standard library ``OrderedDict`` if present (i.e. Python 2.7+), then Raymond Hettinger's backport available on PyPI (in case the Python 2.6 user has that and wants to do class comparisons or something) or finally our bundled back port (one file under ``Bio._py3k``).

Once we drop Python 2.6 support (which is deprecated as of Biopython 1.66 which means its days are numbered but we're supporting it a little while longer), that problem goes away :)

peterjc:
Otherwise this looks good - I don't think I ever tested this on large enough Stockholm files to notice the slowdown.

Would you be happy to be thanked by name in the ``CONTRIB`` and ``NEWS`` file?

wwood:
Thanks for the feedback - exactly the kind I was after.

>Would you be happy to be thanked by name in the CONTRIB and NEWS file?

More than happy. I also added a comment at the top of `StockholmIO.py`, though I'm not sure what convention is. Please let me know if I've messed up.

peterjc:
Good point, there's no copyright holder named on this file - an oversight on my part. See https://github.com/biopython/biopython/blob/master/Bio/AlignIO/PhylipIO.py for a good example; I'll have to go over the history for this file and see if anyone else should be named...

Issue #600, opened 2015-08-28
SeqFeature.Reference comparison


kblin:
Similar to the SeqRecord comparison mentioned in issue #559, comparing references is currently using Python object comparison, which makes the following code fail:

```
from Bio import SeqIO

rec1 = SeqIO.read('foo.gbk', 'genbank')
rec2 = SeqIO.read('foo.gbk', 'genbank')

assert rec1.annotations['references'] == rec2.annotations['references']
```

I'd argue that comparing references is a bit easier than comparing whole SeqRecord objects, so I'd like to make this work.

@peterjc mentioned on twitter that we'd have to decide on how we'd handle cases where the DOI was identical, but e.g. capitalisation of the title differed. Unless other people feel strongly about this, I'd go with a strict comparison for now.

Thoughts?

peterjc:
My example was intended as a corner case, but in practice I think strict comparison of all the reference attributes should work fine.

vincentdavis:
What is the status of this and #559? I see the pull request to Raise exceptions which makes sense to me.
Adding the ability for limited comparison looks interesting to me.

peterjc:
Comparing the references is just one part of comparing SeqRecord objects. The later are so complex that only the default ``id(...)`` comparison makes sense, but it is not intuitive thus the suggestion of a warning or exception in #559.

Comparing references is much simpler. Equality based on testing all the attributes would probably make sense (despite corner cases like capitalisation of the title, or partial information, etc). That seems safe, PR welcome.

peterjc:
I've flagged this for wider comment on the biopython-dev mailing lists, general on the mailing list, http://lists.open-bio.org/pipermail/biopython-dev/2015-November/021169.html

kblin:
A thing to note with the current proposed implementation is that the current comparison ignores the location of the Reference object. That's because equality isn't defined on SeqFeature.Location objects either. And comparing two locations is probably another can of worms to open, and maybe something for another discussion.
My reasoning for ignoring the location was that if the references refer to the same publication, then they should compare to true anyway.

peterjc:
Oh. I'd missed that. At very least the docstring should talk about that behaviour.

We can probably define equality on location objects, but as you say, there are pitfalls here. Perhaps a re-factoring of the reference object would be a better long term idea... even maybe recording the references in the list of ``SeqFeature`` objects?

PR #683, opened 2015-11-25
SeqFeature.Reference: Implement __eq__

kblin:
Allow two reference objects to be checked for equality. This should fix issue #600.
Note that this currently does not check for location identity, as FeatureLocations
also don't implement __eq__ at the moment.

### DIFF
--- a/Tests/test_SeqFeature.py
+++ b/Tests/test_SeqFeature.py
@@ -21,7 +21,7 @@ class TestReference(unittest.TestCase):
 
         self.assertEqual(rec1.annotations['references'][0], rec1.annotations['references'][0])
         cmp1, cmp2 = rec1.annotations['references'][0], rec2.annotations['references'][0]
-        self.assertEqual(cmp1, cmp2, "{} vs {}".format(cmp1, cmp2))
+        self.assertEqual(cmp1, cmp2)
         self.assertNotEqual(rec1.annotations['references'][0], rec1.annotations['references'][1])
         self.assertNotEqual(rec1.annotations['references'][0], rec2.annotations['references'][1])
         self.assertEqual(rec1.annotations['references'][1], rec1.annotations['references'][1])
--- a/Tests/test_SeqFeature.py
+++ b/Tests/test_SeqFeature.py
@@ -3,7 +3,7 @@
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
 
-"""Tests for Reference objects from Bio.SeqFeature.
+"""Tests Bio.SeqFeature.
 """
 import unittest
 from os import path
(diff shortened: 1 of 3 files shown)


peterjc:
Simple and direct - it would need a couple of unit tests though, I suggest adding to ``Tests/test_SeqIO_features.py`` or perhaps starting a new unittest based ``test_SeqFeature.py`` file?

Note I've asked for comments on this and #600 in general on the mailing list, http://lists.open-bio.org/pipermail/biopython-dev/2015-November/021169.html

kblin:
I'm a bit confused. I've added `Tests/test_Reference.py` while writing that code. Do you want those tests to be moved to one of the files you suggested, or do you want additional tests?

peterjc:
Sorry, I missed ``Tests/test_Reference.py``. That would be fine.

At some point I might be tempted to rename it ``test_SeqFeature.py`` and move/add tests for other bits of ``Bio.SeqFeature`` to it.

kblin:
Do you want me to rename it right away?

peterjc:
Sure, then it would be an obvious place to add any new tests specific to ``Bio.SeqFeature`` :)

kblin:
Ok, I just added patches for both the rename and the note on not comparing the FeatureLocations that we discussed in #600.

kblin:
Hmm, the QuantifiedCode thing won't show me which line it doesn't like. Annoying.

Tests/test_SeqFeature.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_eq_identical(self):
+        """Test two identical references eq() to True"""
+        testfile = path.join('GenBank', 'origin_line.gb')
+        rec1 = SeqIO.read(testfile, 'genbank')
+        rec2 = SeqIO.read(testfile, 'genbank')
+
+        self.assertEqual(rec1.annotations['references'][0], rec1.annotations['references'][0])
+        cmp1, cmp2 = rec1.annotations['references'][0], rec2.annotations['references'][0]
+        self.assertEqual(cmp1, cmp2, "%s vs %s" % (cmp1, cmp2))
```
peterjc: Just use ``self.assertEqual(cmp1, cmp2)`` which has a default message which will nicely display the non-equal elements. It handles lists etc nicely too which is a bonus.

This will also stop QualifiedCode nagging about old-style string formatting ;)

kblin:
There we go, fixed that .format, and one more thing my editor's code checker complained about

kblin:
Or not. I don't like the plain assert as it creates pretty unreadable output in python older than 2.7, but I guess we're not worrying about that too much anymore. So I followed @peterjc 's suggestion and just removed the custom printout.

kblin:
Anyway, any thoughts on merging this pull request? Just trying to tie up loose ends before there end of the year.

peterjc:
Sadly no comments on the mailing list about this - I guess that means no objections, so I should go ahead and merge it. Do you mind if squash this into less commits?

kblin:
No, I don't mind at all :)

peterjc:
Applied as squashed commit https://github.com/biopython/biopython/commit/e7d20896b55912f8d3cdf16d04081c985e5d8388 thanks!

Issue #384, opened 2014-10-21
Create a mock-isolated, high-speed test suite


zancas:
As @peterc points out here:  https://github.com/biopython/biopython/issues/374#issuecomment-59647618 the current regression test suite usefully checks for system (and intersystem) functionality (e.g. a running http://togows.org/ instance), however as @peterc also points out in the same comment, there are advantages to having a fully isolated test suite.

These include (at least):
1. no false positives, only the core code within biopython gets exercised
2. no false negatives, no interraction with an external component can "cover" a bug in the core logic
3. speed, by isolating the tests from other components (especially network components) the suite becomes faster
4. the tests can be run more environments, e.g. I can set up a docker instance with biopython and little else installed and run the tests with no network, and perhaps on a resource limited, or sandboxed system

So there are competing features in the different test suites, integration gives a full assurance of production-ready funcationality, unit gives precise information about the contrained subset of functionality in the biopython core code.

The solution is to have both types of tests deployed in the correct contexts.

Setting up the exact right configuration for that is beyond the scope of the next step we propose, which is to write a set of mock objects with the overarching goal of isolating the core code from external systems (and indeed isolating bits of the core code from other bits!).

Feedback?

carlosp420:
I have been exploring the use of the Mock framework to test code that normally requires network calls. Mock has been included in Python 3.3+ as part of unittest. For earlier Python versions we would need to install mock as dependency or we can just run the tests only in Python3.

I would like to try using this for some python tests.

peterjc:
New tests using mock or its back port under Python 2.6 & 2.7 would be acceptable I think (i.e. whole new ``test_XXX.py`` files). We should be able to install the backport under TravisCI and the buildbot slaves.

zancas:
Folks, I've been experimenting with pytest:  https://pytest.org/latest/index.html

I recommend it.

PR #684, opened 2015-11-27
Bio.Entrez: avoid network calls in unit tests.

carlosp420:
As discussed in #384, I have used the library ``mock`` to implement fake network calls for the tests in ``test_Entrez_online.py``.

Some tests were actually testing for two things: (i) constructing the URL to request from, and (ii) reading and parsing the data returned from the network call.
So I have split the tests, one for the URL and another one to test reading and parsing the data.

Doing this was tricky as the function ``_open`` in Bio.Entrez.\_\_init\_\_ also does two things, i.e. construction of the URL and network call via ``_urlopen`` (which belongs to the Python standard library). So I refactored ``_open`` to isolate the URL construction and network calling. In this way it was easier to write tests for both functions.

Thus, ``_open`` is left to take care of the network call (via ``_urlopen``). As we don't really need to test for ``_urlopen``, I used the library ``mock`` to pass binary data (read from disk) to the functions that we want to test, such as ``efetch``, ``einfo``, etc.

I wrote a short blog post explaining some rationale behind the use of ``mock``: <http://mycalesis.me/mocking-network-calls-unittests/>


Tests/test_Entrez_online.py
```diff
@@ -41,13 +50,20 @@
...
+        params = Entrez._construct_params(params=None)
+        options = Entrez._encode_options(ecitmatch=False, params=params)
+        result_url = Entrez._construct_cgi(cgi, post=False, options=options)
+        self.assertTrue(result_url.startswith(URL_HEAD + "einfo.fcgi?"), result_url)
+        self.assertTrue(URL_TOOL in result_url)
+        self.assertTrue(URL_EMAIL in result_url)
 
-    def test_read_from_url(self):
+    @patch("Bio.Entrez._open", return_value=_binary_to_string_handle(open("Entrez/einfo1.xml", "rb")))
+    def test_read_from_url(self, mock_open):
```
peterjc: I can only see you importing ``patch`` from the ``mock`` library - where/how is ``mock_open`` defined?

peterjc:
That blog post makes this one of the best prepared pull requests I can remember looking at for Biopython! :)

I have a high level comment: ``test_Entrez_online.py`` has at least two jobs, one is testing our code (in particular URL construction, I would say here parsing expected responses is secondary since it is primarily tested off-line via ``test_Entrez.py``), but the second is to actively check the NCBI Entrez API is still doing what we expect it too. It really is useful to keep the online test for occasional use, as this can alert us to API changes etc (at the cost of regular test "failures" due to temporary network issues or service outages).

So, hypothetically, could we split this into ``test_Entrez_online.py`` and  ``test_Entrez_mock.py`` which run the same set of tests (via a shared ``Tests/common_Entrez.py`` or something), but in one case truly online and in the other via mock?

That aside, thanks to your re-factoring of the private methods in ``Bio.Entrez``, even without using mock, many of the (new/re-factored) tests like ``test_construct_cgi_einfo``, ``test_construct_cgi_efetch``, ... for URL construction could simply be run offline and moved from ``test_Entrez_online.py`` into ``test_Entrez.py`` (which is currently just parser tests; maybe we should rename that to ``test_Entrez_parser.py`` instead?). This seems less controversial and ought to give immediate coverage improvement.

I don't have any experience using ``mock``, so the other changes are harder for me to judge. What about you @mdehoon?

carlosp420:
Thanks for feedback Peter. I also noticed that "unittest.mock" is not included in PYPY :-(

peterjc:
TravisCI worked on Python 2.6 so the mock backport can work.

TravisCI failed on Python 2.7 and PyPy, likely the version of mock already installed was too old? If so ask pip to update it?

```
$ if [[ $PY3 == 'no' ]]; then pip install mock; fi
You are using pip version 6.0.7, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Requirement already satisfied (use --upgrade to upgrade): mock in /home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages
```

For PyPy3, you might also have to explicitly install mock via PyPI https://pypi.python.org/pypi/mock - I'm guessing their standard library isn't quite up to date here?

Tests/test_Entrez_online.py
carlosp420: I think it ``mock_open`` is defined and used at runtime. Actually instead of ``mock_open`` it could be any name. The test will run if the argument is present regardless of name, otherwise it will fail. Mock is weird.

carlosp420:
I had imported the mock library in the test only when running Python2.6. It now loads mock also in Python2.7 and PyPy.

carlosp420:
I renamed the file ``test_Entrez.py`` to ``test_Entrez_parser.py`` and moved in all the tests that check for the URL construction. 
In order to increase tests coverage I refactored the code to check whether Entrez should do GET or POST request. After this, it was easy to write a test for it (currently in ``test_Entrez_parser.py``).

So the only tests left in ``test_Entrez_online.py`` are those that check that NCBI is returning the data that Biopython expect and have not changed their response. Thus, mocking the network calls is not useful. I think I should remove all the mocking in ``test_Entrez_online.py``.

Test coverage has gone up from 59% to 65% for Bio.Entrez.

peterjc:
I was thinking we'd rename ``test_Entrez.py`` to ``test_Entrez_parser.py`` for the parser test, and add a new file, maybe ``test_Entrez_offline.py`` (or just ``test_Entrez.py``?) to hold your new tests for the URL construction.

carlosp420:
yes, I think ``test_Entrez_parser.py`` for the parser and ``test_Entrez.py`` for the url construction should be ok.

QuantifiedCode is complaining that this pull request adds 45 new issues due to the URL tests that need to access the private functions (such as ``Entrez._constuct_params``) that I created during the refactoring:

```python
    def test_construct_cgi_einfo(self):
        """Test constructed url for request to Entrez."""
        cgi = 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/einfo.fcgi'
        params = Entrez._construct_params(params=None)
        options = Entrez._encode_options(ecitmatch=False, params=params)
        result_url = Entrez._construct_cgi(cgi, post=False, options=options)
        self.assertTrue(result_url.startswith(URL_HEAD + "einfo.fcgi?"), result_url)
        self.assertTrue(URL_TOOL in result_url)
        self.assertTrue(URL_EMAIL in result_url)
```

peterjc:
I would ignore the Qualified Code report here - we're using a private function API in order to test internal details not exposed as a public API.

carlosp420:
@peterjc I think this is ready.

peterjc:
Sadly rebasing this to the current master branch was quite tricky. Instead, I've just cherry-picked the easy and uncontentious bits and renamed ``test_Entrez.py`` to ``test_Entrez_parser.py`` on the master, see 697416639a358ab61168c03b5beec2bfa6db20e9 5ea0f3f742f075233110334536835e41b7395d5a 8a675676f957e3726cbbdc6336f098b88f17bdc5 and 9323d0934bbda6a5797162e8436b67dd1a0e5e13

My intension is this will simplify applying your remaining changes as a manually squashed single commit - which I will look at next - and perhaps prepare this as a new pull request for comment.

peterjc:
I'm closing this pull request in favour of #859 which was based on this work.

PR #685, opened 2015-11-28
PDBParser: uppercase element names

wojdyr:
Avoid problems with lower/mixed case element names, as discussed on ML:
http://mailman.open-bio.org/pipermail/biopython/2015-May/015619.html

This could be fixed in other ways, or fixed only if `self.PERMISSIVE`.
But if the pdb format definition doesn't say anything about lower/upper case, it shouldn't harm to accept lowercase even in strict mode.

### DIFF
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -201,7 +201,7 @@ class PDBParser(object):
                                                global_line_counter)
                     bfactor = 0.0  # The PDB use a default of zero if the data is missing
                 segid = line[72:76]
-                element = line[76:78].strip()
+                element = line[76:78].strip().upper()
                 if current_segid != segid:
                     current_segid = segid
                     structure_builder.init_seg(current_segid)


JoaoRodrigues:
A safer alternative would actually be to uppercase the entire line as it's
read. The PDB format is generally upper case. But this will do as well.
Thanks for bringing this up again.

2015-11-28 14:26 GMT+00:00 Marcin Wojdyr <notifications@github.com>:

peterjc:
It may be a quick fix, but I was worried that making the element names upper case is the wrong thing to do. Calcium is Ca, not CA. Recall the extra element column is partly to avoid confusion with CA meaning alpha Carbon.

The latest version of the format from http://www.wwpdb.org/documentation/file-format is 4.0 but its not very human friendly.

The previous release 3.3 is easier to browse, but http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM sadly does not talk about the case when defining the ``ATOM`` line, and all the examples are single letter elements so don't help.

However, the ``HETATM`` examples from the 3.3 version of the standard do use all upper case elements http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#HETATM in the examples:

```
HETATM 8237 MG    MG A1001      13.872  -2.555 -29.045  1.00 27.36          MG 
 
HETATM 3835 FE   HEM A   1      17.140   3.115  15.066  1.00 14.14          FE
```

JoaoRodrigues:
Well, that last element would be C for an alpha carbon. Also, although not
in the documentation, a CA and Ca (ion) differ in the atom name field. The
ion (as all heteroatoms) has the first letter of the atom name on the first
character of the name field, while the alpha carbon (as all protein and I
think nucleic acid atoms) start on the 2nd. Fancy undocumented features...

Also, the format is usually (i.e. judging from RCSB PDB structures) all
uppercase, no doubt because of historical reasons. I only saw lowercase
letters on modelling programs that anyway disrespect the format
altogether...

2015-12-02 10:15 GMT+00:00 Peter Cock <notifications@github.com>:

peterjc:
We're digressing, but often tools don't properly do the column alignment within the name field (I can't remember now where I ran into that, or if the parser copes).

@JoaoRodrigues is this OK to merge? Do you think it needs a unit test? Should we mention this in the ``NEWS`` or is the change too minor?

@wojdyr would you be happy to be named in the ``NEWS`` and ``CONTRIB`` file as contributing?

JoaoRodrigues:
I think it's fine. As a unit test, we can simply add one element as lower
case in one of the test PDB files and see if it corrects upon parsing.

Tool should align names properly, never saw one doing that wrong. If it
does, it's a really bad tool :) The parser will not cope actually, it will
misread the name and it will attribute the element wrong..

wojdyr:
@peterjc while I'm happy to be named in any file, this change was copied from an old mailing list thread, so no need to add me as a contributor. This PR was actually only a reminder about the known issue.

peterjc:
Applied as cherry-picked commit a00b74d3f4a54d943d07ca3349d7a8638907a624, closing pull request. Thanks for your input here @wojdyr & @JoaoRodrigues.

@JoaoRodrigues can you add a test for this later?

JoaoRodrigues:
Will do.​

PR #687, opened 2015-12-01
Fixes silent bug with truncated '\n'-terminated ATOM/HETATM lines and PDBIO()

JoaoRodrigues:
Some PDB files have truncated lines (less than 80 characters), which is not according to the format specifications. Nevertheless, this is quite prevalent, often because text editors or some programs trim empty spaces at the end of a line. If a newline character is added in such a truncated line, PDBParser will never detected a problem and will actually parse the structure appropriately. Only when writing back with PDBIO will this problem be detected and cause broken lines in the resulting file. 

The fix is very simple and was tested on a few cases. This is not a case where a unit test is necessary in my opinion. I added a 'rU' mode to the as_handle function, to convert any newline characters from Windows to '\n' and then added a call to the .rstrip() function when first reading the line in the coordinate parsing loop to chomp only a terminal '\n' character and nothing else.


peterjc:
Could this break column based parsing when the final field is empty?

Should we issue a parser warning for these truncated lines?

Why not add a (small) test case with a truncated line? Or if there is an already edited PDB file in the tests, edit it further to add a white space truncation?

JoaoRodrigues:
@peterjc It won't break the parser, that's why I took care of stripping only that newline character (if it exists). Truncated lines are a bit too common and usually not a problem for modelling or visualization software. Thus, I'm not sure it's worth to issue warnings (and suffer the accompanying performance hit, which is substantial).

I'll add a (very small) test case then.

JoaoRodrigues:
Additional info for this problem. It happens only if the segid field (which is optional) is shorter than 4  and the line is truncated. I added a small test.

Issue #516, opened 2015-04-15
Per-column annotations for multiple-sequence alignments


peterjc:
This is needed for #357 where we currently discard per-column annotation when parsing Stockholm files.

Like the existing `SeqRecord` per-letter annotations dictionary this needs to be sliced in `__getitem__` to pull out the right columns, and similarly be supported when adding alignments.

This replaces https://redmine.open-bio.org/issues/3388 on the old issue tracker (currently offline).

peterjc:
Work in progress here: https://github.com/peterjc/biopython/tree/per-column-annotation

peterjc:
I've updated  https://github.com/peterjc/biopython/tree/per-column-annotation and the associated pull request #528

PR #692, opened 2015-12-03
Added score/keys parsing to EmbossIO

tcyb:
Needed EmbossIO to retain alignment scores from `needle` for a project of mine, was surprised that this wasn't already supported. Found this discussion thread (http://article.gmane.org/gmane.comp.python.bio.general/7979/match=emboss+alignio+score) and figured that this might be a useful addition.

I'm new, so I'm not sure about the grand architecture of this kind of thing. Maybe its proper home is in SearchIO? But it seems like it would be useful in AlignIO as well.

Actual change just reads the values for the tags 'identity', 'similarity', 'gaps', and 'score' into a dictionary, and sets the 'annotations' field of the returned MultipleSeqAlignment object to be that dictionary.

### DIFF
--- a/Bio/AlignIO/EmbossIO.py
+++ b/Bio/AlignIO/EmbossIO.py
@@ -95,6 +95,7 @@ class EmbossIterator(AlignmentIterator):
         number_of_seqs = None
         ids = []
         seqs = []
+        header_dict = {}
 
         while line[0] == "#":
             # Read in the rest of this alignment header,
@@ -116,7 +117,6 @@ class EmbossIterator(AlignmentIterator):
                 length_of_seqs = int(parts[1].strip())
                 
             # Parse the rest of the header
-            header_dict = {}
             if key == 'identity':
                 header_dict['identity'] = int(parts[1].strip().split('/')[0])
             if key == 'similarity':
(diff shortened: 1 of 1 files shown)


peterjc:
Yes, the pair-wise EMBOSS output at least could be handled nicely under ``Bio.SearchIO``. I think @bow and I talked about this before.

In the meantime, now that the ``MultipleSeqAlignment`` class finally got an annotations dictionary (back in Biopython 1.61), that is a good place to put this information within the ``Bio.AlignIO`` framework. See also #516 for per-column annotation.

tcyb:
Sorry, don't think I checked this code as well as I should have. I'll close this and resubmit when it's ready.

peterjc:
OK - but the general idea of using the annotation dictionary seems very sensible +1

tcyb:
Had the dict initialization inside the read loop, dumb bug. This one should work.

peterjc:
Sorry, this issue and your proposed fix fell off my radar.

The change looks sensible - with a minimal test it would be good to go. Any comments @bow ?

bow:
@peterjc and @tcyb,

This looks ok. SearchIO uses something similar on its `HSPFragment` object to annotate alignments as well. But for now, being a tool that outputs (mostly? / all?) single-alignment results, I think EmbossIO fits well inside AlignIO.

peterjc:
Looks like I missed @bow's reply, sorry. We should squash-and-merge this.

@tcyb: Are you happy to dual licence your contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

tcyb:
Hi all, sorry, missed all this as well. @peterjc, dual license and thanks both sound great.

PR #693, opened 2015-12-03
open files using "with" in Tests/test_Entrez.py

vincentdavis:
Issues reported by https://www.quantifiedcode.com
I left the first case alone in GeneralTests line 30, not sure what that is doing.

### DIFF
--- a/Tests/test_Entrez.py
+++ b/Tests/test_Entrez.py
@@ -27,7 +27,7 @@ class GeneralTests(unittest.TestCase):
     def test_closed_handle(self):
         '''Test parsing closed handle fails gracefully
         '''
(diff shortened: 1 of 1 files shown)


peterjc:
I think originally this was not using the with statement due to the network handles not being context manager aware in all supported versions of Python... I'm waiting to see what TravisCI says.

peterjc:
Regarding the test at line ~30, ``test_closed_handle`` gives the parser a closed handle and checks it raises an ``IOError`` exception. That looks fine as it is, the with statement can't really be used here.

peterjc:
The with commits are merged directly. I ended up rewriting the test name fixes though.

Thank you!

PR #702, opened 2015-12-04
Support PhyML testing on Windows

benfulton:


peterjc:
Thank you - I'd forgotten the ternary conditional operator was new way back in Python 2.5, I really should use it more.

peterjc:
As with #610 before, this needed a small tweak to work on my old Windows XP machine: https://github.com/biopython/biopython/commit/3873d2c818c66b4be37ec68d09a8557c3635eddb

(Spotted via the buildbot which currently has Windows XP as the only Windows platform being tested)

PR #706, opened 2015-12-04
Remove unnecessary "global" keyword usage.

cbrueffer:
It's only needed when writing to a global variable.

Identified via landscape.io


peterjc:
Can you tell I learnt C before Python? I was using global here even for reading a variable - stylistically I find this far more explicit but it seems the Python conventions are against me, www.quantifiedcode.com grumbled about this too.

cbrueffer:
Indeed; in principle I like it too since it makes clear where the variable comes from. Oh well, the things we do to get a higher score... ;-)

peterjc:
I'm going to close this for now. There have been slight changes to ``Tests/common_BioSQL.py`` which still has lots of global statements so I'm not sure what it should look like.

PR #707, opened 2015-12-04
Fix and enable doctests in Phylo

vincentdavis:


peterjc:
Remaining changes for the doctests applied as https://github.com/biopython/biopython/commit/4444c88dee251d385fea9ec6fbe4aa933f5a002e - thank you.

PR #708, opened 2015-12-04
Phylo doc tests

vincentdavis:
All doctest in Phylo now work. 
This pull request is for TreeConstruction.py. I did not enabled these because of the warnings below.
Warnings
* I assume the print(tree) doctests will not be stable
* This includes a hack on matrix __str__ using spaces rather than tabs in commit e597af6

### DIFF
--- a/Bio/Phylo/BaseTree.py
+++ b/Bio/Phylo/BaseTree.py
@@ -334,6 +334,7 @@ class TreeMixin(object):
 
         Example
         -------
+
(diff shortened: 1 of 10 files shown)


vincentdavis:
That last commit 70b5bcf should fix everything. We will see what Travis CI says.
Please give feedback on fix for matrix __str__ in ad54896 I have some ideas to make it a little better.

PR #712, opened 2015-12-04
More basic "with open" fixes

vincentdavis:


peterjc:
I merged the first few commits directly (``git merge --ff-only ...``) pending word back from Tiago about the PopGen code.

peterjc:
Got OK from Tiago, all merged - thanks again Vincent.

Issue #633, opened 2015-10-09
biopython/Bio/Nexus/Nexus.py _unique_label() eval error


biologyguy:
Hi there,
The following function in Nexus.py is throwing exceptions when non-unique taxa are present in an input file:

```python
def _unique_label(previous_labels, label):
    """Returns a unique name if label is already in previous_labels."""
    while label in previous_labels:
        if label.split('.')[-1].startswith('copy'):
            label = '.'.join(label.split('.')[:-1]) \
                  + '.copy' + str(eval('0' + label.split('.')[-1][4:]) + 1)
        else:
            label += '.copy'
    return label
```

The traceback:

```
  File "__init__.py", line 591, in parse
    for r in i:
  File "__init__.py", line 585, in <genexpr>
    i = (r for alignment in AlignIO.parse(fp, format,
  File "__init__.py", line 374, in parse
    for a in i:
  File "NexusIO.py", line 42, in NexusIterator
    n = Nexus.Nexus(handle)
  File "Nexus.py", line 593, in __init__
    self.read(input)
  File "Nexus.py", line 644, in read
    self._parse_nexus_block(title, contents)
  File "Nexus.py", line 685, in _parse_nexus_block
    getattr(self, '_' + line.command)(line.options)
  File "Nexus.py", line 909, in _matrix
    id = _unique_label(list(self.matrix.keys()), id)
  File "Nexus.py", line 286, in _unique_label
    new_label += '.copy' + str(eval('0' + label.split('.')[-1][4:]) + 1)
```

My suggested solution:

```python
def _unique_label(previous_labels, label):
    """Returns a unique name if label is already in previous_labels."""
    while label in previous_labels:
        label_split = label.split('.')
        if label_split[-1].startswith('copy'):
            copy_num = 1
            if label_split[-1] != "copy":
                copy_num = int(label_split[-1][4:]) + 1
            new_label = "%s.copy%s" % ('.'.join(label_split[:-1]), copy_num)
            label = new_label
        else:
            label += '.copy'
    return label
```

etal:
Could you show us an example Nexus file that triggers this error? And, what was the actual error? I wasn't able to replicate it myself by duplicating lines in the MATRIX section of an existing Nexus file (and updating NTAX).

I do find your suggested `_unique_label` more readable, so I'm tempted to just use it anyway.

biologyguy:
Hi Eric,
The actual error was `SyntaxError: invalid token (<string>, line 1)` followed by the trace above.
[...]

PR #715, opened 2015-12-05
test and fix for issue #633

vincentdavis:
test and fix for issue  "_unique_label() eval error" #633 as submitted by @biologyguy in issue comments

### DIFF
--- a/Bio/Nexus/Nexus.py
+++ b/Bio/Nexus/Nexus.py
@@ -282,9 +282,13 @@ def _make_unique(l):
 def _unique_label(previous_labels, label):
     """Returns a unique name if label is already in previous_labels."""
     while label in previous_labels:
-        if label.split('.')[-1].startswith('copy'):
-            label = '.'.join(label.split('.')[:-1]) \
-                  + '.copy' + str(eval('0' + label.split('.')[-1][4:]) + 1)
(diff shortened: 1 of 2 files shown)


peterjc:
We have an OK from @etal in the discussion on #633.

Also bonus points for getting rid of a use of ``eval(...)`` :+1:

If there was no particular reason for the try/except in the test, I/Eric can just remove it when doing the merge.

vincentdavis:
The reason for the try/except in the test is that I was trying to catch and print the error. I am not sure the proper way to do this in a test. Is it necessary to catch errors with try/except in a test?
Let me know I can update also.

biologyguy:
Hi Vincent,
The error should no longer occur, so like Peter said, there's no reason to try and catch it. That's what the test suite is there to do. If, however, you want to catch an expected exception in the future, you can use assertRaises() from unittest.
Thanks for picking up this issue, btw, the bug was causing me some headache :)
-Steve

peterjc:
Squashed into commit https://github.com/biopython/biopython/commit/4a644df02bd4e4ec731d625b706f894040f60481 with minor cleanup (removed the try/except, removed unwanted file ``Tests/temp.del``, checked ``SeqIO.write(...)`` returned four).

Thank you Vincent.

Issue #721, opened 2015-12-09
Rename EvilHandleHack to something less scary


anntzer:
Yes, I know it's just a temporary measure and people are talking about dropping Python3.3 support.  Still it's a bit awkward when you teach biopython to beginning pythoneers (on an interactive console) and you're like "uh you got the result of the query, it says EvilHandleHack but don't worry we'll parse it and everything is fine."

peterjc:
What would you suggest instead of ``EvilHandleHack``?

And/or, how about we only use it on Python 3.3? It should no longer be needed on Python 3.4 onwards (and was never used on Python 2).

anntzer:
HandleWrapper or BiopythonHandleWrapper?
And yes, using it only on Python3.3 would also be a good alternative (because I don't think anyone is teaching beginners using 3.3(?)).

peterjc:
I hope we have hardly anyone using Python 3.3, and probably just people using a cluster or other Linux system with a relatively old distribution provided Python 3. I'll look at making that change...

peterjc:
Using ``io.TextIOWrapper`` on Python 3.4 and 3.5 isn't quite as the drop-in replacement I was hoping for based on ``test_Entrez_only.py``, but this is something we've essentially postponed fixing properly.

At very least we'll have to manually propagate the useful ``.url`` attribute, but that's easy enough.

peterjc:
Marking as fixed now that this will only be used on Python 3.3.

We're deprecating Python 3.3 support as of Biopython 1.67 https://github.com/biopython/biopython/commit/d183e6ed5e23dd6399759c0f73011e84ebe03954 so I think on balance there is little to gain from also renaming ``EvilHandleHack`` since hopefully hardly anyone will see it as Python 3.3 usage dwindles.

Also it is likely that ``Bio.Entrez`` will start returning an ``UndoHandle`` wrapper anyway in order to catch some of the more common poorly-handled NCBI errors, see #515.

Issue #713, opened 2015-12-05
Return bytes or unicode from Bio.SeqIO index get_raw


tiagoantao:
There are 2 problems with that file:

- The code does not work on Python 3

```python
for acc in ["P33487", "P19801", "P13689", "Q8JZQ5", "Q9TRC7"]:
     handle.write(uniprot.get_raw(acc))
```

This would work:

handle.write(uniprot.get_raw(acc).decode('utf-8'))

- The required download and decompression is massive 

ftp://ftp.uniprot.org/pub/databases/uniprot/current_release/knowledgebase/complete/uniprot_sprot.dat.gz

peterjc:
Yes, the sample data is large - but it is a real world example. The full snippet:

```
\begin{verbatim}
>>> from Bio import SeqIO
>>> uniprot = SeqIO.index("uniprot_sprot.dat", "swiss")
>>> handle = open("selected.dat", "w")
>>> for acc in ["P33487", "P19801", "P13689", "Q8JZQ5", "Q9TRC7"]:
...     handle.write(uniprot.get_raw(acc))
>>> handle.close()
\end{verbatim}
```

There is a design problem here: We index both binary and text files, so is it acceptable under Python 3 that ``.get_raw(...)`` return bytes for binary files (e.g. ``format="sff"``) and unicode for text files (e.g. ``format="swiss"``)? Or, should it always return bytes? It looks like right now it may be inconsistent - we need more test coverage here.

Likewise for the ``SeqRecord`` object ``.format(...)`` method - should ``.format("sff")`` return bytes while ``.format("swiss")`` returns unicode?

I'm going to retitle the issue to focus on this.

peterjc:
See also #723 which touches on the bytes/unicode problem from a more general perspective: Most of our parsers are text based and so want unicode strings under Python 3. We have a few explicitly binary formats like the SFF parser. However, even on non-binary formats, there are a few cases like the ``Bio.Entrez`` expat based XML parser, and the ``SeqIO`` indexing which still use byte handles.

peterjc:
Thanking about this I think we need two district methods here in the ``Bio.SeqIO`` and ``Bio.SearchIO`` indexers, what do you think of these options @tiagoantao and @bow:

* Deprecate ``get_raw(key)``, add ``get_raw_bytes(key)`` (available for all file formats) and ``get_raw_text(key)`` (only for plain text formats).
[...]

PR #723, opened 2015-12-09
Only use EvilHandleHack on Python 3.3

peterjc:
As discussed on #721, we should phase out the ``EvilHandleHack`` which was created to work around bugs in the Python 3.0 to 3.3 ``io.TextIOWrapper`` class. We still currently support Python 3.3, which means it is the only version of Python where we need this hack.

I've been primarily using ``test_Entrez_online.py`` for this - note the TravisCI tests are run with ``--offline`` so will not cover these changes.

### DIFF
--- a/Bio/Entrez/Parser.py
+++ b/Bio/Entrez/Parser.py
@@ -207,6 +207,8 @@ class DataHandler(object):
         # expects binary data
         if handle.__class__.__name__ == 'EvilHandleHack':
             handle = handle._handle
+        if handle.__class__.__name__ == 'TextIOWrapper':
+            handle = handle.buffer
         if hasattr(handle, "closed") and handle.closed:
             # Should avoid a possible Segmentation Fault, see:
             # http://bugs.python.org/issue4877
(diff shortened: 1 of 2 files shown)


peterjc:
I regard the QuantifiedCode complaint about an empty exception handler as harmless, i.e. this bit:

```
            try:
                # If wrapping an online handle, this this is nice to have:
                wrapped.url = handle.url
            except AttributeError:
                pass
```

We do that already in the ``EvilHandleHack``.

peterjc:
@tiagoantao @mdehoon Could one of you review this please? Its important for ``Bio.Entrez`` and Python 2/3 and bytes/unicode (which I know Tiago has been looking at).

peterjc:
I am happy to ignore the QualifiedCode complaint here.

peterjc:
Belatedly close this issue, changes applied to master - see #713

peterjc:
It looks like I must have merged this locally, but never pushed it to the official repository (in any case, the fix is not on the master branch yet)...

peterjc:
Actually applied to the official repository now:
59c765604c4eb38dd5fd806c195fc49f039f61b9
a6df8f57f929f556f06fcc6b0c90a3818a53623d
4bd897281900e2129a5dc5d042f6a8b81593913f

PR #724, opened 2015-12-09
fix misspellings as spotted by codespell.py

carlosp420:
https://github.com/lucasdemarchi/codespell


peterjc:
Merged as https://github.com/biopython/biopython/commit/561d506db4e5b8895d470e26f3cdc90581968303 - thanks.

PR #725, opened 2015-12-10
'dumb_consensus' conservation fix at AlignInfo.py

fjossandon:
Hello,
Currently, 'dumb_consensus' counts atoms just after skipping gaps, so if there are 100 sequences in an alignment, and in a given position there is an insertion in only one sequence (meaning 1 residue and 99 gaps), the position will be treated as 100% conserved instead of just 1%, and will pass any threshold. Since I'm looking for strings with different degrees of conservation, this is creating strings with some residues that are not as conserved as expected.

To fix it I moved the atom counting just after the length check, before the gap skipping, and did the same for 'gap_consensus' for consistency. Hope you can merge it.

Best regards,

### DIFF
--- a/Bio/Align/AlignInfo.py
+++ b/Bio/Align/AlignInfo.py
@@ -84,13 +84,14 @@ class SummaryInfo(object):
                 # make sure we haven't run past the end of any sequences
                 # if they are of different lengths
                 if n < len(record.seq):
+                    # Count atoms here or a position with 1 residue and 99 gaps
+                    # will be treated as 100% conserved instead of 1%
+                    num_atoms += 1
                     if record.seq[n] != '-' and record.seq[n] != '.':
                         if record.seq[n] not in atom_dict:
                             atom_dict[record.seq[n]] = 1
                         else:
                             atom_dict[record.seq[n]] += 1
-
-                        num_atoms = num_atoms + 1
 
             max_atoms = []
             max_size = 0
@@ -142,12 +143,11 @@ class SummaryInfo(object):
                 # make sure we haven't run past the end of any sequences
                 # if they are of different lengths
                 if n < len(record.seq):
+                    num_atoms += 1
                     if record.seq[n] not in atom_dict:
                         atom_dict[record.seq[n]] = 1
                     else:
                         atom_dict[record.seq[n]] += 1
-
-                    num_atoms += 1
 
             max_atoms = []
             max_size = 0


fjossandon:
I finished checking the failed test on "test_align.py":
```
ValueError: 
Output  : 'X  0.0 0.0 0.0 4.0'
Expected: 'T  0.0 0.0 0.0 4.0'
```

It was caused because it relied on the wrong behaviour of not counting gaps to get position conservation. In the test, the "dumb_consensus" is used with a "threshold=0.6", but the segment that fails contain this alignment:
```
gi|6273285|gb|AF191659.1|AF191      TATATA----------ATATATTTCAAATTTCCTTATATACCCAAATATA
gi|6273284|gb|AF191658.1|AF191      TATATATA--------ATATATTTCAAATTTCCTTATATACCCAAATATA
gi|6273287|gb|AF191661.1|AF191      TATATA----------ATATATTTCAAATTTCCTTATATATCCAAATATA
gi|6273286|gb|AF191660.1|AF191      TATATA----------ATATATTTATAATTTCCTTATATATCCAAATATA
gi|6273290|gb|AF191664.1|AF191      TATATATATA------ATATATTTCAAATTCCCTTATATATCCAAATATA
gi|6273289|gb|AF191663.1|AF191      TATATATATA------ATATATTTCAAATTCCCTTATATATCCAAATATA
gi|6273291|gb|AF191665.1|AF191      TATATATATATATATAATATATTTCAAATTCCCTTATATATCCAAATATA
                                    ******          ********  **** ********* *********
```
As you can see after the first 6 residues, in the following 10 positions at the "TATATATATA" sequence, conservation is T 4/7 (57%), A 4/7 (57%), T 3/7 (43%), A 3/7 (43%), and the rest "TATATA" is 1/7 for each position (14%); all below the 60% threshold. So with the new behaviour, in the position specific score matrix generated by the test the 10 positions are replaced by the ambiguous character 'N' for not being above threshold in the "with a supplied consensus sequence..." and the "defaulting to a consensus sequence..." parts:
```
A  7.0 0.0 0.0 0.0
X  0.0 0.0 0.0 4.0
X  4.0 0.0 0.0 0.0
X  0.0 0.0 0.0 3.0
X  3.0 0.0 0.0 0.0
X  0.0 0.0 0.0 1.0
X  1.0 0.0 0.0 0.0
X  0.0 0.0 0.0 1.0
X  1.0 0.0 0.0 0.0
X  0.0 0.0 0.0 1.0
X  1.0 0.0 0.0 0.0
A  7.0 0.0 0.0 0.0
```

I haven't found where the "expected" results are read since I don't see it in "test_align.py", so it would be very appreciated if a dev can fix the expected values, or I could try to do it and incorporate it to the pull request if somebody tells the location.

Best regards,

peterjc:
Currently ``test_align.py`` is one of our old print-and-compare tests with the output in ``output/test_align``. See the testing chapter in the tutorial for more details, but you can use ``python run_tests.py -g test_align.py`` to generate (and replace) the expected output file.

However, it sounds like your change is breaking existing behaviour. Can you preserve the current behaviour by default, and make your preference available as an option or new method name instead?

fjossandon:
Hi Peter,
Yes, I think that the change will have impact for low conserved regions, but I think is a good impact. Currently, what 'dumb_consensus' is really doing, is printing the the most frequent residue only among those without gaps there, regardless of the real conservation of the residue when all sequences are considered, which can produce counter intuitive results. For example, in a 10 sequence-alignment with a threshold of 80%, a very conserved position with 8 Adenines and 2 Timines will appear the same than an unconserved position with 1 Guanine and 9 gaps.

This defeats the purpose of a multiple alignment, which highlights the most and least conserved regions. With the current algorithm, a single insertion region artificially appears as a consensus with high threshold just because there are no more competing sequences there. Also the behaviour is inconsistent with 'gap_consensus' where all sequences are counted.

I was using the 'dumb_consensus' with a high threshold to look for potential primers that could detect the majority of the sequences in an alignment, expecting high conservation. But now I realize that many candidates were sequence-specific instead of global. If all sequences are considered instead in the percentage, I could create multiple consensus from 100% to 1% and reliably distinguish between global and specific sequences.

Still, I could try to generate a new argument to select between the 2 behaviours, but I really think that all sequences should be counted by default.

peterjc:
Maybe something like ``simple_consensus`` would be better as a name for the new behaviour, and we can then deprecate ``dumb_consensus`` to discourage people from using it.

(It would be bad to change the default behaviour - especially as people might follow the new documentation using an older version of Biopython from their Linux distribution package etc).

fjossandon:
Ok then, I will close this pull request and later open a new one with the new `simple_consensus`.

PR #726, opened 2015-12-11
Bio.Phylo JSON tree format

iosonofabio:
JSON is a standard file format for web application and it supports tree structures well.

This PR is aimed at developing import/export of JSON format for phylogenetic trees in Bio.Phylo. The basic idea is to have recursive "children" nodes with metadata and an optional top layer with tree metadata:

``` JSON
{
 "name": "myTree",
 "tree": {
  "name": "root",
  "children": [
   {"name": "leaf1", "branch_length": 0.01},
   {"name": "leaf2", "branch_length": 0.02}
  ]
 }
}
```
## Rationale

The main target for JSON export is web applications and JavaScript in general. Because web applications tend to be time-consuming enterprises, people need maximal flexibility for our export to JSON. For this reason:
- the key name "children" can be changed at runtime to whatever (e.g. "clades")
- the top layer can be stripped away
- arbitrary metadata can be attached to nodes and to the top tree object
## Code description

Instead of reinventing the wheel, I used the built-in Python `json` module that seamlessly converts between Python dictionaries and JSON. So for export I first convert a `Bio.Phylo.BaseTree` into a nested dictionary and then to JSON via the `json` module. Importing works in the opposite way.

**Note**: the code is not complete yet, please comment so we can polish it.

### DIFF
--- a/Bio/Phylo/JSONIO.py
+++ b/Bio/Phylo/JSONIO.py
@@ -1,0 +1,139 @@
+# Copyright (C) 2015 by Fabio Zanini (fabio.zanini __AT__ fastmail.fm)
+# Based on Bio.Phylo.NewickIO, Copyright (C) 2009 by Eric Talevich (eric.talevich@gmail.com)
+
+from __future__ import print_function
+from Bio.Phylo import Tree
+
+
+# ---------------------------------------------------------
+# Public API
+def parse(handle, **kwargs):
+    """Parse a tree in JSON format
+
+    :returns: a generator to a Bio.Phylo.Tree object.
+    """
+    return Parser(handle).parse(**kwargs)
+
+
+def write(tree, handle, indent=2, **kwargs):
+    """Write a tree in JSON format to the given file handle."""
+    return Writer(tree).write(handle, indent, **kwargs)
+
+
+# ---------------------------------------------------------
+# Input
+class Parser(object):
+    """Parse a JSON tree given a file handle.
+
+    Based on the Newick parser.
+    """
+
+    def __init__(self, handle):
+        self.handle = handle
+
+    @classmethod
+    def from_string(cls, treetext):
+        handle = StringIO(treetext)
+        return cls(handle)
+
+
+    def parse(self, **kwargs):
+        yield self.read(**kwargs)
+        
+
+    def read(self,
+             children_attrname="children",
+             metadata_nodes={'name': str, 'default': str},
+             metadata_tree={'name': str, 'default': str}):
+        import json
+
+        def node_from_json(node_dict, node):
+            for attr in node_dict:
+                if (children_attrname != 'clades') and (attr == 'clades'):
+                    print('"Clades" metadata conflicts with Biopython, skipping')
+                    continue
+
+                if attr != children_attrname:
+                    if attr in metadata_nodes:
+                        fmt = metadata_nodes[attr]
+                        setattr(node, attr, fmt(node_dict[attr]))
+                        continue
+
+                    fmt = metadata_nodes['default']
+                    try:
+                        setattr(node, attr, fmt(node_dict[attr]))
+                    except:
+                        setattr(node, attr, node_dict[attr])
+
+                else:
+                    for child_dict in node_dict[attr]:
+                        child = Phylo.BaseTree.Clade()
+                        node.clades.append(child)
+                        node_from_json(child_dict, child)
+
+        tree_dict = json.load(handle)
+        tree = Phylo.BaseTree.Tree()
+
+        if 'tree' in tree_dict:
+            root_dict = tree_dict['tree']
+            for attr in tree_dict:
+                if attr == 'tree':
+                    continue
+                if attr in metadata_tree:
+                    fmt = metadata_tree[fmt]
+                    setattr(tree, attr, fmt(tree_dict[attr]))
+                    continue
+                fmt = metadata_tree['default']
+                try:
+                    setattr(tree, attr, fmt(tree_dict[attr]))
+                except:
+                    setattr(tree, attr, tree_dict[attr])
+
+        else:
+            root_dict = tree_dict
+
+        node_from_json(root_dict, tree.root)
+        return tree
+
+
+# ---------------------------------------------------------
+# Output
+class Writer(object):
+
+    def __init__(self, tree):
+        self.tree = tree
+
+
+    def write(self, handle, indent, **kwargs):
+        """Write this instance's tree to a file handle."""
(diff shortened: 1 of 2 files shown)


peterjc:
See discussion on the mailing list, http://lists.open-bio.org/pipermail/biopython-dev/2015-December/021235.html - in particular my concern shared by @cjfields and @etal that we do not want to try to unilaterally invent a Biopython specific JSON phylogenetic representation (see eg http://www.slideshare.net/dullhunk/the-seven-deadly-sins-of-bioinformatics/ Carole Goble (BOSC 2007) *The Seven Deadly Sins of Bioinformatics*). If there is an existing JSON representation we should follow that.

This was also forwarded to the BioJS group https://groups.google.com/forum/#!topic/biojs/gcHHjq1jV1s but thus the most useful replies have been off list.

Fabio started a new thread at http://lists.open-bio.org/pipermail/biopython-dev/2015-December/021244.html

iosonofabio:
Thanks Peter for coordinating this. Nobody wants to reinvent anything. AFAIK the format specified above is the only "standard", but I might be wrong.

I am opinionated about only one thing: I refuse to force metadata/attributes such as names, branch lengths, or similia. One of JSON's advantages is that it's human readable, and if someone just wants a sleek, minimal tree they should not be forced to wade across jungles of empty attributes, e.g.:
```JSON
{
 "name": "myTree",
 "creationDate": "undefined",
 "modifyDate": "undefined",
 "rooted": "undefined",
 "tree": {
  "name": "undefined",
  "branch_length": 0,
  "isSpeciation": "undefined",
  "children": [
   {"name": "undefined", "branch_length": 0.01, "isSpeciation": "undefined"},
   {"name": "undefined", "branch_length": 0.02, "isSpeciation": "undefined"}
  ]
 }
}
```
instead of
```JSON
{"children": [
 {"branch_length": 0.01},
 {"branch_length": 0.02}
 ]
}
```

rneher:
https://github.com/blab/nextflu (http://nextflu.org) uses pretty much exactly what @iosonofabio is proposing.

peterjc:
Would these representations have the same ambiguities about branch support (associating with the split, versus a node) seen in Newick? See #745.

iosonofabio:
not quite sure what you mean here.

Although trees can be rooted or unrooted, most practical implementation are de facto rooted (including Bio.Phylo.BaseTree) for convenience. Given a rooted tree, for each node except the root there is one and only one edge that runs *into* the node (from the root, that is), so associating properties with the node or with the unique edge running into it is the same exact thing...

peterjc:
That's exactly the problem raised in http://biorxiv.org/content/early/2015/12/31/035360 - any edge has two nodes associated with it. As soon as you reroot a tree, the node "into" an edge can change into the "out" node, meaning the bootstrap could not be associated with the wrong edge/split.

If the bootstraps can be recorded in the file format against the edge itself (or split, if you prefer that terminology) this ambiguity goes away.

rneher:
Thanks for clarifying this. My take is that this is a problem with with rerooting (i.e. flipping edges) and the internal representation of the tree in the viewer or biopython, rather than with the file format. I think Bio.Phylo.BaseTree.root_with_outgroup fails to do this correctly (at least confidence values aren't touched as far as I can tell). 

When saving a rerooted tree as newick, labels would need to change such that support values are associated with the correct edge. When saving a rerooted as json, the support field would also need to change, but one wouldn't need to mess with the node names/labels.

iosonofabio:
I agree with Richard. Of course, storing edges and nodes separately would be a cleaner design, but `Bio.Phylo.BaseTree` only stores nodes - and luckily so: I have worked with explicit edges in the past, the bookkeeping rapidly becomes a nightmare.

Now given this informal (but convenient and efficient) representation, it is the job of the rerooting algorithm to swap all edge-related properties such as support. Of course, it would be nice if:
- Biopython had a way of flagging such properties that need swapping under rerooting, or
- that could be an option to the reroot function,

but either way that's not the topic of this PR.

So overall JSON or newick are neither more nor less informal than the underlying `Bio.Phylo.BaseTree` representation. A Biopython tree is internally rooted and keeps so during file I/O.

peterjc:
The topic here is a JSON based tree file format - and I'm asking if that could avoid the boot strap support ambiguity problem in Newick (regardless of what ``Bio.Phylo.BaseTree`` currently does)?

rneher:
short of coming up with a flat format of separate nodes and edges, one could do one of the following to resolve the ambiguity.

  * include an optional field that contains names of properties of the edges, which could default to `branch_length`.   
```
{
 "name": "myTree",
 "edge_properties": ["branch_length", "support", "color",...],
 "tree": {
    "name": "root",
    "children": [
            {"name": "leaf1", "branch_length": 0.01},
            {"name": "leaf2", "branch_length": 0.02}
          ]
      }
}
```
 * have explicit edge properties for each node. 
```
{
 "name": "myTree",
 "tree": {
    "name": "root",
    "edge_properties": {"branch_length":0.01, "support":0.99, "color",...},
    "children": [
            {"name": "leaf1", "edge_properties": {"branch_length":0.01, "support":0.99, "color",...},},
            {"name": "leaf2", "edge_properties": {"branch_length":0.01, "support":0.99, "color",...},}
          ]
      }
}
```
The second likely produces all kinds of incompatibilities. I'd prefer the former.

cjfields:
I think the distinction @peterjc is making is important.  Essentially this is a new format, if I'm not mistaken. Should it be driven by fitting to a specific implementation (which based on comments may have underlying problematic assumptions) or try to deal with other issues such as the one he raised, and have the implementation adapt to it?

If the latter it could start with a biopython implementation but probably have a life outside biopython as well.  Tying it to a specific implementation OTOH may limit its use.

iosonofabio:
I agree with @peterjc and @cjfields: this is a valid point which is currently broken in `Bio.Phylo.BaseTree` and ambiguous in newick, and we should not propagate that kind of issue to a new format.

@rneher 's first proposal is easy to implement, fast to parse, and human-readable, I am in favour. I also think `Bio.Phylo.BaseTree` could follow the same principle to guide the rerooting function, i.e. a tree-wide list of "edge properties" that need to be switched on rerooting. Sounds simple and functional. In this sense, we would solve the issue in the JSON format first and the Biopython implementation would follow.

I am in general against storing separate nodes and edges because, although it is the best format on paper, it is a nightmare for both computers and humans to parse and write.

peterjc:
As far as I know, a JSON phylogeny tree format never caught on (but I could easily have missed such news), so I am going to close this stale PR.

PR #727, opened 2015-12-14
Added 'simple_consensus' as a modification of 'dumb_consensus' with tests

fjossandon:
As discussed in the previous pull request #725 , I created `simple_consensus` at AlignInfo, which count all sequences when getting residue percents, so single insertions are not treated the same way than very conserved strings by the threshold value. Also updated test_align.py to include a test for it.

Cheers,


fjossandon:
I wanted to comment that the issues that QuantifiedCode complains is "Prefer `format()` over string interpolation operator", which is everywhere in the "test_align.py" file and not just my additions. Also saw that other pull requests are marked as fail ('X') only for that, maybe that rule is too strict?

On the other hand, the build and tests pass fine.

peterjc:
Don't worry too much about QuantifiedCode - as you've noticed, it complains a lot about the existing code too. We've only recently started looking at this tool...

Issue #714, opened 2015-12-05
Update tutorial SeqIO.index_db GenBank example gbvlr01.seq etc


tiagoantao:
The code will not work:

 print(gb_vrl["GQ333173.1"].description)

as the database on genbank is now a different one

tiagoantao:
This would probably work if the complete database is downloaded (not only 16 files)

peterjc:
i.e. This example:

```

\begin{verbatim}
>>> from Bio import SeqIO
>>> files = ["gbvrl%i.seq" % (i+1) for i in range(16)]
>>> gb_vrl = SeqIO.index_db("gbvrl.idx", files, "genbank")
>>> print("%i sequences indexed" % len(gb_vrl))
958086 sequences indexed
\end{verbatim}

...

\begin{verbatim}
>>> print(gb_vrl["GQ333173.1"].description)
HIV-1 isolate F12279A1 from Uganda gag protein (gag) gene, partial cds.
\end{verbatim}
```

There are currently 38 chunks, not 16. This ought to be a small change to the example, and the wording could be clarified to help the reader take the likely growth of the database into consideration.

(Is there a smaller more static set of GenBank files we could use instead?)

PR #728, opened 2015-12-14
Update GenBank virus SeqIO.index_db example in Tutorial

peterjc:
This should address GitHub issue #714.

This dataset has roughly doubled in size since I wrote the example, meaning it has become uncomfortably large for use as an example.

Changed the example to use just the first four chunks as suggested by Tiago (@tiagoantao), currently the test record is in the first chunk.

On the downside this is now longer, but previously there was no guidance on how to download the files.

Also I am using glob to build the file list rather than hard coding the number of files expected.


peterjc:
How does this look @tiagoantao ?

peterjc:
Cherry-picked as https://github.com/biopython/biopython/commit/694f0b0befda3448d3c1dba32c4b2d198781e478

PR #734, opened 2015-12-18
Document and test that get_raw returns bytes

### LINKED ISSUE #713: Return bytes or unicode from Bio.SeqIO index get_raw

peterjc:
This would mostly resolve #713 (see discussion there about possible extensions).

@tiagoantao & @bow could you review this please?

### DIFF
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -229,7 +229,12 @@ class _IndexedSeqFileProxy(object):
         raise NotImplementedError("Subclass should implement this")
 
     def get_raw(self, offset):
-        """Returns bytes string (if implemented for this file format)."""
+        """Return the raw record from the file as a bytes string (if implemented).
+
+        If the key is not found, a KeyError exception is raised.
+
+        This may not have been implemented for all file formats.
+        """
         # Should be done by each sub-class (if possible)
         raise NotImplementedError("Not available for this file format.")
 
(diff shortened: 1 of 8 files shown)


bow:
Looks good, @peterjc. I only see that since you are also updating some of the docstrings to mention the bytes string return type, there are some `get_raw` methods that may need to be updated/added as well:

* ``SearchIO/ExonerateIO/exonerate_vulgar.py``
* ``SearchIO/ExonerateIO/exonerate_text.py``
* ``SearchIO/HmmerIO/hmmer3_tab.py``
* ``SearchIO/HmmerIO/_base.py``
* ``SearchIO/BlastIO/blast_xml.py``
* ``SearchIO/BlastIO/blast_tab.py``

It would be nicer if a child method's docstring can fall back on its parent's when not defined. Not without some dirty tricks, I guess :).

peterjc:
Thanks @bow - rebased and applied with https://github.com/biopython/biopython/commit/28b8a1a7bd1d0b84efa4dced8c5db39cffc1ea20 updating the docstrings you mentioned. Closing.

PR #740, opened 2015-12-30
BioSQL ncbi taxonomy fetch bug

ctSkennerton:
I've encountered two bugs when trying to populate the taxonomy information from NCBI when loading a record into a BioSQL database. The first was that the `DatabaseLoader` class (sometimes) automatically converts an NCBI taxonomy ID into an integer, however `Entrez.efetch` expects this value as a string (or a list of strings). I've modified  `Entrez.efetch` such that it automatically converts integers into strings to solve this. The second bug was in the `DBSeq` class removes any taxonomic name that has a rank of 'no rank'. This can change the returned taxonomy of the sequence significantly and is incorrect with what is reported at NCBI. It is further confusing as the taxonomic names are stored correctly in the database but are filtered out after their retrieval.

I've written in a new test to `common_BioSQL.py` to check that the taxonomy info returned from the database is the same as obtained by using the `Entrez` module

### DIFF
--- a/BioSQL/BioSeq.py
+++ b/BioSQL/BioSeq.py
@@ -464,11 +464,8 @@ def _retrieve_taxon(adaptor, primary_id, taxon_id):
             # load_ncbi_taxonomy.pl this is how top parent nodes are stored.
             # Personally, I would have used a NULL parent_taxon_id here.
             break
-        if rank != "no rank":
-            # For consistency with older versions of Biopython, we are only
-            # interested in taxonomy entries with a stated rank.
-            # Add this to the start of the lineage list.
-            taxonomy.insert(0, name)
+
+        taxonomy.insert(0, name)
         taxon_id = parent_taxon_id
 
     if taxonomy:
--- a/Tests/common_BioSQL.py
+++ b/Tests/common_BioSQL.py
@@ -620,16 +620,27 @@ class LoaderTest(unittest.TestCase):
     def test_load_database_with_tax_lookup(self):
         """Load SeqRecord objects and fetch the taxonomy information from NCBI.
         """
-
-        from Bio import Entrez
-        Entrez.email = "biopython-dev@biopython.org"
-
-        handle = Entrez.efetch(db="taxonomy", id=3702, retmode="XML")
-
-        taxon_record = Entrez.read(handle)
-        entrez_tax = []
-        for t in taxon_record[0]['LineageEx']:
-            entrez_tax.append(t['ScientificName'])
+        tax = ['cellular organisms',
+               'Eukaryota',
+               'Viridiplantae',
+               'Streptophyta',
+               'Streptophytina',
+               'Embryophyta',
+               'Tracheophyta',
+               'Euphyllophyta',
+               'Spermatophyta',
+               'Magnoliophyta',
+               'Mesangiospermae',
+               'eudicotyledons',
+               'Gunneridae',
+               'Pentapetalae',
+               'rosids',
+               'malvids',
+               'Brassicales',
+               'Brassicaceae',
+               'Camelineae',
+               'Arabidopsis',
+               'Arabidopsis thaliana']
 
         self.db.load(self.iterator, True)
 
@@ -645,8 +656,7 @@ class LoaderTest(unittest.TestCase):
         self.assertEqual(test_record.annotations['ncbi_taxid'], 3702)
         # make sure that the taxonomic lineage is the same as reported
         # using the Entrez module
-        self.assertEqual(test_record.annotations['taxonomy'],
-                entrez_tax)
+        self.assertEqual(test_record.annotations['taxonomy'], tax)
 
 
 class DeleteTest(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


ctSkennerton:
:bump: 
(Dec 29th probably wasn't the best time to open a PR)

peterjc:
The Entrez change to handle integer arguments looks sensible, I'd be happy to merge that as a separate pull request with an extra test in ``test_Entrez_online.py`` - do you agree @mdehoon?

I'm less clear about the "no rank" issue. As I recall, the old code was mimicking what the NCBI would do for the taxonomy lineage in their GenBank files. Perhaps a specific example would help here?

peterjc:
The TravisCI failure is unrelated, and would be fixed by https://github.com/biopython/biopython/commit/dd233650512cd0cadc637ac37d377141125f4059 - are you comfortable with rebasing a branch and then force pushing it to GitHub? That would give us a TravisCI report against the current master branch.

ctSkennerton:
The example I was using was [X55053.1](http://www.ncbi.nlm.nih.gov/nuccore/X55053), which is from Arabodopsis, where the taxonomy in the genbank record contains levels with no rank, for example [Tracheophyta](http://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=58023&lvl=3&lin=f&keep=1&srchmode=1&unlock). However I also just looked at the [Desulfovibrio desulfuricans genome](http://www.ncbi.nlm.nih.gov/nuccore/NC_011883.1), which does not contain any of the levels without a rank. This is super annoying cause in the [taxonomy database](http://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=525146) it shows those levels - this is where biopython gets its taxonomy info to be added into the database. Why can't NCBI just be consistent! 

My personal preference is that biopython return the hierarchy as it is stored in the database as I feel that consistent and it's less confusing to people who might also access the database info with SQL statements. However this will bring up surprises to people who are expecting the genbank record they put in to be the same as the one that the API spits out (and that record happens to be one where NCBI strip out some of the ranks in the taxonomy). Perhaps we can note this in the documentation somehow and let the user decide how to present the taxonomy.

ctSkennerton:
I'll split out the Entrez change as a separate PR too

peterjc:
Note that traditionally BioSQL was used with a pre-imported copy of the full NCBI taxonomy - the fact that Biopython can incrementally populate the taxonomy tree via Entrez on demand is a relatively recent innovation. I don't know if any of the other Bio* interfaces for BioSQL do this.

The "no rank" issue is tricky. My understanding is BioSQL was designed really as a database schema for GenBank flat files, which is what I was trying to mimic for the output - but the details are not fresh in my mind. Hilmar mights have some thoughts (@hlapp), depending what he thinks we might to file an issue at https://github.com/biosql/biosql/issues  to document the expected use in the tables.

ctSkennerton:
Whether it is pre-populated or is stored incrementally doesn't really change things, it still comes from the taxonomy database. I think the heart of the problem is that the NCBI taxonomy database is not always consistent with what is shown in the genbank flatfile. It does seem like a lose-lose situation since some uses will expect those ranks to be removed, whereas others will not.

peterjc:
Can you rebase this to the current master?

Returning all the taxonomy as you propose seems reasonable...

ctSkennerton:
Ok I rebased onto master. I hope I did it right

peterjc:
Can you try this from your ``biosql_ncbi_tax_fetch_bug`` branch?

```
git fetch origin
git rebase origin/master
git push conner biosql_ncbi_tax_fetch_bug --force
```

Modify the first line ``origin`` if you used something else like ``upstream`` for the Biopython repository, and modify the the last line to replace ``conner`` with however you named your forked Biopython on GitHub.

If it goes horribly wrong, ``git rebase --abort`` should reset things back to has they are now.

peterjc:
Wait until you've got the rebase working, and then GitHub should refresh this pull request with just your changes. We can then go into fixing the tests (hopefully easy) and the style check (looks trivial).

Since you started on this we've tightened up the automatic code style checks. For that it will be helpful to do ``pip install flake8`` and then run it locally.

peterjc:
There's still a (trivial) style issue from flake8,

```
$ flake8 Tests/
Tests/common_BioSQL.py:622:1: E303 too many blank lines (3)
```

(See also #493 and #883 for our evolving git pre-commit hook and TravisCI style checks)

Note to self: This will need additional testing locally in order to include the online tests (which TravisCI currently skips)

ctSkennerton:
Thanks. How do I actually run the pre-commit hook?

peterjc:
If the git pre-commit hook is installed correctly, on a test branch try deliberately breaking the style guidelines in a file (say ``Bio/Seq.py``) and then attempt to commit it (``git commit Bio/Seq.py -m "Breaking the coding style"). The check should run and stop you, unless you over-ride it using ``git commit  ... --no-verify``.

The idea of our hook was that it will be even stricter than the TravisCI style check (eg it will complain about lambda functions), so that you can be aware of some issues but opt to commit them anyway. The style checks on TravisCI are currently more relaxed (there is still plenty of older code in Biopython using lambda functions for example), but we insist those pass.

If you want to talk more about this, let's do it on #493 and #883

peterjc:
Thanks Connor.

(Once a git pre-commit hook is installed, doing ``git commit ...`` will run the checks automatically)

PR #759, opened 2016-01-26
Support RaxML testing on Windows

benfulton:


peterjc:
I just looked at ``raxml-win32-100315.tar.gz`` which is currently the latest provided Windows executable, and its just contains a ``release-win32`` folder with ``raxmlHPC.exe`` and ``rxmlHPC-PTHREADS.exe`` - no installer, and also no instructions.

Does the RaxML documentation give any guidance on installation, such as a suggested path to put the binaries at? Otherwise your assumption that the binary will be on the ``$PATH`` on Windows seems fine.

peterjc:
P.S. I'm happy with the empty exception handler which QuantifiedCode is complaining about here.

benfulton:
No, RaxML for Windows is pretty much an afterthought. The Linux version
just suggests making sure the executable is on the path.

I am inclined to make the $PATH assumption in general for Windows binaries
in the same way as it is for Linux. The tests jump through a lot of hoops
to try to find Windows executables, and it might be all for naught if a
user tries to run the application with the wrong path after successfully
running the tests.

peterjc:
OK. Some of the windows tools like ClustalW used to come with Windows Installers which put the binaries in predictable places (but sadly not on the ``$PATH``), leading to some "heuristics" in our test suite - but as you say, the end user still would have to set the full path to the binaries themselves.

Issue #757, opened 2016-01-25
"No such file or directory" error by Bio.PDB.DSSP


rraadd88:
I have 1MOT.pdb in the working directory. When I run the example bit taken from [here](http://biopython.org/DIST/docs/api/Bio.PDB.DSSP%27.DSSP-class.html). 

<pre><code>
p = PDBParser()
structure = p.get_structure("1MOT", "1MOT.pdb")
model = structure[0]
dssp = DSSP(model, "1MOT.pdb")
</code></pre>

I get this error :

<pre><code>
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-12-e056ad12049f> in <module>()
----> 1 dssp = DSSP(model, "1MOT.pdb")

/lib/python2.7/site-packages/Bio/PDB/DSSP.pyc in __init__(self, model, pdb_file, dssp)
    252         """
    253         # create DSSP dictionary
--> 254         dssp_dict, dssp_keys = dssp_dict_from_pdb_file(pdb_file, dssp)
    255         dssp_map = {}
    256         dssp_list = []

/lib/python2.7/site-packages/Bio/PDB/DSSP.pyc in dssp_dict_from_pdb_file(in_file, DSSP)
    107     # gives unicode handles rather than bytes handles.
    108     p = subprocess.Popen([DSSP, in_file], universal_newlines=True,
--> 109                          stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    110     out, err = p.communicate()
    111 

/lib/python2.7/subprocess.pyc in __init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags)
    708                                 p2cread, p2cwrite,
    709                                 c2pread, c2pwrite,
--> 710                                 errread, errwrite)
    711         except Exception:
    712             # Preserve original exception in case os.close raises.

/lib/python2.7/subprocess.pyc in _execute_child(self, args, executable, preexec_fn, close_fds, cwd, env, universal_newlines, startupinfo, creationflags, shell, to_close, p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite)
   1333                         raise
   1334                 child_exception = pickle.loads(data)
-> 1335                 raise child_exception
   1336 
   1337 

OSError: [Errno 2] No such file or directory 
</code></pre>

Any help?
My guess is this bit in DSSP.py 
</code></pre>
[...]

PR #761, opened 2016-01-26
Added check for DSSP executable

JoaoRodrigues:
As suggested in issue #757 .


JoaoRodrigues:
@peterjc After spending some time looking at possible ways of properly fixing this, I think it's better to adopt the AbstractCommandline wrapper for DSSP (and NACCESS) as it provides a much nicer interface and ties in better with the rest of the codebase. Will work on it. Thanks for the extra pair of eyes here..

PR #767, opened 2016-02-09
BioSeqDatabase __getitem__ bug

ctSkennerton:
I discovered that there was no check in the `__getitem__` method of `BioSeqDatabase` to make sure that the identifier was actually part of the current namespace (biodatabase) and thus it was possible to retrieve any sequence that is in the SQL server. This pull request adds this check, raising a KeyError where appropriate, and adds in some new test functionality for databases with multiple namespaces.

### DIFF
--- a/BioSQL/BioSeqDatabase.py
+++ b/BioSQL/BioSeqDatabase.py
@@ -629,8 +629,10 @@ class BioSeqDatabase(object):
         return list(self.keys())
 
     def __getitem__(self, key):
-        if key not in self:
-            raise KeyError("Entry %r cannot be found or is invalid" % key)
+        record = BioSeq.DBSeqRecord(self.adaptor, key)
+        if record._biodatabase_id != self.dbid:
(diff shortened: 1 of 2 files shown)


peterjc:
Do you think a more explicit message from the ``KeyError`` would be useful here? e.g.

```
raise KeyError("Entry %r does exist, but not in current name space" % key)
```

ctSkennerton:
Hi @peterjc, I've updated the error messages and changed the test to make it compatible with python 2.6.

peterjc:
Green light from TravisCI, and Codecov. None of the QuantifiedCode points seem worth worrying about.

ctSkennerton:
Hi @peterjc,

Thought I'd just bump this PR, I've addressed all of you comments so I think that this probably just got left behind with all the other issues that come up with this repo

peterjc:
Thanks for the reminder, yes - this was ready to include. Applied as squashed commit https://github.com/biopython/biopython/commit/6e069690483961723d547aefcff3ff89edc5b630 - thank you!

PR #769, opened 2016-02-11
Initial changes for Issue #768

frubino:
I've made some changes, to read/write the Stockholm alignment format, to include GF and GC features in the *annotations* attribute and write them back to file.

### DIFF
--- a/Bio/AlignIO/StockholmIO.py
+++ b/Bio/AlignIO/StockholmIO.py
@@ -141,22 +141,29 @@ from .Interfaces import AlignmentIterator, SequentialAlignmentWriter
 from Bio._py3k import OrderedDict
 
 
+PFAM_GR_MAPPING = {
+    "secondary_structure": "SS",
+    "surface_accessibility": "SA",
+    "transmembrane": "TM",
+    "posterior_probability": "PP",
+    "ligand_binding": "LI",
+    "active_site": "AS",
+    "intron": "IN"
+}
+# Following dictionary deliberately does not cover AC, DE or DR
+PFAM_GS_MAPPING = {
+    "organism": "OS",
+    "organism_classification": "OC",
+    "look": "LO"
+}
+
+
 class StockholmWriter(SequentialAlignmentWriter):
     """Stockholm/PFAM alignment writer."""
 
-    # These dictionaries should be kept in sync with those
-    # defined in the StockholmIterator class.
-    pfam_gr_mapping = {"secondary_structure": "SS",
-                       "surface_accessibility": "SA",
-                       "transmembrane": "TM",
-                       "posterior_probability": "PP",
-                       "ligand_binding": "LI",
-                       "active_site": "AS",
-                       "intron": "IN"}
-    # Following dictionary deliberately does not cover AC, DE or DR
-    pfam_gs_mapping = {"organism": "OS",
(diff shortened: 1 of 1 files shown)


frubino:
On the first check: I've seen string interpolation used, instead of "format". Which way is the preferred in Biopython?

On the second... I still have to check the test suite for it.

peterjc:
We still have loads of string interpolation, and I personally prefer it. QuantifiedCode does aggressively push the format method (I wonder if there is a simple setting to turn that off?).

The test failures from TravisCI do look real and relevant though.

frubino:
Hi, yes. I only tested it yesterday using "format" on a stockholm file. I've noticed today that when using `AlignIO.write` instead, an error came out. I'll fix it soon, but I wanted to make sure that some of the code changes were what was expected.

frubino:
The last commits should do the trick. The tests are passed for most python versions, except for PyPy.

frubino:
@peterjc: do you think more changes should be made to the code?

peterjc:
Apologies for not getting back to this work. We've adjusted the QuantifiedCode settings (they've recently added this functionality), so I think it is happy now.

Other than the docstring doctest, did you have any additional tests in mind for this?

frubino:
At the moment I have other tests to add.

Bio/AlignIO/StockholmIO.py
```diff
@@ -389,7 +407,13 @@ def __next__(self):
                 elif line[:5] == '#=GC ':
                     # Generic per-Column annotation, exactly 1 char per column
                     # Format: "#=GC <feature> <exactly 1 char per column>"
-                    pass
+                    feature, text = line[5:].strip().split(None, 1)
+                    # Each feature key could be used more than once,
+                    # so store the entries as a list of strings.
```
iyak: Is this correct? Wikipedia says "Do not use multiple lines with the same #=GC label."
So isn't it something like
```python
if feature in gc:
    raise ValueError("Duplicate feature for #=GC label:\n" + line)
else:
    gc[feature] = text
```

peterjc:
Hello again @frubino - apologies but I forgot about this pull request until #3314 was filed.

Are you willing to dual license your contributions as per #898? If not, then unfortunately I will have to close this.

Did @iyak's proposal make sense to raise an exception on invalid duplicated GF lines? Or it could trigger a warning, and ignore additional entries?

frubino:
> Hello again @frubino - apologies but I forgot about this pull request until #3314 was filed.
> 
> Are you willing to dual license your contributions as per #898? If not, then unfortunately I will have to close this.
> 
> Did @iyak's proposal make sense to raise an exception on invalid duplicated GF lines? Or it could trigger a warning, and ignore additional entries?

Honestly I forgot about it as well, @peterjc . As per the licensing, I accept the dual licensing. I'll give a look at @iyak suggestion in a moment.

Bio/AlignIO/StockholmIO.py
frubino: Yes, that change should be implemented. I think my problem was with some files that didn't follow the specs.

frubino (commented):
Implemented change

peterjc:
Cross reference #4504 in ``Bio.Align`` (different parser, not ``Bio.AlignIO``)

PR #792, opened 2016-03-25
Entrez Efetch using int identifier

ctSkennerton:
Some NCBI identifiers are integers, such as taxonomy ids or GI
numbers. It is natural then that a user may code them as integers
in python. This additional test automatically converts an ID given
to efetch as an integer to a string, which is required by the
entrez API.

As suggested by @peterjc in pull #740 I've moved these changes
into their own PR


peterjc:
Thank you!

PR #794, opened 2016-03-30
added support for mkdssp cmd line argument in Bio.PDB.DSSP

andrewguy:
The recent version of the external DSSP package, which is available in the standard Ubuntu (or Debian) repositories, has renamed the executable as `mkdssp`, rather than `dssp`. Under Debian there is apparently a symlink created that enables calling `dssp` as well, but this doesn't exist under Ubuntu (at least not 14.04). This has created issues as the Bio.PDB.DSSP module calls `dssp` by default. While it is possible to work around this, this is probably needless debugging that users would have to go through.

To make for a painless user experience, and to preserve backwards compatibility with scripts using DSSP, I propose a few small changes to the Bio.DSSP module to call `mkdssp` when a call to `dssp` fails. I've updated the unit test accordingly.

Thanks,

Andrew


peterjc:
Hmm. I'm not sure about this automatic command substitution, it feels too magic - and Debian/Ubuntu specific.

I'd be more comfortable with just updating ``test_DSSP_tool.py`` to work out if the binary is ``dssp`` or ``mkdssp`` and then pass this as an option to the functions.
 
Maybe we should ask on the development mailing list for more opinions?

andrewguy:
Hi Peter,

Yes, in retrospect I think you are right - both in terms of being a little deceptive by substituting commands under the covers, and being specific to a particular OS. Although from a quick look I don't think it is specific to Ubuntu - the binary in the Redhat package is also `mkdssp`.

Perhaps all that is needed (apart from updating the tests) is a little more clarity in the documentation? Even just a quick mention that users of some systems will have to pass `dssp='mkdssp'` as an additional argument due to changes to the external DSSP program?

Another option would be to leave the default `dssp` parameter as `None`. If no specific path is specified, then the module tries both `"dssp"` and `"mkdssp"`. If a user specifies a `dssp` argument, then that's it, no under-the-cover magic. Might require more extensive changes, and I'm not sure what that would do to backwards compatibility.

peterjc:
How about making ``test_DSSP_tool.py`` use either name, don't change the default but as you suggest explicitly document this by showing an example with ``dssp='mkdssp'`` to call the tool when the binary is ``mkdssp`` instead of ``dssp``?

I'd initially assumed this code followed our ``Bio.Application`` command line wrapper approach where it felt wrong to magically detect the binary. Given the DSSP wrapper design, perhaps auto-detection isn't so bad?

CC @JoaoRodrigues @etal @lennax as people likely to have used DSSP from ``Bio.PDB``.

peterjc:
P.S. I am not worried about the QualifiedCode complaint about the new test name method (switching it to all lower case might make it happy though).

andrewguy:
Yep, that sounds reasonable. I'll make the changes to the test case and documentation and update the pull request, leaving the actual code alone.

From a user perspective, it'd be nice not to worry about which particular executable you're using, but it depends what other people think about changing the default value of the `dssp` parameter.

And thanks for the reassurance about the QuantifiedCode complaint - I was a bit worried until I saw what it was objecting to. Can always put DSSP in lower case.

JoaoRodrigues:
`mkdssp` is also the name of the binary built by compiling the source code from DSSP 2.0 (CMBI). I'm not very comfortable with changing the default parameter, surely better documentation would be a plus. I'd be OK with trying `dssp` and then `mkdssp` if the first one fails, as these seem to be two 'default' names. 

:+1:

andrewguy:
I have added some more detail to the documentation, and updated some sections (`DSSP` class) that didn't reflect the current code, especially with regard to expected output. Things should hopefully be a bit clearer now.

I've also added a simple `try: except` statement within the main code that will try `mkdssp` if the original call fails, but doesn't change the default parameter. I've removed all other changes I had previously made to the main code. Hope everyone is ok with this.

krdav:
How come this minor, but relevant commit has not gotten merged? With the try/expect it looks okay and not to Debian specific. Indeed I have also encountered DSSP binaries named mkdssp.

peterjc:
Time :(

@JoaoRodrigues: were you happy with later changes from @andrewguy?

This will now need a manual rebase/merge but I am willing to do that.

JoaoRodrigues:
Yes, all fine by me. Thanks!

peterjc:
I'm trying to deal with some of the open pull requests in the hope of doing a release soon.

@andrewguy Are you happy to dual licence your contributions under both the _Biopython License Agreement_ *and* the _BSD 3-Clause License_? See #898.

Also, are you happy to be thanked by name (Andrew Guy?) in the ``NEWS`` and ``CONTRIB`` files?

andrewguy:
Yep, all fine by me, both in regard to the dual licence and being thanked by name. Cheers!

mathuin:
@peterjc I noticed that this PR is complete excepting a manual rebase.  Is this something that either @andrewguy or I could do to help add this to the next release?

andrewguy:
@mathuin I'm happy to do it, but not quite sure how (still a bit of a git newbie). Got a quick rundown on what that would require?

mathuin:
If the conflicts are straightforward, there should be a button next to "This branch has conflicts that must be resolved" that says "Resolve conflicts".  More info here:  https://github.com/blog/2293-resolve-simple-merge-conflicts-on-github  If they're more complex, you'll have to do it by hand, and that's harder to explain in a comment. :-)

peterjc:
When there are no merge conflicts, we'd normally do the "merge and rebase" via the GitHub interface, so in that case there is no need to do a manual merge.

Unfortunately here there is a merge conflict in ``Bio/PDB/DSSP.py`` which needs to be resolved. If you were more experienced with git, I would ask you to do rebase onto the current master branch. Instead I'll do this now at the command line...

andrewguy:
Thanks Peter, I'm happy to do a rebase onto the current master - was just confused as to what a 'manual' rebase was. Didn't realise you could do it via the github interface. Let me know if you haven't done it yet.

peterjc:
The first conflict is from https://github.com/biopython/biopython/commit/b6f542309a767e75f48602be83e0b3c849b772a6

Regarding the GitHub merging options, see https://github.com/blog/2243-rebase-and-merge-pull-requests - using squash or rebase gives a simple linear history (matching how Biopython used to be developed back on CVS), rather than the more complicated history a branch and merge gives.

peterjc:
Proposed rebased branch: https://github.com/peterjc/biopython/tree/dssp-fix

Currently waiting for TravisCI tests: https://travis-ci.org/peterjc/biopython/builds/196895554

andrewguy:
Cheers Peter, that looks great. Thanks for doing the rebase. :)

peterjc:
Changes applied to master (commits leading up to 3d6f4e9fbeb8f1a03afaa1e147214100421f6690), closing pull request.

Thank you Andrew, Joao for reviewing, and Kristian and Jack for encouragement and also looking at this.

PR #799, opened 2016-04-05
add support for parsing PAML/codeml M2a_rel model results

brandoninvergo:
I was completely unaware of another model used by codeml, added a couple years ago, so parsing its output was failing. This pull request includes the really simple fix and updates to the test suite (and thus also adds a few new files used in the testing).


peterjc:
One of the minor issues flagged by QuantifiedCode  is to switch to using ``test_`` rather than just ``test`` as the prefix in your unit tests - also applies to ``test_SCOP_*.py`` etc. Probably worth a separate pull request...

```
$ grep "def test" *.py | grep -v "def test_" | wc -l
     145
$ grep "def test_" *.py | wc -l
    1868
```

peterjc:
Committed as https://github.com/biopython/biopython/commit/4031b632b8e34dffc29c358cd80713a55258e140 - thank you!

Issue #779, opened 2016-03-03
MMCIF vs. PDB file information


JoaoRodrigues:
This is somewhat related to #778 and #775.

The current mmCIF parser uses `_atom_site.label_asym_id` to attribute chain identitifers in the SMCRA object hierarchy. This is not necessarily equivalent to the chain identifier field in the PDB format, e.g. `4q94` or `1brs`. On the other hand, the `auth_asym_id` field is a direct match to the PDB chain identifier (when it exists).

The practical consequence of this difference is that when parsing the mmCIF and PDB files of the same structure, we get different chain identifiers and different numbers of chains (see below).

```
Str. Name	#M	#C	#R	#A
4q94.cif	1	30	388	2337
4q94.pdb	1	 4	388	2337
```

I suggest we change the assignment of chain IDs to use `auth_asym_id`, if it exists, to have a better match to the PDB files. Also, because it is assigned by the _authors_ of the structure. We should store (in segid for example) the `label_asym_id` information because it is always useful to point to.

What would be your opinion (@ofajardo, @peterjc)?

ofajardo:
Hi, 

I am comparing the output of a couple of programs of mine (based on biopython) when parsing pdb files and their mmCif counterpart, and I have already observed this chain renaming. The chain renaming is confusing, because for example the information on the webpage (http://www.rcsb.org/pdb/home/home.do) for a given structure refers to the chains in the pdb file and not in the mmCif file. Also, HETATOMS are manytimes re-assigned to new chains that did not exist in the pdb file (for example there is only one A chain in a structure with 3 Pr HETATOMS in the pdbfile but in the mmcifFile, we have the 3 PR atoms assigned to chains B,C and D, which did not existed before). I like very much your solution of taking the author information if available because it would do my code, and anybody else's code work equivalently for both pdb and mmCif, and having to adapt ourselves for every situation.

peterjc:
I've not done enough with structures recently to give much weight to my comments (I've never used mmCif for actual work). If you can use the author-assigned identifiers that does seem best as it would match the associated publication, and be consistent between PDB and mmCif.

fomightez:
Not much to add either except maybe here cross referencing with the Jmol community could help? The have been dealing with reading mmCIF files the last few months too. There was recently a discussion on how the EBI PDBe is modifying assembly mmCIF files in ways that don't even match regular mmCIF files, see [here](https://sourceforge.net/p/jmol/mailman/message/34803015/).

[...]

Issue #778, opened 2016-03-02
Exception while parsing mmCif files


ofajardo:
I get an error while parsing some mmCif files. Would it be possible to fix it?
I am using the latest version of biopython in github (1.66+), python 2.7.10, both in linux and windows.

When does the error happens?
Just doing this:
MMCIF_parser = MMCIFParser()
MMCIF_structure = MMCIF_parser.get_structure(protein_ID, pdb_file_path)

structures affected: 4Q96 and 4Q94

**Structure: 4Q96**
traceback: Traceback (most recent call last):
MMCIF_structure = MMCIF_parser.get_structure(protein_ID, pdb_file_path)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/MMCIFParser.py", line 65, in get_structure
self._build_structure(structure_id)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/MMCIFParser.py", line 181, in _build_structure
structure_builder.init_residue(resname, hetatm_flag, int_resseq, icode)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/StructureBuilder.py", line 157, in init_residue
% (resname, field, resseq, icode))
PDBConstructionException: Blank altlocs in duplicate residue ARG (' ', 55, ' ')

**Structure 4Q94**
traceback: Traceback (most recent call last):
MMCIF_structure = MMCIF_parser.get_structure(protein_ID, pdb_file_path)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/MMCIFParser.py", line 65, in get_structure
self._build_structure(structure_id)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/MMCIFParser.py", line 181, in _build_structure
structure_builder.init_residue(resname, hetatm_flag, int_resseq, icode)
File "/homebasel/biocomp/fajardoo/python_test_modules/biopython_repo/Bio/PDB/StructureBuilder.py", line 157, in init_residue
% (resname, field, resseq, icode))
PDBConstructionException: Blank altlocs in duplicate residue SER (' ', 2, ' ')

JoaoRodrigues:
@ofajardo Thanks again!

I actually get a different error when parsing both those structures. Which version of Biopython are you using exactly? I took the structures directly from the RCSB PDB (not biological unit).

```python
[...]

PR #800, opened 2016-04-05
Several improvements and fixed to MMCIFParser and related test suite

JoaoRodrigues:
These commits address #779 and #778 and fix a number of other small issues with MMCIF parsing. The biggest change is the usage of `auth_sym_id` instead of `label_asym_id`, which makes the objects parsed with MMCIFParser and PDBParser identical (regularizes the number of chains). The unit tests for MMCIF were also updated, mostly to remove noisy warnings, and a new one was added that compares objects between different format parsers.


peterjc:
I'm happy to ignore the QuantifiedCode complain about no doc-string for the setup method in the unit test.

peterjc:
Thank you!

Issue #797, opened 2016-04-04
Unable to slice a DBSeqRecord


ctSkennerton:
I came across the following bug when trying to slice a `DBSeqRecord`

```
  File "/Users/connor/anaconda/lib/python2.7/site-packages/Bio/SeqRecord.py", line 432, in __getitem__
    description=self.description)
TypeError: __init__() got an unexpected keyword argument 'id'
```

Having a look at the relevant line (below) you can see that it is trying to construct and new record, however the `__init__` signature is different between [`SeqRecord`](https://github.com/biopython/biopython/blob/master/Bio/SeqRecord.py#L155) and [`DBSeqRecord`](https://github.com/biopython/biopython/blob/master/BioSQL/BioSeq.py#L508), which is causing the TypeError.
```
answer = self.__class__(self.seq[index],
                                    id=self.id,
                                    name=self.name,
                                    description=self.description)
```

I'd like your opinion on how to resolve this and I'll be happy to change the code and open a PR. My thought was to enforce that the new record is a `SeqRecord` rather than using `self.__class__` to dynamically select which record class since this new extracted subsequence wouldn't be found in the database. But maybe you think another way would be better?

peterjc:
The ``DBSeq`` class should be able to slice itself and still avoid pulling the full sequence from the database until needed - and I believe that has unit tests.

It looks like an oversight that ``DBSeqRecord`` never got its own ``_getitem__`` implementation, but that seems the way forward.

It should follow the same logic used in the ``SeqRecord`` about what annotations etc should be preserved (many will be dropped), but this might be best handled as a ``SeqRecord`` using a ``DBSeq`` as the ``.seq``? Hmm. This needs some experimentation...

ctSkennerton:
I think keeping as much of the code as possible in `SeqFeature` is probably more manageable in the long run as there would be less duplicated code. I'm not particularly sure on how to do that but we could change the signature of the `__init__` to allow `kwargs` which would alleviate the current error and still allow a `SeqRecord` or `DBSeqRecord` to be generated.

peterjc:
I don't think changing the signature of the ``DBSeqRecord`` ``__init__`` will help - it is not designed to accept an identifier/name/description etc.

[...]

PR #801, opened 2016-04-06
Solution to DBSeqRecord slicing error

ctSkennerton:
Implements a solution to #797 by checking for `DBSeqRecord` and explicitly returning `SeqRecord`


peterjc:
The latest TravisCI had failures related to dependency installation - not from your work. I've requested a rebuild in case that fixes it (I want to be sure Python 2 still works).

I'd probably switch the import from:

```
import BioSQL.BioSeq
if isinstance(self, BioSQL.BioSeq.DBSeqRecord):
```

to:

```
from BioSQL.BioSeq to DBSeqRecord
if isinstance(self, DBSeqRecord):
```

Otherwise that looks good - although I will most likely squash this down to one or two commits when I merge it to keep the main history simple.

ctSkennerton:
I'm going to close this cause I can't for the life of me figure out how to revert changes in git to make that doctest pass, so I just started again on a new branch

Issue #747, opened 2016-01-06
Option to ignore format GenBank specs/warnings


kescobo:
I'm trying to add some information to a collaborators' genbank files by reading in the file, altering certain fields (like `record.annotations['source']` and then re-writing them. Unfortunately, the LOCUS definition is too long and BioPython keeps spitting out warnings, eg:

```
File "/usr/local/lib/python3.5/site-packages/Bio/SeqIO/InsdcIO.py", line 555, in _write_the_first_line
    raise ValueError("Locus identifier %r is too long" % str(locus))
ValueError: Locus identifier 'BW77_ACAGTG.R1_(paired)_contig_28' is too long
```
(that's line 616 in the current `master` branch)

I'd rather not change the name to keep consistent with collaborator's files, and the files are parsed just fine. If I go in and suppress all of the `assert` statements and the `ValueError`, nothing seems to break. I don't want to get rid of these checks completely, but it would be nice to be able to set a flag or something to ignore these assertions in cases where the file doesn't need to meet strict standards.

peterjc:
You probably got a parser wanting when loading these files in Biopython?

The NCBI does seem to be working towards longer LOCUS names, having relaxed the length limit slightly in recent years. Until they change the LOCUS line format though, being column based all the additional fields are a problem if you want to use an overly long identifier. For now breaking the spec is just going to cause trouble with interoperability with other tools.

kescobo:
>You probably got a parser wanting when loading these files in Biopython?

Actually no - they load in just fine (I'm using v1.66 and python3). I don't get any warnings or errors until I go to write the file. If this is a problem, let me know if you'd like the code/file I'm using to diagnose it. 

>The NCBI does seem to be working towards longer LOCUS names

This is actually the output of some annotation software (RAST I think), and it's really long (you can see it above, 33 chars), I doubt it would fit any of the standards. But it seems like there are plenty of non-standard files out there. Obviously supporting all of them would be a nightmare though... 

>Until they change the LOCUS line format though, being column based all the additional fields are a problem if you want to use an overly long identifier. For now breaking the spec is just going to cause trouble with interoperability with other tools.

[...]

PR #802, opened 2016-04-06
Handle GenBank LOCUS names up to 26 letters if space allows

peterjc:
This pull request implements the relaxation discussed on #747 to better handle the special case where a GenBank LOCUS name is over 16 characters (current official limit) by stealing space from the length field, and thus not disrupting the strict column based layout:

* Add a warning to the parser where currently it silently accepts these longer identifiers;

* Modify the writer to accept long identifiers provided the sequence length is short enough to squeeze in both, but issue a warning in this case.

Includes some testing, although not explicitly verifying when the warnings are triggered.

@kblin could you test this please?

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1096,8 +1096,15 @@ class GenBankScanner(InsdcScanner):
             # Should be possible to split them based on position, if
             # a clear definition of the standard exists THAT AGREES with
             # existing files.
-            consumer.locus(name_and_length[0])
-            consumer.size(name_and_length[1])
+            name, length = name_and_length
+            if len(name) > 16:
+                # As long as the sequence is short, can steal its leading spaces
+                # to extend the name over the current 16 character limit.
+                # However, that deserves a warning as it is out of spec.
+                warnings.warn("GenBank LOCUS line identifier over 16 characters",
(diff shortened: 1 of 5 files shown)


peterjc:
Looks like I need to review ``test_SeqIO_write.py`` and ``test_SeqIO.py`` for these changes...

kblin:
Changes look good to me in principle, will try to test this on real data.

peterjc:
Still something not quite right with the BioSQL tests...

peterjc:
I think it was just the variable typo I just fixed in 7e7dd9311176b44bfa0bc4602e3e6983ddf7d37f which was breaking the tests. Spotted easily with fresh eyes, but not in time for Biopython 1.67 sadly.

@kblin could you throw some real data at this?

kblin:
I'll have to dig up the file again, but I'll give this a whirl.

peterjc:
Sorry, force pushed again to fix the last fix. Using perfect string matching in tests is a pain and I'd skipped running the file suite locally. Thank you TravisCI.

kblin:
Works for my test case :+1:

peterjc:
Thank you @kblin - rebased and applied as the commits leading up to and included 52bd2a2af86284d438cbe64ce5ed1051843ebffe

PR #803, opened 2016-04-06
Issue 797

ctSkennerton:
Same changes as #801 minus the change in doctest that I couldn't figure out how to fix

Implements a solution to #797 by checking for `DBSeqRecord` and explicitly returning `SeqRecord`

### DIFF
--- a/Bio/SeqRecord.py
+++ b/Bio/SeqRecord.py
@@ -429,10 +429,17 @@ class SeqRecord(object):
             if self.seq is None:
                 raise ValueError("If the sequence is None, we cannot slice it.")
             parent_length = len(self)
-            answer = self.__class__(self.seq[index],
-                                    id=self.id,
-                                    name=self.name,
-                                    description=self.description)
+            from BioSQL.BioSeq import DBSeqRecord
+            if isinstance(self, DBSeqRecord):
+                answer = SeqRecord(self.seq[index],
+                                        id=self.id,
+                                        name=self.name,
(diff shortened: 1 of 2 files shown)


peterjc:
Alternative solution which does not change the ``SeqRecord`` but duplicates code in ``DBSeqRecord``:
https://github.com/peterjc/biopython/commits/biosql_slicing

What do you think @ctSkennerton? Yours seems a better long term choice as avoids code duplication.

ctSkennerton:
I'm against duplicated code as the number of people that a looking at `BioSQL` is going to be less than for `SeqRecord` so there is a greater chance of things breaking.

Having said that, I think adding `__getitem__` to `BioSQL.BioSeq` would be worth it if there are additional features that make use of the lazy loading from the database. For example instead of iterating over all of the features and then appending those that are inside the slice, a specific SQL query could be used to only return those features from the database, which would boost performance. (This is something I've thought about doing recently)

I see your comment in `Bio.SeqRecord` about keeping the api in sync - can this be enforced automatically using tests?

peterjc:
A more elegant fix would be to extend the ``DBSeqRecord.__init__`` to take start and end parameters (defaulting to everything) and using these for the sequence and feature lookup. But that is more work.

ctSkennerton:
Yeah I think that is a nice approach but would require rewriting a bunch of the SQL statements. I think for now it would be best to keep it simple and not duplicate the code and then down the track think about adding it in.

PR #807, opened 2016-04-12
Added alternative Biopython logo images

widdowquinn:
Added three .png files:

- transparent background
- black background
- an accidental (almost) monochrome, but quite cool, version

and the Photoshop .psd file that generated the first two.


peterjc:
Applied to the website repository instead as https://github.com/biopython/biopython.github.io/commit/342efd672c846f1ebf075e38535ce322ea217f8d - thank you!

PR #809, opened 2016-04-13
Set taxon left right values

ctSkennerton:
This pull request adds the ability to set/update the left and right values of the taxonomy table when a new taxonomy is added in. This is useful when the database has not been created using the `load_ncbi_taxonomy.pl` script or when a new species (or other taxon) is added into a database with an outdated copy of the NCBI taxonomy loaded into it. As a side effect to this I also added in an `executemany` method to the database adaptor as it makes the code for updating the left and right values much easier. I think that executemany is required in the python DBI 2.0 spec so there shouldn't be a problem with the underlying DBI adaptors not implementing it.

### DIFF
--- a/BioSQL/BioSeqDatabase.py
+++ b/BioSQL/BioSeqDatabase.py
@@ -489,6 +489,13 @@ class Adaptor(object):
             sql = sql.replace("%s", "?")
         self.dbutils.execute(self.cursor, sql, args)
 
+    def executemany(self, sql, args):
+        """Just execute an sql command.
+        """
+        if os.name == "java":
+            sql = sql.replace("%s", "?")
+        self.dbutils.executemany(self.cursor, sql, args)
+
     def get_subseq_as_string(self, seqid, start, end):
         length = end - start
         # XXX Check this on MySQL and PostgreSQL. substr should be general,
--- a/BioSQL/BioSeqDatabase.py
+++ b/BioSQL/BioSeqDatabase.py
@@ -311,12 +311,18 @@ class _CursorWrapper(object):
         self.real_cursor = real_cursor
 
     def execute(self, operation, params=None, multi=False):
+        """Execute a sql statement
+        """
         self.real_cursor.execute(operation, params, multi)
 
     def executemany(self, operation, params):
+        """Execute many sql statements
+        """
         self.real_cursor.executemany(operation, params)
 
     def _convert_tuple(self, tuple_):
+        """Decode any bytestrings present in the row
+        """
         tuple_list = list(tuple_)
         for i, elem in enumerate(tuple_list):
             if type(elem) is bytes:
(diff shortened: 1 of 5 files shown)


BioSQL/Loader.py
```diff
@@ -329,6 +329,46 @@ def add_space(letter):
         assert answer == answer.lower()
         return answer
 
+    def _update_left_right_taxon_values(self, left_value):
+        """ update the left and right values in the table
```
peterjc: Remove leading space on docstring?

BioSQL/DBUtils.py
```diff
@@ -40,6 +40,11 @@ def execute(self, cursor, sql, args=None):
         """
         cursor.execute(sql, args or ())
 
+    def executemany(self, cursor, sql, seq):
+        """
```
peterjc: No docstring?

peterjc:
This looks sensible, not sure who else could review it - @hlapp any comments?

ctSkennerton:
updated the docstrings 😄

hlapp:
I'm not sure I can comment competently. The enumeration implementation seems to work a little differently from the default, so for example I don't know where the increment by +2 is coming from, and other little details. In theory well written tests should show that it's correct though 😃

ctSkennerton:
I believe that all of the tests now pass, however it took a number of additional changes:
1. I had to remove an assertion that had compatibility problems between python 3 and python 2 as it was checking for the long type.
2. Provide an `executemany` method to the `_CursorWrapper` class, which is specific to the mysql-connector 
3. Change one of the SQL statements in the test file that was causing problems for Postgres

peterjc:
One TravisCI build failed due to a dependency issue, rerunning it now.

QuantifiedCode looks OK, other than on ``execturemany`` method still missing a docstring.

BioSQL/BioSeqDatabase.py
```diff
@@ -313,6 +313,9 @@ def __init__(self, real_cursor):
     def execute(self, operation, params=None, multi=False):
         self.real_cursor.execute(operation, params, multi)
 
+    def executemany(self, operation, params):
+        self.real_cursor.executemany(operation, params)
```
peterjc: Missing docstring e.g.

```Execute many SQL commands.```

File: BioSQL/BioSeqDatabase.py [hunk 4]
```diff
@@ -489,6 +492,13 @@ def execute(self, sql, args=None):
             sql = sql.replace("%s", "?")
         self.dbutils.execute(self.cursor, sql, args)
 
+    def executemany(self, sql, args):
+        """Just execute an sql command.
```
peterjc: Many commands? e.g.

```Execute many SQL commands.```

ctSkennerton:
fixed up the docstrings as requested!

peterjc:
Thanks @ctSkennerton, and @hlapp for your input. I've applied this as a squashed commit which also fixed the PEP8 whitespace complaints.

Issue #363, opened 2014-09-12
Genbank LOCUS "circular" topology and molecule type ignored


pgfoster:
I first noticed this when reading in a genbank file, writing it out again, and then looking at the diff.  All was well, except that the word "circular" from the LOCUS line had disappeared.

This is what I think is going on.  When reading genbank files that have "circular" in the first line, eg
LOCUS       NC_014275              15386 bp    DNA     circular INV 01-JUL-2010
using SeqIO we eventually get to 
    `Bio/GenBank/__init__.py`
and the `_FeatureConsumer.residue_type(self, type)` method.
At that point, `type` contains the string "DNA     circular".   The method contains only one line ---
    `self._seq_type = type.strip()`  
However, putting it there makes it get lost.  
I suspect the info about whether it is circular should be put in `self.data`, eg in the `self.data.annotations` dictionary.  The reason is because the `Bio.GenBank.Scanner.InsdcScanner.parse()` method returns `consumer.data` (ie `self.data` above), and so does not `return consumer._seq_type` containing the circular topology info.

kblin:
I almost have a patch for it, just got some of the padding wrong.

peterjc:
Reposting these from the old BugZilla/Redmine tracker regarding how ``is_circular`` is/should be stored in BioSQL, see these old threads:

http://lists.open-bio.org/pipermail/biosql-l/2011-July/001774.html
http://lists.open-bio.org/pipermail/bioperl-l/2011-July/035433.html

kblin:
Wow, eerie. A colleague of mine also just talked to be about this bug. So what would be needed to progress on this? Change the patch to work with a global SeqRecord feature `is_circular` like BioPerl?

peterjc:
The pragmatic short term solution is just add some code in our BioSQL wrapper to ignore the new annotation. Real solution requires mapping to however BioPerl etc all have stored the circular attribute in the database schema - I filed https://github.com/biosql/biosql/issues/5

kblin:
Not sure if I want to follow BioPerl here, though. BioPerl basically ignores whatever the chromosome topology is, unless it's circular.

kblin:
With a binary flag, that's your only choice, of course. But personally I'd prefer if
```python
from Bio import SeqIO
recs = SeqIO.parse(infile, 'genbank')
SeqIO.write(recs, outfile, 'genbank')
```
would leave me with an `outfile` that is identical to my `infile`.

peterjc:
[...]

PR #812, opened 2016-04-21
GenBank: Don't discard 'linear' or 'circular' topology from LOCUS line

kblin:
Just ran into issue #363 myself, so here's a fix.

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -604,6 +604,19 @@ class GenBankWriter(_InsdcWriter):
         assert len(division) == 3
         return division
 
+    def _get_topology(self, record):
+        """Set the topology to 'circular', 'linear' if defined"""
+        max_topology_len = len('circular')
+
+        # return an empty placeholder string if not given
+        if 'topology' not in record.annotations or \
+            record.annotations['topology'] == 'unspecified':
+            return ' ' * max_topology_len
+
+        template = '%%-%ds' % max_topology_len
+
+        return template % record.annotations['topology'][:max_topology_len]
+
     def _write_the_first_line(self, record):
         """Write the LOCUS line."""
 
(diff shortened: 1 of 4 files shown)


kblin:
Ah, durn, BioSQL. I'll have a look.

Bio/SeqIO/InsdcIO.py
```diff
@@ -604,6 +604,18 @@ def _get_data_division(self, record):
...
 
+    def _get_topology(self, record):
+        """Set the topology to 'circular', 'linear' if defined"""
+        max_topology_len = len('circular')
+
+        # return an empty placeholder string if not given
+        if 'topology' not in record.annotations:
+            return ' ' * max_topology_len
+
+        template = '%%-%ds' % max_topology_len
```
peterjc: Would the string justification method be easier? i.e. ``.ljust(max_topology_len)``

Bio/SeqIO/InsdcIO.py
```diff
@@ -647,17 +659,20 @@ def _write_the_first_line(self, record):
...
+        topology = self._get_topology(record)
+
         division = self._get_data_division(record)
 
         assert len(units) == 2
         assert len(division) == 3
         # TODO - date
         # TODO - mol_type
-        line = "LOCUS       %s %s %s    %s           %s %s\n" \
+        line = "LOCUS       %s %s %s    %s  %s %s %s\n" \
```
peterjc: Why the double space before the topology? Which columns does the spec say are allocated here?

peterjc:
I was wondering about tackling this for GenBank/EMBL input/output myself - I have some EMBL tweaking to do for work for an ENA submission, so should be in a position to actively test this next week :)

It was the BioSQL stuff that has held this back - really need to check where BioPerl (as the defacto BioSQL reference implementation) puts the topology field, but a short term hack would be to ignore the topology in the ``SeqRecord`` when writing to the database?

Also do we need ``melanin.gbk`` - lots of the existing ``Tests/GenBank/*.gb`` have useful ``LOCUS`` lines.

Bio/SeqIO/InsdcIO.py
kblin: Possibly. I can give that a spin.

Bio/SeqIO/InsdcIO.py
kblin: That's how I could make it align to the LOCUS line from the GenBank headers I looked at. I didn't find the exact definition for this in the spec. Then again, the spec is very confusing, which is the reason why we have that much fun with that format in the first place.

kblin: Arguably, I could increase the `ljust()` for the molecule type by 1 as well instead.

kblin:
I agree on the `melanin.gbk`. I had that lying around in my Tests/GenBank directory, so I just used it, and then later realized that it was missing. I'll refactor that to use another record.

Bio/SeqIO/InsdcIO.py
peterjc: That sounds sensible.

peterjc:
Re BioSQL, as per my comments on #812, I have filed https://github.com/biosql/biosql/issues/5

peterjc:
Rebased and with changes discussed above as https://github.com/peterjc/biopython/tree/topology - I hope to get the BioSQL side passing our tests next.

peterjc:
Work in progress at https://github.com/biopython/biopython/pull/903 which includes the initial work from here by @kblin

Issue #367, opened 2014-10-07
Partial codons in SearchIO exonerate parser


pschwede:
I'm trying to convert [exonerate](http://www.ebi.ac.uk/~guy/exonerate/) output to psl using a simple script:

``` python
from Bio import SearchIO
QRESULTS = SearchIO.parse("my_exonerate.out", 'exonerate-text')
SearchIO.write(QRESULTS, "my_psl.out, 'blat-psl')
```

The error message depends on the output format. In the case above, it reads

``` plain
Traceback (most recent call last):
  File "./vulgar2psl.py", line 14, in <module>
    SearchIO.write(QRESULTS, ARGS.outfile, 'blat-psl')
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/__init__.py", line 620, in write
    writer.write_file(qresults)
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/BlatIO.py", line 602, in write_file
    for qresult in qresults:
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/__init__.py", line 316, in parse
    for qresult in generator:
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/ExonerateIO/_base.py", line 236, in __iter__
    for qresult in self._parse_qresult():
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/ExonerateIO/_base.py", line 364, in _parse_qresult
    hsp = _create_hsp(prev_hid, prev_qid, prev['hsp'])
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/ExonerateIO/_base.py", line 188, in _create_hsp
    frags = _adjust_aa_seq(frags)
  File "/usr/lib/python2.7/dist-packages/Bio/SearchIO/ExonerateIO/_base.py", line 43, in _adjust_aa_seq
    assert len(frag) % 3 == 0
AssertionError
```

I get other assertion errors with "exonerate-vulgar" and "exonerate-cigar". In none of the cases, the script finishes successfully.

Is this a bug? Am I doing something wrong? Are there any restrictions in using the exonerate parameters?

**edit:**

Here is the exonerate result for now: https://gist.githubusercontent.com/pschwede/df803361b7f6647183c0/raw/6529e84ba6c13a97a50f6a9712ce34f6e8cee28a/Alignments%20Sample

``` plain
$ exonerate --version
exonerate from exonerate version 2.2.0
Using glib version 2.31.12
[...]

PR #815, opened 2016-04-26
Partial codons for SearchIO when parsing Exonerate

bow:
This was reported first in #367 some time ago by @pschwede. Based on the discussion and a suggestion by @zbwrnz, this pull request was made. It treats the partial codons (on each of the HSP fragment) as the amino acid `X`. The actual codon sequence that is split is still kept in the fragment's `.aln_annotation_all` attribute. Additionally, there is now also a `phase` attribute for HSP fragments where this adjustment is made.

I did not add any new exonerate outputs since this was a known issue when the Exonerate parser was written (which i realized as I wrote the pull request). So there was already an unused test case in our test suite (the `exn_22_m_protein2genome.exn` file) where we observe Exonerate splitting a codon. This file was never actually tested, but now that we can handle such scenarios, I also added some tests based on that file.

### DIFF
--- a/Bio/SearchIO/ExonerateIO/_base.py
+++ b/Bio/SearchIO/ExonerateIO/_base.py
@@ -59,8 +59,24 @@ def _make_triplets(seq, phase=0):
 
 
 def _get_fragments_coord(frags):
-    """Given a list of three-letter amino acid fragments, return each
-    fragment's letter position in the full, flattened sequence."""
+    """Returns the letter coordinate of the given list of fragments.
+
+    This function takes a list of three-letter amino acid sequences and
+    returns a list of coordinates for each fragment had all the input
+    sequences been flattened.
+
+    This is an internal private function and is meant for parsing Exonerate's
+    three-letter amino acid output.
+
+    >>> from Bio.SearchIO.ExonerateIO._base import _get_fragments_coord
+    >>> _get_fragments_coord(['Thr', 'Ser', 'Ala'])
+    [0, 3, 6]
+    >>> _get_fragments_coord(['Thr', 'SerAlaPro', 'GlyLeu'])
+    [0, 3, 12, ]
+    >>> _get_fragments_coord(['Thr', 'SerAlaPro', 'GlyLeu', 'Cys'])
+    [0, 3, 12, 18]
+
+    """
     if not frags:
         return []
     # first fragment always starts from position 0
--- a/Bio/SearchIO/ExonerateIO/_base.py
+++ b/Bio/SearchIO/ExonerateIO/_base.py
@@ -30,9 +30,25 @@ def _set_frame(frag):
 
 
 def _make_triplets(seq, phase=0):
-    """Given a sequence of three-letter amino acid codes and its phase,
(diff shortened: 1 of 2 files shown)


Bio/SearchIO/ExonerateIO/_base.py
```diff
@@ -28,40 +29,80 @@ def _set_frame(frag):
     frag.query_frame = (frag.query_start % 3 + 1) * frag.query_strand
 
 
-def _make_triplets(seq):
-    """Splits a string into a list containing triplets of the original
-    string."""
-    return [seq[3 * i:3 * (i + 1)] for i in range(len(seq) // 3)]
+def _make_triplets(seq, phase=0):
+    """Given a sequence of three-letter amino acid codes and its phase,
```
peterjc: This ought to have a one line description for PEP257 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings.

e.g.

``` python
def _make_triplets(seq, phase=0):
    """Split sequence into partial, complete, and partial codons (PRIVATE).

    Given a sequence of three-letter amino acid codes and its phase,
    return the letters discarded to reach the first full triplet, the
    triplets, and any remaining non-triplets.
    """

File: Bio/SearchIO/ExonerateIO/_base.py [hunk 2]
```diff
@@ -28,40 +29,80 @@ def _set_frame(frag):
...
+    np_seq = seq[phase:]
+    non_triplets = len(np_seq) % 3
+    post = "" if not non_triplets else np_seq[-1 * non_triplets:]
+    intacts = [np_seq[3 * i:3 * (i + 1)]
+               for i in range(len(np_seq) // 3)]
+    return pre, intacts, post
+
+
+def _get_fragments_coord(frags):
+    """Given a list of three-letter amino acid fragments, return each
```
peterjc: Again, this ought to have a one line description then body as per PEP257

peterjc:
I noted some minor docstring formatting issues, but otherwise is there any reason not to merge this?

bow:
Thanks for checking, Peter :). I updated the docstrings with another two commits. I think this is ready for merging.

peterjc:
There's a trivial PEP8 issue which TravisCI now checks:

```
Tests/test_SearchIO_exonerate.py:1498:5: E303 too many blank lines (2)
```

peterjc:
Also did my suggestion for the ``_get_fragments_phase`` docstring make sense?

bow:
Ah yes, sorry for missing the `_get_fragments_phase` docstring ~ it's added now. I've also fixed the pep8 style error.

On a side note, would it be an idea to add a `setup.cfg` for these PEP8 error settings? We could have one each in `Bio`, `Tests` (since they have slightly different settings).

peterjc:
PEP8 setup logged as #883. Sure, please explore these options. Right now the TravisCI/Tox settings differ as some parts of the code base are more conforming than others, and I wanted to be as strict as possible.

bow:
Thank you :) ~ I will be closing the issue linked to this PR and also have a look at #883.

PR #818, opened 2016-05-04
Fix in UniProt.GOA

idoerg:
Fixed a bug in GOA GAF parser. The parser defaulted to GAF 1.0 if it didn't recognizes the GAF version. Since there are now 3 GAF versions (2.1, 2.0 and historically 1.0), the code makes an explicit choice when parsing the gene associate file header and raises an exception if the GAF version is not recognized.

### DIFF
--- a/Bio/UniProt/GOA.py
+++ b/Bio/UniProt/GOA.py
@@ -179,7 +179,7 @@ def gpi_iterator(handle):
         sys.stderr.write("gpi 2.1\n")
         return _gpi20iterator(handle)
     else:
-        raise ValueError('Unknown GPI version %s\n' % inline)
+        raise ValueError('Unknown GPI version {0}\n'.format(inline))
 
 
 def _gpa10iterator(handle):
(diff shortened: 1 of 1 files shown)


peterjc:
QuantifiedCode is only complaining about string formatting style, which I think is harmless.

idoerg:
Sorry, I'll fix that.

On Wed, May 4, 2016 at 2:18 PM, Peter Cock <notifications@github.com> wrote:

> QuantifiedCode is only complaining about string formatting style, which I
> think is harmless.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/biopython/biopython/pull/818#issuecomment-216972071>
>



-- 
Iddo Friedberg
http://iddo-friedberg.net/contact.html
++++++++++[>+++>++++++>++++++++>++++++++++>+++++++++++<<<<<-]>>>>++++.>
++++++..----.<<<<++++++++++++++++++++++++++++.-----------..>>>+.-----.
.>-.<<<<--.>>>++.>+++.<+++.----.-.<++++++++++++++++++.>+.>.<++.<<<+.>>
>>----.<--.>++++++.<<<<------------------------------------.

peterjc:
Can you add a test case? i.e. One or two test files using the other version numbers?

idoerg:
I'd like to revisit this.. I see that this hasn't been added yet. UiProt can't handle GAF2.1 at the moment.

peterjc:
Applied as a squashed commit https://github.com/biopython/biopython/commit/409ebea045eba6f596dfcbe350e2707974d74a23 - although a few more tests would still be welcome.

Thanks Iddo, good to have your name against a release again 👍

PR #822, opened 2016-05-11
Testing for missing dependencies and returning an error

tiagoantao:
This will allow the test system to fail if there are missing dependencies.

I had a look at the dependencies and none that were proprietary were found on the test system. I was pretty sure that those existed, but maybe I was wrong....


tiagoantao:
ping @peterjc

peterjc:
I don't care about the minor issues noted by QuantifiedCode here (we should look at how to configure this to be less fussy).

tiagoantao:
@peterjc Should I merge this? Or do you want something changed? If you can think of a proprietary dependency, tell me

peterjc:
None of the tools are using the new exception yet, are they? I'd like to try this on some of our buildbot machines (etc) to see what it is missing - and ideally enable this setting on the buildslaves. I'm wondering if things like Windows-only and Linux-only dependencies would render the new mode actually impossible to pass?

This would also need a mention in the testing chapter of the tutorial.

tiagoantao:
No, none. This is because all the tools I have checked are free. I must be missing something, because I remember at least one being proprietary.

I like the idea of sending this to buildbot. If there are OS-specific extensions, what would be the correct thing to do? In my mind it is not clear.

For example. we might add an exception list to run_tests.py this would be obvious...

..but we could also make the individual tests skip if they are on the wrong operating system. I think this makes more sense, because it is not a test failure, but a lack of tool in the OS...

peterjc:
Some of the tests use command line tools which are Linux/Mac only, and currently just raise ``MissingExternalDependencyError`` which currently lets them be skipped on Windows (or vice versa), e.g. ``test_TCoffee_tool.py``

This becomes a problem with ``--check-dependencies`` or ``--check-free-dependencies``.

Maybe we need something more subtle like ``ExternalDependencyUnavailableError`` (suggestions for clearer naming welcome) for this cross-platform kind of missing dependency where its missing because it doesn't exist on the current OS? For these ``run_tests.py`` should skip the tests, even if the check dependencies options are used?

tiagoantao:
@peterjc Should I merge or not? In order to see buildbot behavior....

peterjc:
Hold off the merging? I can log into our linux64 and win32 buildbots and try it by hand.

tiagoantao:
OK, on a related issue: What should we do with tests that are not supposed to happen? I mean tests of applications that do not exist on a specific operating system? I suggest changing each test to just return OK if the OS is not supported by the dependency...

peterjc:
That's what I was suggesting ``ExternalDependencyUnavailableError`` for.

tiagoantao:
Sorry, I skipped that comment. I will change the PR to reflect that.

tiagoantao:
On the name front, I would recommend `ExternalDependencyUnavailableException` as this is not an Error in any way (of our part)

tiagoantao:
@peterjc I have added `ExternalDependencyUnavailableException` and put a case with TCoffee

peterjc:
Good idea, ``ExternalDependencyUnavailableException`` is better than `` ExternalDependencyUnavailableError``.

peterjc:
Hi Tiago, I was going to suggest rebasing this to check it looks OK with changes to the master.

However I just noticed the proposed deprecations - can I cherry-pick that straight to the master (with a tweak to the Biopython version numbers to reflect the next release)?

peterjc:
Never mind, the deprecation did make it to the master - see #1139 for actually dropping those modules.

PR #823, opened 2016-05-14
Split biosql tests

ctSkennerton:
As discussed in #819, I've split out the `TaxonomyTest` class which requires the internet into a separate file.

PR #826, opened 2016-05-19
BioSQL test configuration

peterjc:
This moves the BioSQL database settings used in the test suite into a new INI file, currently kept in ``test_BioSQL_*.py`` files themselves - which has a number of drawbacks (see e.g. refactoring in #823).

Under each buildslave for buildbot we can manually add a locally appropriate ``Tests/biosql.ini`` file (which is NOT under git version control).

Under TravisCI we setup suitable values for all the BioSQL tests to pass (via ``Tests/biosql.ini.sample`` which is under git version control).

@tiagoantao and @ctSkennerton could you look over this please?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,9 @@ Tests/Graphics/*.eps
 Tests/Graphics/*.svg
 Tests/Graphics/*.png
 
+#Ignore the local BioSQL test settings:
+Tests/biosql.ini
+
 #TODO - The unit tests shouldn't leave temp files after running:
 Tests/BioSQL/temp_sqlite.db
 Tests/BioSQL/temp_sqlite.db-journal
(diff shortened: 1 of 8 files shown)


ctSkennerton:
I like where this is heading, definitely a better approach to write the configuration into a separate file. The variables are still global though, right? Do you think there is a way to implement the tests without them being global? You might run into the same problem that I was having in #823 where the global configuration values weren't being visible by multiple modules when I tried to split things into `common_BioSQL` and `common_BioSQL_online`. But maybe I was just importing the functions wrong and you won't run into this problem.

peterjc:
Thanks for your feedback - I'll merge this as an incremental improvement, and then take a look at applying your initial changes from #823. I'm pretty sure that can still be done with globals, but agree it isn't the cleanest way of doing this.

peterjc:
Rebased and applied as the commits leading up to https://github.com/biopython/biopython/commit/7dc359f38cc36ab04aabd46de0d170b69a06a799

PR #755, opened 2016-01-20
Add the phenotype module

mgalardini:
Sorry, to have a cleaner history and avoid any mistake I preferred doing a new PR.

PR #828, opened 2016-05-19
Bio.phenomics

peterjc:
Based on #755 from @mgalardini


NEWS
```diff
@@ -22,6 +22,10 @@ comparing all of the attributes would be too complex. As of this release
 attempting to compare SeqRecord objects should raise an exception instead. If
 you want the old behaviour, use id(record1) == id(record2) instead.
 
+New experimental module Bio.phenotype is for working with Phenotype Microarray
+plates in JSON and the machine vendor's CVS format (contributed by Marco
```
mgalardini: It's actually CSV format :)

mgalardini:
As far as I can see all the changes look good (except a typo in the NEWS file); thank you very much for your help on this.
Please let me know if I can do any more work on this.

Marco

NEWS
peterjc: Well spotted, CVS and CSV are rather different things :)

peterjc:
Can you easily speed up the tests? e.g. using a smaller dataset?

Also the test coverage for ``pm_fitting.py`` as measured by CodeCov is very low.

mgalardini:
Ah yes, absolutely, will try to have a smaller input file.

I think that CodeCov is not covering pm_fitting because it lacks scipy; on line 224 of test_phenotype we try to import it before testing it. Is there anything that can be done there?

peterjc:
Superseded by #830 with faster tests.

peterjc:
Regarding the SciPy tests, we could split that out into a separate ``test_XXX.py`` file, making the fact it gets skipped if NumPy or SciPy is missing more obvious.

Installing SciPy under TravisCI had previously been too slow to justify, but doing so for at least one version of Python would be the only way to improve this for CodeCov.

Issue #832, opened 2016-05-24
KEGG REST API return types in Python 3


widdowquinn:
Python 3 makes a distinction between string types (`String` and `Byte`) that was not made in Python 2:

```
Python 2.7.11 (default, Mar 22 2016, 08:29:04) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'A' == b'A'
True
```

```
Python 3.5.1 (default, Dec 26 2015, 18:11:22) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'A' == b'A'
False
```

The `Bio.KEGG.REST` functions return Byte types, rather than String types:

```
>>> from Bio.KEGG.REST import *
>>> type(kegg_info("kegg").read())
<class 'bytes'>
>>> kegg_info("eco").read()
b'T00007           Escherichia coli K-12 MG1655 KEGG Genes Database\neco              Release 78.0+/05-21, May 16\n                 Kanehisa Laboratories\n                 4,497 entries\n'
```

These give unintuitive results with `print()`:

```
>>> print(kegg_info("eco").read())
b'T00007           Escherichia coli K-12 MG1655 KEGG Genes Database\neco              Release 78.0+/05-21, May 16\n                 Kanehisa Laboratories\n                 4,497 entries\n'
```

and when attempting to parse sequence data returned from KEGG with `Bio.SeqIO`:

```
>>> from Bio import SeqIO
>>> print(kegg_get("ece:Z5100", "ntseq").read())
[...]

PR #833, opened 2016-05-24
Some KEGG REST queries now return String not Byte in Python3

widdowquinn:
See issue #832.

Python 3 makes a distinction between string types (String and
Byte) that was not made in Python 2. This gave unintuitive results
with print() and SeqIO.read() in Python3. This short fix returns
String rather than Byte handles (as with Bio.Entrez), for all
requests through the KEGG REST API, other than those for image
data, which is still returned as a Byte handle.

### DIFF
--- a/Bio/KEGG/REST.py
+++ b/Bio/KEGG/REST.py
@@ -29,6 +29,8 @@ Nucleic Acids Res. 28, 29-34 (2000).
 
 from Bio._py3k import urlopen as _urlopen
 
+from Bio._py3k import _binary_to_string_handle
+
 __docformat__ = "restructuredtext en"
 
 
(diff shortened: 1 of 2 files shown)


peterjc:
I presume the TravisCI results won't be any help here as we're not using it for the online tests...

widdowquinn:
You're right - I wouldn't have thought they'd be any help.

I tested manually using this notebook: https://github.com/widdowquinn/Notebooks-Bioinformatics/blob/master/Biopython_KGML_intro.ipynb, which gave identical output for me with Python 2 both with and without this change relative to v1.66+, and with Python 3/v1.66+ only after the change.

peterjc:
Can you add something to ``test_KEGG_online.py`` which looks like it hardly parses anything at all right now?

widdowquinn:
The new commit 8c10a24 adds `Bio.SeqIO` parsing to the `aaseq` and `ntseq` tests in `test_KEGG_online.py` - these should fail if a `Byte` handle is returned by `kegg_get()`, and succeed if a `String` handle is returned.

peterjc:
Cherry-picked with minor merge conflict dealt with, and mentioned in the NEWS file - see 24d6ff768748900168bd47cac44a2603a539e466 and the commits preceding it. Thank you!

PR #835, opened 2016-05-26
Added FastMMCIFParser class.

JoaoRodrigues:
@peterjc: this is a repetition of #580, updated to include the recent changes to mmCIF parsers and organized a bit better (the fast parser is in the same module as the regular mmCIF parser). I added tests to the test_mmCIF suite, essentially duplicating some lines or comparing to the original parsed objects.


peterjc:
Merged, thank you.

PR #841, opened 2016-06-04
Small changes to PDB/__init__.py to allow direct import of FastMMCIFParser

JoaoRodrigues:
* Removed try/except from old mmCif parser
* Added explicit import of FastMMCIFParser

Only noticed this now. In my previous commit (and tests) there were some leftover files from a previous build that made me think everything was fine, when it was actually quite complicated to import the new parser..


peterjc:
Well spotted - that try/except should have been removed when we dropped the C dependency for the MMCIF parser.

PR #843, opened 2016-06-06
Setup TravisCI build via matrix

peterjc:
This replaces lots of complicated if statement logic with per-build target environment variables.

I think this is much clearer, but would like a second opinion before applying this to the master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,28 @@
 # Note when testing Python 3, the 'python' command will invoke Python 3
 # and similarly for PyPy too.
 
+# Environment variables setup via the matrix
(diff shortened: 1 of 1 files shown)


cbrueffer:
I like it. What about the PY3 variable though, does the matrix code take care of that automagically from the configured python version?

peterjc:
TravisCI will ensure than "python" and "pip" match the version of Python being tested, including python3, pypy and pypy3.

After the change, there is no ``$PY3`` variable - that was just part of our if-based system.

cbrueffer:
Cool, please commit :-)

peterjc:
Done. Thank you for your comments @cbrueffer

PR #854, opened 2016-06-15
Add mmtf support for biopython

abradle:
MMTF is a new binary file format designed and supported by the RCSB PDB:
http://mmtf.rcsb.org/

This patch adds MMTF parsing support to Biopython via the mmtf-python library

1) Added parser for the data into BioPython structure
2) Added tests and test data for comparison to mmCIF
3) Include mmtf-python in travis.yml (I'm not clear this will work first time).


peterjc:
Thanks @abradle - I look forward to you tweaking this to get the tests passing, and then resubmitting as per our earlier discussion.

Regarding the current TravisCI failure, I think you'd need to add the new module in ``setup.py`` (or perhaps use ``Bio/PDB/MMTF.py`` instead)?

```
ERROR: test_MMTF
  File "run_tests.py", line 420, in runTest
  File "/opt/python/2.6.9/lib/python2.6/unittest.py", line 576, in loadTestsFromName
  File "/home/travis/build/biopython/biopython/Tests/test_MMTF.py", line 2, in <module>
    from Bio.PDB.MMTF import MMTFParser
ImportError: No module named MMTF
```

As an aside, I would lean towards ``mmtf`` (lower case) for the module name given PEP8 guidance, but we have a legacy for mixed and upper cases names under ``Bio.PDB`` so this isn't critical.

abradle:
Cool - I've altered these.

peterjc:
You'll need to rename the files to use lower case ``mmtf`` as well, e.g. ``git mv Tests/test_MMTF.py Tests/test_mmtf.py`` and ``git mv Bio/PDB/MMTF/__init__.py Bio/PDB/mmtf/__init__.py`` should work (although if your Mac has the default case insensitive filesystem that could make life a little harder...)

abradle:
Silly mac case insensitivity on that refactor...

peterjc:
Don't worry about the git history - I'll squash this into one commit once its ready (and having learnt the hard way that ``git rebase`` does not work properly on a case insensitive file system, I intend do that on Linux).

peterjc:
The TravisCI tests passed, although there is a warning:

```
PDBConstructionWarning: WARNING: Chain A is discontinuous at line 0.
```

If that is harmless, you can silence it with something like this: 7c0a50e9b5ad9662870ea8b35450e35e4ed8b454

peterjc:
Something QuantifiedCode picked up on is not using new-style classes, e.g. use

```
class StructureDecoder(object):
```

not:

```
class StructureDecoder():
```

See https://www.quantifiedcode.com/app/issue_class/6VzbBIEp

abradle:
Cool - I've made some changes to comply with quantified code. Quite like that actually (have suggested we try it on Biojava).

JoaoRodrigues:
Hi there,

Awesome, thanks for adding this! However, I am a bit skeptical about having another dependency.. What is the drawback of including the mmtf-python core code in Biopython?

abradle:
Hi Joao,

Thanks! 

So what we want to provide with the mmtf-python project is a clean API to the data.

The benefit of having it separate to the BioPython code is that we can update the format or speed up the parser - but BioPython code doesn't have to change at all. We look after the mmtf-python code.

If you include the core code in Biopython - you guys would then need to maintain that code, keep track of any format updates or changes in decoder implementation. Also mmtf-python also depends on msgpack-python - so you would still need to add that dependency for it to work.

I'd be happy to chat over Skype about this if you'd like?

JoaoRodrigues:
Hi Anthony,

Perfect justification. I also went over the code and indeed, it's way more
complex than I thought when I first looked at it. Makes perfect sense to
keep it like this then. Thanks again for integrating it here!

Cheers





2016-06-16 8:17 GMT-07:00 Anthony Bradley <notifications@github.com>:

abradle:
Great! Thanks.

Tests/test_MMTF.py
```diff
@@ -1,64 +1,78 @@
...
-                             mmCIF_atom.full_id)  # (structure id, model id, chain id, residue id, atom id)
-            self.assertEqual(mmtf_atom.id, mmCIF_atom.name)  # id of atom is the atom name (e.g. "CA")
-            # self.assertEqual(mmtf_atom.serial_number,mmCIF_atom.serial_number) # mmCIF serial number is none
\ No newline at end of file
+                             mmcif_atom.full_id)  # (structure id, model id, chain id, residue id, atom id)
+            self.assertEqual(mmtf_atom.id, mmcif_atom.name)  # id of atom is the atom name (e.g. "CA")
+            # self.assertEqual(mmtf_atom.serial_number,mmcif_atom.serial_number) # mmCIF serial number is none
+
+
+def test_parser():
```
peterjc: Does this do anything? Its not a class so won't unittest ignore it?

peterjc:
@JoaoRodrigues are you happy with the implementation as is? If so I can take care of the merge (and making sure it plays nicely with our test framework, e.g. when missing the dependency).

I'm also comfortable with the proposed soft dependency (needed at run time only).

JoaoRodrigues:
Looks perfect for me, didn't test it myself but if all tests pass it should be fine. Super job @abradle @peterjc !

Tests/test_MMTF.py
```diff
@@ -0,0 +1,64 @@
...
+            self.assertAlmostEqual(mmtf_atom.coord[0], mmCIF_atom.coord[0], places=3)
+            self.assertAlmostEqual(mmtf_atom.coord[1], mmCIF_atom.coord[1], places=3)
+            self.assertAlmostEqual(mmtf_atom.coord[2], mmCIF_atom.coord[2], places=3)
+            self.assertEqual(mmtf_atom.bfactor, mmCIF_atom.bfactor)
+            self.assertEqual(mmtf_atom.occupancy, mmCIF_atom.occupancy)
+            self.assertEqual(mmtf_atom.altloc, mmCIF_atom.altloc)
+            self.assertEqual(mmtf_atom.full_id,
+                             mmCIF_atom.full_id)  # (structure id, model id, chain id, residue id, atom id)
+            self.assertEqual(mmtf_atom.id, mmCIF_atom.name)  # id of atom is the atom name (e.g. "CA")
+            # self.assertEqual(mmtf_atom.serial_number,mmCIF_atom.serial_number) # mmCIF serial number is none
```
peterjc: Missing final carriage return

Bio/PDB/MMTF/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+def get_from_decoded(decoder):
+    structure_decoder = StructureDecoder()
+    decoder.pass_data_on(structure_decoder)
+    return structure_decoder.structure_bulder.get_structure()
+
+class MMTFParser():
+
+    def get_structure_from_url(self, pdb_id):
+        """Get a structure from a URL - given a PDB id
```
peterjc: This is not a PEP257 style docstring - should be a single line, blank line, etc.

https://www.python.org/dev/peps/pep-0257/

I've not double checked this renders nicely as RST (reStructuredText) for our API docs.

peterjc:
@etal, @JoaoRodrigues, @lennax - Do you have any comments on this (assuming it will be a squash commit to deal with the messy file renaming)?

e.g. Does the ``MMTFParser`` class structure in make sense given the rest of ``Bio.PDB``?

Bio/PDB/MMTF/__init__.py
```diff
@@ -0,0 +1,24 @@
+from mmtf import fetch,parse
```
peterjc: There should be a space after the comma for PEP8 conformance (see also 

However, I would replace this with something like this:

```
try:
   from mmtf import fetch, parse
except ImportError:
   raise MissingPythonDependencyError("Install mmtf to use Bio.PDB.mmtf "
                                      "(e.g. pip install python-mmtf)")
```

This serves two goals - first some user guidance when the import fails (this new exception subclasses ``ImportError`` so is still easy to catch), and second it will let ``run_tests.py`` skip this smoothly if the dependency is missing.

peterjc:
@abradle you'd also need to rename ``test_MMTF.py``, this will work on a Mac:

```
cd Tests/
git mv test_MMTF.py x.py
git mv x.py test_mmtf.py
git commit -m "Renaming test to match case of Bio.PDB.mmtf"
```

Otherwise you'll get:

```
$ git mv test_MMTF.py test_mmtf.py
fatal: destination exists, source=Tests/test_MMTF.py, destination=Tests/test_mmtf.py
```

Bio/PDB/mmtf/DefaultParser.py
```diff
@@ -0,0 +1,156 @@
+from Bio.PDB.StructureBuilder import StructureBuilder
+
+class StructureDecoder(object):
+    """
```
peterjc: Minor, but PEP257 would put that on one line as:

```
"""Class to pass the data from mmtf-python into a BioPython data structure."""
```

See https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

Bio/PDB/mmtf/DefaultParser.py
```diff
@@ -0,0 +1,156 @@
...
+	"""
+        # A Bradley - chose to use chain_name (auth_id) as it complies with current BioPython. Chain_id might be better.
+        self.structure_bulder.init_chain(chain_id=chain_name)
+        if self.chain_index_to_type_map[self.chain_counter] == "polymer":
+            self.this_type = ""
+        elif self.chain_index_to_type_map[self.chain_counter] == "non-polymer":
+            self.this_type = "H"
+        elif self.chain_index_to_type_map[self.chain_counter] == "water":
+            self.this_type = "W"
+        self.chain_counter+=1
```
peterjc: Again minor, but PEP8 would use more spaces:

```
self.chain_counter += 1
```

See also https://github.com/biopython/biopython/issues/493

abradle: Fixed these
https://github.com/abradle/biopython/commit/153b343329a06189617046fd110c4df4c8caad9f

peterjc:
Merged as squashed commit 0f3f3ac04679b0dc8cba77c5af281e13f4ef2de7 - thank you @abradle, and also @JoaoRodrigues for looking at this.

P.S. Note http://mmtf.rcsb.org is currently offline :(

abradle:
Yes - we're having some trouble with out cloud this week. Hope to have it
sorted today or early next week.

peterjc:
Final touch would be an example or two in the Tutorial, i.e. additions to ``Doc/Tutorial/chapter_pdb.tex``

Issue #849, opened 2016-06-10
Review status of Scripts/SeqGui/SeqGui.py


peterjc:
The sequence GUI tool ``Scripts/SeqGui/SeqGui.py`` does not have a dedicated maintainer and has only really seen trivial updates for style or compatibility since it was contributed back in 2000 by Cayte.

See also http://biopythongui.blogspot.co.uk/2008/08/gui-seq.html and https://sites.google.com/site/biopythongui/download from 2008 by Wubin Qu (@quwubin)

Based on http://wiki.wxpython.org/How%20to%20install%20wxPython I was able to install wx on my Mac, and run it with the Apple provided Python 2.7

```
$ sudo pip2.7 install --upgrade --trusted-host wxpython.org --pre -f http://wxpython.org/Phoenix/snapshot-builds/ wxPython_Phoenix 
Password:
The directory '/Users/xxx/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
The directory '/Users/xxx/Library/Caches/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
Collecting wxPython_Phoenix
  Downloading http://wxpython.org/Phoenix/snapshot-builds/wxPython_Phoenix-3.0.3.dev2076+9cbca77-cp27-cp27m-macosx_10_6_intel.whl (30.0MB)
    100% |████████████████████████████████| 30.0MB 12.4MB/s 
Installing collected packages: wxPython-Phoenix
Successfully installed wxPython-Phoenix-3.0.3.dev2076+9cbca77
$ python2.7 -c "import wx; print(wx.__version__)"
3.0.3.dev2076+9cbca77
$ python2.7 Scripts/SeqGui/SeqGui.py 
Scripts/SeqGui/SeqGui.py:87: wxPyDeprecationWarning: Call to deprecated item __call__. Use :meth:`EvtHandler.Bind` instead.
  wx.EVT_BUTTON(self, ID_CLOSE, self.OnClose)
Scripts/SeqGui/SeqGui.py:88: wxPyDeprecationWarning: Call to deprecated item __call__. Use :meth:`EvtHandler.Bind` instead.
  wx.EVT_BUTTON(self, ID_APPLY, self.OnApply)
[...]

PR #856, opened 2016-06-17
Re-write of SeqGui using tkinter as GUI

MarkusPiotrowski:
See #849 
- switched from wxPython to tkinter
- selection of codon tables is dynamically loaded from `Bio.Data.CodonTables`


peterjc:
Looks good. There is plenty that could be added (e.g. loading sequencing via SeqIO; listing the NCBI codon tables by number - useful as often all you know is the number not the name; options for the translate button like full CDS or to stop codon), but refreshing this is a good step forward.

PR #857, opened 2016-06-19
test_pairwise2: Add test for print_matrix

MarkusPiotrowski:
With this, testing of pairwise2 is pretty much complete


peterjc:
Merged, see also https://github.com/biopython/biopython/commit/84140e610a8834a512dce89533ca37f7ef11591d

It is a shame that the ``print_matrix`` function does print directly to stdout, it makes the test harder (and prevents using the function in other ways). Also it outputs lines with trailing white space which is a little odd.

I digress - it is great to see the test coverage for the updated pairwise2 module greatly improved :)

PR #858, opened 2016-06-20
Remove deprecated .sub_features from SeqFeature

peterjc:
This removes the remains of the ``.sub_feature`` attribute in the ``SeqFeature`` object, which had been maintained for legacy scripts still using this to access complex joins etc from our GenBank/EMBL parser or BioSQL. The ``CompoundLocation`` object was introduced to replace this.

As written this completely removes the attribute. We could alternatively leave the get/set methods in place but instead of issuing a warning on use (as they did in Biopython 1.67), make them raise an exception?

I'd like a second set of eyes on this, e.g. @kblin (as a heavy user of the the GenBank parsing code), or @chapmanb (who I know looked at this general area for GFF parsing).

If the order of commits seems odd, this was deliberate so that the test suite passed at each step.


kblin:
After a first glance, the patches look reasonable. I just don't think that the b0dc5305788 commit message explains why the `ref` argument is added to `_loc()`. I'll try to make some time to test this branch with antiSMASH, but I'm really busy this week, so please don't wait for this. The unit tests for GenBank are pretty good, and those pass. :)

peterjc:
The ``_loc(..., ref)``` change was to refactor this into one line:

```
                cur_feature.location = _loc(location_line, self._expected_size, strand)
                cur_feature.location.ref = location_ref
```

Fair point, I could explain that in the commit comment (which I can change when rebasing/merging this). Thanks!

peterjc:
Rebased with slight refactoring based on comments from @kblin

peterjc:
Rebased pending merging direct to master after TravisCI green light...

peterjc:
Merged to master branch

PR #859, opened 2016-06-21
Refactoring Bio.Entrez to test internal details

peterjc:
This is based on pull request #684 from @carlosp420 (after I had cherry-picked the simple changes directly to the master, including renaming the old ``test_Entrez.py`` to ``test_Entrez_parser.py`` in order to allow adding this new ``test_Entrez.py`` file).

I believe this captures the important overall changes from #684, except that I have *excluded* the changes to the post logic. i.e. This pull request should have no functional impact.

@carlosp420 does this look OK to you? e.g. did I miss anything?

@mdehoon could you look this over please?

### DIFF
--- a/Bio/Entrez/__init__.py
+++ b/Bio/Entrez/__init__.py
@@ -2,7 +2,7 @@
 # Copyright 2008-2013 by Michiel de Hoon.  All rights reserved.
 # Revisions copyright 2011-2016 by Peter Cock. All rights reserved.
 # Revisions copyright 2015 by Eric Rasche. All rights reserved.
-# Revisions copyright 2015 by Carlos Peña. All rights reserved.
+# Revisions copyright 2015 by Carlos Pena. All rights reserved.
 # This code is part of the Biopython distribution and governed by its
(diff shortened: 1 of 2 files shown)


carlosp420:
Looks good @peterjc

peterjc:
@carlosp420 Any preference on how we write your name con the copyright lines in the source code? We could use the accented character via PEP263 https://www.python.org/dev/peps/pep-0263/ by adding:

```
# -*- coding: utf-8 -*-
```

There is precedent with ``Bio/motifs/applications/_xxmotif.py`` although we've generally avoided this. See also PEP3120 about Python using URF-8 as the default https://www.python.org/dev/peps/pep-3120/ on Python 3 onwards.

carlosp420:
Hi @peterjc. Using "Pena" is fine, besides being simpler and having zero risk for encoding problems.

peterjc:
Great - applied as f525476b0b4e23313d17ffe8e0565a0f5e17ea8b  and dcb71a5651fb5f1c4bc9722de5d385c5dca9ae31 - thank you!

PR #860, opened 2016-06-22
Use HTTPS to access NCBI Entrez and QBLAST

peterjc:
See http://www.ncbi.nlm.nih.gov/mailman/pipermail/utilities-announce/2016-June/000104.html

As of 1 September 2016, the NCBI seem to be planning to automatically redirect HTTP connections to HTTPS. From initial testing HTTPS support is already live, so we can make the switch now.

### DIFF
--- a/Bio/Blast/NCBIWWW.py
+++ b/Bio/Blast/NCBIWWW.py
@@ -9,8 +9,7 @@
 """Code to invoke the NCBI BLAST server over the internet.
 
 This module provides code to work with the WWW version of BLAST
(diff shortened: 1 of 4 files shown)


peterjc:
TODO: Try this for QBLAST as well

peterjc:
Does anyone object to merging this pull request now? Should we seek clarification from the NCBI about their HTTPS transition plans first?

See also this thread: http://lists.open-bio.org/pipermail/biopython-dev/2016-June/021441.html

PR #861, opened 2016-06-23
Bio.SeqUtils: use enumerate instead of range(len(x))

carlosp420:
Enumerate is more "Pythonic".


peterjc:
I like the change at a code level, but am not so keen the variable names - ``i`` and ``idx`` for two different dictionary indexes.

Can you thing of anything clearer (I'm not overly familiar with the codon usage calculation so northing immediately came to my mind).

carlosp420:
Updated the PR :point_up: @peterjc

PR #862, opened 2016-06-24
Adding an MMTF example to the tutorial.

abradle:
I've tried to follow the formatting of the others. Let me know if you'd like any changes.

Thanks for all your help! 

Anthony

### DIFF
--- a/Doc/Tutorial/chapter_pdb.tex
+++ b/Doc/Tutorial/chapter_pdb.tex
@@ -93,24 +93,28 @@ Example: get the list of the $y$ coordinates of all atoms
 \subsection{Reading files in the MMTF format}
 
 You can use the direct MMTFParser to read a structure from a file:
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> from Bio.PDB.mmtf import MMTFParser
 >>> structure = MMTFParser.get_structure("PDB/4CUP.mmtf")
 \end{verbatim}
 
 Or you can use the same class to get a structure by it's PDB ID:
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> structure = MMTFParser.get_structure_from_url("4CUP")
 \end{verbatim}
 
 This gives you a Structure object as if read from a PDB or mmCIF file.
 
-You can also have access to the underlying data using the MMTF module.
+You can also have access to the underlying data using the external MMTF library which Biopython is using internally:
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> from mmtf import fetch
 >>> decoded_data = fetch("4CUP")
 \end{verbatim}
 For example you can access just the X-coordinate.
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> print(decoded_data.x_coord_list)
 \end{verbatim}
--- a/Doc/Tutorial/chapter_pdb.tex
+++ b/Doc/Tutorial/chapter_pdb.tex
@@ -89,6 +89,32 @@ Example: get the list of the $y$ coordinates of all atoms
 >>> y_list = mmcif_dict['_atom_site.Cartn_y']
 \end{verbatim}
 
+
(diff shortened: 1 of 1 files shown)


peterjc:
MMTF probably needs a mention further up in the context of PDF and mmCIF formats.

Also could you try adding this magic line (first argument is the relative path from ``Docs/``, then ``lib:mmtf`` makes the test conditional on being able to import ``mmtf``):

```
%doctest ../Tests lib:mmtf
\begin{verbatim}
# example code to treat as a doctest
\end{verbatim}
```

then run ``Tests/test_Tutorial.py``?

Doc/Tutorial/chapter_pdb.tex
```diff
@@ -89,6 +89,32 @@ \subsection{Reading an mmCIF file}
...
+\end{verbatim}
+
+Or you can use the same class to get a structure by it's PDB ID:
+\begin{verbatim}
+>>> structure = MMTFParser.get_structure_from_url("4CUP")
+\end{verbatim}
+
+This gives you a Structure object as if read from a PDB or mmCIF file.
+
+You can also have access to the underlying data using the MMTF module.
```
peterjc: Potentially confusing to the reader, perhaps change to something like this?

...using the external MMTF library which Biopython is using internally:

Note this test would require internet access, and we currently don't have a mechanism in ``test_Tutorial.py` to handle that, so we can't use it as a doctest (yet).

peterjc:
I realise now that we've not documented how ``test_Tutorial.py`` works with the special markup in the Tutorial's LaTeX source, thus #863. Apologies.

Doc/Tutorial/chapter_pdb.tex
```diff
@@ -93,24 +93,28 @@ \subsection{Reading an mmCIF file}
...
 
-You can also have access to the underlying data using the MMTF module.
+You can also have access to the underlying data using the external MMTF library which Biopython is using internally:
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> from mmtf import fetch
 >>> decoded_data = fetch("4CUP")
 \end{verbatim}
 For example you can access just the X-coordinate.
+%doctest ../Tests lib:mmtf
```
peterjc: This can't be a doctest as the output of the print is missing. It would probably be too long to use as a doctest.

Doc/Tutorial/chapter_pdb.tex
```diff
@@ -93,24 +93,28 @@ \subsection{Reading an mmCIF file}
 \subsection{Reading files in the MMTF format}
 
 You can use the direct MMTFParser to read a structure from a file:
+%doctest ../Tests lib:mmtf
 \begin{verbatim}
 >>> from Bio.PDB.mmtf import MMTFParser
 >>> structure = MMTFParser.get_structure("PDB/4CUP.mmtf")
```
peterjc: Unfortunately this triggers a warning:

```
$ python3.5
Python 3.5.1 (default, Jun 23 2016, 12:06:32) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
>>> from Bio.PDB.mmtf import MMTFParser
>>> structure = MMTFParser.get_structure("PDB/4CUP.mmtf")
/Users/xxx/lib/python3.5/site-packages/Bio/PDB/StructureBuilder.py:85: PDBConstructionWarning: WARNING: Chain A is discontinuous at line 0.
  PDBConstructionWarning)
```

Is there a better behaved MMTK file we could use for this test (a second small MMTK file)?

peterjc:
I've applied the changes as a squashed commit ec2fd23e386a670a4cbbec51818bd01ef41be7f5 thank you, but with the doctests commented out -- pending an example which does not trigger a scary warning, and us extending the ``test_Tutorial.py`` framework to allow tagging tests which require the internet.

Adding a another (small) mmtk file which loads without warnings would be ideal - perhaps the equivalent of the PDF or mmCIF files already used in the Tutorial?

PR #876, opened 2016-07-18
Updated ASA values from newer publication

krdav:
Updated, presumably better, ASA values from recent publication. Can be accessed by the "acc_array" key which is set default to the original "Sander" ASA values for backwards compatibility.

### DIFF
--- a/Bio/PDB/DSSP.py
+++ b/Bio/PDB/DSSP.py
@@ -40,32 +40,32 @@ _dssp_cys = re.compile('[a-z]')
 # Maximal ASA of amino acids
 # Used for relative accessibility
 
-residue_max_acc = dict(
+residue_max_acc = {
     # Miller max acc: Miller et al. 1987 (http://dx.doi.org/10.1016/0022-2836(87)90038-6)
     # Wilke: Tien et al., 2013 (http://dx.doi.org/10.1371/journal.pone.0080635)
     # Sander: Sander & Rost, (1994), Proteins, 20:216-226
-    Miller={
+    'Miller': {
         'ALA': 113.0, 'ARG': 241.0, 'ASN': 158.0, 'ASP': 151.0,
         'CYS': 140.0, 'GLN': 189.0, 'GLU': 183.0, 'GLY': 85.0,
         'HIS': 194.0, 'ILE': 182.0, 'LEU': 180.0, 'LYS': 211.0,
         'MET': 204.0, 'PHE': 218.0, 'PRO': 143.0, 'SER': 122.0,
         'THR': 146.0, 'TRP': 259.0, 'TYR': 229.0, 'VAL': 160.0
     },
-    Wilke={
+    'Wilke': {
         'ALA': 129.0, 'ARG': 274.0, 'ASN': 195.0, 'ASP': 193.0,
         'CYS': 167.0, 'GLN': 225.0, 'GLU': 223.0, 'GLY': 104.0,
         'HIS': 224.0, 'ILE': 197.0, 'LEU': 201.0, 'LYS': 236.0,
         'MET': 224.0, 'PHE': 240.0, 'PRO': 159.0, 'SER': 155.0,
         'THR': 172.0, 'TRP': 285.0, 'TYR': 263.0, 'VAL': 174.0
     },
-    Sander={
+    'Sander': {
         'ALA': 126.0, 'ARG': 248.0, 'ASN': 157.0, 'ASP': 163.0,
         'CYS': 135.0, 'GLN': 198.0, 'GLU': 194.0, 'GLY': 84.0,
         'HIS': 184.0, 'ILE': 169.0, 'LEU': 164.0, 'LYS': 205.0,
         'MET': 188.0, 'PHE': 197.0, 'PRO': 136.0, 'SER': 130.0,
         'THR': 142.0, 'TRP': 227.0, 'TYR': 222.0, 'VAL': 142.0
     }
-)
+}
 
 
 def ss_to_index(ss):
@@ -252,8 +252,12 @@ class DSSP(AbstractResiduePropertyMap):
 
         @param dssp: the dssp executable (ie. the argument to os.system)
         @type dssp: string
-        """
         
+        @param acc_array: Accessible surface area (ASA) from either Miller et al. 1987, \
+        Sander & Rost, (1994) or Wilke: Tien et al. 2013
+        @type acc_array: string Sander/Wilke/Miller
+        """
+
         self.residue_max_acc = residue_max_acc[acc_array]
         
         # create DSSP dictionary
--- a/Bio/PDB/DSSP.py
+++ b/Bio/PDB/DSSP.py
@@ -38,29 +38,34 @@ from Bio.PDB.PDBParser import PDBParser
 _dssp_cys = re.compile('[a-z]')
 
 # Maximal ASA of amino acids
-# Values from Sander & Rost, (1994), Proteins, 20:216-226
 # Used for relative accessibility
-MAX_ACC = {}
-MAX_ACC["ALA"] = 106.0
-MAX_ACC["CYS"] = 135.0
-MAX_ACC["ASP"] = 163.0
-MAX_ACC["GLU"] = 194.0
-MAX_ACC["PHE"] = 197.0
-MAX_ACC["GLY"] = 84.0
-MAX_ACC["HIS"] = 184.0
-MAX_ACC["ILE"] = 169.0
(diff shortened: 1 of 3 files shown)


Bio/PDB/DSSP.py
```diff
@@ -235,7 +240,7 @@ class DSSP(AbstractResiduePropertyMap):
     -42.399999999999999)
     """
 
-    def __init__(self, model, pdb_file, dssp="dssp"):
+    def __init__(self, model, pdb_file, dssp="dssp", acc_array = 'Sander'):
```
peterjc: PEP8 style would use no spaces for `` = `` here.

peterjc:
The TravisCI failure looks like a false positive, something subtle has changed in a Python dependency triggering this which I will address separately:

```
html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 12.0.5)
```

I don't think we have any dependency-free tests for DSSP - and currently ``test_DSSP_tool.py`` does not get run on TravisCI which is why CodeCov doesn't know about it.

In any case, ``test_DSSP_tool.py`` would need updating to try the other scores. Is that something you could work on?

peterjc:
Correction: There are a few dependency-free DSSP tests in ``test_PDB.py``

peterjc:
TravisCI updated in https://github.com/biopython/biopython/commit/cf32fbbb96c7b0208ce5db25c16e531fd5f8a4a7 which ought to fix future pull requests

krdav:
I see that `test_DSSP_tool.py` doesn't do much. In fact it only verifies that DSSP will run and the output can be parsed to the same number of lines. A bit more detailed test would be helpful and adding tests for the DSSP output derived features, like RASA, would also be useful. I would like to work on such a test but I will not get time to do so before next week.

However my small change will not have any effect unless you explicitly call the dssp class with `acc_array='Something'`. Default behaviour is the same.

krdav:
Do you have a guideline for writing biopython unit tests? Or can I just write it up and make a pull request to discuss?

peterjc:
I'm suggesting there should be new tests explicitly using ``acc_array='Something'``

Here ``test_DSSP_tool.py`` will be skipped if the DSSP binary is missing, while ``test_PDB.py`` should always be run (assuming NumPy is present). If your tests don't need the binary, put them in ``test_PDB.py`` please.

There's a chapter in the Tutorial http://biopython.org/DIST/docs/tutorial/Tutorial.html but in general try to follow the style of whichever existing ``Tests/test_XXX.py`` you are adding things to.

krdav:
Cool,  I will do that as soon as time allows me.

Bio/PDB/DSSP.py
```diff
@@ -38,29 +38,34 @@
...
-MAX_ACC["THR"] = 142.0
-MAX_ACC["VAL"] = 142.0
-MAX_ACC["TRP"] = 227.0
-MAX_ACC["TYR"] = 222.0
+
+residue_max_acc = dict(
+    # Miller max acc: Miller et al. 1987 (http://dx.doi.org/10.1016/0022-2836(87)90038-6)
+    # Wilke: Tien et al., 2013 (http://dx.doi.org/10.1371/journal.pone.0080635)
+    # Sander: Sander & Rost, (1994), Proteins, 20:216-226
+    Miller={
```
JoaoRodrigues: This is not proper Python syntax: dictionaries are defined with { key: value, key: value, etc }. The key must be a string or a tuple or any other immutable type, while the value can be anything. Does this even run?

krdav: Clearly you are not used to using the dict() constructor:
https://docs.python.org/3/library/stdtypes.html#dict

I imagine that you would rather have something like this:
```
residue_max_acc = {
    # This dict is taken from the Wilke Lab's DSSP wrapper.
    # The code for the wrapper can be found at: http://wilke.openwetware.org/Software.html ("Program for computing relative solvent accessibilities for a protein that uses the DSSP program of Kabsch and Sander")
    # The values in the dict are presented in Tien et al., 2013 (http://dx.doi.org/10.1371/journal.pone.0080635)
    # and seem to be taken from the proposed values for ASA normalization (in A^2) from Miller et al., 1987 (http://dx.doi.org/10.1016/0022-2836(87)90038-6).
    'Miller': {
        'A': 113.0, 'R': 241.0, 'N': 158.0, 'D': 151.0,
        'C': 140.0, 'Q': 189.0, 'E': 183.0, 'G': 85.0,
        'H': 194.0, 'I': 182.0, 'L': 180.0, 'K': 211.0,
        'M': 204.0, 'F': 218.0, 'P': 143.0, 'S': 122.0,
        'T': 146.0, 'W': 259.0, 'Y': 229.0, 'V': 160.0
    },
    'Wilke': {
        'A': 129.0, 'R': 274.0, 'N': 195.0, 'D': 193.0,
        'C': 167.0, 'Q': 225.0, 'E': 223.0, 'G': 104.0,
        'H': 224.0, 'I': 197.0, 'L': 201.0, 'K': 236.0,
        'M': 224.0, 'F': 240.0, 'P': 159.0, 'S': 155.0,
        'T': 172.0, 'W': 285.0, 'Y': 263.0, 'V': 174.0
    },
}
```

Which is fine with me, as long as the ASA values get updated - one way or the other. And btw the code runs fine and it would indeed have failed a unit test if I had used an unhashable type as keys, but that just isn't the case.

JoaoRodrigues: Clearly not. I'd keep it consistent with the rest of the codebase, which mostly uses {'key': val} syntax. 

Also, what is the rationale behind including these extra datasets of ASAs? In other words, why these specifically?

krdav: The rationale can be found in the article that I refer to:
http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0080635

This will prevent boundary problems like this:
```
if rel_acc > 1.0:
    rel_acc = 1.0
```

Bio/PDB/DSSP.py
```diff
@@ -248,6 +253,9 @@ def __init__(self, model, pdb_file, dssp="dssp"):
         @param dssp: the dssp executable (ie. the argument to os.system)
         @type dssp: string
         """
```
peterjc: You have not explained the new ``acc_array`` argument in the docstring (which is written using RST markup), which should list the three values. The docstring should also include the reference and explanation currently only in the code comments.

Bio/PDB/DSSP.py
peterjc: @JoaoRodrigues is right that we've not historically used this dictionary construction syntax in Biopython, but it is clear enough that I didn't really notice when reading the pull request. His suggestion would fit better with the existing code base.

I see the PLOS ONE paper talks about this boundary problem with the score exceeding one. Do you think Biopython cap this?

krdav: @peterjc Yes there is a logical cap of 1 which looks like this:
```
if rel_acc > 1.0:
    rel_acc = 1.0
```
This makes sense because the relative solvent accessible surface should not be able to go above 1. I argue that the best solution is to use new ASA values rather than capping, but I also would like to let the user decide, likewise for comparative purposes the old ASA values should be accessible.

peterjc:
@JoaoRodrigues are you happy for me to apply these changes (as a squashed commit)?

I've filed issue #878 about dealing with the broken RST markup.

JoaoRodrigues:
@peterjc I'm not sure about the capping. These 'default' values are calculated using conformations that maximize ASA, but there are always abnormal conditions that exceed these values, i.e., terminal amino acids. I think we should not cap, that should be up to the user to post-process. It's the same when calculating RSAs in NACCESS: often you see values above 100% for termini, which is very easily explained since the default values are calculated using tri-peptides.

Other than this, this seems a good addition. Always good to support more datasets.

peterjc:
@JoaoRodrigues capping at 1.0 was already there in the original code. We could make that optional?

As that was your only worry, I'm going to go ahead and merge this (which I will have to do manually due to #878 and #882 updating the DSSP docstrings)...

peterjc:
@krdav are you happy to be listed under your full name in our ``NEWS`` and ``CONTRIB`` file as contributor? You don't list your name on your GitHub profile.

krdav:
@peterjc Sure. My full name is Kristian Davidsen.

peterjc:
Thank you - applied as https://github.com/biopython/biopython/commit/1b704d1182ce0a1c0ace531532b1776dbb33a5f6 with acknowledgement in https://github.com/biopython/biopython/commit/67218b211b01f79cf2aa35ae7ca9df24bd148e44

PR #877, opened 2016-07-19
Use tox under Travis to include pep8, sdist, etc checks

### LINKED ISSUE #840: Enforce basic PEP8 style with TravisCI
See also #493 for using a git pre-commit hook.

I would like to have the git pre-commit hook be quite strict, requiring explicit by-passing with ``--no-verify`` on a case by case basis, while the TravisCI check should be less onerous but mandatory. Right now something like this in ``.travis.yml`` could work:

```
 - pip install pep8
 - pep8 --ignore E123,E126,E128,E501 BioSQL/
 - pep8 --ignore E123,E126,E127,E128,E402,E501 Scripts/
 - pep8 --ignore E127,E128,E501 Doc/examples/
 - pep8 --ignore E121,E122,E123,E124,E125,E126,E127,E128,E129,E131,E201,E202,E231,E241,E302,E402,E501,E731,W291,W293,W391,W503 Bio/
```

I have no particular preference over using the command line tool ``pep8`` versus ``flake8`` or anything else - whatever is quickest would be practical if they check the same things.

Medium term I'd like us to work towards removing as many of these PEP8 violations as possible and make the TravisCI checking stricter over time.

This adds about 20 to 25s to the TravisCI run, but only really needs to be done once rather than for every version of Python (possibly twice for Python 2 and Python 3?). An alternative might by combining TravisCI with Tox along the lines used in https://github.com/galaxyproject/planemo/blob/master/.travis.yml and https://github.com/galaxyproject/planemo/blob/master/tox.ini

peterjc:
This would close #840 by including a pep8 target on TravisCI (the exact options to the pep8 tool and those for the git pre-commit hook on #493 can be refined).

It also adds TravisCI targets to check ``python setup.py sdist --manifest-only`` and ``python setup.py sdist --formats=gztar,zip`` works (as used in our build process) which is how I came to commit d90f53b8ab91fcaf71679898f86bb649b969bd59 which was a regression from getting ``python setup.py bdist_wheel`` to run (which would also now be tested on TravisCI).

The new build targets (``pep8``, ``sdist``, and ``bdist_wheel``) are all quite quick (1.5 mins, 1 min, 1 min) and so do not unduly add to the TravisCI load compared to our main test suite.

Note I am deliberately avoiding using ``tox.ini`` in order to allow this to co-exist with any user's own Tox configuration on their own machine (see #50). We could explicitly add that to the ``.gitignore`` file?

CC @cbrueffer @tiagoantao @msabramo @bow

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -1,0 +1,94 @@
+# This is a configuration file for tox, used to test
+# Biopython on various versions of Python etc under
+# the Travis Continous Integration service which is
+# configured in the file .travis.yml
+#
+# By default tox will look for tox.ini, so this file
+# will not conflict with any personal tox setup.
+#
+# You can explicitly use this tox configuration on your
+# own machine (via the -c setting), PROVIDED you have
+# all the relevant versions of Python installed. e.g.
+# specifying a specific target envronment (via -e):
+#
+# $ pip install tox
+# $ tox -c .travis-tox.ini -e py35-nocov
+#
+# Or with test coverage:
+#
+# $ pip install tox coverage
+# $ tox -c .travis-tox.ini -e py35-cover
+#
+# Or to run the pep8 Python coding style checks:
+#
+# $ pip install tox pep8
+# $ tox -c .travis-tox.ini -e pep8
+#
+# See the envlist setting for other valid arguments.
+
+[tox]
+minversion = 2.0
+skipsdist = True
+envlist =
+    pep8
+    sdist
+    {py26,py27,py33,py34,py35,pypy,pypy3}-cover
+    {py26,py27,py33,py34,py35,pypy,pypy3}-nocov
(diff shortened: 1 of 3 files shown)


peterjc:
The fact that CodeCov reports no change in test coverage suggests that I didn't accidentally reduce the coverage by changing the dependencies installed on each version of Python :)

cbrueffer:
Looks great, thanks!

peterjc:
If there are no objections, I plan to apply this by the end of the week.

peterjc:
One downside as written (compared to the current ``.travis.yml``) is this does not include the online tests, currently setup for Python 3.3 only as an allowed failure.

That leaves me slightly puzzled as to why no change in test coverage was report - perhaps the online tests had been failing?

PR #884, opened 2016-07-22
Replacing SeqRecord .seq when have .letter_annotation

### LINKED ISSUE #880: Allow modifications of sequence in SFF objects
Migrated from https://redmine.open-bio.org/issues/3438 (*Note - don't write # number as GitHub will auto-link that as if it was a GitHub issue; Peter*)

Original discussion from @mmokrejs:

> I find it a bit awkward but namely, with unnecessary overhead, to edit a sequence in a SFF object. I am fine with the requirement that the length of the sequence must stay same is the length of qualities and other annotation lists. However, to edit a sequence I have to do now:

```
    if _was_modified:
        _letter_annotations = _record.letter_annotations
        _annotations = _record.annotations
        _record.letter_annotations = {}
        _record.annotations = {}
        _record.seq = Seq(_sequence, generic_dna)
        _record.letter_annotations = _letter_annotations
        _record.annotations = _annotations

        _new_record = SeqRecord(Seq(_sequence, generic_dna), id=_record.id, name=_record.name, description=_record.description, annotations=_record.annotations, letter_annotations=_record.letter_annotations)

        _wrote = SeqIO.write(_new_record, _fh, 'sff')
    else:
        _wrote = SeqIO.write(_record, _fh, 'sff')
```

> The whole work in backup&restore of annotation lists is not necessary in my eyes. I think providing _record.rewrite_sequence('tcagnnnnnnnn') would be quite helpful here.

peterjc:
In the special case where the new sequence is the same length, and there was per-letter-annotation, rather than a ``ValueError`` allow the sequence to be replaced and preserve the existing per-letter-annotation.

This should resolve #880 - could you test this @mmokrejs please?

### DIFF
--- a/Bio/SeqRecord.py
+++ b/Bio/SeqRecord.py
@@ -292,19 +292,28 @@ class SeqRecord(object):
         >>> sub_record.letter_annotations = {}
         >>> sub_record.letter_annotations
         {}
+
+        Note that if replacing the record's sequence with a sequence of a
+        different length you must first clear the letter_annotations dict.
         """)
 
     def _set_seq(self, value):
         # TODO - Add a deprecation warning that the seq should be write only?
         if self._per_letter_annotations:
-            # TODO - Make this a warning? Silently empty the dictionary?
-            raise ValueError("You must empty the letter annotations first!")
-        self._seq = value
-        try:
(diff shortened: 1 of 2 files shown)


peterjc:
Thank you for testing this @mmokrejs  (see comment on #880), change applied.

PR #899, opened 2016-07-25
Wrong values for alanine ASA, ooops

krdav:
While fiddling around with the unit test i discovered this small mistake that should be corrected ASAP.

PR #900, opened 2016-07-26
Unit test for the DSSP object

krdav:
Adding unit test for the DSSP object. The purpose is two fold:
1. No existing unit test on the DSSP object, only on the parsing of a file created by DSSP.
2. Making a unit test on the newly implemented ASA values.


Bio/PDB/DSSP.py
```diff
@@ -78,7 +78,7 @@
         'THR': 172.0, 'TRP': 285.0, 'TYR': 263.0, 'VAL': 174.0
     },
     'Sander': {
-        'ALA': 126.0, 'ARG': 248.0, 'ASN': 157.0, 'ASP': 163.0,
+        'ALA': 106.0, 'ARG': 248.0, 'ASN': 157.0, 'ASP': 163.0,
```
peterjc: Was this change an accident?

peterjc: I see https://github.com/biopython/biopython/pull/899 now, thanks!

Bio/PDB/DSSP.py
```diff
@@ -267,27 +267,37 @@ class DSSP(AbstractResiduePropertyMap):
     -42.399999999999999)
     """
 
-    def __init__(self, model, pdb_file, dssp="dssp", acc_array="Sander"):
+    def __init__(self, model, in_file, dssp="dssp", acc_array="Sander", file_type='PDB'):
```
krdav: I had to add this to the DSSP class to allow for using a pre-generated DSSP output file. If not adding it I don't know how to test the it without having to call dssp externally. The defaults are of course set to reproduce the old behaviour.

Tests/test_PDB.py
```diff
@@ -1081,6 +1083,21 @@ def test_entitity_copy(self):
             self.assertFalse(e.get_list()[0] is ee.get_list()[0])
 
 
+def eprint(*args, **kwargs):
```
krdav: I found this function pretty useful when debugging my code since stdout is blocked when running the unit test.

peterjc:
@etal @JoaoRodrigues etc - who is best to look at this pull request?

JoaoRodrigues:
I personally don't find the 'in_file' solution very elegant but I understand it. All fine by me, good job @krdav !

krdav:
No neither did I think adding the 'in_file' option was pretty, but I could not come up with any better solution at the time.

peterjc:
Merged, thank you @krdav and @JoaoRodrigues for reviewing this.

PR #903, opened 2016-07-27
Topology in GenBank LOCUS and EMBL ID lines

peterjc:
Work in progress, replaces #812 from @kblin to address #363.

Still need to address if and how to store this in BioSQL... https://github.com/biosql/biosql/issues/5

### DIFF
--- a/Bio/GenBank/__init__.py
+++ b/Bio/GenBank/__init__.py
@@ -1241,6 +1241,12 @@ class _FeatureConsumer(_BaseGenBankConsumer):
                 raise ValueError("Could not determine alphabet for seq_type %s"
                                  % self._seq_type)
 
+            # Also save the chomosome layout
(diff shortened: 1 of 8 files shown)


peterjc:
Also the EMBL parser needs to record the topology as done for the GenBank parser

peterjc:
@kblin what do you think? I don't think I ever got a clear answer from BioPerl/BioSQL on how the topology ought to be stored (I've added some links on https://github.com/biosql/biosql/issues/5 to old email discussions), and the BioPerl optional approach was described as a hack - so perhaps we don't need to force ourselves to match that?

kblin:
Looks good to me.

peterjc:
Great, merged - thank you.

I'm happy to revisit the BioSQL side of things if we can some feedback on https://github.com/biosql/biosql/issues/5 before the next Biopython release.

PR #905, opened 2016-08-02
Parsing value-less qualifiers in EMBL/GenBank like /pseudo

### LINKED ISSUE #902: value-less qualifiers get an empty value in EMBL export
When reading/writing EMBL files, qualifiers without values like `/pseudo` get an empty value after exporting. Example:

```Python
#!/usr/bin/env python
from Bio import SeqIO
import sys

handle = open(sys.argv[1], "rU")
records = []
for record in SeqIO.parse(handle, "embl"):
    records.append(record)

SeqIO.write(records, sys.stdout, "embl")
```

```
$ grep pseudo test.embl
FT                   /pseudo
$ python pseudo_test.py test.embl | grep pseudo
FT                   /pseudo=""
```
on the following example file:

[test.embl.zip](https://github.com/biopython/biopython/files/386526/test.embl.zip)

peterjc:
This is intended to fix #902, and explicitly tests features parsed with qualifiers like ``/pseudo`` are written out like that and NOT as currently happens ``/pseduo=""`` (empty string).

I have not committed this yes as right now this breaks the BioSQL tests. See also https://github.com/biosql/biosql/issues/6


peterjc:
Rebased to solve merge conflict

Carambakaracho:
@peterjc drew my attention to this pull request. This will fix a format error in the genbank write routine for Boolean genbank qualifiers without values like somewhat rare `/pseudo` qualifier. Due to the low number of features affected, the error is easy to miss. The wrong qualifier can cause downstream tools to fail or act unpredictable.
Thank you for your consideration

Issue #848, opened 2016-06-09
Deprecate (or remove?) Scripts/xbbtools/xbb_*.py


peterjc:
The code under `Scripts/xbbtools/` is an old GUI tool, which I have updated bits of over the years but hasn't had any maintainer since 2001 when Thomas Sicheritz Pontén wrote it.

Using `pip install Pmw` to get Python megawidgets it would run under Python 2.7, but (_updated_) until 2813643007851723e5b9c522fda8c66f3bd34847 it wouldn't even start on Python 3.

Unless someone volunteers to take over this, I propose to deprecate it - or maybe even simply delete it for the next release?

peterjc:
Asking on the mailing lists too:

http://lists.open-bio.org/pipermail/biopython/2016-June/015955.html
http://lists.open-bio.org/pipermail/biopython-dev/2016-June/021427.html

MarkusPiotrowski:
It's seems to be using wxPython.
What I can think of is translating it for using `tkinter` with modern `ttk` widgets. This should make it running on Python 2 and 3 and would remove the additional dependency.

peterjc:
Getting it to start on Python 3 again turned out to be trivial: 2813643007851723e5b9c522fda8c66f3bd34847

The only bit of Pwm it seems to use is ``Pmw.ComboBox``, the rest is ``tkinter`` (which I remember was a slight pain as the imports are different under Python 3).

peterjc:
There is (now) a Tk combobox it might be able to use instead, https://docs.python.org/2/library/ttk.html#ttk.Combobox and https://docs.python.org/3/library/tkinter.ttk.html#tkinter.ttk.Combobox

That should allow getting rid of the Pwm dependency as suggested by @MarkusPiotrowski

MarkusPiotrowski:
Sorry, I got lost in `/Scripts/SeqGui...  this requires wxPython...

peterjc:
It might be worth deprecating/removing ``Scripts/SeqGui/SeqGui.py`` as well, unless you want to adopt it? Shall we file a separate issue for that?

MarkusPiotrowski:
In contrast to `SeqGui`, `Xbbtools` looks ~~terrible~~ colorful... and it also throws some errors. On the other hand, `SeqGui` has minimal functionality that is already included in `Xbbtools`, so if we deprecate/remove the latter we should also do the same with the other.
[...]

PR #914, opened 2016-08-22
Update xbbtools

MarkusPiotrowski:
Addresses issue #848.
`xbbtools` should now run under Python 2 and Python 3 and without additional dependencies. Although I added some functionality, the main purpose was to make this script running; it has still very limited capabilities.

Solved bugs:

 - Problem with unicode error when running under Python 2
 - Help didn't open

Improvements:

 - Removed dependency from `pmw` library
 - Switched to tkinter's ttk widgets
 - Added single frame translation for all six frames and three frame translation for both directions
 - Added some message windows to make the blast option more self-explaining, added a paragraph to the Help window
 - Used Biopython's `NcbiblastxxxCommandline` instead of `os.sytem()`

Others:

 - Removed some leftover methods
 - Some work on coding style, e.g., removed star imports, made `try...except` clauses more explicit etc.


peterjc:
Looks good - can you add something to NEWS file about this too (next the paragraph about SeqGui)?

MarkusPiotrowski:
Updated NEWS file.

MarkusPiotrowski:
...and rebased with upstream master.

peterjc:
Cherry-picked as https://github.com/biopython/biopython/commit/ed93c0090e89a32102ed40a5cefd3cd64b1ff705 thank you :)

PR #917, opened 2016-08-25
Adds license information to the setup function

anweshadas:
As suggested in https://packaging.python.org/distributing/#license


peterjc:
Biopython does not use the MIT license, see discussion on #898

anweshadas:
Thank you for the update. I will add myself to that ticket for updates.

peterjc:
See also https://pypi.python.org/pypi?%3Aaction=list_classifiers - the current Biopython License Agreement does not fit nicely into this list.

PR #927, opened 2016-09-08
updatind SamTools API for sort - 1.3.1

NikiB:
I made the changes for SamTools sort according to the new samtools version 1.3.1
all test_samtools_tool.py worked

### DIFF
--- a/Bio/Sequencing/Applications/_samtools.py
+++ b/Bio/Sequencing/Applications/_samtools.py
@@ -9,6 +9,8 @@ from Bio.Application import _Option, _Argument, _Switch
 from Bio.Application import AbstractCommandline, _ArgumentList
 from Bio.Application import _StaticArgument
 
+import subprocess
+
 
 class SamtoolsViewCommandline(AbstractCommandline):
 
@@ -321,6 +323,7 @@ class SamtoolsSortCommandline(AbstractCommandline):
     Concatenate BAMs, equivalent to::
 
     $ samtools sort [-no] [-m maxMem] <in.bam> <out.prefix>
+    $ samtools sort [-n] [-T FREFIX] [-o file] [-I INT] [-m maxMem] <in.bam>
 
     See http://samtools.sourceforge.net/samtools.shtml for more details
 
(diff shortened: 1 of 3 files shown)


NikiB:
travis is running the check on an old samtools program, is there a chance to update for the new one?

peterjc:
Do you have any thoughts on how we should try to support both versions of the samtools command line API? In other tools like BLAST+ where an option as was added or removed, we have offered both (with comments in the docstrings) - but it makes writing the unit tests a little harder.

peterjc:
@saketkc what do you think as you wrote most of the samtools wrappers?

saketkc:
I would personally want to support both versions. I will take a look at it in the next couple of days.
I haven't personally moved to `samtools` 1.3 though.

NikiB:
I can do the support for both versions in samtools sort as it done with BLAST+. My only concern is for the options with the same name and different meanings.

NikiB:
I added the ability to use both versions of samtools, the 0.1.x and 1.3.x

Bio/Sequencing/Applications/_samtools.py
```diff
@@ -341,17 +341,26 @@ def __init__(self, cmd="samtools", **kwargs):
...
-            _Switch(["-o", "o"], """Output the final alignment
-                                    to the standard output"""),
             _Switch(["-n", "n"], """Sort by read names rather
                                     than by chromosomal coordinates"""),
+            _Option(["-o", "o"], """(file) Write the final sorted output to FILE,
+                    rather than to standard output""",
+                    equate=False, checker_function=lambda x: isinstance(x, str)),
+            _Option(["-O", "O"], """(FORMAT) Write the final output as sam, bam, or cram""",
+                    equate=False, checker_function=lambda x: isinstance(x, str)),
+            _Option(["-T", "T"], """(FREFIX) Write temporary files to PREFIX.nnnn.bam, or if the specified PREFIX
```
peterjc: Typo, F should be P in ``(PREFIX)``

Bio/Sequencing/Applications/_samtools.py
```diff
@@ -321,47 +323,95 @@ class SamtoolsSortCommandline(AbstractCommandline):
...
-                                                    out_prefix=out_prefix)
-    >>> print(samtools_sort_cmd)
-    samtools sort /path/to/input_bam /path/to/out_prefix
+    >>> import subprocess
+    >>> child = subprocess.Popen(\
+                str("samtools"), \
+                stdout=subprocess.PIPE, \
+                stderr=subprocess.PIPE, \
+                universal_newlines=True)
+    >>> stdoutdata, stderrdata = child.communicate()
```
peterjc: This has effectively made the ``samtools`` binary a dependency for running the Biopython doctests - something we've avoided until now.

peterjc:
Auto-detecting the samtools version won't work for some use cases like preparing a script on one machine (building command line strings) and then submitting it to a cluster.

I would much prefer this be explicit and in the hands of the user - probably the simplest way to do this is two command line wrapper classes, avoiding code duplication by subclassing (see the BLAST+ wrappers for a good example of this - we avoid repeating many many common BLAST+ parameters that way).

NikiB:
I divided the sort by versions since both versions are totally different. and now there is no dependency

peterjc:
Sorry for the delay in getting back to you. With one minor change I think this is ready to go. Right now this removes the old ``SamtoolsSortCommandline`` class, which is a backwards incompatible change.

I suggest renaming ``SamtoolsVersion0xSortCommandline`` back to ``SamtoolsSortCommandline``, and leaving the new class as ``SamtoolsVersion1xSortCommandline``.

Also, are you happy to dual license your contributions under both the Biopython Licence Agreement AND the 3-Clause BSD license? See #898.

And finally, are you happy to be thanked by name in the ``NEWS`` and ``CONTRIB`` files?

NikiB:
Yes to both your questions about contribution.

For the legacy name SamtoolsSortCommandline I just added the import in init as 
"import SamtoolsVersion0xSortCommandline as SamtoolsSortCommandline".
This way people who need to be sure of the of the version have good reference as well.

peterjc:
What do you think @saketkc (of the changes as they stand, but in particular the class naming)?

saketkc:
I like @NikiB's idea of aliasing `SamtoolsSortCommandline` `SamtoolsVersion0xSortCommandline`.
All changes look good to me. Thanks @NikiB for handling this!

saketkc:
@peterjc This should be good to merge?

peterjc:
Thanks for chasing this up @saketkc.

Other than two unused ``import subprocess`` lines, this should be fine.

@NikiB: Are you happy to dual licence your contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

NikiB:
yes, no problem with any of it

peterjc:
Merged, thank you.

@NikiB are you happy for me to thank you under your real name (which is visible along with your email address via the commit itself), or just as the alias `NikiB`? I'm double checking as you have not filled in a name on your GitHub profile.

Note should you wish to, it is possible to hide your email address on GitHub - see https://help.github.com/articles/keeping-your-email-address-private/

NikiB:
No problem :) 
I will update my profile here later

PR #932, opened 2016-09-11
README: drop 2.6 from list of supported Python versions

carlosp420:
As stated in the News, the next version of Biopython 1.69 will not support
Python 2.6.

This version has also been removed from the Travis configuration file.
So maybe is good to remove it from the README as well.


peterjc:
Well spotted, I agree, thank you. Do you want to be added to the NEWS file for the next release?

carlosp420:
That would be nice. thanks!

PR #935, opened 2016-09-13
remove importing unittest2 needed for python2.6

carlosp420:
We don't need to import unittest2 (needed for Python2.6) in tests as the new Biopython version will drop support for this version.


peterjc:
Good work, thank you!

PR #936, opened 2016-09-13
Parse and write structured GenBank comments

biologyguy:
- The original implementation was not robust against malformed structured comments, causing crashes
- Any structured comment information was previously discarded from the record when written
    -> Both of these issues are now fixed
- Also alpha ordered the Record class attributes


peterjc:
This looks very interesting, but before accepting it needs a new unit test, probably best in ``Tests/test_GenBank_unittest.py``

Also none of the existing example files are relevant, you could add another (small) example file under ``Tests/GenBank`` and then also update ``Tests/test_SeqIO.py`` and ``Tests/test_GenBank.py``. The later are older print-and-compare tests, see the chapter in our tutorial for background on how to update these (``run_tests.py -g test_GenBank.py`` etc).

Bio/GenBank/Record.py
```diff
@@ -165,33 +165,33 @@ class Record(object):
...
         self.taxonomy = []
-        self.references = []
-        self.comment = ''
-        self.features = []
-        self.base_counts = ''
-        self.origin = ''
-        self.sequence = ''
-        self.contig = ''
-        self.primary = []
+        self.version = ''
```
peterjc: Did you actually change anything here, or just re-sort the attributes alphabetically? That makes sense, but would have been clearer as a self-contained commit is it is unrelated to the comment parsing work. If you're comfortable with git and are happy to revise the branch history, that would be great - otherwise don't worry about it.

peterjc:
We will also need your agreement to dual licensing this contribution under both the history Biopython license and the 3-clause BSD license, as per #898.

biologyguy:
Thanks Peter,
I'll revise the pull request at some point (hopefully this week) to separate out the reordered list of parameters and to add some unit tests.
-Steve

peterjc:
Great - thanks :)

biologyguy:
Hi again Peter,
I've been picking away at this the past few days, and I feel like the GenBank.Record class isn't actually used anywhere. As far as I can tell, GenBank parsing/writing is completely handled by Scanner and InsdcIO, so should the Record.py script be deprecated? Am I missing something?
Thanks,
-Steve

biologyguy:
Replaced by pull request #945

peterjc:
The ``Bio.GenBank`` parser is a scanner/consumer system, and was originally written with two consumers - one recreating the very GenBank-specific object (``GenBank.Record``) and one creating our more generic ``SeqRecord`` object which became the core of ``Bio.SeqIO``.

Importantly ``str(gb_record)`` using ``GenBank.Record`` should give a valid GenBank format string.

So while it isn't officially deprecated, the ``GenBank.Record`` approach is less used - but ought to be a more faithful representation of the raw data. An additional advantage is the ``GenBank.Record`` works for the fake-sequence whole genome sequencing (WGS) Genbank records (rather than a length in ``aa`` or ``bp``, they have a read count ``rc``).

PR #938, opened 2016-09-15
Add 'complement' as Seq method, add tests and do PEP8

MarkusPiotrowski:
* The `Seq` __module__ has methods `transcribe(dna)`, `back_transcribe(rna)`, `translate(seq, ...)` and  `reverse_complement(seq)`, but misses `complement(seq)`. This is sometimes confusing since this method   is available under the `Seq` __class__. E.g. `Seq.reverse_complement(seq)` and `seq.reverse_complement()` are both working, but `Seq.complement(seq)` fails.

 * Added tests to `tests_seq.py` to include `Seq.complement(seq)`
 * PEP8 revision for `Seq.py` and `test_seq.py` (actually much more work than the other two tasks)

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2192,7 +2192,8 @@ def reverse_complement(sequence):
     """Returns the reverse complement sequence of a nucleotide string.
 
     If given a string, returns a new string object.
-    Given a Seq or a MutableSeq, returns a new Seq object with the same alphabet.
+    Given a Seq or a MutableSeq, returns a new Seq object with the same
+    alphabet.
 
     Supports unambiguous and ambiguous nucleotide sequences.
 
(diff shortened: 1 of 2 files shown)


Bio/Seq.py
```diff
@@ -28,7 +28,8 @@
 from Bio import BiopythonWarning
 from Bio import Alphabet
 from Bio.Alphabet import IUPAC
-from Bio.Data.IUPACData import ambiguous_dna_complement, ambiguous_rna_complement
+from Bio.Data.IUPACData import (ambiguous_dna_complement,
+                                ambiguous_rna_complement)
```
peterjc: Is this a PEP8 change? The brackets are something I don't recall seeing elsewhere in the Biopython codebase?

Bio/Seq.py
```diff
@@ -770,8 +783,8 @@ def complement(self):
             ttable = _rna_complement_table
         else:
             ttable = _dna_complement_table
-        # Much faster on really long sequences than the previous loop based one.
-        # thx to Michael Palmer, University of Waterloo
+        # Much faster on really long sequences than the previous loop based
+        # one. Thx to Michael Palmer, University of Waterloo
```
peterjc: I'd have spelt out "Thanks" in full while changing the line wrapping, but its still fairly clear.

peterjc:
This looks sensible - unless you are comfortable rewriting the history, don't worry about this but for next time, I think a separate commit for all the PEP8 changes would have been much clearer. Like this the core functional change is a little hidden.

MarkusPiotrowski:
I'll give it a try. Do you mean different __commits__ for code and PEP8 or different __PRs__?

Bio/Seq.py
MarkusPiotrowski: It's valid (PEP328) and PEP8 compliant. I prefer it over having either two separate imports or using a backslash. This is even more useful in `test_seq.py` which has 4 imports from `Bio.Data.IUPACData`

Bio/Seq.py
MarkusPiotrowski: I do so

peterjc:
Different commits as part of the same PR is fine, thanks.

peterjc:
Lovely - I've added something to the NEWS file along with your name - you can include that next time 👍 

Thanks!

Issue #820, opened 2016-05-11
Improve test coverage by moving self-tests


peterjc:
Many of the Python files in Biopython still have self-tests at the end which are executed if the file is run directly via the idiom:

``` python
if __name__ == "__main__":
    print("Running a quick self-test")
    ...
```

This is bad for several reasons - these tests are not being run routinely via our buildbot or TravisCI (and therefore if the test breaks, we don't know about it), they are not counted on the coverage report https://codecov.io/github/biopython/biopython/ (and in fact actively reduce the coverage score), and moreover with Python 3 making subtle changes to imports, some of these self tests have accidentally become Python 2 only.

This is a tracking issue to encourage more commits like these which move these self-tests under ``Tests/test_XXX.py`` using the unittest framework instead:

* eb0c7ab903882587f3ece06e51e15c6e68855cc9
* da5a414819fdaef1fc9eb3a713fb00f92030889e

Or, in some cases using a doctest instead makes sense:

* 9242f12a65fb1d6bdc82f8201531662fb63a101b

In those cases, we're left with this stub like this which is useful when working directly on a file's doctests:

``` python
if __name__ == "__main__":
    from Bio._utils import run_doctest
    run_doctest(verbose=0)
```

peterjc:
Here's an indication of some obvious targets to tackle next:

``` bash
$ python count_main.py Bio BioSQL
Walking Bio
26	726	Bio/bgzf.py
48	258	Bio/MaxEntropy.py
32	194	Bio/NaiveBayes.py
135	212	Bio/_py3k/__init__.py
44	622	Bio/Align/AlignInfo.py
267	371	Bio/Align/Generic.py
246	574	Bio/AlignIO/FastaIO.py
65	179	Bio/AlignIO/NexusIO.py
25	535	Bio/Blast/NCBIXML.py
358	1730	Bio/GenBank/Scanner.py
11	180	Bio/Graphics/GenomeDiagram/_Colors.py
16	184	Bio/Graphics/GenomeDiagram/_FeatureSet.py
8	172	Bio/Graphics/GenomeDiagram/_GraphSet.py
26	302	Bio/Graphics/GenomeDiagram/_Track.py
29	217	Bio/KDTree/KDTree.py
29	161	Bio/KEGG/KGML/KGML_parser.py
57	1760	Bio/Nexus/Nexus.py
6	61	Bio/PDB/Dice.py
15	344	Bio/PDB/DSSP.py
[...]

PR #945, opened 2016-09-28
Parse and write structured GenBank comments

biologyguy:
- The original implementation was not robust against malformed structured comments, causing crashes
  - Structured comments are now parsed if possible, or converted to normal comments if malformed
- Any structured comment information was previously discarded from the record when written 
  - Structured comments are now written by SeqIO
  - Features in these comments are space-padded to line up with each other
- Unit tests added to test_GenBank_unittest.py
  - The example sequence files where modified so a direct comparison could be made between input and output
- Stylistic cleanup in InsdcIO.py, Scanner.py, test_SeqIO_features.py, and test_GenBank.py

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -63,6 +63,7 @@ class InsdcScanner(object):
         assert len(self.FEATURE_QUALIFIER_SPACER) == self.FEATURE_QUALIFIER_INDENT
         self.debug = debug
         self.line = None
+        self.handle = None
 
     def set_handle(self, handle):
         self.handle = handle
@@ -73,6 +74,7 @@ class InsdcScanner(object):
 
         Any preamble (such as the header used by the NCBI on ``*.seq.gz`` archives)
         will we ignored."""
+        line = ""
         while True:
             if self.line:
                 line = self.line
@@ -112,6 +114,7 @@ class InsdcScanner(object):
             "Not at start of record"
 
         header_lines = []
+        line = ""
         while True:
             line = self.handle.readline()
             if not line:
@@ -325,7 +328,7 @@ class InsdcScanner(object):
                     if qualifiers[-1][1] is None:
                         raise StopIteration
                     qualifiers[-1] = (key, qualifiers[-1][1] + "\n" + line)
-            return (feature_key, feature_location, qualifiers)
+            return feature_key, feature_location, qualifiers
         except StopIteration:
             # Bummer
             raise ValueError("Problem with '%s' feature:\n%s"
@@ -344,6 +347,7 @@ class InsdcScanner(object):
 
         assert self.line[:self.HEADER_WIDTH].rstrip() in self.SEQUENCE_HEADERS, \
             "Not at start of sequence"
+        line = ""
         while True:
             line = self.handle.readline()
             if not line:
@@ -352,7 +356,7 @@ class InsdcScanner(object):
             if line == "//":
                 break
         self.line = line
-        return ([], "")  # Dummy values!
+        return [[], ""]  # Dummy values!
 
     def _feed_first_line(self, consumer, line):
         """Handle the LOCUS/ID line, passing data to the comsumer
@@ -505,6 +509,7 @@ class InsdcScanner(object):
             self.parse_header()  # ignore header lines!
             feature_tuples = self.parse_features()
             # self.parse_footer() # ignore footer lines!
+            line = ""
             while True:
                 line = self.handle.readline()
                 if not line:
@@ -610,11 +615,11 @@ class EmblScanner(InsdcScanner):
                 repr(self.line)
             # Remove tailing number now, remove spaces later
             linersplit = line.rsplit(None, 1)
-            if(len(linersplit) > 1):
+            if len(linersplit) > 1:
                 seq_lines.append(linersplit[0])
             line = self.handle.readline()
         self.line = line
-        return (misc_lines, "".join(seq_lines).replace(" ", ""))
+        return misc_lines, "".join(seq_lines).replace(" ", "")
 
     def _feed_first_line(self, consumer, line):
         assert line[:self.HEADER_WIDTH].rstrip() == "ID"
@@ -719,8 +724,7 @@ class EmblScanner(InsdcScanner):
         consumer.size(length_parts[0])
 
     def _feed_header_lines(self, consumer, lines):
-        EMBL_INDENT = self.HEADER_WIDTH
-        EMBL_SPACER = " " * EMBL_INDENT
+        embl_indent = self.HEADER_WIDTH
         consumer_dict = {
             'AC': 'accession',
             'SV': 'version',  # SV line removed in June 2006, now part of ID line
@@ -742,8 +746,8 @@ class EmblScanner(InsdcScanner):
         # We have to handle the following specially:
         # RX (depending on reference type...)
         for line in lines:
-            line_type = line[:EMBL_INDENT].strip()
-            data = line[EMBL_INDENT:].strip()
+            line_type = line[:embl_indent].strip()
+            data = line[embl_indent:].strip()
             if line_type == 'XX':
                 pass
             elif line_type == 'RN':
@@ -903,7 +907,7 @@ class _ImgtScanner(EmblScanner):
         while self.line.rstrip() in self.FEATURE_START_MARKERS:
             self.line = self.handle.readline()
 
-        bad_position_re = re.compile(r'([0-9]+)>{1}')
+        bad_position_re = re.compile(r'([0-9]+)>')
 
         features = []
         line = self.line
@@ -1029,7 +1033,7 @@ class GenBankScanner(InsdcScanner):
 
         self.line = line
         # Seq("".join(seq_lines), self.alphabet)
-        return (misc_lines, "".join(seq_lines).replace(" ", ""))
+        return misc_lines, "".join(seq_lines).replace(" ", "")
 
     def _feed_first_line(self, consumer, line):
         """Scan over and parse GenBank LOCUS line (PRIVATE).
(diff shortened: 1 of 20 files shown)


peterjc:
It looks like you have modified some of the (NCBI provided) GenBank files by running them through Biopython (which has unfortunately introduced changes). I would prefer we keep the raw data unmodified, and if you need an extra test file like ``EU851978.gbk`` (original) and ``EU851978_output.gbk`` (second file).

Tests/test_SeqIO_features.py
```diff
@@ -501,21 +500,19 @@ def test_qualifiers(self):
...
-                qualifiers={"test": ["a test"]})
+        f = SeqFeature(FeatureLocation(10, 20), strand=+1, type="CDS", qualifiers={"test": ["a test"]})
         self.assertEqual(f.qualifiers["test"], ["a test"])
 
 
 class FeatureWriting(unittest.TestCase):
     def setUp(self):
-        self.record = SeqRecord(Seq("ACGT" * 100, generic_dna),
-                                id="Test", name="Test", description="Test")
+        self.record = SeqRecord(Seq("ACGT" * 100, generic_dna), id="Test", name="Test", description="Test")
```
peterjc: Can you revert this - you've made the line overly long.

Tests/test_SeqIO_features.py
```diff
@@ -501,21 +500,19 @@ def test_qualifiers(self):
...
 
-    def write_read_check(self, format):
+    def write_read_check(self, in_format):
```
peterjc: I don't understand this variable change name - the format argument is used to write a file and then read it back - it isn't an input file format.

Tests/test_SeqIO_features.py
```diff
@@ -735,8 +732,7 @@ def test_fuzzy_join(self):
 
         f1 = SeqFeature(FeatureLocation(AfterPosition(310), 320), strand=-1)
         # Note - is one-of(340,337) allowed or should it be one-of(337,340)?
-        f2 = SeqFeature(FeatureLocation(325, OneOfPosition(340, [ExactPosition(340),
-                                                                ExactPosition(337)])),
+        f2 = SeqFeature(FeatureLocation(325, OneOfPosition(340, [ExactPosition(340), ExactPosition(337)])),
```
peterjc: Again, this had been deliberately split to avoid an overly long line.

Bio/SeqIO/InsdcIO.py
```diff
@@ -901,10 +905,10 @@ def _write_contig(self, record):
             self._write_single_line("CO", text)
 
     def _write_sequence(self, record):
-        LETTERS_PER_BLOCK = 10
```
peterjc: These were deliberately upper case as constants https://www.python.org/dev/peps/pep-0008/#constants

Tests/test_SeqIO_features.py
biologyguy: This was to prevent shadowing of the outer scope variable 'format'. Couldn't think of a better word to use than 'in_format'. 'next_format'?

Tests/test_SeqIO_features.py
biologyguy: Sure, no problem. What is your letter count threshold?

Bio/GenBank/Scanner.py
```diff
@@ -352,7 +356,7 @@ def parse_footer(self):
             if line == "//":
                 break
         self.line = line
-        return ([], "")  # Dummy values!
+        return [[], ""]  # Dummy values!
```
peterjc: This has changed the return type from a tuple to a list? How about ``return [], ""`` instead?

Bio/GenBank/Scanner.py
```diff
@@ -505,6 +509,7 @@ def parse_cds_features(self, handle,
             self.parse_header()  # ignore header lines!
             feature_tuples = self.parse_features()
             # self.parse_footer() # ignore footer lines!
+            line = ""
```
peterjc: I would have said that was redundant (or is this one of those subtleties about scope in Python that sometimes catches me out?)

Bio/SeqIO/InsdcIO.py
biologyguy: Should they be defined at the module level then?

Bio/GenBank/Scanner.py
```diff
@@ -719,8 +724,7 @@ def _feed_seq_length(self, consumer, text):
         consumer.size(length_parts[0])
 
     def _feed_header_lines(self, consumer, lines):
-        EMBL_INDENT = self.HEADER_WIDTH
-        EMBL_SPACER = " " * EMBL_INDENT
+        embl_indent = self.HEADER_WIDTH
```
peterjc: As noted elsewhere, all upper case was intended to indicate a constant: https://www.python.org/dev/peps/pep-0008/#constants

Bio/GenBank/Scanner.py
biologyguy: Yes, definitely. Sorry.

Tests/test_GenBank.py
```diff
@@ -187,19 +186,17 @@ def t_write_format():
...
 
     # test for cleaning of translation
     translation_feature = first_record.features[1]
     test_trans = translation_feature.qualifiers["translation"][0]
-    assert ' ' not in test_trans, \
-      "Did not clean spaces out of the translation"
-    assert '\012' not in test_trans, \
-      "Did not clean newlines out of the translation"
+    assert ' ' not in test_trans, "Did not clean spaces out of the translation"
+    assert '\012' not in test_trans, "Did not clean newlines out of the translation"
```
peterjc: Pushes the line length over 80 characters, border line case for taste I guess.

Also might be nice to consistently use double quotes here (pre-existing issue).

Bio/GenBank/Scanner.py
biologyguy: Yep, the variable was defined in a logic block and then used outside of that block. I found a couple of these.

Tests/test_GenBank.py
biologyguy: Okay, 80 is the magic number. Got it.

Tests/test_SeqIO_features.py
peterjc: Ideally as per PEP8 https://www.python.org/dev/peps/pep-0008/#maximum-line-length aim for 79 characters, but we and PEP8 allow flexibility here as needed to prioritised readability.

Note that our TravisCI/Tox setup currently runs ``pep8 --max-length ...`` explicitly on some parts of the code, but ``Bio/`` in particular as a whole still has some extremely long lines so we can't enforce this globally.

Tests/test_SeqIO_features.py
peterjc: OK, that makes sense. How about ``intermediate_format`` or ``tmp_format`` then?

Bio/SeqIO/InsdcIO.py
peterjc: In theory perhaps, but e.g. the ``EmblWriter`` and its subclass ``ImgtWriter`` deliberately use the same class-level constant names with different values.

Bio/GenBank/Scanner.py
peterjc: Easy mistake to make - and it would have been unlikely to break anything.

Tests/test_GenBank.py
peterjc: I should have said aim for 79 characters as per PEP8, https://www.python.org/dev/peps/pep-0008/#maximum-line-length

Running the ``pep8`` tool locally would be helpful here.

Bio/GenBank/Scanner.py
peterjc: In fact according to CodeCov, this line isn't covered by any of our tests at all :(

peterjc:
I noticed you did some work updating the "self-test" in ``Bio/GenBank/Scanner.py``, at some point we need to move that into a proper unit test - see #820.

peterjc:
Overall this is looking very good. Other than the query about test files and the relatively minor style changes, it looks good to go.

The code coverage numbers reported via CodeCov are very slightly down, but I think this is in part things like explicitly declaring variables before logic blocks which clearer stylistically. I'm not worried about that.

biologyguy:
Great :)
Thanks for the code review Peter. I'll make the necessary adjustments that you've pointed out and make a new pull request.

peterjc:
If that's easier for you, that's fine - but it doesn't have to be a new pull request - you can update the same branch (even force push to the same branch), and GitHub will update the pull request.

biologyguy:
Neat. I didn't actually know that GitHub updated pull requests like that.
What a wonderful world we live in :D

peterjc:
I didn't get to finish going over this today, but did cherry-pick and apply some of your PEP8/cleanup work to the master branch - see the commits leading up to https://github.com/biopython/biopython/commit/49abb609b4d9d8c3f0a26b257136e06ab0d5ea45

Don't worry about rebasing this branch or merging - I'm happy to work from this pull request as it is - thanks

biologyguy:
Thanks Peter! I'll carve out some more time in the coming weeks to fix the other edge cases.

peterjc:
For next time, could you make separate pull requests for code cleanup and new functionality - so that reviewing one need not block the other. I'm not going to merge your ``Bio/GenBank/Scanner.py`` constant changes, but have tried to pull out the other cleanup suggestions.

However, I would still like to review and merge the structured comment work. Are you able to make a cleaned up branch (or fresh pull request)?

Otherwise, the fix from @tetromino on #951 applies nicely, but does not look at writing the structured comments, and would need tests added.

biologyguy:
Hi Peter,
This slipped off my radar, sorry about that. And ya, I should have split
the cleanup from the functionality; they were both happening concurrently
as I tried to make the new function work, but I could have rolled things
back. For next time.
Unfortunately I'm racing some deadlines at the moment, so if this can't
wait another month, than go ahead and use tetromino's fix. Otherwise I will
have time over the x-mas break.
Thanks,
-Steve

peterjc:
That's great Steve - I'll leave this PR open to revisit later to rescue the structured comment output work, and focus on merging #951 for now. Thanks!

biologyguy:
Hi Peter,
I've created a new pull request (#1029) to add the structured comment writing functionality without all the other formatting changes.
-Steve

PR #947, opened 2016-09-29
Move self-test Sffio

vincentdavis:
Not ready to merge! This commit only fixes file example file paths for doc tests and add doc test to self test.


peterjc:
You shouldn't need to change those paths - the way we run the Biopython doctests via ``Tests/run_tests.py`` all the paths should be relative to ``Tests/`` (not relative the Python file the doctest is in).

For self-testing via the ``__main__`` trick the helper function ought to take care of this - see https://github.com/biopython/biopython/blob/master/Bio/_utils.py#L112

PR #948, opened 2016-09-29
SffIO selftest work

vincentdavis:
Not ready to merge. Moved selftest and code for making test files to test_SffIO. Added doctest to _main_


vincentdavis:
Converted selftest code to test_self test class/case. Need comments/feedback, I did a quck conversion.
New test_SffIO coverage is 75% up from 57%
[cov_test_SffIO_after_selftest_move.pdf](https://github.com/biopython/biopython/files/501934/cov_test_SffIO_after_selftest_move.pdf)

Tests/test_SffIO.py
```diff
@@ -368,6 +367,209 @@ def test_index2(self):
...
+        filename = "Roche/E3MFGYR02_random_10_reads.qual"
+        qual_trim = list(SeqIO.parse(filename, "qual"))
+
+        for s, sT, f, q, fT, qT in zip(sff, sff_trim, fasta_no_trim, qual_no_trim, fasta_trim, qual_trim):
+            assert s.id == f.id == q.id
+            assert str(s.seq) == str(f.seq)
+            assert s.letter_annotations["phred_quality"] == q.letter_annotations["phred_quality"]
+            assert s.id == sT.id == fT.id == qT.id
+            assert str(sT.seq) == str(fT.seq)
+            assert sT.letter_annotations["phred_quality"] == qT.letter_annotations["phred_quality"]
```
peterjc: You should use ``self.assertEqual(..., ....)`` here, although I'm not sure if there is a multiple version for three or four entries.

Tests/test_SffIO.py
```diff
@@ -368,6 +367,209 @@ def test_index2(self):
...
+                print(record.id)
+        with open(filename, "rb") as handle:
+            index1 = sorted(_sff_read_roche_index(handle))
+        with open(filename, "rb") as handle:
+            index2 = sorted(_sff_do_slow_index(handle))
+        assert index1 == index2
+
+    def test_read_wrong(self):
+        filename = "Roche/greek.sff"
+        try:
```
peterjc: Probably you can just use ``self.assertRaises(ValueError, ReadRocheXmlManifest, handle)`` here.

Tests/test_SffIO.py
```diff
@@ -368,6 +367,209 @@ def test_index2(self):
...
+                # print(ReadRocheXmlManifest(handle))
+                r = ReadRocheXmlManifest(handle)
+            assert False, "Should fail!"
+        except ValueError:
+            pass
+
+        with open(filename, "rb") as handle:
+            for record in SffIterator(handle):
+                pass
+            try:
```
peterjc: Probably you can use ``self.assertRaises(ValueError, next, SffIterator(handle))`` here?

Tests/test_SffIO.py
```diff
@@ -368,6 +367,209 @@ def test_index2(self):
...
+try:
+    print(ReadRocheXmlManifest(
+        open("Roche/E3MFGYR02_alt_index_at_end.sff", "rb")))
+    assert False, "Should fail!"
+except ValueError:
+    pass
+k = list(_sff_do_slow_index(
+    open("Roche/E3MFGYR02_alt_index_at_end.sff", "rb")))
+print("Done")
+    """
```
peterjc: Can un-indent the final triple quote, but perhaps it would be clearer to put the whole lump of code as:

```
if False:
    # Ugly code to make test files...
    ...
```

What do you think? We don't count the test files when calculating coverage so that's not a downside.

Bio/SeqIO/SffIO.py
```diff
@@ -1306,4 +1311,5 @@ def write_record(self, record):
 
 if __name__ == "__main__":
     from Bio._utils import run_doctest
-    run_doctest(verbose=0)
\ No newline at end of file
+
```
peterjc: Extra blank line?

(But putting in the trailing newline at the end of the file is good)

Tests/test_SffIO.py
```diff
@@ -367,91 +367,93 @@ def test_index2(self):
...
-#         try:
-#             for record in SffIterator(handle):
-#                 print(record.id)
-#             assert False, "Should have failed"
-#         except ValueError as err:
-#             print("Checking what happens on re-reading a handle:")
-#             print(err)
+
+class TestSelf(unittest.TestCase):
+    """
```
peterjc: Use a single line as per PEP257,

```
"""These tests were originally selftests run in SffIO.py"""
```

Also there was a spelling error.

If you think if a more meaningful class name, that'd be even better.

vincentdavis:
@peterjc Ready for another review.

Tests/test_SffIO.py
```diff
@@ -368,181 +368,104 @@ def test_index2(self):
 
 
 class TestSelf(unittest.TestCase):
-    """
-    These tests were originally seltests run in SffIO.py
-    """
+    """ These tests were originally seltests run in SffIO.py """
```
peterjc: The typo is still there, seltests [sic] rather than selftests, but if that's the only remaining issue I'll fix it as part of the merge...

Tests/test_SffIO.py
```diff
@@ -368,181 +368,104 @@ def test_index2(self):
...
-            assert s.id == f.id == q.id
-            assert str(s.seq) == str(f.seq)
-            assert s.letter_annotations["phred_quality"] == q.letter_annotations["phred_quality"]
-            assert s.id == sT.id == fT.id == qT.id
-            assert str(sT.seq) == str(fT.seq)
-            assert sT.letter_annotations["phred_quality"] == qT.letter_annotations["phred_quality"]
+        with open(filename, "rb") as a_handle, open(filename, "rb") as b_handle:
+            sff = list(SffIterator(a_handle))
+            sff_trim = list(SffIterator(b_handle, trim=True))
+
```
peterjc: At this point (because we used list), both the iterators should have parsed the entire file, and we can close the handle. That was the goal with the original structure:

```
        with open(filename, "rb") as handle:
            sff = list(SffIterator(handle))
        with open(filename, "rb") as handle:
            sff_trim = list(SffIterator(handle, trim=True))
```

Also there is no reason to have the rest of the code indented as part of this with clause.

Tests/test_SffIO.py
```diff
@@ -368,181 +368,104 @@ def test_index2(self):
...
-        handle = BytesIO()
-        w = SffWriter(handle, xml=metadata)
-        w.write_file(sff)  # list
-        data = handle.getvalue()
-        # And again with an iterator...
-        handle = BytesIO()
-        w = SffWriter(handle, xml=metadata)
-        w.write_file(iter(sff))
-        assert data == handle.getvalue()
+        with open(filename, "rb") as a_handle, open(filename, "rb") as b_handle:
```
peterjc: Again, same comment as before - the with clauses should be as small as possible - as per the original.

Tests/test_SffIO.py
```diff
@@ -368,181 +368,104 @@ def test_index2(self):
...
+            def fileiter(handle):
                 for record in SffIterator(handle):
                     # print(record.id)
                     i = record.id
-                assert False, "Should have failed"
-            except ValueError as err:
-                pass
-                # print("Checking what happens on re-reading a handle:")
-                # print(err)
+            self.assertRaises(ValueError, fileiter, handle)
```
peterjc: Do we really need to define ``fileiter`` here? Does ``next(...)`` or even ``list(...)`` work by trying to start iteration through the file?

Tests/test_SffIO.py
```diff
@@ -368,181 +368,104 @@ def test_index2(self):
...
-    assert str(old.seq)==str(new.seq)
-j = list(_sff_do_slow_index(
-    open("Roche/E3MFGYR02_alt_index_in_middle.sff", "rb")))
-
-# Ugly bit of code to make a fake index at end
-records = list(SffIterator(
-    open("Roche/E3MFGYR02_random_10_reads.sff", "rb")))
-with open("Roche/E3MFGYR02_alt_index_at_end.sff", "w") as out_handle:
+if False:
+    """
```
peterjc: By putting this within a large ``if False:`` block, I mean you could remove the triple-quotes for a multi-line comment. Apologies if I was unclear.

Tests/test_SffIO.py
peterjc: How strange. I'd have to play with that to try to work out why. I presume the original try/except structure would work within the unit test framework, so maybe just use that for now?

vincentdavis: OK, last pull leaves this as I have it. @peterjc Thanks for your time reviewing.

peterjc:
Now that the code used to write the test indexes is "live", and not a big comment, it shows up some PEP8 issues. If you want to fix that too, great - otherwise I'll do it (it was my non-standard code in the first place - wink)

vincentdavis:
@peterjc I'll let you fix that as you wish or I wonder if it is worth testing the code to help prevent bit rot.

peterjc:
The fact it was a big comment meant it escaped the recent PEP8 clean up. Also just looking at the code, it probably needs some work for Python 3 (as it is all about manipulating a binary file). As to testing the code, well maybe: I wanted to record how I created the non-trivial binary test SFF files.

peterjc:
Thank you Vincent :)

PR #959, opened 2016-10-10
modify the method to_dict, add file to parse and handle duplicate key

wuxue:
When use SeqIO.to_dict method, found use SeqIO.to_dict(SeqIO.parse(file, 'fasta')) high frequency. So modify the method to_dict, add file to parse and handle duplicate key error.

### DIFF
--- a/Bio/SeqIO/__init__.py
+++ b/Bio/SeqIO/__init__.py
@@ -675,13 +675,17 @@ def read(handle, format, alphabet=None):
     return first
 
 
-def to_dict(sequences, key_function=None):
+def to_dict(sequences, key_function=None, idcheck='max'):
     """Turns a sequence iterator or list into a dictionary.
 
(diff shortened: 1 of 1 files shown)


peterjc:
Sorry, I'm not going to merge this. The ``SeqIO.to_dict`` function is deliberately simple. Rather than ``SeqIO.to_dict(SeqIO.parse(file, 'fasta'))`` our documentation tries to recommend you use ``SeqIO.index(file, 'fasta')`` or for a stored-and-reusable index, ``SeqIO.index_db(index_filename, file, 'fasta')``. Here the entire indexing design is based on unique identifiers (with some flexibility offered by the ``key_function``).

In your situation there is a potentially serious problem with your sequence naming (duplicate names are bad), but if you really need to do this, write your own function instead (based on the ``to_dict`` example).

PR #960, opened 2016-10-10
Misc fixes code quality

vincentdavis:
The last test for restrictions fails. 
```python
self.assertEqual(Ana.with_name([EcoRI]), {EcoRI: [33]})
```
Based on the [cookbook](http://biopython.org/DIST/docs/cookbook/Restriction.html#4.5)
It should pass. It looks to me like it should pass but can't really say.

Cookbook
```python
>>> new_seq = Seq('TTCAAAAAAAAAAAAAAAAAAAAAAAAAAAAGAA', IUPACAmbiguousDNA())
>>> rb = RestrictionBatch([EcoRI, KpnI, EcoRV])
>>> Ana = Analysis(rb, new_seq, linear=False)
>>> Ana
Analysis(RestrictionBatch(['EcoRI', 'EcoRV', 'KpnI']),Seq('TTCAAAAAAAAAAAAAAAAAA
AAAAAAAAAAGAA', IUPACAmbiguousDNA()),False)
>>> Ana.blunt()  # output only the result for enzymes which cut blunt
{'EcoRV': []}
>>> Ana.full()  # all the enzymes in the RestrictionBatch
{'KpnI': [], 'EcoRV': [], 'EcoRI': [33]}
>>> Ana.with_sites()  # output only the result for enzymes which have a site
{'EcoRI': [33]}
>>> Ana.without_site()  # output only the enzymes which have no site
{'KpnI': [], 'EcoRV': []}
>>> Ana.only_between(1, 20)  # the enzymes which cut between position 1 and 20
{}
>>> Ana.only_between(20, 34)  # etc...
{'EcoRI': [33]}
>>> Ana.only_outside(20, 34)
{}
>>> Ana.with_name([EcoRI])
{'EcoRI': [33]}
>>>
```


Tests/test_Restriction.py
```diff
@@ -222,6 +222,23 @@ def test_batch_analysis(self):
         self.assertEqual(hits[EcoRV], [8])
         self.assertEqual(hits[EcoRI], [16])
 
+    def test_analysis_restrictions(self):
+        """Test Fancier restriction analysis
```
peterjc: One line docstring as per PEP257 would be:

``` python
   def test_analysis_restrictions(self):
       """Test Fancier restriction analysis."""
```

peterjc:
I've cherry-picked the easy bits, thanks.

For the restriction code and #965 can you get the test suite to pass solving the first bug?

vincentdavis:
@peterjc by first bug do you mean the "with_name" issue?
self.assertEqual(Ana.with_name([EcoRI]), {EcoRI: [33]})
Based on the [cookbook](http://biopython.org/DIST/docs/cookbook/Restriction.html#4.5)
I am having trouble detrmining the correct behavoure. Hard coding into with_names the following.
```python
print('test search: {}'.format(RestrictionBatch([EcoRI]).search(self.sequence)))
```
{EcoRI: []}

If this is not the correct result then the problem is with the search function of which there are several tests. 
I guess I am asking for confirmation. What is the correct result, as it is in the cookbook or as it is in the current test?

peterjc:
Looking back that was poorly worded, sorry.

I meant can you fix the straightforward bits now (keeping the test suite passing), and then think about the open question about the restriction behaviour raised here and on #965?

vincentdavis:
see commit 0016d57 for fix.

vincentdavis:
All the tests pass on my local setup. This is frustrating.

peterjc:
As per the example on #961, you'll need to add this to use ``.assertWarns`` on Python 2.7:

``` python
# Remove unittest2 import after dropping support for Python 2
if sys.version_info[0] < 3:
    try:
        import unittest2 as unittest
    except ImportError:
        raise MissingPythonDependencyError("Under Python 2 this test needs the unittest2 library")
else:
    import unittest
```

I've asked for comments on #961 via the mailing list http://mailman.open-bio.org/pipermail/biopython-dev/2016-October/021554.html

MarkusPiotrowski:
@vincentdavis For me it looks like that the keyword argument `linear` must be explicitely handed over. In the cookbook example, the sequence is circular but the call in the `with_name` method did not pass this argument to the `search` function.

I tried adding this in `with_name`
```python
if not dct:
            return RestrictionBatch(names).search(self.sequence, self.linear)
```` 
and it seemed to work:
```bash
>>> Ana.with_name([EcoRI])
{EcoRI: [33]}
```

vincentdavis:
Thanks for taking a look at that @MarkusPiotrowski !

peterjc:
Thank you both - assuming there are no objections with using ``unittest2`` as proposed on #961 this looks good - although I'll squash it down to one commit on merging.

vincentdavis:
Just s quick note, Fixes issue #965

peterjc:
Thank you both, merged.

PR #961, opened 2016-10-10
Use assertWarns via unittest2 under Python 2

peterjc:
Some of our unit tests try to check if a warning is raised (some go further and check the number of warnings). The simple case is now easy to do via the unittest library since Python 3.2 which added ``assertWarns`` as a context manager.

We can use this on Python 2 by requiring the unittest2 backport (which we previously required on some of the Biopython tests under Python 2.6).

See also work by @carlosp420 on d8e242aa4f70d4562fe4a14d7e0e7acafdfcaace and e9306d1a4422024a64c97b42d126fdc4a55bf8a3

Assuming there are no objections to once again requiring unittest2 for some of our tests under Python 2, then we should update other tests as well to use this cleaner idiom.


peterjc:
Dev mailing list pinged for comment: http://mailman.open-bio.org/pipermail/biopython-dev/2016-October/021554.html

Issue #883, opened 2016-07-22
Refine PEP8 standards for TravisCI and git pre-commit hook


peterjc:
Issue #840 (via pull request #877) added pep8 checks to TravisCI, issue #493 describes a git pre-commit hook to enforce pep8 style guidelines.

Currently the TravisCI setup applies different pep8 strictness to the top level folders (out of practicality), but medium term we should try to agree on and fix remaining issues in order to have a single strictly enforced set of pep8 checks. I would suggest having the git pre-commit hook being even stricter (but optionally enforced).

As suggested by @bow on #840, we could then record these settings in a `setup.cfg` or `.pep8` file which would be used by most of the PEP8 testing tools (including local testing and some editors).

This information also needs to be added to our contribution guidelines, currently at http://biopython.org/wiki/Contributing

bow:
Hi @peterjc,

This isn't exactly the refined PEP8 standard we are aiming for, but what do you think about moving the per-directory style settings out of the `.travis-tox.ini` file and into separate config in the respective directories (i.e. something like [this](https://github.com/bow/biopython/commit/2d56380fbcfd42ed744dede3ff1f38d81f1b5908))?

Would this have any downside compared to the current setup that I may be missing? The upside is that now some editors can read and apply the style checks on the fly already :).

peterjc:
Good plan. 👍 

Since I transitioned from the command line tool ``pep8`` to ``flake8`` that makes a lot of sense. When we used the command line tool ``pep8`` (now renamed ``pycodestyle``), per-folder settings did not seem to be easily possible with hidden files https://pycodestyle.readthedocs.io/en/latest/intro.html#configuration

If we need to use per-folder ``. pydocstyle`` files, according to http://www.pydocstyle.org/en/latest/snippets/config.html that would work too.

Do you want to apply that change directly and/or make a pull request?

bow:
I can apply this directly sometime later today, if that's ok.

On second look, what do you think about using `setup.cfg` instead of `.pycodestyle` in the root folder? The goal is then to gradually remove the `.flake8` files in the subdirectories to have just a single `[flake8]` entry in the root `setup.cfg` (in other words, to have a single style setting for the entire Biopython).

peterjc:
I certainly would hope we can get to a single flake8 configuration for the whole of Biopython.

I'm a little less settled on yet another non-hidden file in the project root, but that does make sense as a longer term goal covering all the style settings.

bow:
[...]

PR #968, opened 2016-10-12
Added a feature for changing the id of an PDB.Entity's child.

Bernhard10:
When changing the id of an `PDB.Entity`, it has to be changed at 2 places:
*) The element's attribute `id`
*) The parent's `child_dict`.
Further more, the `full_id` has to be invalidated recursively for all children.

I do not know biopython well enough to know if there are many usecases for this feature and if there are other pitfalls that I missed, but this is a first reference implementation how changing an entity's id could work.

A probably better implementation would be making `Entity.id` a property with a costum setter method that uses the `.parent` attribute to acces the parent's `child_dict`
If this feature makes sense, I could implement this alternative version.

### DIFF
--- a/Bio/PDB/Entity.py
+++ b/Bio/PDB/Entity.py
@@ -92,45 +92,46 @@ class Entity(object):
     def reset_full_id(self):
         """Reset the full_id.
 
-        Sets the full_id of this entity and 
+        Sets the full_id of this entity and
         recursively of all its children to None.
-        This means that it will be newly generated 
+        This means that it will be newly generated
         at the next call to get_full_id.
         """
         for child in self:
             try:
                 child.reset_full_id()
-            except AttributeError: pass #Atoms do not cache their full ids.
+            except AttributeError: 
+                pass  # Atoms do not cache their full ids.
         self.full_id = None
 
-    def change_child_id(self, old_id, new_id, rename_structure = "_modified"):
+    def change_child_id(self, old_id, new_id, rename_structure="_modified"):
         """Change the id of a child.
 
         @param rename_structure: Append this string to the id
                 of the corresponding (parent) Structure instance.
-                This is to guarantee adherence to point one of 
-                the pdb advisory 
+                This is to guarantee adherence to point one of
+                the pdb advisory
                 (http://www.rcsb.org/pdb/static.do?p=general_information/about_pdb/pdb_advisory.html)
         @type rename_structure: string
         """
         entity = self[old_id]
-        #Pdb does not want people to distribute modified
-        #data under the original id. Changing the id of an
-        #entity can be seen as a modification of the pdb data.
-        #We thus change the id of the Structure object to be on the save side.
+        # Pdb does not want people to distribute modified
+        # data under the original id. Changing the id of an
+        # entity can be seen as a modification of the pdb data.
+        # We thus change the id of the Structure object to be on the save side.
         if rename_structure:
             parent = self
             while parent is not None:
                 if parent.level == "S":
-                    assert parent.get_parent() is None #Structure Entity should not have a parent.
-                    parent.id+=rename_structure
+                    assert parent.get_parent() is None  # Structure Entity should not have a parent.
+                    parent.id += rename_structure
                     break
                 parent = parent.get_parent()
         entity.id = new_id
         self.child_dict[new_id] = entity
         del self.child_dict[old_id]
         entity.reset_full_id()
-            
+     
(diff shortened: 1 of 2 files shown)


peterjc:
@Bernhard10 I recommend installing the ``pep8`` command line tool via ``pip install pep8`` to help ensure your changes will pass on TravisCI. We also have a git pre-commit hook you could use which does a similar check, see #493 and #883 

@etal @JoaoRodrigues @lennax your thoughts on the PDB changes please?

JoaoRodrigues:
Hi there,

If I understand correctly, the problem is that if you say, e.g. a chain ID,
you want to propagate that change to its children.

It's fine as it is, bar the occasional pep8 issues. However, why not do
this directly when you set the id, by defining a @id.setter decorator?

Cheers,

João

Bernhard10:
@JoaoRodrigues I only thought of the `@id.setter` solution when I was done coding the current version.
But yeah, the setter version is better, so if this is feature is going to be merged, I will change it to the setter version.

peterjc:
Given @JoaoRodrigues agrees in principle, please work on the setter version.

This will also need a new test in ``tests_PDB.py`` as well. Thanks!

Tests/test_PDB.py
```diff
@@ -995,6 +995,78 @@ def test_get_atoms(self):
...
+    def setUp(self):
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", PDBConstructionWarning)
+            self.struc = PDBParser(PERMISSIVE=True).get_structure(
+                                                  'X', "PDB/a_structure.pdb")
+
+    def test_change_model_id(self):
+        """Change the id of a model"""
+        for model in self.struc:
+            break  # Get first model in structure
```
peterjc: Assuming this works it would be more elegant (here and the other examples below):

```
model = next(self.struc)
```

This will also give a clear error if there were ever no models.

Tests/test_PDB.py
```diff
@@ -995,6 +995,78 @@ def test_get_atoms(self):
...
+        self.assertEqual(chain[(' ', 1, ' ')], res)
+
+    def test_full_id_is_updated(self):
+        """
+        Invalidate cached full_ids if an id is changed.
+        """
+        for atom in self.struc.get_atoms():
+            break  # First atom
+        original_id = atom.get_full_id()  # ('X', 0, 'A', ('H_PCA', 1, ' '), ('N', ' '))
+        if original_id != ('X', 0, 'A', ('H_PCA', 1, ' '), ('N', ' ')):
```
peterjc: This should be:

```
self.assertEqual(original_id, ('X', 0, 'A', ('H_PCA', 1, ' '), ('N', ' ')))
```

You can include an error message as well but it seems unnecessary.

Bio/PDB/Entity.py
```diff
@@ -89,6 +115,21 @@ def add(self, entity):
         self.child_list.append(entity)
         self.child_dict[entity_id] = entity
 
+    def reset_full_id(self):
```
peterjc: Do you think this should be a new public method, or denote it as private using a leading underscore? i.e. ``def _reset_full_id(self):``

Tests/test_PDB.py
Bernhard10: I did not use `assertEqual` on purpose, because I am asserting something about the file I load for the test, not about the functionality I am testing, so I thought letting the test result in an Error not a Failure is clearer. (There is currently no test addressing the caching of full_ids)

The way I wrote the tests (depending on a loaded pdb file instead of creating an instance of Entity and mocking all external dependencies) is in accordance with the other tests in test_PDB and somewhat more practical/ easier, but it brings clear disadvantages, such as depending on the content of the file.

Tests/test_PDB.py
Bernhard10: I guess it would have to be

 ```
next(iter(self.struc))
```

Tests/test_PDB.py
peterjc: Most of our tests depend on sample files - relatively few create test data structures in code.

peterjc:
This looks good. Two questions before I squash-and-mege this:

Are you happy to be named in the NEWS and CONTRIB files?

Are you happy to dual licence your Biopython contributions under both the current Biopython Licence Agreement and the 3-Clause BSD license? See #898 

Thanks!

Bernhard10:
Yeah, thanks, I am happy with both.
My name is Bernhard Thiel, in case it doesn't show up on github.

PR #970, opened 2016-10-12
Update Polypeptide.py

amelieH:
add try/except in get_ca_list

_Update_: This is a pull request to address #969

### DIFF
--- a/Bio/PDB/Polypeptide.py
+++ b/Bio/PDB/Polypeptide.py
@@ -197,8 +197,11 @@ class Polypeptide(list):
         """
         ca_list = []
         for res in self:
-            ca = res["CA"]
-            ca_list.append(ca)
+            try:
+                ca = res["CA"]
+                ca_list.append(ca)
+            except:
+                ca_list.append("None")
         return ca_list
 
     def get_phi_psi_list(self):


peterjc:
Why do you want to make this change? Do you have a test case this fixes? 

As an aside I worry that PDB files need the field to be two letters, but I've not looked at PDB files recently enough to comment. @etal or @JoaoRodrigues?

Bio/PDB/Polypeptide.py
```diff
@@ -197,8 +197,11 @@ def get_ca_list(self):
         """
         ca_list = []
         for res in self:
-            ca = res["CA"]
-            ca_list.append(ca)
+            try:
```
JoaoRodrigues: You can also just use the `dict.get` syntax to to this without the try/except. I would go for the try/except if you were to issue a warning.

amelieH: I'm not sure what you mean.
I try to do "ca = res.get("CA")" instead and I got

    ca = res.get("CA")
AttributeError: 'Residue' object has no attribute 'get'

So what do you mean by "dict.get syntax" ?

JoaoRodrigues: Nevermind, my bad :) Thought `res['CA']` was accessing `res.child_dict` and there you'd have a `.get` method.

peterjc:
Are you happy in principle for this to be merged then @JoaoRodrigues?

It will need a unit test added to ``Tests/test_PDB.py`` though - hopefully one of the existing PDB files under ``Tests/PDB/`` would be suitable.

JoaoRodrigues:
@peterjc Sure, as long as the current tests pass and there's a new unit test for this, all good! Thanks @amelieH  !

peterjc:
@amelieH can you use one of the problematic PDB files listed on #969 (ideally a small one) to make a unit test for this fix? Ask you you need more guidance - there's a chapter in the Biopython Tutorial about the test framework.

peterjc:
Closing due to no follow up on the licensing etc.

PR #975, opened 2016-10-26
Fix some parsing issues with MMTF files

StefansM:
The MMTF spec states that empty [altloc](https://github.com/rcsb/mmtf/blob/v1.0/spec.md#altloclist) and [insertion code](https://github.com/rcsb/mmtf/blob/v1.0/spec.md#inscodelist) fields should be represented with a NUL (`\x00`) character. The MMTF files included with biopython for testing used a space instead of NUL. Since this is the same character used by `StructureBuilder` to indicate an empty field the altloc and inscode fields could be passed directly to `StructureBuilder`.

When parsing conforming MMTF files with a NUL character, the NUL field was passed to `StructureBuilder` and every (non-disordered) atom was listed as disordered with an altloc of `\x00`. Similarly, every residue without an insertion code was assigned an insertion code of '\x00'.

This pull request converts the NUL characters in these fields to spaces so that `StructureBuilder` understands them. It also replaces the default blank hetero-flag with a space for consistency with the PDB and MMCIF parsers. I updated the test files with the latest available versions. These have the same `mmtfVersion` of 0.2 but actually follow the spec.


peterjc:
@abradle - Could you review this please?

Does the specification need adjusting? I presume the test files you contributed to Biopython were at one stage live on the PDB website?

Does your MMTF library need fixing (as well as, or instead of fixing Biopython)?

abradle:
Looks great to me! Thanks @StefansM 

I maybe misunderstood. But from my point of view - this fixes stuff and no further action is need.

Best wishes,

Anthony

peterjc:
@abradle regarding your API, are you happy with the MMTF library returning low level nulls, rather than perhaps the Python ``None`` object?

peterjc:
@StefansM are you happy to be named in the ``NEWS`` and ``CONTRIB`` file?

Are you happy to dual licence your contribtions under both the Biopython Licence Agreement AND the 3-Clause BSD license (see issue #898)? Thanks!

StefansM:
No problem being named and no problem with the licensing. Thanks!

PR #985, opened 2016-11-09
Setup

abdounasser202:
I've improve the setup.py so that numpy is installed automatically. No, to install Biopython manually just do
```
python setup.py install
```
This command will install all dependencies that are required.


setup.py
```diff
@@ -477,30 +453,55 @@ def is_Numpy_installed():
...
          },
-   }
-
-try:
-    setup(**setup_args)
+        classifiers=[
+            'Development Status :: 4 - Beta',
+            'Environment :: Web Environment',
+            'Intended Audience :: Developers',
+            'License :: OSI Approved :: BSD License',
```
peterjc: Biopython does not (yet) use a BSD license.

setup.py
```diff
@@ -430,32 +403,35 @@ def is_Numpy_installed():
...
-                  include_dirs=[numpy_include_dir],
-                  ))
-    EXTENSIONS.append(
-        Extension('Bio.PDB.QCPSuperimposer.qcprotmodule',
-                  ["Bio/PDB/QCPSuperimposer/qcprotmodule.c"],
-                  include_dirs=[numpy_include_dir],
-                  ))
-
+except ImportError as e:
+    subprocess.call(["pip", "install", "numpy"])
```
peterjc: This will fail if pip is not installed.

peterjc:
Which operating systems and versions of Python have you tried this on? Jython?

It appears it would add pip as a new requirement of Biopython.

peterjc:
See also discussion on #978

setup.py
Cediddi: This will also fail if pip has an alias. It might be better to import pip.

    try:
        import pip
    except ImportError:
        print("pip is not found. Install numpy manually")
    else:
        pip.main(["install", "numpy"])

peterjc: I'm uncomfortable with trying to execute pip like this in the first place. Telling pip about the dependency seems far safer.

peterjc:
Thanks for closing this - I forgot to do that after 363bc0c

PR #986, opened 2016-11-09
[IMP] Add requirements.txt

abdounasser202:
No we can install all requirements by typing
```
pip install -r requirements.txt
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,8 @@
+NumPy
+ReportLab
+matplotlib
+networkx
+pygraphviz
+rdflib
+psycopg2
+MySQLdb


peterjc:
See discussion on #978. This needs testing on Jython and PyPy.

Only NumPy is a "hard" requirement (and NOT for Jython), so this should be conditional on the Python implementation.

Everything else is a soft optional module - is there a convention for those?

peterjc:
In 363bc0c8c080a1546271bf42e2cd8a456766a9b0 I have added a minimal ``requirements.txt`` as follows:

```
# We want NumPy under CPython or PyPy, but not Jython
numpy; platform_python_implementation != 'Jython'
```

That's the only "hard" dependency in Biopython. See also #978.

It seems ``requirements-dev.txt`` is quite common for development requirements - anyone wanting to contribute to Biopython development would likely want all these soft dependencies installed for full test coverage.

Note we would need to add restrictions to some of them (e.g. some are Python 2 only, others may not work on PyPy, etc).

If there is a convention for "soft" dependencies or recommended modules, we could use that as well/instead.

PR #987, opened 2016-11-09
Use OrderedDict for SeqFeature qualifiers

peterjc:
My goal is to help round-trip EMBL/GenBank files by preserving the feature qualifier ordering explicitly using an ``OrderedDict``.

This means we can stop explicitly sorting the qualifiers on output (which produced stable but arbitrary ordering, rather than attempting to match the original order).

Some of this behaviour will happen by default as of Python 3.6 because the default ``dict`` will preserve order (see https://mail.python.org/pipermail/python-dev/2016-September/146327.html and https://stackoverflow.com/questions/39980323/dictionaries-are-ordered-in-python-3-6 etc).

Comments welcome, e.g. @kblin @chapmanb @idoerg @widdowquinn as I know you all use the GenBank parser

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -1,6 +1,6 @@
 # Copyright 2000-2003 Jeff Chang.
 # Copyright 2001-2008 Brad Chapman.
-# Copyright 2005-2015 by Peter Cock.
+# Copyright 2005-2016 by Peter Cock.
 # Copyright 2006-2009 Michiel de Hoon.
 # All rights reserved.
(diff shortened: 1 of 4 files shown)


widdowquinn:
Sounds like a reasonable thing to do. I've no complaints.

kblin:
Awesome! :+1:

peterjc:
Not sure why GitHub didn't pick this up, but I applied this last week as https://github.com/biopython/biopython/commit/c1f93f3b870b48c3483724abb1b045967feaae84

Thanks for the positive feedback, closing this PR.

PR #1005, opened 2016-11-23
Explicitly record EMBL/GenBank molecule type

### LINKED ISSUE #363: Genbank LOCUS "circular" topology and molecule type ignored

peterjc:
This builds on my recent commits to improve the EMBL/GenBank topology parsing (see fd7b17199329d9cda4c0b55a62a93dc25533391d and the subsequent commits which refactored to cover some corner cases, and add more tests), and will really close #363 

Note this means ``record.annotations["topology"]`` and ``record.annotations["molecule_type"]`` together essentially replace the legacy handling of the composite residue type dating from early GenBank/EMBL files where this was one field.

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -636,6 +636,10 @@ class EmblScanner(InsdcScanner):
             raise ValueError('Did not recognise the ID line layout:\n' + line)
 
     def _feed_first_line_patents(self, consumer, line):
+        # Old style EMBL patent records where ID line ended SQ
+        # Not 100% sure that PRT here is really molecule type and
+        # not the data file division...
+        #
         # Either Non-Redundant Level 1 database records,
         # ID <accession>; <molecule type>; <non-redundant level 1>; <cluster size L1>
         # e.g. ID   NRP_AX000635; PRT; NR1; 15 SQ
@@ -647,7 +651,7 @@ class EmblScanner(InsdcScanner):
         fields = [data.strip() for data in line[self.HEADER_WIDTH:].strip()[:-3].split(";")]
         assert len(fields) == 4
         consumer.locus(fields[0])
-        consumer.residue_type(fields[1])
+        consumer.residue_type(fields[1])  # semi-redundant
         consumer.data_file_division(fields[2])
         # TODO - Record cluster size?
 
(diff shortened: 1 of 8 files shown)


peterjc:
TravisCI is happy except for pypy3 which is failing in general.

peterjc:
@kblin could you cast your eyes over this please?

kblin:
Sorry, missed that notification again. I really need a better setup where my GitHub notifications go in my mail client, as GitHub is unable to send them to the right address.

The only worry I have is that I do see quite a bunch of GenBank files that have neither "linear" nor "circular", but the way I read the check in the GenBank parser, that would trigger the `ParserFailureError("Unexpected topology %r should be linear or circular" % topology)` error.

peterjc:
Sorry, I could have waited but ended up working on a bunch of related EMBL/GenBank bits at the end of last week and wanted to merge it sooner rather than later.

Regarding the error for values other than ``linear`` and ``circular``, that should also allow blank - which would mean nothing in the annotations dictionary at all. Certainly none of the test cases are triggering that which was me being paranoid about bad data. Perhaps I need to reword the exception slightly...

kblin:
Nah, it's fine. I just looked over the code too quickly. Looking good. Belated :+1: from me :)

peterjc:
Thanks :)

PR #1011, opened 2016-11-27
CEL parser for CEL file version 4

picrin:
Backwards compatibility with version 3 is maintained.

A "strict" switch is added. By default strict=False. Setting strict=True may break backwards compatibility.

Test cases for package Affy are added. Both new and old code is now tested.

affy_v4_example.CEL is added. It is artifically generated and small.


Bio/Affy/CelFile.py
```diff
@@ -185,7 +318,3 @@ def read(handle):
         else:
             continue
     return record
-
-if __name__ == "__main__":
-    from Bio._utils import run_doctest
-    run_doctest()
```
peterjc: Did you remove this deliberately? TravisCI etc will run this via the listing inside ``run_tests.py`` but having this ``__main__`` trick is usually very helpful for checking the doctests during development.

Tests/test_Affy.py
```diff
@@ -0,0 +1,102 @@
...
+    # tests the new Affymetrix v4 parser
+    def testAffy4(self):
+        with open(self.affy4, "rb") as f:
+            record = CelFile.read(f)
+            assert(record.intensities.shape == (5, 5))
+            assert(record.intensities.shape == record.stdevs.shape)
+            assert(record.intensities.shape == record.npix.shape)
+            assert(record.ncols == 5)
+            assert(record.nrows == 5)
+            numpy.allclose(record.intensities, [[  0.,   1.,   2.,   3.,   4.],
```
peterjc: Don't you need to check the return value from ``numpy.allclose`` to see if it is ``True`` or ``False``?

Tests/test_Affy.py
```diff
@@ -0,0 +1,102 @@
...
+    def testAffyStrict(self):
+        with self.assertRaises(IOError) as context:
+            with open(self.affy3, "rb") as f:
+                record = CelFile.read("hello", strict=True)
+
+    # tests if the new strict mode is backwards compatible
+    def testAffyStrict(self):
+        with open(self.affy3, "rb") as f:
+            record = CelFile.read("hello")
+        assert record.DatHeader is None
```
peterjc: You open a handle to the affy3 file, but then don't use it. It looks like a mistake.

Bio/Affy/CelFile.py
picrin: I have migrated some of these test cases to test_Affy.py. I didn't realise these were getting run by CI, and I'll put this back in.

Tests/test_Affy.py
picrin: ACK

Tests/test_Affy.py
picrin: ACK

Tests/test_Affy.py
```diff
@@ -0,0 +1,102 @@
...
+                                           [-15., -16., -17., -18., -19.],
+                                           [-20., -21., -22., -23., -24.]])
+            numpy.allclose(record.npix, [[9, 9, 9, 9, 9],
+                                         [9, 9, 9, 9, 9],
+                                         [9, 9, 9, 9, 9],
+                                         [9, 9, 9, 9, 9],
+                                         [9, 9, 9, 9, 9]])
+
+# Writes a small example Affymetrix V4 CEL File
+def writeExampleV4():
```
peterjc: A somewhat similar example came to mind from the SFF tests, which we recently moved. Originally my "Ugly code to make test files" was in ``Bio/SeqIO/SffIO.py`` (as a comment), but we moved to a "footnote" in ``Tests/test_SffIO.py`` wrapped in an ``if False:`` block: https://github.com/biopython/biopython/commit/6b2b125d56b93f10f3c590590575bf3d1c9f019e

This was for two reasons: Having it as a comment had resulting in bit-rot, for instance PEP8 checks were not enforced, and it was never checked for Python 3 compatibility. Also having it under ``Bio/`` counted against us for the code coverage metrics.

Bio/Affy/CelFile.py
picrin: I've put it back in, but I can't get it to work neither under python2 nor python3 due to import issues. (python2 tries to import a version of biopython I installed through pip, and python3 can't import Bio).

The docstrings work through CI though.

Tests/test_Affy.py
picrin: ACK

Tests/test_Affy.py
```diff
@@ -0,0 +1,102 @@
...
+class AffyTest(unittest.TestCase):
+    def setUp(self):
+        self.affy3 = "Affy/affy_v3_example.CEL"
+        self.affy4 = "Affy/affy_v4_example.CEL"
+    def tearDown(self):
+        pass
+    
+    # tests if the new strict mode complains about passing a non-file
+    def testAffyStrict(self):
+        with self.assertRaises(IOError) as context:
```
picrin: ACK

Tests/test_Affy.py
```diff
@@ -0,0 +1,102 @@
...
+            with open(self.affy3, "rb") as f:
+                record = CelFile.read("hello", strict=True)
```
picrin: ACK

Tests/test_Affy.py
picrin: I moved it to be run every time a testcase is run. Every time a testcase is run the test file is generated anew.

test.py
```diff
@@ -0,0 +1,3 @@
+from Bio.Affy import CelFile
+with open("Affy/406335477_B.CEL", "rb") as f:
+    record = CelFile.read(f, strict=True)
```
picrin: ACK

peterjc:
We're using TravisCI to run the command-line tool pep8 via another tool called Tox, but to mimic this on your computer with our current settings try:

```
pip install pep8
pep8 --max-line-length 92 BioSQL/
pep8 --ignore E402 --max-line-length 90 Scripts/
pep8 --max-line-length 90 Doc/examples/
pep8 --ignore E122,E123,E126,E127,E128,E129,E501,E731 Bio/
```

If you want to run this via tox, using our current settings, try:

```
pip install tox
tox -c .travis-tox.ini -e pep8
```

Issue #883 has links to some of this for background reading (and the git commit hook). I see this as a work in progress and would like to gradually enforce more of the PEP8 style checks.

Bio/Affy/CelFile.py
```diff
@@ -81,35 +83,49 @@ def __init__(self):
...
+                          "a file-like object by using `open`.")
     if magicNumber != 64:
-        # In v4 we're always strict, as we don't have to worry about backwards compatibility
+        # In v4 we're always strict, as we don't have to worry about backwards
+        # compatibility
         if strict and mode != "r":
-            raise IOError("You're trying to open an Affymetrix v3 CEL file. You have to use a read mode, like this open(filename, \"r\").")
+            raise IOError("You're trying to open an Affymetrix v3 CEL file."
+                          "You have to use a read mode, like this"
+                          "`open(filename, \"r\")`.")
```
peterjc: You need spaces here at the string split points, e.g.:

``` python
            raise IOError("You're trying to open an Affymetrix v3 CEL file. "
                          "You have to use a read mode, like this "
                          "`open(filename, \"r\")`.")
```

However, it might be too strict to check for "r" only - using "rU" should work too.

Bio/Affy/CelFile.py
```diff
@@ -142,61 +158,69 @@ def read4(f):
...
         # For debugging
-        #print([i for i in char], end="")
+        # print([i for i in char], end="")
         if char == b'\x04':
             break
         if i == safetyValve:
             raise(ValueError("Parse Error"))
 
     # After that there are precisely 15 bytes padded. Again, undocumented.
     padding = f.read(15)
```
peterjc: Are these fifteen nulls? Should you check, and at least issue a warning if the padding is non-null values? See similar cases in the SFF parser, file ``Bio/SeqIO/SffIO.py``

Bio/Affy/CelFile.py
```diff
@@ -77,7 +90,164 @@ def __init__(self):
...
+        # compatibility
+        if mode != "rb":
+            raise IOError("You're trying to open an Affymetrix v4 CEL file. "
+                          "You have to use a read binary mode, like this "
+                          "`open(filename \"rb\")`.")
+        return read4(handle)
+
+
+# read Affymetrix files version 4.
+def read4(f):
```
peterjc: Missing a docstring

Bio/Affy/CelFile.py
```diff
@@ -77,7 +90,164 @@ def __init__(self):
         self.modified = None
 
 
-def read(handle):
+def read(handle, strict=False):
```
peterjc: This is missing a docstring.

peterjc:
Good work - the automated tests are passing :)

Looking at this, you are renaming the original v3 ``read`` to ``read3``, adding ``read4``, and adding a new ``read`` function which does version detection.

I think it might be simpler to remove the ``strict`` option and instead encourage people to use ``read3`` or ``read4`` directly if they know which file format they have... but before asking you to make any further changes I'm hoping for some other perspectives and have asked on the development mailing list for comments:

http://mailman.open-bio.org/pipermail/biopython-dev/2016-November/021608.html

P.S. do you think ``read_v3`` and ``read_v4`` would be clearer names?

picrin:
I did have a feeling that I was overcomplicating things with strict. I agree it's better the way you're suggesting, and I'll have a go to see if it can be done this way. `read_v3` and `read_v4` would be more pythonic, but then the whole module uses CamelCase (a weird dialect of CamelCase, with CelFile instead of CELFile), so that would be inconsistent. For the sake of consistency readV3 and readV4 might be better.

Ideally, strict wouldn't be necessary if both `read3` and `read4` were strict to start with. At the moment `read4` is fail-fast and `read3` is garbage-in-garbage-out, and we're stuck with it for the sake of backwards compatibility.

One could argue that the fail-fast aspect of `read4` is consistent with Python's Zen **Errors should never pass silently.**.

One could argue that `read3` is consistent with Postel's Law **Be conservative in what you do, be liberal in what you accept from others** , but I don't think that's a valid argument. I read the intention of Postel's Law to be that if presented with *technically* invalid input, one should make an attempt to interpret it, but if the input is *completely* invalid, one should still fail.

peterjc:
We have lots of parsers in Biopython from different authors which take slightly different approaches.

I personally prefer a warning (via ``warnings.warn(msg, BiopythonParserWarning)`` or similar) for ambiguous data where the parser can still continue safely. This allows the user to be strict if they want to by catching the warnings and upgrading them to exceptions.

If the file is truly ambiguous and cannot be passed clearly, I would expect the parser to abort with a clear exception. Sometimes if only a small part of the file is broken (e.g. one line in a text format) it can be more practical to omit that value or use a sentinel value like ``None``, issue a warning, and continue.

The PDB parser is an example with an explicit strict mode (which I understand made sense because at the time there were many problematic PDB files in the database).

Perhaps it is time to debate and document an official Biopython policy here?

picrin:
I've added dosctrings for all methods defined in the module & removed the `strict` flag.

The only remaining issue I'm aware of is the naming convention. Should we break the convention of the module and use `read_3`, or should we keep the convention of the module and go for `readV3`?

Bio/Affy/CelFile.py
```diff
@@ -1,10 +1,10 @@
-# Copyright 2004 by Harry Zuzan.  All rights reserved.
+# Copyright 2004 by Harry Zuzan and Adam Kurkiewicz. All rights reserved.
```
peterjc: Your work is from 2016, so I'd use a second line following the style elsewhere in Biopython:

``` python
# Copyright 2004 by Harry Zuzan.  All rights reserved.
# Copyright 2016 by Adam Kurkiewicz. All rights reserved.
```

Bio/Affy/CelFile.py
```diff
@@ -78,8 +91,198 @@ def __init__(self):
...
+    """ Reads Affymetrix CEL file, version 4, and returns a corresponding Record object.
+
+    Most importantly record.intensities correspond to intensities from the CEL file.
+
+    record.mask and record.outliers are not set.
+
+    Example Usage:
+
+    >>> from Bio.Affy import CelFile
+    >>> with open('Affy/affy_v4_example.CEL', "rb") as handle:
```
peterjc: Mixed quotes.

Bio/Affy/CelFile.py
```diff
@@ -78,8 +91,198 @@ def __init__(self):
...
+    record.noutliers = None
+    record.outliers = None
+    record.modified = None
+
+    # Real data never seems to have anything but zeros here, but we don't want
+    # to take chances. Raising an error is better than returning unreliable
+    # data.
+
+    if headersMap["Axis-invertX"] != "0":
+        raise(ValueError("Parse Error"))
```
peterjc: You have unneeded brackets, and the exception message  is too vague - how about:

``` python
    raise ValueError("Expected Axis-invertX to be x0x, not %r" %
                     headersMap["Axis-invertX"])
```

(I have previously wondered if we should define ``from Bio import BiopythonParserError`` for this kind of situation, perhaps subclassing ``ValueError``?)

Bio/Affy/CelFile.py
```diff
@@ -78,8 +91,198 @@ def __init__(self):
...
+
+    if headersMap["AxisInvertY"] != "0":
+        raise(ValueError("Parse Error"))
+
+    if headersMap["AxisInvertY"] != "0":
```
peterjc: Is this a typo, or an accidental duplication of the y-axis inversion check?

Tests/test_Affy.py
```diff
@@ -0,0 +1,139 @@
...
+                        b"=" +
+                        headers[header] +
+                        b"\n")
+            except TypeError:
+                f.write(header + b"=" + headers[header] + b"\n")
+        f.write(prePadding)
+        f.write(b"\x00" * 15)
+        for i in range(25):
+            f.write(packData(float(i), float(-i), 9))
+    if __name__ == "__main__":
```
peterjc: This should not be indented (and should have some blank lines above it too).

Bio/Affy/CelFile.py
```diff
@@ -1,10 +1,10 @@
...
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
 
 """
-Classes for accessing the information in Affymetrix cel files.
+Classes for accessing the information in Affymetrix cel files version 3 and 4
 
 Functions:
 read      Read a cel file and store its contents in a Record
```
peterjc: Docstring needs to include the new version  and 4 functions for completeness.

Bio/Affy/CelFile.py
```diff
@@ -1,10 +1,10 @@
-# Copyright 2004 by Harry Zuzan.  All rights reserved.
+# Copyright 2004 by Harry Zuzan and Adam Kurkiewicz. All rights reserved.
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
 
 """
-Classes for accessing the information in Affymetrix cel files.
+Classes for accessing the information in Affymetrix cel files version 3 and 4
```
peterjc: Missing trailing full stop, see PEP257 https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

peterjc:
I've just noted some very minor things, but was interrupted before finishing this review. This is now very nearly ready to merge :)

Regarding the function names, I would still go for ``read_v3`` and ``read_v4`` which is consistent with PEP8 the existing all lower case function ``read``. Class names should be title case, and you are right ``CelFile`` could have been `CELFile``, but we're not going to change that as it would break backward compatibility.

Are you happy to dual licence your contributions under both the _Biopython License Agreement_ *and* the _BSD 3-Clause License_? See #898.

Do you want to add a short (approx two line) paragraph to the ``NEWS`` file, along with your name - and add yourself to the ``CONTRIB`` file too? Or should I do that?

Bio/Affy/CelFile.py
picrin: I was meant to check for other weird fields. Will fix.

Bio/Affy/CelFile.py
picrin: Sorry, I don't understand. CEL files come in 2 versions (v3 and v4). The old parser supported v3. The new parser supports v4. Both parsers support version 3 and 4, and the documentation reflects that.

Do you mean that 3 functions need to be documented for completeness:

1. read
2. readV3
3. readV4

picrin:
That should be all done.

picrin:
Should we remove the information about functions and classes from the module-level docstring? It kind of looks horrible in the HTML documentation:

http://biopython.org/DIST/docs/api/

peterjc:
Certainly the API docs could be prettier. The docstrings are now rendered as RST, so you can use bullet point lists e.g. https://github.com/biopython/biopython/blob/master/Bio/Blast/Applications.py

Bio/Affy/CelFile.py
peterjc: Sorry, I missed a letter out: I meant: "Docstring needs to include the new version 3 and 4 functions for completeness."

i.e. Listing ``read``, ``readV3`` and ``readV4`` or whatever they get called.

Bio/Affy/CelFile.py
```diff
@@ -78,8 +94,204 @@ def __init__(self):
...
+    # data.
+    def raiseBadHeader(field, expected):
+        actual = int(headersMap[field])
+        message = "The header {field} is expected to be 0, not {value}".format(value=actual, field=field)
+        if actual != expected:
+            raise ValueError(message)
+
+    raiseBadHeader("Axis-invertX", 0)
+
+    raiseBadHeader("AxisInvertY", 0)
```
peterjc: Is it really ``Axis-invertX`` (with a hyphen/minus) and ``AxisInvertY`` (without)?

Bio/Affy/CelFile.py
```diff
@@ -78,8 +94,204 @@ def __init__(self):
...
+
+    # for debugging
+    # pp.pprint("preHeadersMap")
+    # pp.pprint(preHeadersMap)
+    # pp.pprint("headersMap")
+    # pp.pprint(headersMap)
+
+    record.version = preHeadersMap["version"]
+    if record.version != 4:
+        raise ValueError("You are trying to parse CEL file version 4. This file violates the structure expected from CEL file version 4")
```
peterjc: Ideally split this over multiple lines aiming for about 78 characters, e.g.:

``` python
if record.version != 4:
         raise ValueError("You are trying to parse CEL file version 4. This "
                          "file violates the structure expected from CEL "
                          "file version 4")```

File: NEWS [hunk 21]
```diff
@@ -37,6 +37,8 @@ patent data files, and the related IMGT parser was updated to cope with
 IPD-IMGT/HLA database files after release v3.16.0 when their ID line changed.
 The GenBank output now uses colon space to match current NCBI DBLINK lines.
 
+The Bio.Affy package supports Affymetrix version 4 of the CEL file format, in addition to version 3.
```
peterjc: Again, line wrap as:

```
The Bio.Affy package supports Affymetrix version 4 of the CEL file format, in
addition to version 3.
```

NOTE to self: Can we add a quick one line Unix command to the TravisCI/Tox settings along side running PEP8 to check this file?

NEWS
```diff
@@ -60,6 +62,7 @@ Sourav Singh (first contribution)
 Steve Bond
 Uri Laserson
 Vincent Davis
+Adam Kurkiewicz (first contribution)
```
peterjc: I forget why now, but we list the contributors in the NEWS file alphabetically by first name - so you'd currently be first :)

Bio/Affy/CelFile.py
picrin: Yes it is, both in documentation and in real-life: http://www.affymetrix.com/estore/support/developer/powertools/changelog/gcos-agcc/cel.html.affx

peterjc:
Good work. Sadly there's a conflict with the ``NEWS`` file now (I was merging some other work), so I will merge this manually at the command line (rather than in the GitHub interface), most likely as a squashed commit....

peterjc:
Thanks for spotting the mis-sorting in the NEWS file too 👍

picrin:
I can fix the conflict by merging with master first.

picrin:
That should be the conflicts fixed.

peterjc:
No need - I've done it locally and am just running the tests now.

For future reference, I'd much prefer a rebase over a merge.

picrin:
OK, Sorry

picrin:
There is a good but little known feature of git called `merge --squash`. It's like rebase, but much less fragile. I've once given a talk about it:
https://github.com/picrin/gitAsYouDontKnowIt/blob/master/GitAsYouDontKnowIt.pdf

peterjc:
Yep, squash is very handy. We've got GitHub setup with the big green button for "Squash and merge" or "Rebase and merge".

picrin:
with `merge --squash` you can maintain linear history, but you have to fix the conflicts only once.

peterjc:
Applied as https://github.com/biopython/biopython/commit/6c14a26dda32ad6d3147036a01f3d0c4d306c647 - thank you!

PR #1012, opened 2016-11-28
Record missing data in Bio.Phylo.PAML.codeml

### LINKED ISSUE #483: Bio.Phylo.PAML.codeml's results parser quietly fails to read all the data
For a prettier version of the following go here: http://stackoverflow.com/posts/20872441/edit, or here: https://www.biostars.org/p/89848/.

**Biopython** comes with methods to interface with the **PAML** package for phylogenetic analysis.

In particular I am using [**Bio.Phylo.PAML**](http://biopython.org/wiki/PAML) to run analyses using PAML's codeml.exe program which in my case does Ka/Ks (dN/dS) ratio analysis on pairs of orthologous gene sequences.

After running the analysis using `results = cml.run()` I can see it has successfully generated `result.out` files that look about right. Most importantly the final line of the file is what I need to parse into Python:

```
t= 0.2173  S=   703.9  N=  1489.1  dN/dS=  0.2247  dN = 0.0344  dS = 0.1529
```

What I am most interested in is `dN/dS = 0.2247`

According to Biopython's [**PAML wiki**](http://biopython.org/wiki/PAML) this value can be obtained from Python by doing `results = cml.run()` which generates a dictionary with a set of values I am interested in after running the analysis. The wiki claims I can find the values I need in a key called `'parameters'`. But this only returns one of the values I need `t= 0.2173`, look: 

```
>>> results.values()
['Fcodon', 'One dN/dS ratio for branches, ', '4.7b', {0: {'description': 'one-ratio', 'parameters': {'t': 0.1982}}}, {'htlv': {}, 'stlv': {}}]
```

Notice, that my parameters key only contains the `t= 0.2173` and has omitted `S=   703.9  N=  1489.1  dN/dS=  0.2247  dN = 0.0344  dS = 0.1529`

Could anybody with `codeml` experience explain to me why the parser fails to yield most of the parameters (values) I am interested in?

**Extra details**
- Using Python2.7, PAML4.7a
- Running on Windows 7
- I will readily edit in any data or info you require to help me fix this

alpae:
The long existing issue #483 was also a problem in one of my analysis. I've added a fix and a unittest. The codeml parser parses now all site parameters of a pairwise codeml analysis.

### DIFF
--- a/Bio/Phylo/PAML/_parse_codeml.py
+++ b/Bio/Phylo/PAML/_parse_codeml.py
@@ -5,6 +5,7 @@
 
 import re
 
+import collections
 
 line_floats_re = re.compile("-*\d+\.\d+")
 
@@ -411,7 +412,7 @@ def parse_pairwise(lines, results):
     #
     # t= 0.0126  S=    81.4  N=   140.6  dN/dS= 0.0010  dN= 0.0000  dS= 0.0115
     pair_re = re.compile("\d+ \((.+)\) ... \d+ \((.+)\)")
-    pairwise = {}
+    pairwise, seq1, seq2 = collections.defaultdict(dict), None, None
     for line in lines:
         # Find all floating point numbers in this line
         line_floats_res = line_floats_re.findall(line)
@@ -420,21 +421,17 @@ def parse_pairwise(lines, results):
         if pair_res:
             seq1 = pair_res.group(1)
             seq2 = pair_res.group(2)
-            if pairwise.get(seq1) is None:
-                pairwise[seq1] = {}
-            if pairwise.get(seq2) is None:
-                pairwise[seq2] = {}
-            if len(line_floats) == 1:
-                pairwise[seq1][seq2] = {"lnL": line_floats[0]}
-                pairwise[seq2][seq1] = pairwise[seq1][seq2]
-            elif len(line_floats) == 6:
-                pairwise[seq1][seq2] = {"t": line_floats[0],
-                                        "S": line_floats[1],
-                                        "N": line_floats[2],
-                                        "omega": line_floats[3],
-                                        "dN": line_floats[4],
-                                        "dS": line_floats[5]}
-                pairwise[seq2][seq1] = pairwise[seq1][seq2]
+        if len(line_floats) == 1 and seq1 is not None and seq2 is not None:
+            pairwise[seq1][seq2] = {"lnL": line_floats[0]}
+            pairwise[seq2][seq1] = pairwise[seq1][seq2]
+        elif len(line_floats) == 6 and seq1 is not None and seq2 is not None:
+            pairwise[seq1][seq2].update({"t": line_floats[0],
+                                         "S": line_floats[1],
+                                         "N": line_floats[2],
+                                         "omega": line_floats[3],
+                                         "dN": line_floats[4],
+                                         "dS": line_floats[5]})
+            pairwise[seq2][seq1] = pairwise[seq1][seq2]
     if pairwise:
         results["pairwise"] = pairwise
     return results
--- a/Bio/Phylo/PAML/_parse_codeml.py
+++ b/Bio/Phylo/PAML/_parse_codeml.py
@@ -412,7 +412,9 @@ def parse_pairwise(lines, results):
     #
     # t= 0.0126  S=    81.4  N=   140.6  dN/dS= 0.0010  dN= 0.0000  dS= 0.0115
     pair_re = re.compile("\d+ \((.+)\) ... \d+ \((.+)\)")
-    pairwise, seq1, seq2 = collections.defaultdict(dict), None, None
+    pairwise = collections.defaultdict(dict)
+    seq1 = None
+    seq2 = None
     for line in lines:
         # Find all floating point numbers in this line
         line_floats_res = line_floats_re.findall(line)
--- a/Bio/Phylo/PAML/_parse_codeml.py
+++ b/Bio/Phylo/PAML/_parse_codeml.py
@@ -5,8 +5,6 @@
 
 import re
 
-import collections
-
 line_floats_re = re.compile("-*\d+\.\d+")
 
 try:
@@ -412,7 +410,7 @@ def parse_pairwise(lines, results):
     #
     # t= 0.0126  S=    81.4  N=   140.6  dN/dS= 0.0010  dN= 0.0000  dS= 0.0115
     pair_re = re.compile("\d+ \((.+)\) ... \d+ \((.+)\)")
-    pairwise = collections.defaultdict(dict)
+    pairwise = {}
     seq1 = None
     seq2 = None
     for line in lines:
@@ -423,6 +421,10 @@ def parse_pairwise(lines, results):
         if pair_res:
             seq1 = pair_res.group(1)
             seq2 = pair_res.group(2)
+            if seq1 not in pairwise:
+                pairwise[seq1] = {}
+            if seq2 not in pairwise:
+                pairwise[seq2] = {}
         if len(line_floats) == 1 and seq1 is not None and seq2 is not None:
             pairwise[seq1][seq2] = {"lnL": line_floats[0]}
             pairwise[seq2][seq1] = pairwise[seq1][seq2]
--- a/Tests/test_PAML_codeml.py
+++ b/Tests/test_PAML_codeml.py
@@ -481,8 +481,11 @@ class ModTest(unittest.TestCase):
                         % version.replace('_', '.')
             results_path = os.path.join(res_dir, results_file)
             results = codeml.read(results_path)
-            for seq2 in results['pairwise'].values():
-                for data in seq2.values():
+            for seq in results["pairwise"]:
+                seq_vals = results["pairwise"][seq]
+                self.assertTrue(len(seq_vals) > 0,
+                                "no parsed parameters for '%s'" % seq)
+                for data in seq_vals.values():
                     for param in ("t", "S", "N", "omega", "dN", "dS", "lnL"):
                         self.assertTrue(param in data, version_msg +
                                         ": '%s' not in parsed parameters" % param)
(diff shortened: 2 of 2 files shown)


peterjc:
@brandoninvergo could you look at this please? There isn't a test file to look at as you requested on #483, but @alpae has included a unit test.

brandoninvergo:
At first glance it seems OK, although it's been a while since I looked at the code.

The patch changes the code to use collections.defaultdict, which I don't think was an option when I wrote the module due to backwards-compatibility problems.  Is it still an issue? Also, line 415 in the patch performs multiple definitions of different types on the same line, which I think is bad form. How does it fit with the Biopython coding standards?

Anyway, it seems to me that it could have just as easily been fixed by unindenting lines 423-437 by one level. That looks like it was the source of the bug (ah that Python "feature" of indentation-as-structural-delimiter can really bite you if you miss something).  But if we go that route, please change 423 and 425 to read `if seq1 not in pairwise` and `if seq2 not in pairwise`, which is more Pythonic!

Bio/Phylo/PAML/_parse_codeml.py
```diff
@@ -411,7 +412,7 @@ def parse_pairwise(lines, results):
     #
     # t= 0.0126  S=    81.4  N=   140.6  dN/dS= 0.0010  dN= 0.0000  dS= 0.0115
     pair_re = re.compile("\d+ \((.+)\) ... \d+ \((.+)\)")
-    pairwise = {}
+    pairwise, seq1, seq2 = collections.defaultdict(dict), None, None
```
peterjc: As noted by @brandoninvergo using three separate lines would be more in line with the rest of Biopython here.

alpae:
@brandoninvergo the change in line 423/425 is not the same whether you test that the seq1/seq2 are in pairwise or None. The logLiklihood value and the other values are not read from the same line.  I'm happy do adapt line 415 to single assignment statements. The defaultdict has been introduced in python2.5. Given that Py2.6 is required for biopython, I guess that is fine.

Bio/Phylo/PAML/_parse_codeml.py
alpae: sure, will do that

peterjc:
We have no dependency problem using defaultdict now (or the ordered dict either).

peterjc:
Biopython 1.68 was our last release to support Python 2.6, so you can now assume Python 2.7 or Python 3.3 onwards.

Tests/test_PAML_codeml.py
```diff
@@ -472,6 +473,24 @@ def testParsePairwise(self):
...
+            version_msg = "Improper parsing for version %s" \
+                        % version.replace('_', '.')
+            results_path = os.path.join(res_dir, results_file)
+            results = codeml.read(results_path)
+            for seq2 in results['pairwise'].values():
+                for data in seq2.values():
+                    for param in ("t", "S", "N", "omega", "dN", "dS", "lnL"):
+                        self.assertTrue(param in data, version_msg +
+                                        ": '%s' not in parsed parameters" % param)
+                        self.assertTrue(isinstance(data[param], numbers.Number))
```
peterjc: I don't think we've ever used the numbers module before in Biopython (at least, searching for ``import numbers`` didn't show any matches). Maybe we should, or would checking against ``int`` and ``float`` work just as well?

peterjc: Thanks. Please open an issue on taking advantage of the numbers module in Biopython if you think it would be useful.

peterjc:
(I've given the pull request a more descriptive name, although I did muddle this with another open issue first - whoops)

peterjc:
@brandoninvergo are you happy for this to be merged now? Is the proposed test enough?

Tests/test_PAML_codeml.py
```diff
@@ -472,6 +472,24 @@ def testParsePairwise(self):
...
+    def testParseSitesParamsForPairwise(self):
+        """Verify that pairwise site estimates are indeed parsed. Fixes #483"""
+        res_dir = os.path.join(self.results_dir, "codeml", "pairwise")
+        for results_file in os.listdir(res_dir):
+            version = results_file.split('-')[1].split('.')[0]
+            version_msg = "Improper parsing for version %s" \
+                        % version.replace('_', '.')
+            results_path = os.path.join(res_dir, results_file)
+            results = codeml.read(results_path)
+            for seq2 in results['pairwise'].values():
```
brandoninvergo: The test doesn't actually catch the problem as written.  Changing 484-485 to this should work:

            for seq in results["pairwise"]:
                seq_vals = results["pairwise"][seq]
                self.assertTrue(len(seq_vals) > 0, "no parsed parameters for '%s'" \
                                % seq)
                for data in seq_vals:

peterjc:
@alpae: Once this is ready to merge, are you happy to dual licence your contributions under both the _Biopython License Agreement_ *and* the _BSD 3-Clause License_? See #898.

Also, would you be happy to be thanked by name in the ``NEWS`` and ``CONTRIB`` files?

Tests/test_PAML_codeml.py
brandoninvergo: Hang on, my suggestion seems to be buggy too.

brandoninvergo:
Here's a simplified version of the test.  Also, the original testPairwise was broken, so it's fixed here. There's a bit of overlap between the two tests. Oh well

    def testParsePairwise(self):
        res_dir = os.path.join(self.results_dir, "codeml", "pairwise")
        for results_file in os.listdir(res_dir):
            version = results_file.split('-')[1].split('.')[0]
            version_msg = "Improper parsing for version %s" \
                        % version.replace('_', '.')
            results_path = os.path.join(res_dir, results_file)
            results = codeml.read(results_path)
            # Pairwise models have an extra top-level item: pairwise
            self.assertEqual(len(results), 5, version_msg)
            self.assertTrue("pairwise" in results, version_msg)
            pairwise = results["pairwise"]
            self.assertEqual(len(pairwise), 2, version_msg +
                             ": should be two species")
            seq1, seq2 = list(pairwise.keys())
            self.assertEqual(len(pairwise[seq1][seq2]), 7, version_msg +
                             ": wrong number of parameters parsed")
            self.assertEqual(len(pairwise[seq2][seq1]), 7, version_msg +
                             ": wrong number of parameters parsed")

    def testParseSitesParamsForPairwise(self):
        """Verify that pairwise site estimates are indeed parsed. Fixes #483"""
        res_dir = os.path.join(self.results_dir, "codeml", "pairwise")
        for results_file in os.listdir(res_dir):
            version = results_file.split('-')[1].split('.')[0]
            version_msg = "Improper parsing for version %s" \
                        % version.replace('_', '.')
            results_path = os.path.join(res_dir, results_file)
            results = codeml.read(results_path)
            self.assertTrue("pairwise" in results)
            self.assertEqual(len(results["pairwise"]), 2)
            seq1, seq2 = list(results["pairwise"].keys())
            params = results["pairwise"][seq1][seq2]
            self.assertEqual(len(params), 7,
                            version_msg + ": wrong number of parsed parameters" +
                             " for %s-%s" % (seq1, seq2))
            for param in ("t", "S", "N", "omega", "dN", "dS", "lnL"):
                self.assertTrue(param in params, version_msg +
                                ": '%s' not in parsed parameters" \
                                % (param))
                self.assertTrue(isinstance(params[param], float))
                if param != "lnL":
                    self.assertTrue(params[param] >= 0)

alpae:
@peterjc dual licence: of course, fine with me, also about being mentioned in news/contribs. my real name is Adrian Altenhoff. ;-)

brandoninvergo:
As for the `defaultdict` question, the problem is that the rest of the code does it the old-fashioned way. I'd rather see all of it updated at once to `defaultdict`.

brandoninvergo:
And what about this for a fix to the `parse_pairwise` function, which retains the current coding style of the rest of the file?  (sorry, I don't know how to properly use Github in cases like this)

```
def parse_pairwise(lines, results):
    """Parse results from pairwise comparisons.
    """
    # Find pairwise comparisons
    # Example:
    # 2 (Pan_troglo) ... 1 (Homo_sapie)
    # lnL = -291.465693
    #  0.01262 999.00000  0.00100
    #
    # t= 0.0126  S=    81.4  N=   140.6  dN/dS= 0.0010  dN= 0.0000  dS= 0.0115
    pair_re = re.compile("\d+ \((.+)\) ... \d+ \((.+)\)")
    pairwise = {}
    seq1 = None
    seq2 = None
    for line in lines:
        # Find all floating point numbers in this line
        if not seq1 and not seq2:
            pair_res = pair_re.match(line)
            if pair_res:
                seq1 = pair_res.group(1)
                seq2 = pair_res.group(2)
                pairwise[seq1] = {seq2: {}}
                pairwise[seq2] = {seq1: {}}
                continue
        else:
            line_floats_res = line_floats_re.findall(line)
            line_floats = [_nan_float(val) for val in line_floats_res]
            if len(line_floats) == 1:
                pairwise[seq1][seq2].update({"lnL": line_floats[0]})
                pairwise[seq2][seq1] = pairwise[seq1][seq2]
            elif len(line_floats) == 6:
                pairwise[seq1][seq2].update({"t": line_floats[0],
                                             "S": line_floats[1],
                                             "N": line_floats[2],
                                             "omega": line_floats[3],
                                             "dN": line_floats[4],
                                             "dS": line_floats[5]})
                pairwise[seq2][seq1] = pairwise[seq1][seq2]
    if pairwise:
        results["pairwise"] = pairwise
    return results
```

alpae:
@brandoninvergo your suggested test changes do not work, as in the test files the pairwise runs are done with more than 2 sequences. then you end up with several pairs, so the len(pairwise)==2 check generally fails. Also, because in a result file generally you find more than one pairwise result, only checking initially if seq1/seq2 are None is not enough: you need to test multiple times, but ignore float-lines until you have the first matching sequence pair...

I kept the spirit of your suggestions for both test and parse code, including the standard dict instead of the defaultdict (valid point that all other functions don't use it). Hope this is ok.
Cheers

brandoninvergo:
Ah right, I see my mistake with the updated tests.  You're correct.  Sorry, I've never used codeml in pairwise mode and I didn't look closely at the test results file.

@peterjc everything looks ok to me now.

PR #1014, opened 2016-11-29
Mark residues as disordered when parsing MMCIF

StefansM:
This commit updates MMCIFParser to match the behaviour of PDBParser (and the documentation) when parsing point mutations. This is done by calling the "init_residue" of StructureBuilder when the residue name changes, as well as when the residue ID changes.

### DIFF
--- a/Tests/test_MMCIF.py
+++ b/Tests/test_MMCIF.py
@@ -203,6 +203,52 @@ class ParseReal(unittest.TestCase):
         structure = parser.get_structure("example", open("PDB/1A8O.cif"))
         self.assertEqual(len(structure), 1)
 
+    def test_point_mutations(self):
+        """Test if the parser can parse point mutations correctly."""
+
+        def _run_parser(parser):
+            structure = parser.get_structure("example", "PDB/3JQH.cif")
+
+            # Residue 1 and 15 should be disordered.
+            res_1 = structure[0]["A"][1]
+            res_15 = structure[0]["A"][15]
+
+            # Cursory check -- this would be true even if the residue just
+            # contained some disordered atoms.
+            self.assertTrue(res_1.is_disordered(), "Residue 1 is disordered")
+            self.assertTrue(res_15.is_disordered(), "Residue 15 is disordered")
+
+            # Check a non-mutated residue just to be sure we didn't break the
+            # parser and cause everyhing to be disordered.
+            self.assertFalse(
+                structure[0]["A"][13].is_disordered(),
+                "Residue 13 is not disordered")
+
+            # Check that the residue types were parsed correctly.
+            self.assertSetEqual(
+                set(res_1.disordered_get_id_list()),
+                {"PRO", "SER"},
+                "Residue 1 is proline/serine")
+            self.assertSetEqual(
+                set(res_15.disordered_get_id_list()),
+                {"ARG", "GLN", "GLU"},
+                "Residue 15 is arginine/glutamine/glutamic acid")
+
+            # Quickly check that we can switch between residues and that the
+            # correct set of residues was parsed.
+            res_1.disordered_select('PRO')
+            self.assertAlmostEqual(
+                res_1["CA"].get_occupancy(),
+                0.83, 2, "Residue 1 proline occupancy correcy")
+
+            res_1.disordered_select('SER')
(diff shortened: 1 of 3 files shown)


peterjc:
@StefansM can you think of an extra test for this change? Even a one line assert added to an existing test method in ``test_PDB.py`` would be great.

@etal @JoaoRodrigues @lennax etc does this look good to merge otherwise?

StefansM:
@peterjc Sorry -- that was lazy of me. I have added some tests. I had to add an extra MMCIF file to the PDB directory because none of the existing ones contained point mutations, but it's only a small (67KiB) file.

Tests/test_MMCIF.py
```diff
@@ -203,6 +203,52 @@ def test_filehandle(self):
...
+            # Check a non-mutated residue just to be sure we didn't break the
+            # parser and cause everyhing to be disordered.
+            self.assertFalse(
+                structure[0]["A"][13].is_disordered(),
+                "Residue 13 is not disordered")
+
+            # Check that the residue types were parsed correctly.
+            self.assertSetEqual(
+                set(res_1.disordered_get_id_list()),
+                {"PRO", "SER"},
```
peterjc: This is fun to see - it might be the first set literal in Biopython as we've only just dropped support for Python 2.6 :)

Tests/test_MMCIF.py
```diff
@@ -203,6 +203,52 @@ def test_filehandle(self):
...
+                0.83, 2, "Residue 1 proline occupancy correcy")
+
+            res_1.disordered_select('SER')
+            self.assertAlmostEqual(
+                res_1["CA"].get_occupancy(),
+                0.17, 2, "Residue 1 serine occupancy correcy")
+
+        _run_parser(MMCIFParser(QUIET=True))
+        _run_parser(FastMMCIFParser(QUIET=True))
+
```
peterjc: Personally I'd restructure this test to have two test methods (normal and fast), something like this (untested):

``` python
    def _run_parser(self, parser):
        """Internal function for common test code."""
        structure = parser.get_structure("example", "PDB/3JQH.cif")
        # ...

    def test_point_mutations_main_parser(self):
        """Point mutations using MMCIFParser"""
        self._run_parser(MMCIFParser(QUIET=True))

    def test_point_mutations_fast_parser(self):
        """Point mutations using FastMMCIFParser"""
        self._run_parser(FastMMCIFParser(QUIET=True))
```

This means ``MMCIFParser`` and ``FastMMCIFParser`` will get tested separately (useful if for example ``MMCIFParser`` was breaking meaning ``FastMMCIFParser`` would not be tested). i.e. General principle of making individual tests smaller

But as it is looks fine.

peterjc:
@etal @JoaoRodrigues @lennax Good to merge? Can you think of a smaller test case?

Should we add a matching test using the PDB and MMTF versions of this file?

StefansM:
@peterjc Good idea to split daf4d17 into two test cases; I just added commit 317850c to do this.

It looks like `test_PDB.py` already does these tests by parsing a custom PDB file.

The MMTF tests just check whether a `Structure` parsed from an MMTF file matches the same structure from an MMCIF file, but  does not actually check the alternate locations of atoms or residues. This doesn't seem like the ideal approach, because a bug in the MMCIF parsers can cause tests for the MMTF parser to fail.

If I were to refactor `test_mmtf.py`, `test_MMCIF.py` and `test_PDB.py`, I would probably try to put structure tests into their own file, so that `Structure` objects built by `MMCIFParser`, `FastMMCIFParser`, `PDBParser` and `MMTFParser` are all tested by common code.

peterjc:
@StefansM that looks great - I'm happy with the pull request now. If you want to look at those further unit test changes later, that would be great.

However, before merging, I'd like to wait for a second opinion as I've never used the disordered residues stuff in PDB. Pinging @JoaoRodrigues @lennax @etal - or @abradle given the MMTF interaction - any comments would be welcome.

abradle:
I have no issues with the PR. I agree with the broad point about the testing for the MMTF parser. Not sure I have bandwidth currently.

peterjc:
Thanks @abradle for looking at this, I've gone ahead and merged it.

Thank you @StefansM :)

Issue #892, opened 2016-07-24
Biopython trie implementation can't load large data sets


twrightsman:
Migrated from https://redmine.open-bio.org/issues/3395

@mnowotka said:

> Imagine I have Biopython trie:
> 
> ```
> from Bio import trie
> import gzip
> 
> f = gzip.open('/tmp/trie.dat.gz', 'w')
> tr = trie.trie()
> #fill in the trie
> trie.save(f, trie)
> ```
> 
> Now /tmp/trie.dat.gz is about 50MB. Let's try to read it:
> 
> ```
> from Bio import trie
> import gzip
> 
> f = gzip.open('/tmp/trie.dat.gz', 'r')
> tr = trie.load(f)
> ```
> 
> Unfortunately I'm getting meaningless error saying:
> "loading failed for some reason"
> 
> Any hints?

twrightsman:
@peterjc replied:
> Can you try the same test case without gzip? i.e. Can you load /tmp/trie.dat rather than /tmp/trie.dat.gz?
> 
> Also I would try explicitly opening the files in binary mode.
> 
> P.S. Which OS, which version of Python, which version of Biopython?

twrightsman:
@mnowotka replied:
> Sure, I'll update this issue as soon as I check that.

twrightsman:
@mnowotka replied:
> OK, I tried using standard python file handler with explicit binary mode and it also failed. The file is now 165.5MB.
> I also tried bz2 and zip compression, without any luck...

twrightsman:
@peterjc replied:
> Well that is progress - it means this isn't a problem coming from reading a compressed file on disk - you've made the test case simpler. Can you actually share a self contained example script? If not, I suggest you try halving the dataset (only record the first half of the tries), and retest. Then repeat - this should tell you if the problem is as you suspect a large dataset, or something specific about a special value.
> 
> Alternatively can you share the (compressed) file? I could at least check if it fails the same way here, and perhaps add some debugging code to get more information.
> 
> The error message itself is coming from some C code, which hasn't changed for some time:
> https://github.com/biopython/biopython/blob/master/Bio/triemodule.c
> 
> The error itself is likely triggered in function _deserialize_transition in trie.c:
> https://github.com/biopython/biopython/blob/master/Bio/triemodule.c
> 
> You still haven't told us the important information of which OS, which version of Python, which version of Biopython. Given it is C code, I'd also like to know how Biopython was installed (e.g. did you compile it from source yourself).

[...]

PR #1015, opened 2016-11-29
Proposed fix Issue #892

noamkremen:
Problem due to PyObject_CallMethod(py_handle, "read", "i", length) at line number 693 returning null when a file with key lengths over 125 chars was opened in 'r' mode instead of 'rb' mode. This did not happen with shorter key lengths. 

[Peter: Edited to avoid hash+number which GitHub treats as a link to an issue or pull request]

### DIFF
--- a/Bio/triemodule.c
+++ b/Bio/triemodule.c
@@ -691,6 +691,12 @@ _read_from_handle(void *wasread, const int length, void *handle)
     }
 
     py_retval = PyObject_CallMethod(py_handle, "read", "i", length);
+    if (!py_retval)
+    {
+        PyErr_SetString(PyExc_RuntimeError, "Failed to read from file. "
+                "Verify that you did not open a binary file('rb') in text mode('r') or vice versa.");
+        goto error;
+    }
 #ifdef IS_PY3K
     if(!PyBytes_Check(py_retval)) {
 #else
--- a/Bio/triemodule.c
+++ b/Bio/triemodule.c
@@ -694,7 +694,7 @@ _read_from_handle(void *wasread, const int length, void *handle)
     if (!py_retval)
     {
         PyErr_SetString(PyExc_RuntimeError, "Failed to read from file. "
-                "Verify that you did not open a binary file('rb') in text mode('r') or vice versa.");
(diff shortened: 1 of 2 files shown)


Bio/triemodule.c
```diff
@@ -691,6 +691,12 @@ _read_from_handle(void *wasread, const int length, void *handle)
     }
 
     py_retval = PyObject_CallMethod(py_handle, "read", "i", length);
+    if (!py_retval)
+    {
+        PyErr_SetString(PyExc_RuntimeError, "Failed to read from file. "
+                "Verify that you did not open a binary file('rb') in text mode('r') or vice versa.");
```
peterjc: Can you put spaces before the open brackets? Otherwise it looks confusingly like calling the ``file`` function, or a hypothetical ``mode`` function.

Tests/test_trie.py
```diff
@@ -129,6 +129,36 @@ def test_with_prefix(self):
         self.assertEqual(set(['ANA', 'ANANA']),
                          set(trieobj.with_prefix("AN")))
 
+    def test_large_save_load(self):
+        """
```
peterjc: See PEP257 https://www.python.org/dev/peps/pep-0257/

The docstring should be something like this with a title line then a blank line:

``` python
    def test_large_save_load(self):
        """Generate random key/val pairs in three length categories,

        Uses 100 in each category. Insert them into a trie and into a reference dict.
        Write the trie to a temp file and read it back, verify that trie entries match
        the reference dict.
        """
```

The unittest module will also pull out the title line and print it at the command line when running ``test_trie.py`` directly (at least on recent versions of Python).

Bio/triemodule.c
noamkremen: Done

Tests/test_trie.py
noamkremen: Done

peterjc:
While this looks sensible to me, I'm curious why we sometimes get NULL returned here, and how the key length of 125 is important.

@mdehoon: As you did the Python 3 conversion, can you review this please?

peterjc:

noamkremen:
Sure, OK on both .

peterjc:
Thank you!

PR #1020, opened 2016-12-09
Replace assertTrue with assertIn/assertNotIn when suitable

### LINKED ISSUE #1006: Use self.assertIn in our unit tests
Until recently we supported Python 2.6, where the ``unittest`` library did not support the ``self.assertIn(search_term, thing_to_search)`` method meaning we used ``self.assertTrue(search_term in thing_to_search)`` instead, often with a message term.

I fixed a couple just now where there was a TODO comment: 0da38441dd3a220d62bd3edf09baf35dc2bd5c32

There are over 200 of these examples remaining, e.g.

```
$ cd Tests/
$ grep assertTrue *.py | grep " in "
common_BioSQL.py:        self.assertTrue("biosql-test" in server)
common_BioSQL.py:        self.assertTrue("biosql-test2" in server)
common_BioSQL.py:        self.assertTrue("biosql-test" in server)
...
```

Fixing these by hand would be extremely tedious, but with a little regular expression knowledge you can probably do this in one line of awk or perl etc. Or write a Python script to do it?

milindl:
This should close #1006 

assertTrue(a in b) is now assertIn(a, b)
assertTrue(a in b, b) is assertIn(a, b)
assertTrue(a in b, message) is assertIn(a, b, message)
The same for 'not in', by using assertNotIn()

In cases where there was any doubt, I just left them unchanged for safety.

Thanks!


Tests/test_Mafft_tool.py
```diff
@@ -87,7 +87,7 @@ def test_Mafft_simple(self):
         # became "Progressive alignment 1/2..." and "Progressive alignment 2/2..."
         self.assertTrue(("Progressive alignment ..." in stderrdata) or
                         ("Progressive alignment 1/" in stderrdata), stderrdata)
-        self.assertTrue("$#=0" not in stderrdata)
+        self.assertNotIn("$#=0" , stderrdata)
```
peterjc: You have an extra space before the comma, which is why TravisCI's pep8 check is failing (here and elsewhere)

Tests/test_Entrez_online.py
```diff
@@ -60,24 +60,24 @@ def test_parse_from_url(self):
...
-        self.assertTrue("id=15718680%2C157427902%2C119703751" in handle.url, handle.url)
+        self.assertIn(URL_TOOL, handle.url)
+        self.assertIn(URL_EMAIL, handle.url)
+        self.assertIn("id=15718680%2C157427902%2C119703751", handle.url)
         recs = list(Entrez.parse(handle))
         handle.close()
         self.assertEqual(3, len(recs))
         # arbitrary number, just to make sure the parser works
-        self.assertTrue(all(len(rec).keys > 5) for rec in recs)
+        self.assertIn(all(len(rec).keys > 5) for rec, recs)
```
peterjc: This is now giving a syntax error and causing TravisCI to fail the test. This is actually a generator expression inside an ``all(...)`` call.

Tests/test_Mafft_tool.py
milindl: Just noticed it, I'm on it. 

There was another fail, [https://travis-ci.org/biopython/biopython/jobs/182623109](https://travis-ci.org/biopython/biopython/jobs/182623109)

I've changed the line to the original version and added a TODO since I was unsure of the cause of the error.

peterjc:
Almost perfect - you need to fix the syntax error in ``Tests/test_Entrez_online.py`` line 70, and remove the extra spaces for PEP8 conformance.

To run the PEP8 check locally, ``pip install pep8`` and then this will match what we're currently enforcing on TravisCI via Tox:

```
```

peterjc:
Once those minor issues are fixed, this should be ready to merge.

Are you happy to dual licence your contributions under both the _Biopython License Agreement_ *and* the _BSD 3-Clause License_? See #898.

Thanks!

milindl:
I've fixed the spacing and the line 70 problem :)

Yes, I'm happy about both those things. 

Thanks!

peterjc:
Thank you very very much.

PR #1021, opened 2016-12-09
Add selftests for FastaIO

souravsingh:


Tests/test_AlignIO_FastaIO.py
```diff
@@ -72,3 +73,261 @@
                 print(" - %s: %r" % (k, record.annotations[k]))
     print("=" * 78)
 print("Finished tested reading files")
+
+class TestSelf(unittest.TestCase):
+    def test_example(self):
+            simple_example = \
```
peterjc: This line should be one level less indented.

Tests/test_AlignIO_FastaIO.py
```diff
@@ -72,3 +73,261 @@
...
+        45119 residues in 180 library sequences
+         Scomplib [34.26]
+         start: Tue May 20 16:38:45 2008 done: Tue May 20 16:38:45 2008
+         Total Scan time:  0.020 Total Display time:  0.010
+        
+        Function used was FASTA [version 34.26 January 12, 2007]
+        """
+
+    
+            alignments = list(FastaM10Iterator(StringIO(simple_example)))
```
peterjc: This whole block should be one level less indented

Tests/test_AlignIO_FastaIO.py
```diff
@@ -72,3 +73,261 @@
...
+            assert len(alignments) == 4, len(alignments)
+            assert len(alignments[0]) == 2
+            for a in alignments:
+                print("Alignment %i sequences of length %i"
```
peterjc: There shouldn't be any print statements in the unittest-based tests. Instead use things like ``self.assertEqual(123, a.get_alignment_length())``

Tests/test_AlignIO_FastaIO.py
```diff
@@ -309,7 +309,7 @@ def test_example(self):
         alignments = list(FastaM10Iterator(StringIO(simple_example)))
         assert len(alignments) == 4, len(alignments)
         assert len(alignments[0]) == 2
-        self.assertEqual(% (len(a), a.get_alignment_length()))
+        self.assertEqual(%(len(a), a.get_alignment_length()))
```
peterjc: What is this percent sign doing? Surely you just want to compare two integers with the syntax ``self.assertEqual(value_one, value_two)`` thus:

``` python
self.assertEqual(len(a), a.get_alignment_length())
```

Tests/test_AlignIO_FastaIO.py
```diff
@@ -72,3 +73,259 @@
...
+        for filename in files:
+            if os.path.splitext(filename)[-1] == ".m10":
+                print("")
+                print(filename)
+                print("=" * len(filename))
+                for i, a in enumerate(FastaM10Iterator(open(os.path.join(path, filename)))):
+                    print("#%i, %s" % (i + 1, a))
+                    for r in a:
+                        if "-" in r.seq:
+                            assert r.seq.alphabet.gap_char == "-"
```
peterjc: Within the unittest framework, you would write this as:

``` python
self.assertEqual(r.seq.alphabet.gap_char, "-")
```

Tests/test_AlignIO_FastaIO.py
```diff
@@ -325,7 +325,6 @@ def test_example(self):
                 for i, a in enumerate(FastaM10Iterator(open(os.path.join(path, filename)))):
                     print("#%i, %s" % (i + 1, a))
                     for r in a:
-                        if "-" in r.seq:
-                            assert r.seq.alphabet.gap_char == "-"
+                        self.assertEqual(r.seq.alphabet.gap_char, "-")
```
peterjc: You lost the if-statement, does it still work?

Tests/test_AlignIO_FastaIO.py
```diff
@@ -325,6 +325,7 @@ def test_example(self):
                 for i, a in enumerate(FastaM10Iterator(open(os.path.join(path, filename)))):
                     print("#%i, %s" % (i + 1, a))
                     for r in a:
+                        if "-" in r.seq:
```
peterjc: You now have invalid indentation because the next line should be indented.

peterjc:
You are making progress. One of the things our TravisCI setup is complaining about is PEP8 style, which you can check locally with the command line tool ``pep8`` which can usually by installed with ``pip install pep8``, then:

```
Tests/test_AlignIO_FastaIO.py:79:1: E302 expected 2 blank lines, found 1
Tests/test_AlignIO_FastaIO.py:87:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:92:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:97:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:104:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:185:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:196:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:243:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:306:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:310:1: W293 blank line contains whitespace
Tests/test_AlignIO_FastaIO.py:311:9: E303 too many blank lines (2)
Tests/test_AlignIO_FastaIO.py:320:1: W293 blank line contains whitespace
```

Right now it is only complaining about whitespace, which is easy to fix.

souravsingh:
The integration tests are throwing an assertion error at line 311:``assert len(alignments) == 4, len(alignments)``

peterjc:
Stylewise replace:

``` python
assert len(alignments) == 4, len(alignments)
```

with:

``` python
self.assertEqual(len(alignments), 4)
```

The most likely explanation is something is wrong with setting up the multi-line string ``simple_example``, and by eye I can see changes to the relative indentation (e.g. the line starting "vs"). Looking back at ``Bio/SeqIO/FastaIO.py`` this is probably why I wrote the example with the multiline string all having no Python indentation. Try copying that text EXACTLY, changing only the indentation of the line:

``` python
        simple_example = \
```

I've noticed something else which will be a problem later: You're adding to ``test_AlignIO_FastaIO.py`` which (was) an old fashioned print-and-compare test. Ideally we'd switch that to a unittest based one, but that is extra work right now. The simple alternative is to put your tests into a new file, named ``test_AlignIO_FastaIO_unittest.py`` perhaps?

souravsingh:
@peterjc The tests are now passing successfully.

peterjc:
Well done, this is a big step getting TravisCI to pass :)

You no longer need these changes to ``test_AligIO_FastaIO.py``,

``` python
from Bio.AlignIO import FastaIO
import unittest
```

peterjc:
Can you add this to the end of your new test file so that it can be run on its own (without ``run_tests.py``), which ought to let you see the exact output - which will show the output from the still undesirable print functions:

```
if __name__ == "__main__":
```

Can you then replace the prints with ``self.assertEqual(..., ...)`` lines?

souravsingh:
Is the PR ready for merging?

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,253 @@
...
+"""
+
+
+class TestSelf(unittest.TestCase):
+    def test_example(self):
+        alignments = list(FastaIO.FastaM10Iterator(StringIO(simple_example)))
+        self.assertEqual(len(alignments), 4)
+        self.assertEqual(len(alignments[0]), 2)
+        for a in alignments:
+            self.assertEqual((len(a), a.get_alignment_length()))
```
peterjc: Too many brackets - you are passing a single argument (a tuple of 2 elements) to the ``assertEqual`` method. Try:

```
self.assertEqual(len(a), a.get_alignment_length())
```

This is why TravisCI is currently failing,

```
ERROR: test_example (test_AlignIO_FastaIO_unittest.TestSelf)
  File "/home/travis/build/biopython/biopython/Tests/test_AlignIO_FastaIO_unittest.py", line 231, in test_example
    self.assertEqual((len(a), a.get_alignment_length()))
TypeError: assertEqual() missing 1 required positional argument: 'second'
```

souravsingh:
The results of the length of alignment don't match.

@peterjc Is this due to a mistake in copying of example or due to some other reason ?

peterjc:
Sorry, its something I really should have noticed last time - the alignment object is matrix like with a row and column count, and they are not equal:

```
FAIL: test_example (test_AlignIO_FastaIO_unittest.TestSelf)
  File "/home/travis/build/biopython/biopython/Tests/test_AlignIO_FastaIO_unittest.py", line 231, in test_example
    self.assertEqual(len(a), a.get_alignment_length())
AssertionError: 2 != 108
```

Likely going back to the original example this would make more sense:

``` python
self.assertEqual(2, len(a))
self.assertEqual(108, a.get_alignment_length())
```

souravsingh:
Another AssertionError is caught- ``AssertionError: 108 != 64``

How did the return value change?

peterjc:
Because this is inside a loop, ``for a in alignments:``, there are four different alignments of different sizes.

I suggest trying something like this:

``` python
rows = (2, 3, 4, 5)   # fill in real numbers here
cols = (108, 64, 123, 456)  # fill in real numbers
for a, rows, cols in zip(alignments, rows, cols):
   self.assertEqual(rows, len(a))
   self.assertEqual(cols, a.get_alignment_length())
```

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -228,8 +228,11 @@ def test_example(self):
         self.assertEqual(len(alignments), 4)
         self.assertEqual(len(alignments[0]), 2)
         for a in alignments:
-            self.assertEqual(2, len(a))
-            self.assertEqual(108, a.get_alignment_length())
+            rows = (2, 3, 4, 5)
+            cols = (108, 64, 123, 456)
+            for a, rows, cols in zip(alignments, rows, cols):
```
peterjc: You've accidentally got a double nested for loop here, there should be just one loop with ``a``.

peterjc:
Are you running ``test_AlignIO_FastaIO_unittest.py`` locally before pushing commits to GitHub? TravisCI reports it fails with:

```
FAIL: test_example (test_AlignIO_FastaIO_unittest.TestSelf)
  File "/home/travis/build/biopython/biopython/Tests/test_AlignIO_FastaIO_unittest.py", line 233, in test_example
    self.assertEqual(rows, len(a))
AssertionError: 3 != 2
```

souravsingh:
@peterjc The last commit I pushed passed the tests on my local machine.

peterjc:
Could you double check ``git status`` and  ``git diff`` in case you've missed something which needs committing, and also make sure you've got the latest Biopython installed (or run via ``run_tests.py``)?

souravsingh:
@peterjc I will be closing the PR and retry again.

peterjc:
OK - since you only changed two files, it ought to be easy to start again in a new branch and re-apply those changes with copy and paste. I appreciate git takes some time and practise to master.

PR #1023, opened 2016-12-12
using Affy.CelFile.read_v4(handle), when handle is not in binary mode results in a non-descriptive error

picrin:
A piece of code like this:

    import Bio.Affy.CelFile as CELFile
    with open("CELFileV4.CEL"), "r") as f:
        record = CELFile.read_v4(f)

fails in a non-descriptive way with `UnicodeDecodeError`. Instead, an error should be raised, which informs the user **what** went wrong, in this case, opening the file in a wrong mode.

This has already been in place with the version-detecting function `read`.

This pull request proposes a unified approach, which shares error handling between `read` and `read_v4`.

### DIFF
--- a/Bio/Affy/CelFile.py
+++ b/Bio/Affy/CelFile.py
@@ -19,17 +19,18 @@ except ImportError:
     raise MissingPythonDependencyError(
         "Install NumPy if you want to use Bio.Affy.CelFile")
 
+
+class ParserError(ValueError):
+    def __init__(self, *args):
+        super(ParserError, self).__init__(*args)
+
+_modeError = ParserError("You're trying to open an Affymetrix v4"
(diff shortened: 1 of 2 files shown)


picrin:
Just to clarify: I've used rebase-like workflow for this second commit in a row, as requested by Peter. I've also tried to treat exception handling a little bit more seriously, by introducing module's own ParserError.

The biggest weakness of this PR is the discrepancy in module's behaviour between python2 and python3 (just in exception handling), but then this discrepancy was already there, now it's just understood and exposed in a test case.

The biggest advantage is sensible error messages in python3.

peterjc:
Thank you

(Sorry for the slight delay, I'm on holiday and have not been checking GitHub as often as usual).

picrin:
Thank you Peter

Issue #1010, opened 2016-11-26
tests do not work on linux with the most recent master 


picrin:
4 tests never finish on linux:
    Tests/test_Entrez_online.py
    Tests/test_KEGG_online.py
    Tests/test_KGML_nographics.py
    Tests/test_NCBI_qblast.py

6 tests fail (stderr attached)

[failure.txt](https://github.com/biopython/biopython/files/614371/failure.txt)

stdout is:
    running test
    Python version: 2.7.11 (default, Sep  2 2016, 14:51:16) 
    [GCC 5.3.1 20160406 (Red Hat 5.3.1-6)]
    Operating system: posix linux2

peterjc:
I assume ``test_Affy`` is something you are working on?

At least some of your test failures suggest you didn't compile the C bits of Biopython before running the tests (you need to run ``python setup.py build`` first). e.g. ``test_QCPSuperimposer`` and ``_pwm``.

I know some of the TogoWS services seem to be down - I've not yet reported that to them. In general you can skip the network tests with:

```
python setup.py build
cd Tests
python run_tests.py --offline
```

picrin:
Thanks, all the test cases pass.

picrin:
I thought it would be a good idea to include this in the official README.md.

peterjc:
Since this is working, I'm going to close this issue. See discussion on pull request #1024 about how best to clarify the documentation about compiling before testing.

PR #1024, opened 2016-12-13
fix instructions to run tests in README.md

picrin:
Include instructions from #1010 on how to run tests locally.

### DIFF
--- a/README
+++ b/README
@@ -162,8 +162,12 @@ Testing
 Biopython includes a suite of regression tests to check if everything is
 running correctly. To run the tests, go to the biopython source code
 directory and type::
-
-    python setup.py test
+    
+    # build C extensions
+    python setup.py build
+    cd Tests
+    # skip online tests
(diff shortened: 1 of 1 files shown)


peterjc:
We do say earlier to run ``python setup.py build`` but repeating this would be clearer.

I would however prefer we leave ``python steup.py test`` as the recommendation, and only add the ``--offline`` approach as an option (if at all). Perhaps add:

```
If any of the test fail due to network issues, this is often down to
chance or a service outage. If the problem does not go away on
re-running the tests, you can deliberately run the offline test only
with this alternative:

   python setup.py build
   cd Tests
   python run_tests.py --offline
```

```

peterjc:
Alternative clarification to README file committed https://github.com/biopython/biopython/commit/58fb3a4abccf186981641929026ff380e698e9b4

Thank you for your feedback.

PR #1031, opened 2017-01-03
Tidy up GenBank processing code

biologyguy:
Revisiting the cleaning up from commit #945, where changes in functionality were annoyingly mingled with the tidying up.
 
- Remove redundant parentheses
- Decorate static methods as such
- Alpha-order long lists of attributes
- Define 'constant' attributes during object instantiation, instead of repeatedly setting them in methods
- Remove unnecessary 'self' from method variables
- Fix line indents
- Modify variable names to prevent shadowing outer scope


Bio/GenBank/Scanner.py
```diff
@@ -476,16 +477,16 @@ def parse_records(self, handle, do_features=True):
         """
         # This is a generator function
         while True:
-            record = self.parse(handle, do_features)
-            if record is None:
+            next_record = self.parse(handle, do_features)
```
peterjc: Given the loop only deals with a single record, why use ``next_record`` rather than ``record`` as the variable name?

Bio/GenBank/__init__.py
```diff
@@ -466,10 +466,10 @@ def __init__(self, debug_level=0, use_fuzziness=1,
     def parse(self, handle):
         """Parse the specified handle.
         """
-        self._consumer = _FeatureConsumer(self.use_fuzziness,
-                                          self._cleaner)
-        self._scanner.feed(handle, self._consumer)
-        return self._consumer.data
+        _consumer = _FeatureConsumer(self.use_fuzziness,
```
peterjc: I wonder if any existing code relied on the private attribute? Your version does seem cleaner...

Bio/GenBank/__init__.py
```diff
@@ -654,14 +662,14 @@ def size(self, content):
         """Record the sequence length."""
         self._expected_size = int(content)
 
-    def residue_type(self, type):
+    def residue_type(self, res_type):
```
peterjc: Technically this is a change to the method's API, which is why I'd left it as is - but this seems unlikely to cause any real trouble.

Bio/GenBank/Scanner.py
```diff
@@ -522,8 +523,8 @@ def parse_cds_features(self, handle,
                     # SeqRecord objects cannot be created with annotations, they
                     # must be added afterwards.  So create an empty record and
                     # then populate it:
-                    record = SeqRecord(seq=None)
-                    annotations = record.annotations
+                    new_record = SeqRecord(seq=None)
```
peterjc: Yes, ``new_record`` is more explicit, but as ``record`` was the only SeqRecord variable in this method wasn't the shorter variable name clear enough?

Bio/GenBank/__init__.py
```diff
@@ -523,7 +523,8 @@ def _unhandled(self, data):
     def __getattr__(self, attr):
         return self._unhandled
 
-    def _split_keywords(self, keyword_string):
+    @staticmethod
+    def _split_keywords(keyword_string):
```
peterjc: Good idea. As I recall when this was written we didn't have ``@staticmethod`` in Python, but we can use it now.

Bio/GenBank/Scanner.py
```diff
@@ -611,11 +615,11 @@ def parse_footer(self):
                 repr(self.line)
             # Remove tailing number now, remove spaces later
             linersplit = line.rsplit(None, 1)
-            if(len(linersplit) > 1):
+            if len(linersplit) > 1:
```
peterjc: Whomever wrote that must have done a lot of C code ;)

I wonder why we'd not spotted it during the PEP8 cleanup work?

peterjc:
I'm fairly happy with this as-is, with some fairly minor queries noted on specific lines. Don't feel you have to just revert the bits I queried - explaining your thinking might convince me.

Bio/GenBank/Scanner.py
biologyguy: `record` shadows an outer scope variable. I'd rather use the shorter version as well, but it's safer to use different variable names IMO. How about changing `next_record` to `rec`?

Bio/GenBank/__init__.py
biologyguy: Ya, good point. Seeing as the builtin `type` isn't used in the function, I guess is doesn't matter. Revert? Keep?

Bio/GenBank/Scanner.py
biologyguy: See above regrading outer scope shadowing.

Bio/GenBank/__init__.py
biologyguy: Unit tests still passed, so doesn't seem to be affecting much internally. If other people were accessing it, maybe there will be an angry email? We could also just set a null self._consumer in the __init__.

biologyguy:
Thanks for the code review, Peter.

Bio/GenBank/__init__.py
peterjc: I'd revert it just in case - the style benefit seems quite small.

Bio/GenBank/__init__.py
peterjc: It is a private property (leading underscore), so should be fine.

Bio/GenBank/Scanner.py
peterjc: I don't see a problem with reusing ``record`` here, it is short and consise. Perhaps my other programming language background blinds me to the shadowing problem - I am still sometimes caught out by Python's scoping rules.

biologyguy:
All contentious points resolved :)

peterjc:
Thank you :)

Issue #1030, opened 2017-01-03
AbiIO.py issue with reading file from .fsa file: seq referenced before assignment


graph1994:
I'm using SeqIO.read to read a fsa file from an Applied Biosystems 3730xl machine.
Fails to find the key of PBAS2 then doesn't set seq thus causing it to fail when used on line 411. I've found if I hard code the seq, smaple_id and qual array it is able to build with abif_raw intact and usable. 

3730xl does not have PBAS2 listed: [ABI Format docs](http://www6.appliedbiosystems.com/support/software_community/ABIF_File_Format.pdf)
I would like to extend this file to allow for 3730xl 

This is how I'm using it: `record = SeqIO.read(file, 'abi')`

Stack Trace:
```
Traceback (most recent call last):
  File "/Users/kgraff/Projects/fsa_interperter/fsa_reader.py", line 48, in <module>
    result = openFileCreatePlot(fileNameDir)
  File "/Users/kgraff/Projects/fsa_interperter/fsa_reader.py", line 20, in openFileCreatePlot
    record = SeqIO.read(file, 'abi')
  File "/Users/kgraff/anaconda/lib/python2.7/site-packages/Bio/SeqIO/__init__.py", line 664, in read
    first = next(iterator)
  File "/Users/kgraff/anaconda/lib/python2.7/site-packages/Bio/SeqIO/__init__.py", line 600, in parse
    for r in i:
  File "/Users/kgraff/anaconda/lib/python2.7/site-packages/Bio/SeqIO/AbiIO.py", line 411, in AbiIterator
    record = SeqRecord(Seq(seq, alphabet),
UnboundLocalError: local variable 'seq' referenced before assignment
```

peterjc:
@bow could you take a look at this please?

@graph1994 could you share a test file for debugging, ideally something we could include in future releases for our unit tests?

Thanks!

graph1994:
If you don't mind I'd like to try and help solve the problem, I'll try to find a test file that can be used and include that with my PR as well

peterjc:
Sure - Bow wrote the original code so would be the ideal person to review a pull request from you.

bow:
@graph1994, thanks for reporting the bug and feel free to explore how to solve the problem. The current parser does lack tests for .fsa files, so adding them would be much appreciated.

bow:
Closing this issue since this has been addressed in #1037.

PR #1037, opened 2017-01-08
Feature/abi io fsa file support

graph1994:
First take on Issue #1030 

Issue:
- AbiIO.py does not set seq,sample_id and qual properties causing a failure on line 409 when those fields are used. 

FIX:
 - <del>A  potential fix would be to check if the file ends with fsa then save the record with different values that exist in fsa files. </del>
- Added a try block when setting the record on line 409 then if it fails on NameError then treat it as a fsa file type 
- Add test.fsa file to Test directory 
- Changed url for ABI docs to in comments to latest

Needed Fixes:
- [x] Fix test function so that they work well with fsa file type


Bio/SeqIO/AbiIO.py
```diff
@@ -406,11 +406,21 @@ def AbiIterator(handle, alphabet=None, trim=False):
         file_name = basename(handle.name).replace('.ab1', '')
     except AttributeError:
         file_name = ""
-    record = SeqRecord(Seq(seq, alphabet),
-                       id=sample_id, name=file_name,
-                       description='',
-                       annotations=annot,
-                       letter_annotations={'phred_quality': qual})
+    if handle.name.endswith('.fsa'):
```
peterjc: You should not depend on the filename which could have been changed somewhere. Instead I would set each property according to the data you have.

graph1994: Very good point, came up with a different way of setting fsa details, set a try for record and if there is a NameError then try setting fsa

bow:
@graph1994 I just realized a review which I wrote a long time ago has not been submitted :( ~ I hope this is not too stale :).

In any case, @peterjc this PR should address #1062 as well.

graph1994:
Hey @bow, So I'll work on finding a better way to implement this then comment here with my plan of attack for the problem!

graph1994:
Here is the plan:
After looking at the [documention](http://www6.appliedbiosystems.com/support/software_community/ABIF_File_Format.pdf), there is a field that always exists in fsa formats but not in abi. This field is _'SpNm1': 'Fragment Analysis sample name'_  and the 'APFN2' field is on every abi file **_but not_** fsa. 
***
Where should the Check for this field happen: 
* It is failing on Line 409 from having no seq field set so it has to happen before that line.

bow:
@graph1994 Great, this is going in the right direction :).

A couple more comments:

* Since `trim` is a user-defined argument, I prefer to keep it out of the `if` block test to prevent cases where the file is `.fsa` but the user accidentally supplies `trim=True`. In these cases, `trim` should just be ignored or an error should be raised (I'm ok with both, but tend to go with just ignoring). Of course, this means the block containing `yield` must be adjusted as well.

* With `trim` out, I think checking for `SpNm1` alone is enough. If you do want to include `APFN` in the `if` block test, I think it's much clearer to have 3 `if` clauses:

    - 1 for `.fsa`, where `SpNm1` is present and not `APFN`
    - 1 for `.ab1`, where `SpNm1` is not present and `APFN` is
    - the other for all other invalid cases, where a `ValueError` or `AssertionError` is raised

I'm not completely sure about the last case (the invalid, unexpected one). This is where a little testing would give you an idea how conformant the files are. If there are `.ab1` files.

Also, you can move the `.ab1` filename extraction into the block where you are sure the file  is `ab1`.

graph1994:
Alright, Think I fixed it up but I was wondering if there is a better way to have this `if not trim or 'SpNm1' in raw:` maybe I should create boolean `isFsa = 'SpNm1' in raw ` since I'm checking the value twice?

Tests/test_SeqIO_AbiIO.py
```diff
@@ -211,9 +202,11 @@ def test_seqrecord(self):
...
         for trace in test_data:
             record = SeqIO.read(test_data[trace]['handle'], 'abi')
-            self.assertEqual(basename(test_data[trace]["path"][-1]).replace('.ab1', ''), record.name)
-            self.assertEqual(test_data[trace]['seq'], str(record.seq))
-            self.assertEqual(test_data[trace]['qual'], record.letter_annotations['phred_quality'])
+            if(trace != 'test_fsa'):
+                self.assertEqual(basename(test_data[trace]["path"][-1]).replace('.ab1', ''), record.name)
+                self.assertEqual(test_data[trace]['seq'], str(record.seq))
+                self.assertEqual(test_data[trace]['qual'], record.letter_annotations['phred_quality'])
+            print(record)
```
peterjc: Avoid prints in the unit tests.

graph1994:
Okay looks like this is finally in a good state, sorry for the delay and the amount of commits. Please let me know if there are any other ways I can help improve this. 😄

Bio/SeqIO/AbiIO.py
```diff
@@ -401,18 +401,33 @@ def AbiIterator(handle, alphabet=None, trim=False):
...
-        file_name = ""
-    record = SeqRecord(Seq(seq, alphabet),
-                       id=sample_id, name=file_name,
-                       description='',
-                       annotations=annot,
-                       letter_annotations={'phred_quality': qual})
-
-    if not trim:
+    # fsa check
+    if (set(['SpNm1', 'LIMS1', 'CTID1']).issubset(raw)):
```
bow: Since `set(['SpNm1', 'LIMS1', 'CTID1']).issubset(raw)` is used twice, I would put the result in a variable instead.

Also, I take it this means that it is not expected for any ab1 and/or fsa files to have just any of these tags?

Tests/test_SeqIO_AbiIO.py
```diff
@@ -211,9 +202,10 @@ def test_seqrecord(self):
         """Test if the extracted seqrecords data are equal to expected values."""
         for trace in test_data:
             record = SeqIO.read(test_data[trace]['handle'], 'abi')
-            self.assertEqual(basename(test_data[trace]["path"][-1]).replace('.ab1', ''), record.name)
-            self.assertEqual(test_data[trace]['seq'], str(record.seq))
-            self.assertEqual(test_data[trace]['qual'], record.letter_annotations['phred_quality'])
+            if(trace != 'test_fsa'):
```
bow: Minor style comment: this should be `if trace != 'test_fsa':` (no braces).

Bio/SeqIO/AbiIO.py
graph1994: Correct an fsa file should have all three of these tags to be valid, I named the variable is_fsa_file hope thats okay!

bow:
@graph1994 Thanks for making the changes. I think this looks good for merging now :). Any comments @peterjc?

peterjc:
Has this accidentally dropped ``data_3100 `` when adding ``test_fsa`` to ``test_SeqIO_AbiIO.py``?

Otherwise given @bow is the ABI code owner, a squash-merge sounds good.

graph1994:
Oh good catch on the accidental removal of 3100, I'll make sure to add that back in!

bow:
Good catch indeed @peterjc ~ I'll wait for the squash then.

graph1994:
Okay should be good now! Should we update the documentation on the website once this gets merged?

peterjc:
@graph1994: Are you happy to dual licence your contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

@bow: This looks OK now for a squash-merge. Do you want to do that (and add Kurt to the acknowledgements)?

graph1994:
Yes dual license is fine, and yeah that would be cool! Thanks both of you for the help/feedback! 🎉

bow:
@peterjc Sure, sometime later today I will push the squashed to `master` along with the updates in NEWS and CONTRIB.

@graph1994 Thank you for submitting the PR!

bow:
Squashed & merged to `master` in f19382d38 and noted in 5b9fcc712. Thanks again @graph1994!

PR #1039, opened 2017-01-19
Adding parameter to return print_that output instead of printing it

sbassi:
This change allows to return the output of print_that instead of sending to stdout. This is usefull when you need this output for a template system.

### DIFF
--- a/Bio/Restriction/PrintFormat.py
+++ b/Bio/Restriction/PrintFormat.py
@@ -87,8 +87,9 @@ class PrintFormat(object):
 
         return
 
-    def print_that(self, dct, title='', s1=''):
-        """PF.print_that(dct, [title[, s1]]) -> Print dct nicely formatted.
+    def print_that(self, dct, title='', s1='', print_=True):
+        """PF.print_that(dct, [title[, s1[, print_]]]) -> Print dct nicely
+        formatted.
 
         Arguments:
          - dct is a dictionary as returned by a RestrictionBatch.search()
@@ -97,6 +98,8 @@ class PrintFormat(object):
          - s1 is the title separating the list of enzymes that have sites from
            those without sites.
          - s1 must be a formatted string as well.
+         - print_ is a boolean to flag that the method will output to stdout if
+           True. If false, the output will be returned.
 
         The format of print_that is a list."""
         if not dct:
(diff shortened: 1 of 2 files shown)


peterjc:
I don't understand the use-case here versus just doing ```print(example.print_that(...))``` - could you explain what you meant about a template system in this context?

Rather than a new boolean argument, would an optional handle make more sense (e.g. ``sys.stdout`` or ``sys.stderr`` or a text mode file to write to)? (This might be irrelevant but I don't yet understand your use case.)

sbassi:
With template system I refer to jinja2 and similar.
To render the template I need to pass a dictionary with all values to render, but example.print_that() returns None (instead of returning something, it prints to stdout), so I can't grab the output of example.print_that().
In order to make the print_that() method to return something instead of printing it out, I added a new parameter. With the default value so the user can keep using is as it was before and have the same result.

peterjc:
OK - I see what you mean. Frankly I am not keen on methods which directly print, rather than the more flexible option of returning a string.

How about we introduce a new method like ``.format(...)`` (or similar name) returning the string? And then define ``.print_that(...)`` to just do ``print(self.format(...))``, and ideally deprecate ``.print_that(...)`` as well?

Its a bit more hassle, but leaves us with a cleaner API (although separately I'd like to retire the entire restriction module in favour of a re-write, see eg #268).

sbassi:
I agree, that is the way to go. Will work on that.

sbassi:
I called the new method .format_output() since format() was taken in that namespace.

sbassi:
@peterjc The code is ready for review

peterjc:
While ``Bio/Restriction/Restriction.py`` looks great, ``Bio/Restriction/PrintFormat.py`` needs updating to the new API style. Perhaps you forgot to commit changes to that file?

sbassi:
I thought it was not need because print_that inside PrintFormat is mean to be called from print_that in Restriction. But will modify it anyway.

peterjc:
Probably almost no one would be using that method, but it would be consistent to do both. Thanks.

peterjc:
Looks good, thank you - finally two administrative questions:

Are you happy to dual licence your Biopython contributions (including re-licensing your past contributions) under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

Also, are you happy to be thanked by name in the ``NEWS`` file for the next release (you are already listed in the ``CONTRIB`` file)?

sbassi:
Sorry I didn't answer. My answer is yes for both questions.

peterjc:
Thank you 👍

PR #1046, opened 2017-01-23
Add selftest for FastaIO

souravsingh:
Resumes again from PR #1021


Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,256 @@
...
+                    print("")
+                    print(filename)
+                    print("=" * len(filename))
+                    for i, a in enumerate(FastaIO.FastaM10Iterator(open(os.path.join(path, filename)))):
+                        print("#%i, %s" % (i + 1, a))
+                        for r in a:
+                            if "-" in r.seq:
+                                self.assertEqual(r.seq.alphabet.gap_char, "-")
+                            else:
+                                assert not hasattr(r.seq.alphabet, "gap_char")
```
peterjc: Probably use:

``` python
self.assertFalse(hasattr(r.seq.alphabet, "gap_char"))
```

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,256 @@
...
+        rows = (2, 3, 4, 5)
+        cols = (108, 64, 123, 456)
+        for a, rows, cols in zip(alignments, rows, cols):
+            self.assertEqual(rows, len(a))
+            self.assertEqual(cols, a.get_alignment_length())
+            for r in a:
+                print("%s %s %i" % (r.seq, r.id, r.annotations["original_length"]))
+            # print(a.annotations)
+            print("Done")
+            path = "Fasta/"
```
peterjc: Everything after this should be a new test method (and less indented, there is no reason to do this four times as part of the above for loop), e.g. ``def test_m10_files(self):``

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,256 @@
...
+        alignments = list(FastaIO.FastaM10Iterator(StringIO(simple_example)))
+        self.assertEqual(len(alignments), 4)
+        self.assertEqual(len(alignments[0]), 2)
+        rows = (2, 3, 4, 5)
+        cols = (108, 64, 123, 456)
+        for a, rows, cols in zip(alignments, rows, cols):
+            self.assertEqual(rows, len(a))
+            self.assertEqual(cols, a.get_alignment_length())
+            for r in a:
+                print("%s %s %i" % (r.seq, r.id, r.annotations["original_length"]))
```
peterjc: Rather than a print statement, this would ideally be set of comparisons:

``` python
self.assertEqual(r.seq, ...)
self.assertEqual(r.id, ...)
self.assertEqual(r.annotations["original_length"], ...)
```

You'd need to supply the expected values for each row of each of the four alignments, a bit like how the code above provides the expected number of rows and columns for each of the four alignments.

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,256 @@
...
+            print("Done")
+            path = "Fasta/"
+            files = sorted(f for f in os.listdir(path) if os.path.splitext(f)[-1] == ".m10")
+            for filename in files:
+                if os.path.splitext(filename)[-1] == ".m10":
+                    print("")
+                    print(filename)
+                    print("=" * len(filename))
+                    for i, a in enumerate(FastaIO.FastaM10Iterator(open(os.path.join(path, filename)))):
+                        print("#%i, %s" % (i + 1, a))
```
peterjc: Again, no print calls when using the ``unittest`` framework. Here the simplest option would be remove the print statement.

peterjc:
Something is going wrong on TravisCI,

```
FAIL: test_example (test_AlignIO_FastaIO_unittest.TestSelf)
  File "/home/travis/build/biopython/biopython/Tests/test_AlignIO_FastaIO_unittest.py", line 233, in test_example
    self.assertEqual(rows, len(a))
AssertionError: 3 != 2
```

Tests/test_AlignIO_FastaIO_unittest.py
souravsingh: Done

Tests/test_AlignIO_FastaIO_unittest.py
souravsingh: Done

Tests/test_AlignIO_FastaIO_unittest.py
souravsingh: Done

Tests/test_AlignIO_FastaIO_unittest.py
souravsingh: I might have misunderstood the instructions. So I would urge to take a look and suggest any corrections.

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+class TestSelf(unittest.TestCase):
+    def test_example(self):
+        alignments = list(FastaIO.FastaM10Iterator(StringIO(simple_example)))
+        self.assertEqual(len(alignments), 4)
+        self.assertEqual(len(alignments[0]), 2)
+        rows = (2, 3, 4, 5)
+        cols = (108, 64, 123, 456)
+        for a, rows, cols in zip(alignments, rows, cols):
```
peterjc: Apologies, I didn't notice the accidental re-use of the variable names ``rows`` and ``cols`` here. This should be something like:

``` python
row_counts = (2, 3, 4, 5)
col_counts = (108, 64, 123, 456)
for a, rows, cols in zip(alignments, row_counts, col_counts):
```

Tests/test_AlignIO_FastaIO_unittest.py
peterjc: Right now ``rows`` is just a single integer, the number of rows for the current alignment.

You would need to define a (nested) list of expected values for each row, similarly to how we defined the expected number of rows and columns.

peterjc:
I see now why TravisCI is failing - the number of rows `` (2, 3, 4, 5)`` and number of columns ``(108, 64, 123, 456)`` were intended to be obviously fake examples for you to fill in.

Tests/test_AlignIO_FastaIO_unittest.py
souravsingh: So changing the variable names should help in passing the tests, correct?

Tests/test_AlignIO_FastaIO_unittest.py
```diff
@@ -0,0 +1,255 @@
+from Bio._py3k import StringIO
+from Bio.AlignIO import FastaIO
+import os
+import unittest
+
+simple_example = \
```
peterjc: You've lost potentially important white space information compared to the string as used in ``Bio/AlignIO/FastaIO.py``

peterjc:
OK, looking at your proposed test, where I have added some comments.

```python
    def test_example(self):
         alignments = list(FastaIO.FastaM10Iterator(StringIO(simple_example)))
         self.assertEqual(len(alignments), 4)
         self.assertEqual(len(alignments[0]), 2)  # this is redundant with the first loop entry
         row_counts = (2, 3, 4, 5)  # some of these entries are place-holders (i.e. wrong!)
         col_counts = (108, 64, 123, 456)  # some of these entries are place-holders (i.e. wrong!)
         for a, rows, cols in zip(alignments, row_counts, col_counts):
             self.assertEqual(rows, len(a))
             self.assertEqual(cols, a.get_alignment_length())
             for r in a:
                 self.assertEqual(r.seq, rows)  # problematic, see below
                 self.assertEqual(r.id, rows)  # problematic, see below
                 self.assertEqual(r.annotations["original_length"], rows)  # problematic, see below
             print("Done")  # avoid prints in the finished test
```

Right now this fails as follows,

```
FAIL: test_example (test_AlignIO_FastaIO_unittest.TestSelf)
  File "/home/travis/build/biopython/biopython/Tests/test_AlignIO_FastaIO_unittest.py", line 236, in test_example
    self.assertEqual(r.seq, rows)
AssertionError: Seq('SGSNT-RRRAISRPVRLTAEED---QEIRKRAAECGKTVSGFLRAAALGKKVNS...LSR', Gapped(ProteinAlphabet(), '-')) != 2
```

Clearly it is meaningless to compare ``r.seq`` (which is a sequence object) to the ``rows`` variable which is an integer (here ``2``).

Likewise the next line of the test tries to compare ``r.id`` (which is a string) to the number of rows.

Likewise the line after that tries to compare ``r.annotations["original_length"]`` which I expect to be an integer to the number of rows. It might match the number of columns, i.e. variable ``cols`` but I think it more likely to say ``r.annotations["original_length"] <= cols`` as there could be gaps added to the sequence.

For now, remove/comment out this bit and focus on getting the expected number of rows and columns correct:

```python
             for r in a:
                 self.assertEqual(r.seq, rows)
                 self.assertEqual(r.id, rows)
                 self.assertEqual(r.annotations["original_length"], rows)
             print("Done")
```

Issue #980, opened 2016-11-05
Deuterium atoms set atom.element to ""


mathuin:
For historical reasons I'm sure, the PDB files that contain deuterium atoms have "D" in the element column.  

When Bio.PDB.Atom._assign_element() at https://github.com/biopython/biopython/blob/5045a7a3e86d5b32e0eaab941ab35daac86c59f8/Bio/PDB/Atom.py#L73-L100 receives "D" as an argument, it fails the `putative_element.capitalize() in IUPACData.atom_weights` test and so atom.element is set to "".

I expected atom.element to be set to "D" or to "H", not to "".

If "D" is acceptable, then line 90 should be changed to check for the "D" value or presence in the `IUPACData.atom_weights` structure.

If "H" is preferred, an explicit test for "D" (and setting to "H")  just before line 90 would work fine.

The only reason I haven't submitted a pull request is that I don't know which way is preferred.

Tell me, and I'll write it!  Thanks!

JoaoRodrigues:
Hi Jack,

I'd just add the element to the IUPACData.atom_weights dictionary. Can you
show an example of deuterium? How is the naming? Follows regular Hs
convention?

Thanks for this!

João

2016-11-04 17:15 GMT-07:00 Jack Twilley <notifications@github.com>:

peterjc:
According to wikipedia, IUPAC does use D as deuterium so that should be fine...

mathuin:
5AE0 A/204 is a disordered HIS residue with a number of deuterium atoms.

Here are the ones from the A altloc:

```
ATOM    998  D  AHIS A 204      51.975 -10.411 157.129  0.51 11.38           D  
ATOM    999  DA AHIS A 204      54.505 -11.036 157.739  0.51 16.69           D  
ATOM   1000  DB2AHIS A 204      53.389  -8.970 156.146  0.51 17.82           D  
ATOM   1001  DB3AHIS A 204      54.920  -8.918 156.555  0.51 17.82           D  
ATOM   1002  DD1AHIS A 204      52.980  -9.936 153.949  0.51 34.62           D  
ATOM   1003  DD2AHIS A 204      56.305 -11.286 155.446  0.51 23.83           D  
ATOM   1004  DE1AHIS A 204      53.919 -11.317 152.283  0.51 31.17           D  
```

There might be others out there, I only noticed this one because it triggered a corner case in some code I'm working on.  The naming appears to follow the same convention as H.

JoaoRodrigues:
[...]

PR #1049, opened 2017-01-24
Added support for deuterium as an element.

mathuin:
For some proteins in the PDB, it is important for researchers to
distinguish between hydrogen isotopes.  Those researchers will often
populate the element field with 'D' for deuterium.  This usage is
in accordance with IUPAC provisional recommendation IR-3.3.2, but
there is no reference to deuterium in the atom_weights table in the
IUPACData.py file so these values are not recognized.

The value used for the atomic weight of deuterium was taken from
http://www.ciaaw.org/hydrogen.htm.

For the purposes of testing this change, one of the HETATM lines in the
a_structure.pdb file was modified to reference a deuterium atom instead
of the carbon atom it originally represented.  This required minor
changes to the ParseTest class.  A specific test for this change was
added to the Atom_Element class.

### DIFF
--- a/Bio/Data/IUPACData.py
+++ b/Bio/Data/IUPACData.py
@@ -297,6 +297,7 @@ extended_protein_weight_ranges, avg_extended_protein_weights = \
 # Taken from http://www.chem.qmul.ac.uk/iupac/AtWt/ & PyMol
 atom_weights = {
     'H': 1.00794,
+    'D': 2.01410,
     'He': 4.002602,
     'Li': 6.941,
     'Be': 9.012182,
--- a/CONTRIB
+++ b/CONTRIB
@@ -85,6 +85,7 @@ Hongbo Zhu <https://github.com/hongbo-zhu-cn>
 Hye-Shik Chang <perky at domain fallin.lv>
 Iddo Friedberg <idoerg at domain burnham.org>
 Jacek Śmietański <https://github.com/dadoskawina>
+Jack Twilley <https://github.com/mathuin>
 James Casbon <j.a.casbon at domain qmul.ac.uk>
 Jason A. Hackney <jhackney at domain stanford.edu>
 Jeff Hussmann <first dot last at gmail dot com>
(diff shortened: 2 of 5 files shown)


mathuin:
This pull request is intended to close issue #980.

I am happy to dual license my contributions under both the _Biopython License Agreement_ and the _BSD 3-Clause License_ in accordance with #898.

I am also happy to be thanked by name (Jack Twilley) in the `NEWS` and `CONTRIB` files, should this contribution warrant such recognition.

peterjc:
This would deserve having your name in the ``NEWS`` and ``CONTRIB`` files. Note the sort order is now enforces, i.e. Make sure this passes:

```
$ grep -v "^#" CONTRIB | LC_ALL=C sort -u -c
```

(It looks like that is broken on the master, which I will fix now...)

peterjc:
The TravisCI problem with ``CONTRIB`` sorting is fixed on the master now as of 6f7a324c158b0eb1d0ebc03d5d237fdaf984bd88 - other than that the test results for this pull request are all fine.

mathuin:
I updated my branch to include myself in the `NEWS` and `CONTRIB` files.

If you'd like me to merge the current master into the branch, let me know -- I suspect that it may result in a merge conflict or at the least a merge commit, but if you're okay with that I'm okay with that.

Alternatively, I can generate a patch and apply it to current master in a new PR or in a force-push to this one if you'd prefer a cleaner history.  Just let me know!

peterjc:
It is not your fault that TravisCI is failing on pypy3, which on the master branch is currently an allowed fail (there is a tox dependency problem which I hope will be sorted out before too long).

There isn't any need to update the branch at this point. For future reference when there are conflict due to changes on the master branch, rather than a merge, I would suggest a rebase (``git rebase origin/master``) if you are comfortable with that.

@JoaoRodrigues are you happy for this to be committed? Either say say, or go ahead and merge via Github (using the "rebase and merge" action would be fine) please.

JoaoRodrigues:
Only got the notifications from this and the other thread today, sorry.. 

Thanks @mathuin ! Sounds perfect @peterjc , I'll rebase and merge then.

peterjc:
Thank you both 👍

PR #1050, opened 2017-01-27
new options use_sw_tback and comp_based_stats options

benfulton:
As of blast 2.6, use_sw_tback and comp_based_stats options are available for rpstblastn (They are already available for various other versions).


peterjc:
Looks good, waiting for TravisCI to finish...

PR #1055, opened 2017-02-08
MAF AlignIO code (reprise)

peterjc:
This new pull request and branch was created by rebasing #350 from @adamnovak (based on the work by @polyatail and me dating back to 2011) with minor updates like further PEP8 compliance in order to pass our TravisCI configuration checks. @bow had reviewed the original pull request.

After merging this, I would like to include most of the changes from @blaiseli on #503 which was also based on the above branch but didn't include the full history. See discussion on #503 about that.

I would appreciate at least one of you to look at this before I merge it, in case I've made any accidental mistakes - there were a lot of manual merge conflicts to resolve. Thanks!


Bio/AlignIO/MafIO.py
```diff
@@ -0,0 +1,581 @@
...
+        # for now, use ._annotations private property, but restrict keys to those
+        # specifically supported by the MAF format, according to spec
+        try:
+            anno = " ".join(["%s=%s" % (x, y) for x, y in alignment._annotations.items() if x in ("score", "pass")])
+        except AttributeError:
+            anno = "score=0.00"
+
+        self.handle.write("a %s\n" % (anno,))
+
+        recs_out = 0
```
bow: minor tidbit: this block, down to the for loop can be shortened into this:

```
for recs_out, record in enumerate(alignment, start=1):
    self._write_record(record)
```

This is functionally the same with the current code, just more succinct :).

bow:
Everyone,

I've added some comments on the PR. Nothing major (i.e. nothing on the parsing logic itself), since I have not used MAF that much to be familiar with all its nuances. In general, the PR (after adressing the comments) seems ok and ready for merge.

Bio/AlignIO/MafIO.py
peterjc: The explicit counting is consistent with existing usage in ``Bio.SeqIO`` and ``Bio.AlignIO``, so it would be  worth updating all of them at once. But I need to be convinced this is clearer... I failed to find a Python idiom on this - so I've asked on http://stackoverflow.com/questions/42207102/python-idiom-for-counting-loop-execution

peterjc:
Thanks Bow :)

I'm unconvinced about the loop counting style question, but that could be addressed in a follow up pull request as there are lots of other examples already in the code-base.

You said minor comments (plural), but that was the only one I can see...

bow:
Hey Peter,

Ah sorry. The one you saw (about using `enumerate`) is actually a comment (i.e. it's fine as it is; I made the remark mostly to know what you think about using `enumerate` instead).
 The other ones are probably not visible to you since they are the ones I actually marked as 'review'. It seems that for these I need to do an additional submission?

Anyway, I'm doing that now and hopefully you should be able to see them :).

peterjc:
Thanks - all the other comments have come through now. I'm still not quite sure how the (relatively recent) Git review vs comment thing works myself.

Bio/AlignIO/MafIO.py
```diff
@@ -0,0 +1,581 @@
...
+            pass
+        elif not line:
+            break
+
+
+class MafIndex(object):
+    def __init__(self, sqlite_file, maf_file, target_seqname):
+        """Indexes or loads the index of a MAF file"""
+
+        import os
```
peterjc: Certainly lets put ``import os`` at top level.

For SQLite3 there's the corner case of systems like Jython (and self-compiled Python missing library headers) where that import will fail, which we already have to handle for ``Bio.SeqIO.index`` etc.

Bio/AlignIO/MafIO.py
```diff
@@ -0,0 +1,581 @@
...
+
+                records.append(SeqRecord(Seq(sequence, alphabet),
+                               id=line_split[1],
+                               name=line_split[1],
+                               description="",
+                               annotations=anno))
+            elif line.startswith("e") or \
+                 line.startswith("i") or \
+                 line.startswith("q"):
+                # not implemented
```
peterjc: Good question, or as a middle ground:

``` python
import warnings
...
warnings.warn("MAF line type %r not implemented" % line[0], BiopythonParserWarning)
```

If the extra lines are fairly minor then ignoring them silently is probably OK.

Bio/AlignIO/MafIO.py
```diff
@@ -0,0 +1,581 @@
...
+You are expected to use this module via the Bio.AlignIO functions(or the
+Bio.SeqIO functions if you want to work directly with the gapped sequences).
+
+"""
+from Bio.Alphabet import single_letter_alphabet
+from Bio.Seq import Seq
+from Bio.SeqRecord import SeqRecord
+from Bio.Align import MultipleSeqAlignment
+from .Interfaces import SequentialAlignmentWriter
+import shlex
```
peterjc: Good catch. I've recently started using flake8 on the codebase and found (and fixed) other unused imports in the code base.

Bio/AlignIO/MafIO.py
peterjc: http://genome.ucsc.edu/FAQ/FAQformat.html#format5 says

> Some MAF files may contain other optional line types:
> 
> an "i" line containing information about what is in the aligned species DNA before and after the immediately preceding "s" line
> an "e" line containing information about the size of the gap between the alignments that span the current block
> a "q" line indicating the quality of each aligned base for the species

As they are optional an error seems too much. Silently ignoring them (as now) seems reasonable.

Bio/AlignIO/MafIO.py
```diff
@@ -0,0 +1,581 @@
...
+        except ImportError:
+            from Bio import MissingPythonDependencyError
+            raise MissingPythonDependencyError("Requires sqlite3, which is "
+                                               "included Python 2.5+")
+
+        self._target_seqname = target_seqname
+        self._maf_file = maf_file
+
+        # make sure maf_file exists, then open it up
+        if os.path.isfile(self._maf_file):
```
peterjc: In this case I am OK with passing any exception straight back to the user if the file does exist but cannot be opened (e.g. due to file permissions). This also avoids the question of if raising a ``ValueError`` is the best choice.

peterjc:
OK, I think I have addressed @Bow's useful comments, plus added a tiny bit more documentation to the main ``MafIO`` docstring, and updated the ``CONTRIB`` and ``NEWS`` files. The later caused a merge conflict which I have resolved by rebasing this.

bow:
Looks good ~ this seems ok for merging now :).

peterjc:
TravisCI has caught the fact I didn't update ``test_old_file_not_found`` with the exception change. I'll look at that, then merge. Thanks!

peterjc:
The PyPy test is failing, logged as #1059

peterjc:
Thank you everyone 👍 

Next up, closing https://github.com/biopython/biopython/pull/504 with @blaiseli's additions.

peterjc:
I've updated the former wiki page https://github.com/biopython/biopython.github.io/commit/0dee498baba107be49a70393eaf74c01447135b6 now this will be in Biopython 1.69, but have not looked at the actual content.

PR #1056, opened 2017-02-10
Deprecate Bio.DocSQL

peterjc:
See also:

http://lists.open-bio.org/pipermail/biopython/2017-January/016054.html
http://lists.open-bio.org/pipermail/biopython-dev/2017-January/021627.html

To date there have been no objections (or comments) on the mailing list about deprecating this.


peterjc:
(Those style commits were not meant to be on this branch and pull request, but I would have applied them anyway at some point)

PR #1057, opened 2017-02-10
Return FASTQ and SFF quality scores as an array (RFC, WIP)

peterjc:
**Request for comment, work in progress**

Previously the FASTQ and SFF parsers returned a list of integers, here we return a unsigned byte integer array instead.

Note could do this on ABI and QUAL parser too, although the later has a test case which causes an overflow.

This change in itself is a proof of principle, and to allow checking of side effects. Step two would be to use this for speeding up the parser...

### DIFF
--- a/Bio/SeqIO/QualityIO.py
+++ b/Bio/SeqIO/QualityIO.py
@@ -118,7 +118,8 @@ EAS54_6_R1_2_1_413_324 CCCTTCTTGTCTTCAGCGTTTCTCC
 EAS54_6_R1_2_1_540_792 TTGGCAGGCCAAGGCCGATGGATCA
 EAS54_6_R1_2_1_443_348 GTTGCTTCTGGCGTGGGTGGGGGGG
 
-The qualities are held as a list of integers in each record's annotation:
+The qualities are held as an array of signed byte integers in each record's
+annotation:
 
 >>> print(record)
 ID: EAS54_6_R1_2_1_443_348
@@ -128,7 +129,7 @@ Number of features: 0
 Per letter annotation for: phred_quality
 Seq('GTTGCTTCTGGCGTGGGTGGGGGGG', SingleLetterAlphabet())
 >>> print(record.letter_annotations["phred_quality"])
-[26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 24, 26, 22, 26, 26, 13, 22, 26, 18, 24, 18, 18, 18, 18]
+array('b', [26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 24, 26, 22, 26, 26, 13, 22, 26, 18, 24, 18, 18, 18, 18])
 
 You can use the SeqRecord format method to show this in the QUAL format:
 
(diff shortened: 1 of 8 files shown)


lgautier:
The gist of what is in https://github.com/lgautier/fastq-and-furious/issues/2#issuecomment-278835739 is that turning the quality string in a `list` of integer as done in `SeqIO` is taking about as much time as the rest the parsing. Using an `array` **and** moving the conversion down to C can result in a very significant gain or performance.

peterjc:
Looking at this fresh, an "unsigned char" integer would typically be more elegant ("B"), but using a signed integer ("b") is needed for the legacy Solexa FASTQ negative quality scores.

I'm going to rebase this - the recent changes to the ``Seq`` object to use bytes internally may make the quality overhead more obvious.

mdehoon (approved):

peterjc:
Looks like I missed some of the tests, will have to re-rebase or tweak this further.

peterjc:
Rebasing for f-string work, and hopefully fixed the doctest too.

Thinking about this, we can probably use a signed byte integer for Solexa scores (which can be negative), and an unsigned byte integer for PHRED scores (positive or zero) which would allow some head room for higher values in QUAL files etc.

Another question is if numpy arrays are preferable to the standard library array - they may turn out to be better for speed?

peterjc:
Playing with doing FASTQ mapping of a quality string to a bytes array (with the offset mapping done via the ``.translate`` method), it is clearly going to be more efficient to replace ``FastqGeneralIterator`` which works in text mode yielding unicode strings, with an equivalent iterator in binary mode yielding bytes. That change can be done separately, and should show a speed up on top of help of recent ``Seq`` changes to use bytes internally.

peterjc:
I broke a Tutorial example, 

```
  File "/Users/runner/work/biopython/biopython/Tests/test_Tutorial.py", line 244, in test_doctests
    raise ValueError(
ValueError: 1 Tutorial doctests failed: test_chapter_seq_annot_line_00848

```

Need to update this (and others like it):

```
%doctest ../Tests/Quality
\begin{minted}[breaklines=true,breaksymbolleft=]{pycon}
>>> from Bio import SeqIO
>>> record = next(SeqIO.parse("example.fastq", "fastq"))
>>> len(record)
25
>>> print(record.seq)
CCCTTCTTGTCTTCAGCGTTTCTCC
>>> print(record.letter_annotations["phred_quality"])
[26, 26, 18, 26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 26, 22, 26, 26, 26, 26, 26, 26, 26, 23, 23]
\end{minted}
```

The ``repr(...)`` output change here and in the embedded doctests is unavoidable, but I wonder how much real world code this list/array change would actually break?

peterjc:
I was thinking specifically of vectorised operations in C via numpy for the ASCII to Quality Score offset calculation potentially being faster, however using the bytes ``.translate(...)`` method might prove just as good.

lgautier:
If a simple "char as a byte" -> "uint8 as a byte" conversion, `./translate()` might not only be as good but faster.

lgautier:
Speaking about speed, I benchmarked `fastq-and-furious` with its C-extension against `pyfastx`. They are pretty much equivalent in speed and that's likely the limit of what can be done with Python on the parsing side. At least with a single process. `pyfastx` has a low single-digit speed advantage but at the cost of being monolithic. It takes a path to a gzip-compressed FASTQ as string and goes straight down to C whereas `fastq-and-furious` can take any Python `io` object (so works with streams and compression schemes there is a Python wrapper for). If I am correct to think that this is the speed limit, `FastqGeneralIterator` can only hope to get up to 50% faster by moving things to C (from the speed I observed when I ran my benchmark). As you get closer to that number I guess there will be a diminishing return in trying to get faster.

PR #1067, opened 2017-02-23
README.rst, NEWS.rst, CONTRIB.rst, DEPRECATED.rst etc

peterjc:
Renames our legacy extension-less text files to use ``*.rst`` and tweak their contents as needed to render nicely as reStructuredText which looks much nicer under GitHub etc.

Creates symlinks in order to preserve the legacy filenames (although I'm not sure how important that is?).

See also http://mailman.open-bio.org/pipermail/biopython/2017-February/016083.html and http://mailman.open-bio.org/pipermail/biopython-dev/2017-February/021644.html

Pending an enhancement of ``rst-lint`` from https://github.com/twolfson/restructuredtext-lint/ to allow it to ignore "info" level issues, I plan to add checking the markup to our TravisCI setup.


peterjc:
Note this does not introduce back-ticks everywhere, e.g. ``Bio.SeqIO``, which would be a nice further step in future.

It has also made me even more aware that there is no clear ordering convention in the ``DEPRECATED`` file, which might be worth addressing at a later date.

cbrueffer:
Looks fine to me!

peterjc:
Given no objections and one endorsement, applying now...

PR #1068, opened 2017-02-24
remove assertion comparing Count and IdList

jimmyodonnell:
search_result["IdList"] is constrained by the 'retmax' argument to Entrez.esearch, while search_result["Count"] is the total number of records found. Count can exceed the length of the IdList; this assertion can cause trouble or confusion.

### DIFF
--- a/Doc/Tutorial/chapter_entrez.tex
+++ b/Doc/Tutorial/chapter_entrez.tex
@@ -1299,7 +1299,6 @@ additional argument of \verb|usehistory="y"|,
 \begin{verbatim}
 >>> gi_list = search_results["IdList"]
 >>> count = int(search_results["Count"])
->>> assert count == len(gi_list)
 \end{verbatim}
 
 \noindent However, you also get given two additional pieces of information, the {\tt WebEnv} session cookie, and the {\tt QueryKey}:


peterjc:
This would just hide the issue. How about replacing it with an assert like ``len(len(gi_list) == min(count, retmax)`` with an explanation of this in the text?

jimmyodonnell:
That would work, too. I should ask, then - what is the intent of the assert here?

peterjc:
Defensive code by example, making the point that (usually) the id-count should match the count value given. Clearly as you've found this isn't always true - but you've understood why. Its possible that without the assert, you might not have noticed some of the expected results were silently missing due to the default return limit?

jimmyodonnell:
Gotcha; I agree with that motivation. 

Because search_result["Count"] sets the number of records to iterate over, you'll get the total number of records hosted by NCBI, not what is set by retmax. I am certain you know far more about the curiosities of NCBI than I do, but I think retmax will just limit how many records are reported in the original XML retrieval... I could be wrong.

In my case (repeating the example exactly but using query '"Poecilia wingei"[Organism]'), there are a total of 24 records, and 20 GI numbers were returned by the original query (because retmax was left the default of 20). By simply eliminating the assertion, I still retrieved all 24 records.

I think the scenario you are looking to guard against would involve setting some variable that gets passed as retmax *and* is used as the max iterator in the actual record retrieval. Is that right?

jimmyodonnell:
Yep, I agree that briefly pointing out the potential for discrepancy here is a good idea. It could definitely catch users off guard.

I am happy to add that note if you like. I have just *now* read the CONTRIBUTING doc (sorry) -- would you like me to follow protocol and create a branch and make these changes off master?

peterjc:
Ideally yes, I would have said make a new branch. You can do that and open a new pull request (referencing this one in the description).

However, if you prefer to just continue with this pull request using your master branch that's fine for me too (in which case once this work is done, you'd want to delete your master branch and reset it back to the main master branch)

jimmyodonnell:
Jumping back on this now. I'll create a new issue, branch, and pull request that references this one... Just figured I'd stick to best practices!

peterjc:
Closing in favour of #1157

PR #1069, opened 2017-03-01
Update test_NCBIXML.py

Gasta88:
I've inserted new tests for all the other XML files that weren't tested before.

This is my first attempt to increase testing coverage for a script, please point at any possible flaws.

### DIFF
--- a/Tests/test_NCBIXML.py
+++ b/Tests/test_NCBIXML.py
@@ -13,7 +13,7 @@ E_VALUE_THRESH = 1e-10
 class TestNCBIXML(unittest.TestCase):
 
     def test_xml_2212L_blastp_001(self):
-        "Parsing BLASTP 2.2.12, gi|49176427|ref|NP_418280.3| (xml_2212L_blastp_001)"
+       "Parsing BLASTP 2.2.12,gi|49176427|ref|NP_418280.3| (xml_2212L_blastp_001)"
 
         filename = 'xml_2212L_blastp_001.xml'
         datafile = os.path.join("Blast", filename)
@@ -1864,7 +1864,7 @@ class TestNCBIXML(unittest.TestCase):
         # <Iteration_message>CONVERGED</Iteration_message>
         self.assertRaises(StopIteration, next, records)
         handle.close()
-    
+
     def test_xml_2226_blastn_001(self):
         "Parsing BLASTN 2.2.26+, Query_1 (xml_2226_blastn_001)"
 
@@ -1877,7 +1877,7 @@ class TestNCBIXML(unittest.TestCase):
 
         self.assertEqual(record.query_id, "Query_1")
         self.assertEqual(len(alignments), 0)
-        
+
         record = next(records)
         alignments = record.alignments
 
@@ -1904,8 +1904,7 @@ class TestNCBIXML(unittest.TestCase):
         self.assertEqual(alignments[4].length, 1077)
         self.assertEqual(len(alignments[0].hsps), 1)
         self.assertFalse(alignments[4].hsps[0].expect > E_VALUE_THRESH)
-        
-		
+
         record = next(records)
         alignments = record.alignments
         self.assertEqual(record.query_id, "Query_3")
@@ -1935,7 +1934,7 @@ class TestNCBIXML(unittest.TestCase):
 
         self.assertRaises(StopIteration, next, records)
         handle.close()
-		
+
     def test_xml_2226_blastn_002(self):
         "Parsing BLASTN 2.2.26+, Query_1 (xml_2226_blastn_002)"
 
@@ -1951,7 +1950,7 @@ class TestNCBIXML(unittest.TestCase):
 
         self.assertRaises(StopIteration, next, records)
         handle.close()
-    
+
     def test_xml_2226_blastn_003(self):
         "Parsing BLASTN 2.2.26+, Query_1 (xml_2226_blastn_003)"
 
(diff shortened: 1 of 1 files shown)


peterjc:
Looking at the current coverage (prior to your changes), e.g. recent snapshot at https://codecov.io/gh/biopython/biopython/src/b783754f87d13f99c3ed36d61a0bb92fcbf455a7/Bio/Blast/NCBIXML.py - suggests there are some debug statements not being tested (not very important) and a few other corner cases. In particular testing old-style concatenated XML files seems not to be fully tested.

Once TravisCI finishes we should get an updated coverage report. It is also possible to run the ``coverage`` tool locally (look in our ``.travis-tox.ini`` for hints on this).

peterjc:
TravisCI is failing on the style check, you should be able to reproduce this locally with:

```
pip install flake8
flake8 --ignore E122,E123,E126,E127,E128,E241,E402,E501,E731,E101,E305,F401,F405,F821,F841 Tests/
```

Specifically you have lots of trailing whitespace in ``Tests/test_NCBIXML.py``. Some editors give an easy way to remove this, or you could try ``sed`` or another command line tool for this.

peterjc:
On the coverage metrics, according to https://codecov.io/gh/biopython/biopython/commit/ae2671c4e03a43fca4fc6463d836a59911b7bbf5 this didn't actually help - but including these XML files in ``Tests/test_NCBIXML.py`` seems sensible anyway.

Gasta88:
An indentation error caused few failed checks in Travis. I've edited the error but before creating another commit I would like to work on the coverage of this script.

Since it is my first time, have you got any suggestions?

Tests/test_NCBIXML.py
```diff
@@ -13,7 +13,7 @@
 class TestNCBIXML(unittest.TestCase):
 
     def test_xml_2212L_blastp_001(self):
-        "Parsing BLASTP 2.2.12, gi|49176427|ref|NP_418280.3| (xml_2212L_blastp_001)"
+       "Parsing BLASTP 2.2.12,gi|49176427|ref|NP_418280.3| (xml_2212L_blastp_001)"
```
peterjc: That should be eight spaces indented

peterjc:
If you specifically want to improve the coverage for ``NCBIXML.py``, one option would be adding a test trying to parse a non-XML file, and checking we get this exception triggered:

``` python
            raise ValueError("Your XML file did not start with %s... "
                             "but instead %s"
                             % (XML_START, repr(text[:20])))
```

I would do this using ``self.assertRaises(ValueError, ...)``

Gasta88:
Thanks for the suggestion, I've tried to go down that way with 4 new tests.

peterjc:
Cross referencing #1076 #1077 #1078 and #1079

To include the files in this pull request, you need to add the commits to the branch being used on this pull request (which looks like the automatically named branch ``patch-1``). Then push the updated branch to your GitHub, and it will update this repository.

Any extra BLAST test files should be described in ``Tests/Blast/README.txt`` and ideally the git commit adding the file would explain how it was created (i.e. the command line used) and what the point of the test file is. i.e. Why is it interesting, does it have an unusual feature?

Gasta88:
Naively I'm trying to test borderlines situations where the XML is corrupted or edited wrongly.

In the test cases added I've manually edited one of the test to trigger such scenarios (ex: XML starting with a characters different from the standard <?xml..;  query divided by the <BlastOutput_iterations> and <Iteration> tags; missing queries; etc).

I suppose I need to run BLAST and retrieve the result in an XML, hoping that will trigger the exception that I'm trying to test.

peterjc:
Sorry, I assumed this were all "good" BLAST XML files.

Constructing broken test files is a useful idea, often seen in security research where it is called fuzz testing https://en.wikipedia.org/wiki/Fuzzing

In this case making bad XML files to trigger the error handling in Biopython's BLAST XML code makes good sense for testing that error handling. Maybe include something like ``_err_`` in the filenames to make it clearer they are expected to fail to parse?

Lucioric2000:
A special case in broken XML files is where they are a (still) incomplete Blast results file, or a broken download. Blast results XML files can be huge (bigger tan 300 MB), and its computation time may be more than 24 hours, situations that may cause a timeout in some cluster queuing systems, so the file may persist incomplete, not necesariliy at the end of a Iteration object. In these case, the good practice should be to consider as good the records up to the incomplete one, which should raise an exception that should be handled adequately by the NCBIXML caller program.
Best regards.
Lucio Montero.

Enviado desde Correo<https://go.microsoft.com/fwlink/?LinkId=550986> para Windows 10

De: FGastaldello<mailto:notifications@github.com>
Enviado: lunes, 6 de marzo de 2017 01:14 p. m.
Para: biopython/biopython<mailto:biopython@noreply.github.com>
CC: Subscribed<mailto:subscribed@noreply.github.com>
Asunto: Re: [biopython/biopython] Update test_NCBIXML.py (#1069)


Naively I'm trying to test borderlines situations where the XML is corrupted or edited wrongly.

In the test cases added I've manually edited one of the test to trigger such scenarios (ex: XML starting with a characters different from the standard <?xml..; query divided by the <BlastOutput_iterations> and tags; missing queries; etc).

I suppose I need to run BLAST and retrieve the result in an XML, hoping that will trigger the exception that I'm trying to test.

peterjc:
300MB test files are too big for checking into the repository, but truncated XML files would be sensible to test, yes.

peterjc:
Thank you - I'll look at the replacement pull request #1087 instead.

PR #1072, opened 2017-03-03
Test online doctests in the Entrez tutorial

peterjc:
This extends the ``test_Tutorial.py`` dependencies notation for magic comments in the Tutorial LaTeX source to mark tests as online only.

It has highlighted we have some annoying unicode ``u`` prefixes to get rid of as they only appear under legacy Python 2.7 (and block simple doctest style checks).


peterjc:
How does this look @mdehoon (as the Entrez module owner), @jimmyodonnell (since you worked on some of the examples I update here)?

jimmyodonnell:
Looks good to me!

peterjc:
Thanks Jimmy 👍

Issue #787, opened 2016-03-18
Bio.PDB - deepcopy() causes RecursionError on Disordered Entity


andrewguy:
I'm currently using the Python `subprocess` module to perform parallel computations using a Bio.PDB chain object. When running this with some structures, a RecursionError is raised during this process, which is due to the `subprocess` module performing a `deepcopy` operation on the Bio.PDB chain object.

When simply trying to deepcopy, I get the following (with the first two lines repeated _many_ times):

```
 File "/home/user/Programs/anaconda3/lib/python3.5/site-packages/Bio/PDB/Entity.py", line 206, in __getattr__
    if not hasattr(self, 'selected_child'):
RecursionError: maximum recursion depth exceeded while calling a Python object
```

The particular pdb file I was working with at the time was `1zro.pdb`, although this isn't the only file that has this issue.

Any suggestions as to a suitable workaround/fix?

Cheers,

Andrew

peterjc:
Suggested workaround: Pass the filename and chain identifier to each child process instead, and re-parse the PDB file.

Can you cut down your code to make a short self contained example script to help reproduce the problem?

andrewguy:
Hi Peter,

That's a good suggestion, I will reconstruct the PDB object for each child process.

This code should minimally reproduce the problem:

```
import urllib.request
from io import StringIO
from copy import deepcopy
import Bio.PDB

pdb_name = '1zro'
#Download file
pdb_url = "http://www.rcsb.org/pdb/files/" + pdb_name + ".pdb"
pdb_file = urllib.request.urlopen(pdb_url)
#Create string stream for Bio.PDB to read:
pdb_input_stream = StringIO(pdb_file.read().decode('utf-8'))

#Make Bio.PDB structure object
parser = Bio.PDB.PDBParser()
structure = parser.get_structure(pdb_name, pdb_input_stream)

#Attempt to perform a deepcopy on Bio.PDB Chain:
structure_copy = deepcopy(structure[0]['A'])
```

(Apologies, the file was 1zro, not 1zrl...)

JoaoRodrigues:
This has been referenced before in #302 . It seems to be a case only when disordered residues are present.. At the time [this Python bug report](http://bugs.python.org/issue5508) was also linked to the thread. 

As Peter suggested, a simple workaround is to just re-parse the file in each subprocess. I'll have a look in the coming days, see if we can close this for good without a major code overhaul..

andrewguy:
[...]

PR #1075, opened 2017-03-05
Fix recursion error when performing deepcopy on PDB structures.

andrewguy:
Trying to perform a `deepcopy` of a PDB structure with disordered atoms causes recursion errors in Python  3+. See issue #787 

This proposed fix raises an AttributeError when `__getattr__('__setstate__')` is called on a disordered entity to stop `copy.deepcopy` recursing.

### DIFF
--- a/Tests/test_PDB.py
+++ b/Tests/test_PDB.py
@@ -12,6 +12,7 @@
 """Unit tests for the Bio.PDB module."""
 from __future__ import print_function
 
+from copy import deepcopy
 import os
 import sys
 import tempfile
@@ -525,6 +526,11 @@ class ParseTest(unittest.TestCase):
         finally:
             os.remove(filename)
 
+    def test_deepcopy_of_structure_with_disorder(self):
+        """Test deepcopy of a structure with disordered atoms.
+        Shouldn't cause recursion.
+        """
+        _ = deepcopy(self.structure)
 
 class ParseReal(unittest.TestCase):
     """Testing with real PDB files."""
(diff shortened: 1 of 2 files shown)


JoaoRodrigues:
Great! If that simple "hack" fixes the problem without affecting anything else, there we go. Thanks for coming back to this!

Tests/test_PDB.py
```diff
@@ -525,6 +526,12 @@ def test_pdbio_write_truncated(self):
         finally:
             os.remove(filename)
 
+    def test_deepcopy_of_structure_with_disorder(self):
+        """Test deepcopy of a structure with disordered atoms.
+        Shouldn't cause recursion.
```
peterjc: There should be a blank line here as per PEP257 docstring style guidelines.

peterjc:
According to TravisCI,

```
Tests/test_PDB.py:531:1: W293 blank line contains whitespace
```

Are you familiar with git pre-commit hooks? See #493 for what we'd currently recommend in the (relatively new) file https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

peterjc:
Thank you!

The idea of the git commit hook is to quickly run the style style checks on any altered files, so you can be confident that TravisCI will pass. This local check would be even stricter (e.g. complains about lambda functions which our TravisCI configuration currently allows due to historic usage), with the expectation that you may by-pass it.

andrewguy:
Thanks Peter, no I wasn't familiar with git pre-commit hooks, but will be from now on. Cheers!

PR #1076, opened 2017-03-05
Create xml_2226_tblastn_007.xml

Gasta88:
This new file will be used to increase coverage for NCBIXML.py


peterjc:
Can you include this in the main pull request instead, and make sure to update ``Tests/Blast/README.txt`` to explain what the file is.

(The automated check from codecov failed, but a tiny variation in coverage seems to be expected - some of the Biopython tests are stochastic)

Also ideally in the git commit include the command line used to make the example file.

PR #1077, opened 2017-03-05
Create xml_2226_tblastn_008.xml

Gasta88:
This new file will be used to increase coverage for NCBIXML.py

this file is voluntary empty to test the NCBIXML.handle() exception


peterjc:
Can you include this in the main pull request instead, and make sure to update ``Tests/Blast/README.txt`` to explain what the file is.

This looks like an empty file - is this deliberate?

Also ideally in the git commit include the command line used to make the example file.

Issue #1081, opened 2017-03-06
CIF to PDB conversion yields bad PDBs


JoaoRodrigues:
The PDB format is character delimited and the `PDBParser` relies on this spacing to determine  some things from the atom name (e.g. element). MMCIF files on the other hand are designed to be tabulated and therefore we cannot rely on 'spacing'. This is particularly important for the `Atom.fullname` attribute and has implications in writing PDB files from parsed mmCIFs. 

A possible fix is to work on `PDBIO` to pad the full atom name with an empty space for heavy atoms in ATOM records. MDtraj does something like this:

```
if len(atom.name) < 4 and atom.name[:1].isalpha() and (atom.element is None or len(atom.element.symbol) < 2):
    atomName = ' '+atom.name
```

A better fix would be to maybe just remove full name and go with name only, but this will probably break a lot of things.

PR #1082, opened 2017-03-06
Fixed atom name issue when converting CIF to PDB files

JoaoRodrigues:
Addresses and fixes issue #1081 

I explicitly did not create a new unit test because there are too many variables (different elements, different H namings), and I consider it just an added check to PDBIO that should not change the output for the majority of cases. The patch does pass all tests and was tested in crystal and NMR structures (for Hs).

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -13,7 +13,6 @@ from Bio.Data.IUPACData import atom_weights  # Allowed Elements
 
 _ATOM_FORMAT_STRING = "%s%5i %-4s%c%3s %c%4i%c   %8.3f%8.3f%8.3f%s%6.2f      %4s%2s%2s\n"
 
-
 class Select(object):
     """Select everything fo PDB output (for use as a bas class).
 
@@ -69,6 +68,7 @@ class PDBIO(object):
             record_type = "HETATM"
         else:
             record_type = "ATOM  "
+
         if atom.element:
             element = atom.element.strip().upper()
             if element.capitalize() not in atom_weights:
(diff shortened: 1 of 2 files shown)


JoaoRodrigues:
@peterjc @lennax @etal Maybe another pair of eyes should go over this before merging?

peterjc:
I've never worked with CIF files so am unsure of the background to this problem.

lennax:
@JoaoRodrigues Despite the innumerable combinations of atom names, I'd still advocate for at least some very basic unit tests (e.g., doesn't affect hydrogens or single letter atom names). Also, won't this code render an alpha carbon as `_CA` rather than `_CA_`? (underscores to explicitly represent spaces)

JoaoRodrigues:
@lennax You're right, I'll see what I can do to keep tests simple.

>Also, won't this code render an alpha carbon as _CA rather than _CA_? (underscores to explicitly represent spaces)

In `PDBIO` the name is written to a formatted `%-4s` string so the right-side padding is taken care of. Internally though, yes, it is missing that.

peterjc:
Is this good to go once it has a unit test included?

JoaoRodrigues:
@peterjc @lennax Added a unit test that uses `1LCD` (already in the PDB folder), which is a NMR structure (with Hs). I'm reading the CIF file, writing a PDB, re-reading and comparing atom names and elements. If something is off in formatting it should pop either during parsing the PDB file or in assigning elements.

I think it's good to go.

peterjc:
TravisCI is failing due to an external dependency - I saw this yesterday too,

```
curl -L -O ftp://ftp.cmbi.ru.nl/pub/software/dssp/dssp-2.0.4-linux-amd64
```

I'll temporarily comment this out...

peterjc:
@JoaoRodrigues can you try rebasing this branch? That ought to restart TravisCI with 3b03cb4c1e70a1e1ea29729f6ad3765233f6dcd7 to skip the bad dependency URL.

JoaoRodrigues:
@peterjc Done (I think).

peterjc:
Looks good - thank you!

Issue #1083, opened 2017-03-07
MafIO: Duplication of MAF blocks which span two exons?


TBradley27:
Hello,

I have had a problem trying to use biopython/Bio/AlignIO/MafIO.py recently

Version: Cloned from Github repository Feb 14 2017

In particular, I have been having problems using the 'get_spliced' method of the MafIndex class.

Here is my code:

```
import sys
from Bio import AlignIO
from Bio.AlignIO import MafIO

idx = AlignIO.MafIO.MafIndex("/path/chr1.mafindex" "/path/chr1.maf", "hg38.chr1" )

multiple_alignment = idx.get_spliced([11662625, 11662852], [11662731, 11662928], strand = 1)
```

NB: Chr1.maf and its corresponding index refers to the relevant MAF file from UCSC 100-way alignment:

http://hgdownload.cse.ucsc.edu/goldenPath/hg38/multiz100way/maf/

Running this scripts returns the following error message:

```
File "foo.py", line 15, in <module>
    strand = 1)
  File "/path/Bio/AlignIO/MafIO.py", line 538, in get_spliced
    (self._target_seqname, len(split_by_position[self._target_seqname]), total_rec_length))
    ValueError: Target seqname (hg38.chr1) has 372 records, expected 543
```

In order to debug, I amended the original MafIO.py file in order to print out the values of some variables - I printed `multiseq` for each `multiseq` in `fetched`.  I also printed `seqrec` for each `seqrec` in `multiseq` for the reference species only.

I noticed the following output which seemed to be particularly interesting:

```
multiseq: SingleLetterAlphabet() alignment with 12 rows and 188 columns
catctcatcttgaatgg--taatcctgggt---gtttagggaga...-tg hg38.chr1
catctcatcttgaatgg--taatcctgggt---gtttagggaga...-tg panTro4.chr1
catctcatcttgaatgg--taatcctggat---gtttagggaga...-tg gorGor3.chr1
catctcatcttgaatgg--taatcctgggt---gtttagggaga...-tg ponAbe2.chr1
catctcaccttgaatgg--taatcccgggt---gtttaggaaga...-tg nomLeu3.chr24
cacctcatcttgaacag--taatcctgggt---gtttagggaaa...--- rheMac3.chr1
cacctcatcttgaacgg--taatcctgggt---gtttagggaaa...--- macFas5.chr1
cacctCATCTTGAATGG--TAATCCTGGGT---GTTTAGGGaaa...--- papAnu2.chr1
[...]

PR #1086, opened 2017-03-08
Rebased version of Blaise Li's MAF alignment work

peterjc:
This is based on and would replace pull request #504 by @blaiseli, but deliberately avoids any changes to the offset code (including the sample test files).

Includes his proposed fix for #1083.

@blaiseli does this look OK to you? Comments also welcome from @polyatail and @adamnovak. Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,12 +47,15 @@ addons:
     - samtools
     - wise
 
-# There is a phyml Ubuntu package, but currently too old
+# We setup $HOME/bin and add it to the $PATH for extra binaries we're using.
+#
+# There is a phyml Ubuntu package, but currently too old.
+#
 # There are TravisCI provided versions of PyPy and PyPy3, but currently too old.
 # We therefore deactivate that, and download and unzip portable PyPy binaries.
 #
 # We also need DSSP for testing but it is not available in the repositories.
-# Try to download the binary for Linux and place it in /usr/bin
+# Try to download the binary for Linux and place it in $HOME/bin
 # 
 before_install:
   - cd $HOME
--- a/Bio/AlignIO/MafIO.py
+++ b/Bio/AlignIO/MafIO.py
@@ -444,10 +444,9 @@ class MafIndex(object):
         start_bin >>= _bin_first_shift
         end_bin >>= _bin_first_shift
 
-        for i in range(0, len(bin_offsets)):
+        for bin_offset in bin_offsets:
             if start_bin == end_bin:
-                return bin_offsets[i] + start_bin
-
+                return bin_offset + start_bin
             start_bin >>= _bin_next_shift
             end_bin >>= _bin_next_shift
 
--- a/Bio/AlignIO/MafIO.py
+++ b/Bio/AlignIO/MafIO.py
@@ -1,6 +1,7 @@
 # Copyright 2011, 2012 by Andrew Sczesnak.  All rights reserved.
 # Revisions Copyright 2011, 2017 by Peter Cock.  All rights reserved.
 # Revisions Copyright 2014, 2015 by Adam Novak.  All rights reserved.
+# Revisions Copyright 2015, 2017 by Blaise Li.  All rights reserved.
 #
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
@@ -17,9 +18,58 @@ See http://genome.ucsc.edu/FAQ/FAQformat.html#format5
 You are expected to use this module via the Bio.AlignIO functions(or the
 Bio.SeqIO functions if you want to work directly with the gapped sequences).
 
+According to the MAF format documentation
+(https://genome.ucsc.edu/FAQ/FAQformat.html#format5), alignment blocks are
+composed of sequence lines:
+-----
+ Lines starting with 's' -- a sequence within an alignment block
+
+  s hg16.chr7    27707221 13 + 158545518 gcagctgaaaaca
+  s panTro1.chr6 28869787 13 + 161576975 gcagctgaaaaca
+  s baboon         249182 13 +   4622798 gcagctgaaaaca
+  s mm4.chr6     53310102 13 + 151104725 ACAGCTGAAAATA
+
+ The 's' lines together with the 'a' lines define a multiple alignment. The 's'
+ lines have the following fields which are defined by position rather than
+ name=value pairs.
+
+  * src -- The name of one of the source sequences for the alignment. For
+  sequences that are resident in a browser assembly, the form
+  'database.chromosome' allows automatic creation of links to other assemblies.
+  Non-browser sequences are typically reference by the species name alone.
+
+  * start -- The start of the aligning region in the source sequence. This is a
+  zero-based number. If the strand field is '-' then this is the start relative
+  to the reverse-complemented source sequence (see Coordinate Transforms).
+
+  * size -- The size of the aligning region in the source sequence. This number
+  is equal to the number of non-dash characters in the alignment text field
+  below.
+
+  * strand -- Either '+' or '-'. If '-', then the alignment is to the
+  reverse-complemented source.
+
+  * srcSize -- The size of the entire source sequence, not just the parts
+  involved in the alignment.
+
+  * text -- The nucleotides (or amino acids) in the alignment and any insertions (dashes) as well.
+-----
+
+We see that the coordinates in the MAF format are defined in terms of
+zero-based start positions and aligning region sizes.
+
+A minimal aligned region of length one and starting at first position in the
+source sequence would have start == 0 and size == 1.
+
+As we can see on this example, start + size will give one more than the
+zero-based end position. We can therefore manipulate start and start + size as
+python list slice boundaries.
+
+If we want an inclusive end coordinate, we need to use end = start + size - 1:
+a 1-column wide alignment would have start == end.
 """
 import os
-import itertools
+from itertools import islice
 
 try:
     from sqlite3 import dbapi2 as _sqlite
@@ -40,7 +90,6 @@ MAFINDEX_VERSION = 1
 
 class MafWriter(SequentialAlignmentWriter):
     """Accepts a MultipleSeqAlignment object, writes a MAF file"""
-
     def write_header(self):
         """Writes the MAF header"""
         self.handle.write("##maf version=1 scoring=none\n")
@@ -54,6 +103,7 @@ class MafWriter(SequentialAlignmentWriter):
         elif record.annotations.get("strand") == -1:
             strand = "-"
         else:
+            # TODO: issue warning?
             strand = "+"
 
         fields = ["s",
@@ -64,7 +114,7 @@ class MafWriter(SequentialAlignmentWriter):
                   strand,
                   "%15s" % record.annotations.get("srcSize", 0),
                   str(record.seq)]
-        self.handle.write(" ".join(fields) + "\n")
+        self.handle.write("%s\n" % " ".join(fields))
 
     def write_alignment(self, alignment):
         """
@@ -83,7 +133,9 @@ class MafWriter(SequentialAlignmentWriter):
         # for now, use ._annotations private property, but restrict keys to those
         # specifically supported by the MAF format, according to spec
         try:
-            anno = " ".join(["%s=%s" % (x, y) for x, y in alignment._annotations.items() if x in ("score", "pass")])
+            anno = " ".join(["%s=%s" % (x, y)
+                             for x, y in alignment._annotations.items()
+                             if x in ("score", "pass")])
         except AttributeError:
             anno = "score=0.00"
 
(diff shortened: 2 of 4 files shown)


blaiseli:
@peterjc I think your version of the sql query in `MafIndex.search` has a bug:

            result = con.execute("SELECT DISTINCT start, end, offset FROM "
                                 "offset_data WHERE bin IN (%s) AND (end BETWEEN %s AND %s "
                                 "OR %s BETWEEN start AND end) ORDER BY start, end, "
                                 "offset ASC;"
                                 % (possible_bins, exonstart, exonend, exonend))


As I indicated in a comment in my version:

            # We are testing overlap between the query segment and records in
            # the index, using non-strict coordinates comparisons.
            # The query segment end must be passed as end-inclusive

(This was one of the various reasons why I had initially chosen to use end-inclusive coordinates: https://github.com/biopython/biopython/pull/504#issuecomment-151620693)

Like yours, my version now uses end-exclusive coordinates for `MafIndex.search` and `MafIndex.get_spliced`, that's why I have `exonend - 1` in the sql query:

            result = db_con.execute(
                "SELECT DISTINCT start, end, offset FROM "
                "offset_data WHERE bin IN (%s) AND (end BETWEEN %s AND %s "
                "OR %s BETWEEN start AND end) ORDER BY start, end, "
                "offset ASC;" % (possible_bins, exonstart, exonend - 1, exonend - 1))

peterjc:
Thanks - I was trying to leave out all the end point changes in this branch, I will address this...

peterjc:
The version currently on the master branch as of https://github.com/biopython/biopython/blob/49e7da7b44c913143956e3662f4ada36937a3dc8/Bio/AlignIO/MafIO.py is:

``` python
            result = con.execute("SELECT DISTINCT start, end, offset FROM "
                     "offset_data WHERE bin IN (%s) AND (end BETWEEN %s AND %s "
                     "OR %s BETWEEN start AND end) ORDER BY start, end, "
                     "offset ASC;"
                     % (possible_bins, exonstart, exonend, exonend))
```

On this branch at the time of writing this is (deliberately) unchanged other than the indentation:


``` python
            result = con.execute("SELECT DISTINCT start, end, offset FROM "
                                 "offset_data WHERE bin IN (%s) AND (end BETWEEN %s AND %s "
                                 "OR %s BETWEEN start AND end) ORDER BY start, end, "
                                 "offset ASC;"
                                 % (possible_bins, exonstart, exonend, exonend))
```

Are you saying the current code has a problem and sometimes misses entries? If so this was not clear to me and would greatly benefit from being made explicit as a separate commit with an extra test. Would the minimal change to the master be:

``` python
            # We are testing overlap between the query segment and records in
            # the index, using non-strict coordinates comparisons.
            # The query segment end must be passed as end-inclusive
            result = con.execute("SELECT DISTINCT start, end, offset FROM "
                                 "offset_data WHERE bin IN (%s) AND (end BETWEEN %s AND %s "
                                 "OR %s BETWEEN start AND end) ORDER BY start, end, "
                                 "offset ASC;"
                                 % (possible_bins, exonstart, exonend + 1, exonend + 1))
```

Have I missed out one of your tests which covered this?

blaiseli:
After re-reading the code, I think it may be that your version actually works fine because the `__maf_indexer` method (which is used to feed the sqlite index insertion commands) yields end-exclusive end coordinates. (After writing this post, it turns out that this version of the code still seems buggy after all.)

However, I find it very difficult to reason with end-exclusive coordinates when it comes to the part where we use the sql query to retrieve the offset of recorded MAF blocks. Therefore, I'm more confident that my version, which internally uses end-inclusive coordinate behaves correctly.

I'll try to make the problem clearer by taking an example:

1) We settled on the decision that the `MafIndex.search` interface should take exon ends as lists of end-exclusive coordinates. I'm OK with that. But there is a distinct issue: What type of coordinates we should use in more internal parts of the code? (I argued that end-inclusive was easier to deal with: https://github.com/biopython/biopython/pull/504#issuecomment-151620693)

2) `BETWEEN` tests with non-strict inequalities (see the excerpt of the sqlite documentation I had introduced in a comment before the sql query portion of code).

3) Suppose we want to search for MAF blocks overlapping the first 100 positions of the reference sequence. With the agreed-upon "zero-based, end-exclusive" convention, we would have to call `MafIndex.search(starts=[0,], ends=[100,])`.

4) Suppose we have a "forbidden" MAF block starting at position 100 (zero-based inclusive) and having length 1. This would be the position just after what we are interested in (hence the "forbidden"). How would it be represented in the sqlite index?

In my version of `__maf_indexer` (internal use of end-inclusive coordinates):

- I parse the start: `start = int(line_split[2])` (which results in `start == 100`)
- I parse the size: `size = int(line_split[3])` (which results in `size == 1`)
- I compute an end-inclusive end: `end = start + size - 1` (which results in `end == 100`).

In the current master branch version (internal use of end-exclusive coordinates):

- `start` is parsed the same way (therefore `start == 100`)
- `end` is computed as an end-exclusive end coordinate: `end = int(line_split[2]) + int(line_split[3])` (amounting to `end = start + size` if a `size` variable had been explicitly parsed, which results in `end == 101`)

In both versions, these `start` and `end` values are what are inserted in the sqlite index.

5) What happens when the index is queried?

In both versions (same `MafIndex.search` user interface), we have `exonstart == 0` and `exonend == 100`.

The "forbidden" MAF block should not be matched by this sql query portion:
`(end BETWEEN %s AND %s OR %s BETWEEN start AND end)`
where `start` and `end` are as described in 4) and the `%s` are substituted with things based on `exonstart` and `exonend`:

- In my version: `start == 100`, `end == 100` and the substitutions are `exonstart` (0), `exonend - 1` (99), and `exonend - 1`(99). This gives:

`(100 BETWEEN 0 AND 99 OR 99 BETWEEN 100 and 100)`, which is false, as it should be.

- In the current master branch version: `start == 100`, `end == 101` and the substitution are `exonstart` (0), `exonend` (100), and `exonend` (100). This gives:

`(101 BETWEEN 0 AND 100 OR 100 BETWEEN 100 and 101)`, which is true, resulting in a match for the "forbidden" MAF block.

Apart from the issue of whether the substitutions should be done with `exonend` or `exonend - 1`, we can notice a further problem: the `BETWEEN start and end` part of the expression will match for 2 possible exon end values if the index is built using end-exclusive coordinate (`start == 100` and `end == 101`), which is wrong given that the MAF block has a length of only 1.

I think that it is just not possible to get the query correct using `BETWEEN` when the index is built using end-exclusive coordinates. We might get it correct using different sql expressions, but it is much easier, in my opinion, to use end-inclusive coordinates as early as we can in the internal parts of the code.

peterjc:
@blaiseli Are you happy with the (limited) set of changes in the pull request as it stands? i.e. Is it a useful improvement to what is already on the master branch?

Can we then separately consider if there are advantages to changing the internal representation of the end points? If this changes what gets saved in the SQLite index files, then we would also need to bump the table design version number to avoid breaking the code of people using the existing behaviour (even though it has never yet been in a Biopython release, it has been available and used from Andrews' branch for some years).

blaiseli:
I realize that the test case presented in the first post of my pull request <https://github.com/biopython/biopython/pull/504> can actually be used as a test case for the bug arising from using end-exclusive coordinates when building the index. Although the issue was stated in a different manner, the whole point of my pull request was to avoid this problem.

Here is the behaviour with my current version:
```
>>> from Bio.AlignIO.MafIO import MafIndex
>>> idx = MafIndex("MAF/test_coordinates_consistency.mafindex", "MAF/test_coordinates_consistency.maf", "hg19.chrY")
>>> len(list(idx.search([13314], [13315])))
1
>>> len(list(idx.search([13315], [13316])))
1
>>> len(list(idx.search([13316], [13317])))
1
>>> len(list(idx.search([13317], [13318])))
1
>>> len(list(idx.search([13317], [13317])))
  File "/home/bli/.local/lib/python3.6/site-packages/Bio/AlignIO/MafIO.py", line 492, in search
    exonstart, exonend, exonlen))
ValueError: Exon coordinates (13317, 13317) invalid: exon length (0) < 1
```

And here is the behaviour with the version I downloaded this morning from your mafio branch:
```
>>> from Bio.AlignIO.MafIO import MafIndex
>>> idx = MafIndex("MAF/test_coordinates_consistency.mafindex", "MAF/test_coordinates_consistency.maf", "hg19.chrY")
>>> len(list(idx.search([13314], [13315])))
1
>>> len(list(idx.search([13315], [13316])))
2
>>> len(list(idx.search([13316], [13317])))
2
>>> len(list(idx.search([13317], [13318])))
1
>>> len(list(idx.search([13317], [13317])))
  File "/home/bli/.local/lib/python3.6/site-packages/Bio/AlignIO/MafIO.py", line 477, in search
    raise ValueError("Exon coordinates invalid (%s >= %s)" % (exonstart, exonend))
ValueError: Exon coordinates invalid (13317 >= 13317)
```

With my version, when I ask for a length 1 block, I consistently get a length 1 block. Not with the other.

Here is the content of `MAF/test_coordinates_consistency.maf`

```
a score=-11454.000000
s hg19.chrY              13282 34 +  59373566 CCCCTCA-------------CCTTGACCCT------------CCCATTCTGCCCCACCT
s ponAbe2.chrX_random  9952046 34 +  10445060 CCTCTCA-------------CCTTGACCCT------------CCCATTCTGCCGCACCT
i ponAbe2.chrX_random C 0 C 0
s nomLeu3.chrX            1654 34 + 141252148 CCTCTCG-------------CCTTGACCCT------------CCCATTCTGCCCCACCC
i nomLeu3.chrX        C 0 C 0
s falChe1.KB397537        3106 24 -      4194 cccccca------------------agtcc------------cacttttcccca-----
i falChe1.KB397537    C 0 C 0
s ficAlb2.chr10       19660605 51 +  21346708 CTCTtcagggatgtgctggtgaGTGAGCGC---GGCCGGGGACCGGGCCAGCCG-----
i ficAlb2.chr10       C 0 C 0
s taeGut2.chr26        1809911 37 -   4907541 CTTCCCA-----GTGCCTCTCTGTGAATCC------------CAGCACCCTCCA-----
i taeGut2.chr26       C 0 C 0
s pseHum1.KB221448      102653 24 -    117795 cccccca------------------aatct------------gggacccctccc-----
i pseHum1.KB221448    C 0 C 0
s latCha1.JH129912       91397 46 -    184544 CCCCCCG-------------CTGTGCCCTTATGAACCGAGAACACCCCCTGCTACACCC
i latCha1.JH129912    N 0 C 0

a score=-43159.000000
s hg19.chrY               13316 51 +  59373566 ----------GTCAGGATCACAAGGACCCCCAGATCAGCA----GATGGGAACCGGACC------AAAAAG
s ponAbe2.chrX_random   9952080 51 +  10445060 ----------GTCAGGATCACAAGGACCCCCAGCTCAGCA----GATGGGAACCGGACC------AAAAAG
i ponAbe2.chrX_random C 0 C 0
s nomLeu3.chrX             1688 51 + 141252148 ----------GTCAGGATCACAAGGACCCCAAGCTCAGCA----AATGGCAAACGGACC------AAAAAG
i nomLeu3.chrX        C 0 C 0
s monDom5.chr2        536305954 43 + 541556283 ----------GTTCAGAGCAAAAGCACTGCCGGCTCCACG----GCCAGGGACTGGG--------------
i monDom5.chr2        N 0 C 0
s falChe1.KB397537         3130 12 -      4194 ----------caaaactcccct-------------------------------------------------
i falChe1.KB397537    C 0 C 0
s ficAlb2.chr10        19660656 39 +  21346708 ----------GGGGGAGTCACCGGGATCCCTGGG----------GATGTCACCCAGACC------------
i ficAlb2.chr10       C 0 I 7
s taeGut2.chr26         1809948 46 -   4907541 ----------ACGGCAGCTACCCCTACACCACAGTCACCAC---AGTGCAACCCAGTGC------------
i taeGut2.chr26       C 0 T 831
s pseHum1.KB221448       102677 53 -    117795 ----------caaaccgtgacccc--tccccaaatcagGATTGGGGtgggaccccctccccaaac------
i pseHum1.KB221448    C 0 C 0
s latCha1.JH129912        91443 49 -    184544 TTATGAACCAGGAACAACCAACCCTGCAGCGCACTTACAA----ACCAACAAC------------------
i latCha1.JH129912    C 0 C 0

```

blaiseli:
> Can we then separately consider if there are advantages to changing the internal representation of the end points?

@peterjc I think we really should use my version because the one using internal end-exclusive coordinates is bugged and difficult to fix, as shown in my previous two comments (first through reasoning on the code, second through running a test).

peterjc:
Excellent - a clear test case to help support changing the indexing :)

I suggest we merge this pull request as is (small cosmetic changes, bin optimisation, and the duplicate block fix), and then you start a NEW branch for the proposed indexing changes.

blaiseli:
If you think it is clearer to start a new branch to correct the indexing bug, I can try to restart from a merge of the present pull request.

peterjc:
Great :) Given the positive testing for solving #1083, I'm going to merge this now.

I would find a new branch for the indexing issue easier to work with (if you need help doing that, and rescuing the earlier changes, do ask and I can try to help).

PR #1087, opened 2017-03-08
Update test_NCBIXML.py

Gasta88:
I've decide to close #1069 and open a new PR with all the required changes gathered together.


Tests/Blast/xml_2226_tblastn_007_err.xml
```diff
@@ -0,0 +1,23 @@
+this is an empty file to test the method NCBIXML.handle()
```
peterjc: This comment does not make sense - do you mean this is a real XML file but with some unwanted text at the start?

Gasta88: Precisely.
I could have gone with a random string like "aksjdhaskjdnnJadsfaddNASDLNAIDIN", but I choose this. 

Do you think I should change it?

Gasta88:
This PR went under my radar, apologies for that.

I see that Codecov is complaining for a script that I haven't touched.(`Bio/NeuralNetwork/Gene/Schema.py`)

How is that my edited script is involved in this one?

peterjc:
The coverage numbers tracked by CodeCov are a little stochastic (random), so don't worry about tiny drops in coverage from other areas of the code. If I could find the right settings I'd tweak it not to complain about a drop of ``0.03%`` which is most likely noise.

Tests/Blast/xml_2226_tblastn_007_err.xml
peterjc: To me there is a clear difference between an empty file, and a non-empty non-XML file. Your file is a small text file, it is therefore a non-empty non-XML file.

I would remove the file, and instead use an existing file (e.g. a real text BLAST file). In the Python code for the test comment that you are checking error handling with a non-XML file.

peterjc:
My general feeling here is that rather than adding the bad data files to the ``Tests/`` folder, they could be generated at run time.

e.g. Using ``StringIO`` you can easily have a handle to a partial XML file, or an XML file with unwanted text at the start.

Gasta88:
If I use the `io.StringIO` approach, do you suggest that I close this PR and create a new one?

peterjc:
Which ever is easier for you. If we stick with this pull request I would likely squash it when it comes to merging, so having the old commits isn't a problem.

peterjc:
Also use ``from Bio._py3k import StringIO`` to get a text based StringIO handle on either Python 2 or 3.

Using ``io.StringIO`` will give you a unicode string based handle under Python 2 which is different.

Tests/test_NCBIXML.py
```diff
@@ -1752,12 +1746,11 @@ def test_xml_2222_blastp_001(self):
         handle.close()
 
     def test_xml_2218L_rpsblast_001(self):
-        "Parsing PSI-BLASTP 2.2.18, single query which converges in 3 iterations (xml_2218L_rpsblast_001)"
+        """Parse PSI-BLASTP 2.2.18, single query which converges in 3 iterations (xml_2218L_rpsblast_001)."""
```
peterjc: Here the docstring and comment do not match the method name and filename, which both look like RSP-BLAST.

(This is a pre-existing issue but worth fixing now)

Tests/test_NCBIXML.py
```diff
@@ -1746,7 +1747,7 @@ def test_xml_2222_blastp_001(self):
         handle.close()
 
     def test_xml_2218L_rpsblast_001(self):
-        """Parse PSI-BLASTP 2.2.18, single query which converges in 3 iterations (xml_2218L_rpsblast_001)."""
+        """Parse RSP-BLAST 2.2.18, single query which converges in 3 iterations (xml_2218L_rpsblast_001)."""
```
peterjc: Can you delete the comment after this too (the bit about pgpblast vs psiblast, which I assume was copy-and-pasted in error from an older test).

Gasta88:
Apologies Peter, I haven't payed attention to that comment.

Tests/test_NCBIXML.py
```diff
@@ -1866,6 +1854,1681 @@ def test_xml_2218L_rpsblast_001(self):
...
+                          </BlastOutput_iterations>
+                        </BlastOutput>
+                        """)
+
+        self.assertRaises(ValueError, NCBIXML.read, handle)
+        handle.close()
+
+    def test_xml_2226_tblastn_007_err(self):
+        """Parse TBLASTN 2.2.26+."""
+        handle = StringIO("""this is an empty file to test the method NCBIXML.handle()
```
peterjc: You've still got the confusing text here (about it being an empty file).

Also rather than embedding the string I was expecting you to do something like this:

```python
with open(some_good_blast_xml_filename) as handle:
   xml = handle.read()
handle = StringIO("some random text before the XML\n" + xml)
```

Tests/test_NCBIXML.py
```diff
@@ -1866,6 +1854,1681 @@ def test_xml_2218L_rpsblast_001(self):
...
+                              </BlastOutput_iterations>
+                            </BlastOutput>
+                            """)
+        records = NCBIXML.parse(handle)
+        self.assertRaises(ValueError, next, records)
+        handle.close()
+
+    def test_xml_2226_tblastn_008_err(self):
+        """Parse TBLASTN 2.2.26+."""
+        handle = StringIO("")
```
peterjc: The method name and docstring should be something about parsing an empty file - this is not anything to do with tblastn

Tests/test_NCBIXML.py
```diff
@@ -1866,6 +1854,1681 @@ def test_xml_2218L_rpsblast_001(self):
...
+        records = NCBIXML.parse(handle)
+        self.assertRaises(ValueError, next, records)
+        handle.close()
+
+    def test_xml_2226_tblastn_009_err(self):
+        """Parse TBLASTN 2.2.26+."""
```
peterjc: Is this really an error XML file as the name says? If its a good XML file leave it as a file on disk.

Gasta88: This one is tricky.

As fuzzy test I've appended different XML bodies into one source file to test the fact that even multiple output gathered into one single source could be successfully parsed. 

Do you think that I should keep this edited XML and add it to the folder `Tests\PDB`?

peterjc: OK in this case, yes keep ``xml_2226_tblastn_009_err.xml`` as a file under ``Tests/Blast/``

peterjc: However you do need to clarify the docstring for ``test_xml_2226_tblastn_009_err``

peterjc:
You'll need to remove the test files now done with StringIO, and update the README file to match.

peterjc:
In f4a803ddc11287111e589a7aedb48fcf715aaa5c why did you remove the repeated XML headers from the middle of the ``xml_2226_tblastn_009_err.xml`` file?

```xml
...
</BlastOutput>		
 <?xml version="1.0"?>		
 <!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">		
 <BlastOutput>
...
```

Having them present made it look like a concatenation  as might be generated in a naive split/merge job parallelisation pipeline - or like the output for very early versions of legacy BLAST.

Gasta88:
Because, when I was (wrongly) using `StringIO` to have a virtual XML file to test in this function,an error was always appearing during testing regarding the fact that there wasn't a specific <root> in the file.

By removing the XML headers I could successfully test the string parsed and as well the XML file newly created. 
I could put them back to see if it throws again the same error as before. Would this be ok?

peterjc:
Note some minor white space fixes for the style check to pass on TravisCI:

```
$ flake8 Tests/
Tests/test_NCBIXML.py:2779:65: E226 missing whitespace around arithmetic operator
Tests/test_NCBIXML.py:2780:1: W293 blank line contains whitespace
```

peterjc:
Regarding ``xml_2226_tblastn_009_err.xml`` it depends on why you created the test.

The initial CodeCov report is puzzling me - your changes can't have changed things that much...

Gasta88:
This was my first attempt with `unittest` module. I remember that I've made a function to test each missing XML file that wasn't tested in the original version of the testing file. I thought it could improve coverage if every XML file was tested.

As a second objective I looked at the coverage of the file before my modifications and I saw that some functions weren't tested. For this occasion I've used fuzzy testing (all the `XYZ_err.xml` files) to try and cover those as well.

This might not have been the best approach but , as I mentioned before, it was my first attempt and I was at the bottom of the learning curve :)

Regarding `xml_2226_tblastn_009_err.xml`, I'm using this fake file to test how `Bio.NCBIXML` handle of multiple XML bodies merge into a single XML file. If the scenario is unrealistic or of minor interest, I can remove it from the testing file and leave the other functions.

peterjc:
Based on the coverage reports we are not currently testing concatenated XML files - so explicitly testing this is desirable. i.e. with this in the middle of the file:

```
...
</BlastOutput>		
 <?xml version="1.0"?>		
 <!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">		
 <BlastOutput>
...
```

This used to happen pre BLAST 2.2.14, and I've heard of people using this with job splitting and then processing concatenated output.

So +1 on explicitly testing this.

Issue #267, opened 2013-12-05
database created by SeqIO.index_db() function only accessible from working directory


kashyapchhatbar:
If a fasta file is stored as a file on disk (using an SQLite3 databse) using Bio.SeqIO.index_db() function, the absolute path retrieval of the SQLite3 databse does not work.

When we are trying to retrieve sequences from the stored files on disk using Bio.SeqIO.index_db() function, initialization works fine but sequence retrieval does not work because it is looking for the ".fasta" file in the working directory rather than looking in the absolute path where ".idx" file is present.

Currently the workaround is to set the current working directory to the original path of the file location. But changing current working directory inside scripts at multiple times may not be ultra productive after all.

Illustrated example attached in screenshot.

``` shell
bioinfoboy@bifx-cli:~/db/biomart$ pwd
/homes/bioinfoboy/db/biomart
bioinfoboy@bifx-cli:~/db/biomart$ python3
Python 3.3.3 (default, Dec  3 2013, 17:27:12) 
[GCC 4.6.3] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from Bio import SeqIO
>>> dbHandler = SeqIO.index_db("hKC.idx")
>>> dbHandler["ENST00000577736_snoU13.464-201#snoRNA"]
SeqRecord(seq=Seq('GTCCTTTTATAGTTGATGAGCATGATGATTGGGTGTTCACACGCATGTGTGAAA...ACA', SingleLetterAlphabet()), id='ENST00000577736_snoU13.464-201#snoRNA', name='ENST00000577736_snoU13.464-201#snoRNA', description='ENST00000577736_snoU13.464-201#snoRNA', dbxrefs=[])
>>> 
bioinfoboy@bifx-cli:~/db/biomart$ cd
bioinfoboy@bifx-cli:~$ pwd
/homes/bioinfoboy
bioinfoboy@bifx-cli:~$ python3
Python 3.3.3 (default, Dec  3 2013, 17:27:12) 
[GCC 4.6.3] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from Bio import SeqIO
>>> dbHandler = SeqIO.index_db("/homes/bioinfoboy/db/biomart/hKC.idx")
>>> dbHandler["ENST00000577736_snoU13.464-201#snoRNA"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1088, opened 2017-03-09
MAFINDEX_VERSION = 2 (inclusive end coordinates).

blaiseli:
The use of inclusive coodinates for the sqlite MAF index is an attempt to fix issues discussed in the following pull requests:
https://github.com/biopython/biopython/pull/504
https://github.com/biopython/biopython/pull/1086

A test has been added in `test_MafIO_index.py` to check that the number of MAF alignment blocks returned when querying for a single position will be 1 at the boundary between blocks (it could be 2 with the previous MAF index, which is a bug).

The indices built with end-exclusive coordinates will not be compatible with the present version.


peterjc:
@polyatail @adamnovak what do you think? In particular do you agree that the one-column corner case (see ``test_correct_block_boundary`` based on the discussion https://github.com/biopython/biopython/pull/1086#issuecomment-285080702 recently) is correct and exposes a problem, or is @blaiseli not using the API as you intended?

If you agree there is a problem, can you see a fix without altering the offsets recorded in the SQLite index?

polyatail:
I agree this corner case demonstrates the module producing undesirable output. However, I think this will be unlikely to cause any issues in typical use cases. If the consensus is that we fix it, I propose that we add code to recognize this issue specifically rather than overhauling the way the index is stored.

@blaseli - do you know if in this test case the nucleotide position in question genuinely is stored in both records? If so, then it is working correctly. If it's actually stored in only one record, then code to remove the superfluous result ought to be simple--maybe even a single line.

blaiseli:
I believe the issue is not only happening with 1-column, but also with any segment that ends on the boundary between blocks. I don't think it is a good idea to try work around this problem by adding special handling. The way the index is built and used has a fundamental issue and it will lead to healthier code to fix this.

See again https://github.com/biopython/biopython/pull/1086#issuecomment-285069073 for my analysis of the problem.

blaiseli:
@peterjc @polyatail I included tests with more than 1 column (I made a mistake in my commit message should) at the boundary between 2 blocks in `test_correct_block_boundary`. I included a copy of the tested MAF blocks in the docstring so that it is easier to see what should happen.
See https://github.com/biopython/biopython/pull/1088/commits/621664789f74a4636a1f1a1d41be6414272e6aba

I'm quite certain that the current master version would fail at these tests too, not only on the 1-column cases.

polyatail:
@blaiseli If it's the case that it will always be in one block and not the other, it is a simple fix to no longer return the irrelevant block. I disagree that this is a fundamental, case-breaking issue that requires substantial repair. For one, in cases that it returns a second block, it will do so consistently, and the end-user will choose to disregard the returned block if it does not contain the data they are interested in.

blaiseli:
@polyatail I don't get what makes you reluctant to accept a bug fix and prefer to work around the bug by adding detection of special cases.

The fixed code is available in the present pull request, and we will not have to handle special cases and risk further bugs in trying to implement the fix.

Feel free to try, if you think this is a better solution, but I really don't see the point.

The only problem I see with my pull request is that the code will not be compatible any more with older indices, but this will be detected and the user will know he has to remove the old indices.

blaiseli:
I did some tests with the master version.

`search` yields spurious alignments (resulting in a failure at `test_correct_block_boundary`), and these extra alignments are used inside `get_spliced` when building `split_by_position`. However, this doesn't affect the results of `get_spliced` because the final sequences are built using only positions in `range(exonstart, exonend)`. For users using only `get_spliced`, the bug will have no other consequences than a bit of useless computations.

As a side note, during my tests, I came upon a problem with `__check_existing_db` method (which affects both the present pull request and the master version): When the index is loaded using a different working directory than the one from which it was created, the `Index uses a different file` `ValueError` is triggered because the path to the MAF file is not the same. If we are to fix the index, we could perhaps fix this also by using something else as the file name, like a md5sum hash, for instance.

blaiseli:
Probably the computing of a hash of the MAF file would take too much time. Something more simple but less robust would consist in registering only the base name of the MAF file instead of the path.

peterjc:
The relative paths thing is a separate problem already solved for SeqIO/SearchIO's use of SQLite indexes - see #267. Likely the same fix would work here - but please look at that on a separate pull request and/or issue.

polyatail:
@blaiseli I'm not opposed to bug fixes, but I'm concerned about minimizing the impact on anyone who is using the module. Perhaps we can patch this issue for now and save the overhaul for a future release of several changes that truly justify incrementing the index version number.

blaiseli:
@polyatail Do you envision other changes in the index that could be made to justify changing the index?

In my opinion, fixing the present bug is enough justification.

Besides, since the MafIO module has just recently been made available in the master branch, better merge the fix soon than let more users adopt the bugged version and be annoyed by a future change. At least it should be merged before the next Biopython release.

@peterjc If you think it is better, I can revert the `__check_existing_db` code to its previous state in the present pull request.

blaiseli:
@peterjc I tried to implement the relative path fix in https://github.com/biopython/biopython/pull/1101
Not sure I got all subtleties, but it seems to work for the few tests I've done.

blaiseli:
@peterjc I read on the mailing list that a new biopython version is about to be officially released. If I followed correctly, it will be the first official version to include the MafIO module, and it will be without the present pull request and its bug fix. I think it would be better to include the bug fix also, before more users adopt the bugged version and get annoyed by a future index change. The earlier the bug is fixed the better.

peterjc:
@blaiseli Too late, I'm been letting one more thing delay Biopython 1.69 for too long already. But I do hope the next release will be much sooner.

Is it really not possible to fix this edge condition with a ``- 1`` in the end-value query (leaving the contents of the index database files as is)?

peterjc:
Style checks failing on TravisCI,

```
$ flake8 Tests/
Tests/test_MafIO_index.py:331:1: D209 Multi-line docstring closing quotes should be on a separate line
Tests/test_MafIO_index.py:331:1: D205 1 blank line required between summary line and description
Tests/test_MafIO_index.py:350:1: D205 1 blank line required between summary line and description
Tests/test_MafIO_index.py:350:1: D207 Docstring is under-indented
Tests/test_MafIO_index.py:413:1: D205 1 blank line required between summary line and description
Tests/test_MafIO_index.py:413:1: D207 Docstring is under-indented
Tests/test_MafIO_index.py:452:1: D205 1 blank line required between summary line and description
Tests/test_MafIO_index.py:452:1: D207 Docstring is under-indented
Tests/test_MafIO_index.py:487:1: D205 1 blank line required between summary line and description
Tests/test_MafIO_index.py:487:1: D207 Docstring is under-indented
```

See https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst for setting this up on your own machine.

Are you familiar with ``git rebase``? If yes, does this rebase cleanly onto the current master branch? If yes, could you update the pull request to the rebased version.

peterjc:
I don't see  where you check ``MAFINDEX_VERSION`` when opening an existing index. Can version 1 indexes still be used as is? 

Or are they now incompatible, in which case you should raise an exception and ask the user to rebuild the index.

blaiseli:
It seems there's a check in `__check_existing_db`:

```
            if idx_version != MAFINDEX_VERSION:
                raise ValueError("Index version (%s) incompatible with "
                                 "this version of MafIndex" % idx_version)
```

peterjc:
Thanks! It wasn't in the diff because you'd planned ahead with this 👍

blaiseli:
I think I managed to create a branch that seems to rebase cleanly on the current master branch: https://github.com/blaiseli/biopython/tree/fix_mafindex_2

Is there a way I can make this branch the current pull request, or should I create a separate pull request?

peterjc:
Simple low risk answer: Close this pull request, make a new pull request with the new branch.

Risky answer: Make these changes to your ``fix_mafindex`` branch, and force push that here. e.g.

```
git checkout fix_mafindex_2
git branch -D fix_mafindex
git checkout -b fix_mafindex
git push blaiseli fix_mafindex --force
```

This will destroy the old branch (which is the risky aspect), but GitHub would then refresh this pull request with the new version.

blaiseli:
I took the risky way, replacing `fix_mafindex` by `fix_mafindex_2`, but saved my original branch as https://github.com/blaiseli/biopython/tree/old_fix_mafindex

It seems tha travis-ci is stuck since yesterday on "Waiting for status to be reported"

peterjc:
Hmm. Both https://status.github.com/messages and https://www.traviscistatus.com/ look fine.

This looks like the report https://travis-ci.org/biopython/biopython/builds/364079578 (all green), but something went wrong in reporting this here on the pull request?

peterjc:
Merged, thank you!

PR #1089, opened 2017-03-09
Update _Colors.py

Gasta88:
In reference to issue #820, moved self_test into doctest format.
I'm using this as a reference: https://github.com/biopython/biopython/commit/9242f12a65fb1d6bdc82f8201531662fb63a101b

When I run the verbose output for doctests, I get no error but no method is tested. 

What am I doing wrong?

### DIFF
--- a/Bio/Graphics/GenomeDiagram/_Colors.py
+++ b/Bio/Graphics/GenomeDiagram/_Colors.py
@@ -30,6 +30,31 @@ from reportlab.lib import colors
 class ColorTranslator(object):
     """ Class providing methods for translating representations of color into
     """
+
+    """
+    Example:
+
+    >>> from Bio.Graphics import GenomeDiagram
+    >>> gdct=GenomeDiagram._Colors.ColorTranslator()
+    >>> print(gdct.float1_color((0.5, 0.5, 0.5)))
+    Color(.5,.5,.5,1)
+    >>> print(gdct.int255_color((1, 75, 240)))
+    Color(.003922,.294118,.941176,1)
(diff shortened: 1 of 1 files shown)


Bio/Graphics/GenomeDiagram/_Colors.py
```diff
@@ -30,6 +30,31 @@
 class ColorTranslator(object):
     """ Class providing methods for translating representations of color into
     """
+
+    """
```
peterjc: You need to remove the two triple quote lines above - right now your test is not part of the docstring, rather a self contained string in the function.

peterjc:
Try ``run_doctest(verbose=1)`` or ``run_doctest(verbose=2)``, from memory the zero setting is silent by default?

You will also need to add this doctest to the list in ``Tests/run_tests.py`` so that it will be included in the test-suite as run on TravisCI, via ``python run_tests.py`` or specifically via ``python run_tests.py doctest``.

Gasta88:
I've added the change of Tests/run_tests.py in the PR #1092 , I couldn't merge the branch into this one from my repo.

Is the failure in codecode caused by the Verbose=2 or by the fact that the commit happened before I edited Tests/run_tests.py?

peterjc:
I don't know why CodeCov is unhappy.

Regarding multiple pull requests - are you using git at the command line, or simply editing files via GitHub?

Gasta88:
Since I'm using different computers, on the latest commit I've used the GitHub interface. In one of them I have restrictions on what I can install on it (work computer).

I'll use git in the future.

PR #1092, opened 2017-03-10
Update run_tests.py

Gasta88:
This PR must be combined with #1089.

Added Bio.Graphics.GenomeDiagram._Colors,py doctests


peterjc:
It looks like CodeCov has a problem of some kind, otherwise this ought to have shown a small increase to the test coverage numbers. Thank you!

PR #1101, opened 2017-03-17
Stored MAF file path relative to index.

blaiseli:
Based on https://github.com/biopython/biopython/pull/381
See also https://github.com/biopython/biopython/pull/1088#issuecomment-287148840


peterjc:
Does this look MAF indexing change look sensible to you @polyatail @adamnovak? Its based on what we did for the SeqIO and SearchIO indexing to record paths relative to the SQLite index file rather than the current working directory.

polyatail:
Looks OK to me. Tbh, I would've just applied basename() and called it a day, leaving it to the user to make sure they used the right index with the right MAF file. This is much more thorough.

adamnovak:
What happens if you index a MAF file by absolute path, and then later you move it to another directory or machine? Is there a way to override the check and tell the library you really do mean to use this index?

Or are the indexes so small in practice that you can just make people regenerate them?

polyatail:
Not sure about the first part, but the indexes *do* take a while to regenerate from, e.g. the human genome. When I was using this module regularly, I always dreaded having to regenerate the indices and usually left that running overnight.

adamnovak:
Then we probably do want to have a good way to recover from a file/index
move.

I'm thinking about parallel computing systems such as Toil in our lab,
where you might index something in a temp directory on one machine, and
then copy it out to different temp directories on other machines for use in
later processing steps.

polyatail:
Well, I guess for savvy users, they can manually edit the SQLite file. Perhaps we can just include something about it in the Wiki?

polyatail:
Or maybe add an optional argument to check only the basename

peterjc:
Relative paths ought work nicely for the case of putting things in a temp folder (e.g. index and MAF in same folder).

Likewise absolute paths ought to work nicely for a MAF file shared between users, with a (temp) index file located anywhere.

I personally don't see much need for complicated the API with options to override the base path. No one has asked for this in SeqIO or SearchIO, although I suspect those indexes are smaller, and for SearchIO short term.

peterjc:
Merged - Thank you @blaiseli, and thank you @polyatail and @adamnovak for looking at this.

PR #1105, opened 2017-03-21
Updated Bio/PDB/QCPSuperimposer/__init__.py and converted Tests/test_QCPSuperimposer.py to unittests

MaxGreil:
I moved the print test from Bio/PDB/QCPSuperimposer/__init__.py to Tests/test_QCPSuperimposer.py and converted it to a unittests (see testOldTest). 
Additional I tested all public modules of of Bio/PDB/QCPSuperimposer/__init__.py separately in own unittests (see Tests/test_QCPSuperimposer.py).
Eventually I deleted Tests/output/test_QCPSuperimposer as it is no longer needed for the unittests in Tests/test_QCPSuperimposer.py.
This was done according to biopython issue #820.


Bio/PDB/QCPSuperimposer/__init__.py
```diff
@@ -138,55 +138,5 @@ def get_rms(self):
...
-    dif = t1 - t0
-    print("process wall time (msec): %d" % (dif.total_seconds() * 1000))
-
-    print(y_on_x1)
-    print("")
-    print(y_on_x2)
-    print("")
-    print("%.2f" % rms)
+    #moved test to Tests/test_QCPSuperimposer.py
+    # 2017 by Maximilian Greil
```
peterjc: You can remove all of this if statement (and in fact right now it is invalid syntax unless unless you add a ``pass`` command)

Tests/test_QCPSuperimposer.py
```diff
@@ -19,60 +22,141 @@
...
+		              [52.71, -1.18, 49.38]])
+
+		self.sup = QCPSuperimposer()
+		self.sup.set(self.x,self.y)
+
+	#Public methods
+
+	def testSet(self):
+		"""Test method set(x,y)"""
+		assert(array_equal(self.sup.reference_coords, self.x))
```
peterjc: Ideally that would be: ```self.assertTrue(array_equal(self.sup.reference_coords, self.x)``

Tests/test_QCPSuperimposer.py
```diff
@@ -19,60 +22,141 @@
...
+		assert(array_equal(self.sup.coords, self.y))
+		assert(self.sup.transformed_coords == None)
```
peterjc: Ideally that would be ``self.assertIsNone(self.sup.transformed_coords)``.

Note stylistically it is considered better to use ``x is None`` rather than ``x == None``

Tests/test_QCPSuperimposer.py
```diff
@@ -19,60 +22,141 @@
...
+		calc_rot = array([[0.68304939, -0.5227742, -0.51004967],
+ 		                [0.53664482, 0.83293151, -0.13504605],
+ 			        [0.49543503, -0.18147239, 0.84947743]])
+		assert(array_equal(around(self.sup.rot, decimals=3), around(calc_rot, decimals=3)))
+		calc_tran = array([-7.43885125, 36.51522275, 36.81135533])
+		assert(array_equal(around(self.sup.tran, decimals=3), around(calc_tran, decimals=3)))
+		calc_rms = 0.003
+		assert(float(('%.3f' % self.sup.rms)) == calc_rms)
```
peterjc: Ideally this would be ``self.assertEqual(float('%.3f' % self.sup.rms), calc_rms)``

There was also a set of extra brackets for calling ``float``.

peterjc:
This needs a bit more work, but in general its just what I was hoping for. I look forward to reviewing an updated version of this work later.

A general problem I spotted was with the Python indentation which should use four spaces. It is hard to tell from GitHub's display but I suspect you may have some tabs here?

Also, rather than ``assert a ==b``, we would use ``self.assertEqual(a, b)`` etc using the ``unittest`` methods, here https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertEqual

Thank you for looking at this 👍

MaxGreil:
Hello Peter,
thank you for your feedback. 
I will open a new pull request when I updated my code.

Issue #1096, opened 2017-03-14
Explicitly test origin wrapping in test_GenomeDiagram.py


peterjc:
Many bacteria have circular genomes where there is a gene spanning the origin (usually the annotators try to avoid this by shifting the origin), e.g. ``NC_005213`` from *Nanoarchaeum equitans* (which interestingly has been withdrawn from RefSeq because "data validation identified problems with the assembly or annotation.") has:

```
LOCUS       NC_005213             490885 bp    DNA     circular CON 16-DEC-2014
DEFINITION  Nanoarchaeum equitans Kin4-M chromosome, complete genome.
ACCESSION   NC_005213
VERSION     NC_005213.1
...
     gene            complement(join(490883..490885,1..879))
                     /locus_tag="NEQ001"
                     /db_xref="GeneID:2732620"
     CDS             complement(join(490883..490885,1..879))
                     /locus_tag="NEQ001"
                     /note="conserved hypothetical [Methanococcus jannaschii];
                     COG1583:Uncharacterized ACR; IPR001472:Bipartite nuclear
                     localization signal; IPR002743: Protein of unknown
                     function DUF57"
                     /codon_start=1
                     /transl_table=11
                     /product="hypothetical protein"
                     /protein_id="NP_963295.1"
                     /db_xref="GeneID:2732620"
                     /translation="MRLLLELKALNSIDKKQLSNYLIQGFIYNILKNTEYSWLHNWKK
                     EKYFNFTLIPKKDIIENKRYYLIISSPDKRFIEVLHNKIKDLDIITIGLAQFQLRKTK
                     KFDPKLRFPWVTITPIVLREGKIVILKGDKYYKVFVKRLEELKKYNLIKKKEPILEEP
                     IEISLNQIKDGWKIIDVKDRYYDFRNKSFSAFSNWLRDLKEQSLRKYNNFCGKNFYFE
                     EAIFEGFTFYKTVSIRIRINRGEAVYIGTLWKELNVYRKLDKEEREFYKFLYDCGLGS
                     LNSMGFGFVNTKKNSAR"
...
```

Here ``complement(join(490883..490885,1..879))`` is spanning the origin (on the reverse complement strand), source: https://www.ncbi.nlm.nih.gov/nuccore/NC_005213.1?report=genbank

GenomeDiagram should draw these, but ``test_GenomeDiagram.py`` does not explicitly test this kind of feature - ideally with a code-based example without having to add yet another GenBank file to the test suite.

[...]

Issue #1095, opened 2017-03-14
Improving doctest Bio/Graphics/GenomeDiagram/_FeatureSet.py


Gasta88:
Hi all, 

before editing and creating a PR for this script there are few things that I would like to discuss:

1) Testing on this script rely on a GenBank file which seems to be stored locally on the creator machine. I've found a copy of it and I would like to place it in the repo, but I'm not sure about the location. Maybe `Tests/Graphics`?

2) I'm relying on the doctest framework over the unittest because it seems in line with the previous testing that was implemented on the script. Any opinion of the contrary?

Thanks for the inputs.

peterjc:
RE: self test in https://github.com/biopython/biopython/blob/biopython-168/Bio/Graphics/GenomeDiagram/_FeatureSet.py#L198

The basic functionality being tested here is already covered, e.g. https://github.com/biopython/biopython/blob/biopython-168/Tests/test_GenomeDiagram.py#L822

I'm reluctant to add another reasonably sized file to the repository (and thus the files distributed with Biopython) unnecessarily. ``NC_005213`` from *Nanoarchaeum equitans* is a nice test case in that is it a small genome and it has an example of a feature wrapping the origin.

I suggest we remove the self-test in ``_FeatureSet.py``, and consider constructing a simple genome diagram in code which include an origin-wrapping feature, drawn in both linear and circular mode (enhancement filed as #1096).

Gasta88:
Thanks @peterjc.

I'll get my head around this issue then.

peterjc:
I was not clear enough in my comment. As genomes go, NC_005213 is relatively small but I still DO NOT want to include it in the test suite.

Instead, we could recreate an origin spanning feature in code for the test.

peterjc:
Closing, see #1096.

PR #1106, opened 2017-03-21
Update test_GenomeDiagram.py

Gasta88:
As refereed in issue #1096 and #1095, the self-test has been removed from `Bio/Graphics/GenomeDiagram/_FeatureSet.py` and a code base example has been added for `Tests/test_GenomeDiagram.py`.


peterjc:
TravisCI is failing under Python 2.7 due to this:

```
ERROR: test_circular_genome_diagram_NC_005213_pdf (test_GenomeDiagram.DiagramTest)
Construct and draw PDF using NC_005213 genome.
  File "/home/travis/build/biopython/biopython/Tests/test_GenomeDiagram.py", line 19614, in test_circular_genome_diagram_NC_005213_pdf
    handle = io.StringIO(NC_005213_text)
TypeError: initial_value must be unicode or None, not str
```

This is due to a change from Python 2 and 3 about how ``io.StringIO`` works. What you should use here is ``from Bio._py3k import StringIO`` which will take care of it. Outside of Biopython, the library ``six`` has similar mechanisms for writing code which will work under both Python 2 and 3.

peterjc:
I've not actually viewed the changes to ``Tests/test_GenomeDiagram.py`` because GitHub says there are too many for the webview. I'd have to check at the command line. This is odd, perhaps something nasty has happened with Windows vs Unix new lines?

Gasta88:
Thanks for the StringIO head up.  

I've no idea why the file is not visible when you open that for editing, but I've copied the text from my IDE (Spyder). It shouldn't cause any problem.

To be sure I've also checked the script in plain Notepad and it's not showing any funny character.

peterjc:
Note sure what GitHub things, but it says for this file:

> 19,817 additions, 1,026 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

Gasta88:
Not very user friendly, but this should avoid using the command line:
https://patch-diff.githubusercontent.com/raw/biopython/biopython/pull/1106.patch

peterjc:
@Gasta88 I believe you are working on Windows. How did you install git, and do you recall an option from the installer about automatically converting the line endings? I'm almost certain part of the large diff is due to a Windows vs Unix end of line change in ``test_GenomeDiagram.py``.

Separately, please do not embed the large  *Nanoarchaeum equitans* NC_005213 GenBank file into the Python test script. It would have been better to have added this as a file under ``Tests/GenBank/NC_005213.gbk`` (so that it could be used in other tests as well), but I would prefer you not include it at all.

I was not clear enough in https://github.com/biopython/biopython/issues/1095#issuecomment-286391312 about this, added comment https://github.com/biopython/biopython/issues/1095#issuecomment-288422263 - my apologies for the confusion.

Gasta88:
No problem, I'll close this PR.

peterjc:
This has reminded me about getting a Windows development machine setup here as a VM (since I don't currently have anyway to test and debug under 64-bit Windows). Thanks.

PR #1108, opened 2017-03-22
Update test_QCPSuperimposer.py

MaxGreil:
Moved tests from Bio/PDB/QCPSuperimposer/__init__.py to Tests/test_QCPSuperimposer.py, so the if __name__ statement can be removed from Bio/PDB/QCPSuperimposer/__init__.py.
Converted all tests in Tests/test_QCPSuperimposer.py to unittests. Tested all public methods in Bio/PDB/QCPSuperimposer/__init__.py, so Tests/output/test_QCPSuperimposer can be removed, too.
See biopython issue #820

### DIFF
--- a/Tests/test_QCPSuperimposer.py
+++ b/Tests/test_QCPSuperimposer.py
@@ -1,12 +1,15 @@
+# Copyright 2017 by Maximilian Greil. All rights reserved.
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
 
-from __future__ import print_function
+import unittest
 
 try:
     from numpy import array
     from numpy import dot  # missing in old PyPy's micronumpy
+    from numpy import around
+    from numpy import array_equal
 except ImportError:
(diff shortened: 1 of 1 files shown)


peterjc:
The TravisCI check is failing where we check the style:

```
$ flake8 Tests/
Tests/test_QCPSuperimposer.py:25:1: E302 expected 2 blank lines, found 1
Tests/test_QCPSuperimposer.py:27:1: W293 blank line contains whitespace
Tests/test_QCPSuperimposer.py:28:5: E303 too many blank lines (2)
```

You should be able to check this locally if you first do ``pip install flake8``. The good news is that looks easy to fix.

peterjc:
One other small change, and this should be ready for me to merge - can you rename the test methods from things like ``testSet`` to ``test_set`` in order to follow PEP8 ("lowercase with words separated by underscores as necessary to improve readability") https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

(Also right now our ``run_tests.py`` only looks for methods ``test_``).

Issue #1112, opened 2017-03-24
test_NaiveBayes.py fails under PyPy3.5 v5.7 beta


peterjc:
See #953, on a branch I have modified ``setup.py`` to expect NumPy to be present under PyPy. This exposes a problem with hashing NumPy floats under the current latest version of PyPy3.

e.g. https://travis-ci.org/peterjc/biopython/jobs/214366116

```
======================================================================
ERROR: test_calculate_function_with_scale (test_NaiveBayes.NaiveBayesTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/peterjc/biopython/Tests/test_NaiveBayes.py", line 77, in setUp
    self.model = NaiveBayes.train(self.xs, self.ys)
  File "/home/travis/build/peterjc/biopython/build/lib.linux-x86_64-3.5/Bio/NaiveBayes.py", line 204, in train
    nb.p_conditional[i][j] = _contents(values)
  File "/home/travis/build/peterjc/biopython/build/lib.linux-x86_64-3.5/Bio/NaiveBayes.py", line 38, in _contents
    counts[item] = counts.get(item, 0) + term
TypeError: 'numpy.float64' objects are unhashable
======================================================================
ERROR: test_calculate_function_wrong_dimensionality (test_NaiveBayes.NaiveBayesTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/peterjc/biopython/Tests/test_NaiveBayes.py", line 77, in setUp
    self.model = NaiveBayes.train(self.xs, self.ys)
  File "/home/travis/build/peterjc/biopython/build/lib.linux-x86_64-3.5/Bio/NaiveBayes.py", line 204, in train
    nb.p_conditional[i][j] = _contents(values)
  File "/home/travis/build/peterjc/biopython/build/lib.linux-x86_64-3.5/Bio/NaiveBayes.py", line 38, in _contents
    counts[item] = counts.get(item, 0) + term
TypeError: 'numpy.float64' objects are unhashable
======================================================================
ERROR: test_classify_function (test_NaiveBayes.NaiveBayesTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/peterjc/biopython/Tests/test_NaiveBayes.py", line 77, in setUp
    self.model = NaiveBayes.train(self.xs, self.ys)
[...]

Issue #953, opened 2016-10-07
Review status of NumPy in PyPy


peterjc:
Last time we looked at this, PyPy's reimplementation of NumPy was very incomplete - missing things like linear algebra support - but they have been making steady improvement, see http://buildbot.pypy.org/numpy-status/latest.html and https://morepypy.blogspot.co.uk/search/label/numpy

Note according to https://bitbucket.org/pypy/numpy they do not yet support NumPy under PyPy for Python 3.

Part of this would be installing PyPy's NumPy port under TravisCI.

Note it looks like there it now no need to modify the imports,

```
$ pypy
Python 2.7.10 (c95650101a99, Sep 06 2016, 10:37:47)
[PyPy 5.4.1 with GCC 4.2.1 Compatible Apple LLVM 5.1 (clang-503.0.40)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> import numpy
>>>> numpy.__version__
'1.10.0'
>>>> quit()
```

This is still not going to work with our C code using NumPy, so things like `Bio.Cluster` would not be expected to work.

peterjc:
Things are looking pretty good, ``test_PDB.py`` passes for example:

```
$ pypy setup.py test
running test
Python version: 2.7.10 (c95650101a99, Sep 06 2016, 10:37:47)
[PyPy 5.4.1 with GCC 4.2.1 Compatible Apple LLVM 5.1 (clang-503.0.40)]
Operating system: posix darwin
test_Ace ... ok
test_AlignIO ... ok
test_AlignIO_ClustalIO ... ok
test_AlignIO_EmbossIO ... ok
test_AlignIO_FastaIO ... ok
test_AlignIO_PhylipIO ... ok
test_AlignIO_convert ... ok
test_AlignInfo ... ok
test_Application ... ok
test_BWA_tool ... ok
test_BioSQL_MySQLdb ... skipping. BioSQL test configuration file biosql.ini missing (see biosql.ini.sample)
test_BioSQL_MySQLdb_online ... skipping. BioSQL test configuration file biosql.ini missing (see biosql.ini.sample)
test_BioSQL_mysql_connector ... skipping. BioSQL test configuration file biosql.ini missing (see biosql.ini.sample)
test_BioSQL_mysql_connector_online ... skipping. BioSQL test configuration file biosql.ini missing (see biosql.ini.sample)
[...]

PR #126, opened 2012-12-11
pypy support for Bio.PDB

diegozea:
I'm pypy you can't import Bio.PDB after import numpypy because linalg is not implemented. This lines solves the problem (for using the rest of the module until linalg support in numpypy):

```
try:
    # Superimpose atom sets
    from Superimposer import Superimposer
    # Kolodny et al.'s backbone libraries
    from FragmentMapper import FragmentMapper
except:
    # Problem with pypy, don't support for numpy.linalg
    pass
```

For example, this is one of the errors (from FragmentMapper) before the try block:

```
dzea@deepthought:~/databases/Uniprot$ pypy2.0 
Python 2.7.3 (7e4f0faa3d51, Nov 16 2012, 15:53:25)
[PyPy 2.0.0-beta1 with GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
And now for something completely different: ``PyPy 1.3 released''
>>>> import numpypy
>>>> from Bio import PDB
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/dzea/bin/pypy-2.0-beta1/site-packages/Bio/PDB/__init__.py", line 67, in <module>
    from FragmentMapper import FragmentMapper
  File "/home/dzea/bin/pypy-2.0-beta1/site-packages/Bio/PDB/FragmentMapper.py", line 35, in <module>
    from Bio.SVDSuperimposer import SVDSuperimposer
  File "/home/dzea/bin/pypy-2.0-beta1/site-packages/Bio/SVDSuperimposer/__init__.py", line 8, in <module>
    from SVDSuperimposer import *
  File "/home/dzea/bin/pypy-2.0-beta1/site-packages/Bio/SVDSuperimposer/SVDSuperimposer.py", line 7, in <module>
    from numpy.linalg import svd, det
ImportError: No module named numpy.linalg

```

PR #1116, opened 2017-03-24
Enable NumPy code and most C code under PyPy

peterjc:
This would address #126 (``Bio.PDB`` under PyPy) and #953 (our NumPy code in general under PyPy).

Known issues which I believe are PyPy limitations:

- #1114 ``Bio.trie`` does not compile (and therefore still excluded in ``setup.py``)
- #1112 ``test_NaiveBayes.py`` fails under PyPy3.5 v5.7 beta

I have been testing with both PyPy2 v5.6 and the new release v5.7, plus also PyPy3.5 v5.7 beta.

I would like a second set of eyes on this before merging it, and thoughts on what to do with the Navie Bayes test? The simple answer is mark PyPy3 as an allowed failure under TravisCI...

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -50,7 +50,8 @@ whitelist_externals =
     bash
     echo
 # Want to avoid overhead of compiling numpy or scipy:
-install_command = pip install --only-binary=numpy,scipy {opts} {packages}
+# (But must compile numpy for PyPy right now)
+install_command = pip install --only-binary=scipy {opts} {packages}
 deps =
     #Lines startings xxx: are filtered by the environment.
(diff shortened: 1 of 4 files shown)


peterjc:
It seems without the binary-only restriction, under Python 3.3 pip will fetch the latest NumPy which does not support Python 3.3.... simple answer is test that without NumPy. As an aside, we could look at dropping Python 3.3 support in Biopython too.

peterjc:
It was easy to specifically install NumPy 1.11.3 for Python 3.3.

With the ``test_NaiveBayes.py`` now being skipped if the ``numpy.float64`` is unhashable (#1112), this is quite a clean set of changes.

PR #1118, opened 2017-03-25
add the default radius for the element

autodataming:
add the default radius for the element

### DIFF
--- a/Bio/PDB/radius.py
+++ b/Bio/PDB/radius.py
@@ -1,0 +1,30 @@
+#!python2 radius
+
+#refer: http://www.crystalmaker.com/support/tutorials/crystalmaker/atomic-radii/index.html
+#VDW radius
+radius_dict={
+'H':0.53,
+'He':0.31,
+'Li':1.67,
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
I'm sorry, but I have no idea what you want to achieve with this and the other two pull requests: #1119 and #1120 . What is your motivation? What use case would you have for this?

Also, you should have a look at other files in the Bio.PDB module to get a sense of how they are structure. Namely, the copyright notice at the top of the file, naming conventions, etc.

peterjc:
@autodataming could you explain what it is you are trying to do here?

autodataming:
@peterjc  I want to calculate the surface of the protein.

JoaoRodrigues:
@autodataming There are at least three modules that do so already: `ResidueDepth` (uses MSMS), `NACCESS` (uses NACCESS), and `HSAExposure` (does not have dependencies). 

Why would you want to store this data (radii alone) without any other accessory code to calculate surfaces?

PR #1119, opened 2017-03-25
add atom radius

autodataming:
add atom radius


peterjc:
I'm going to close this (and #1120) and leave the discussion on #1118 only

PR #1120, opened 2017-03-25
add atom default radius

autodataming:
add atom default radius


peterjc:
I'm going to close this (and #1119) and leave the discussion on #1118 only

PR #1123, opened 2017-03-26
Add selftest for PDBIO

souravsingh:


Tests/test_PDB.py
```diff
@@ -1332,6 +1332,25 @@ def test_NACCESS_asa_file(self):
...
+class TestSelf(unittest.TestCase):
+    
+    def test_example(self):
+        from Bio.PDB.PDBParser import PDBParser
+        import sys
+        p = PDBParser(PERMISSIVE=True)
+        s = p.get_structure("test", sys.argv[1])
+        io = PDBIO()
+        io.set_structure(s)
+        io.save("out1.pdb")
```
peterjc: See the ``tempfile`` module for how to generate a temporary file - and then after this test please delete it.

Tests/test_PDB.py
```diff
@@ -1332,6 +1332,25 @@ def test_NACCESS_asa_file(self):
             naccess = process_asa_data(asa)
         self.assertEqual(len(naccess), 524)
 
+class TestSelf(unittest.TestCase):
+    
+    def test_example(self):
+        from Bio.PDB.PDBParser import PDBParser
+        import sys
```
peterjc: These imports should be a the top of the file.

Tests/test_PDB.py
```diff
@@ -1332,6 +1332,25 @@ def test_NACCESS_asa_file(self):
...
+        p = PDBParser(PERMISSIVE=True)
+        s = p.get_structure("test", sys.argv[1])
+        io = PDBIO()
+        io.set_structure(s)
+        io.save("out1.pdb")
+        with open("out2.pdb", "w") as fp:
+            s1 = p.get_structure("test1", sys.argv[1])
```
peterjc: There won't be any command line arguments when run via the test framework. You should see if the test make sense with some of the example files we include under ``Tests/PDB/`` here.

Tests/test_PDB.py
```diff
@@ -1332,6 +1332,25 @@ def test_NACCESS_asa_file(self):
             naccess = process_asa_data(asa)
         self.assertEqual(len(naccess), 524)
 
+class TestSelf(unittest.TestCase):
```
peterjc: The fact that this was once a self test is not really of interest. Can you give this class and the test a more meaningful name? Something about PDB input/output?

Tests/test_PDB.py
```diff
@@ -19,6 +19,7 @@
 import unittest
 import warnings
 from Bio._py3k import StringIO
+from Bio.PDB.PDBParser import PDBParser
```
peterjc: This import needs to be after the check for if numpy is installed (so that ``run_tests.py`` can skip the whole test if NumPy is missing).

peterjc: More importantly, it clashes with this existing import later on:

```python
from Bio.PDB import PDBParser, PPBuilder, CaPPBuilder, PDBIO, Select
```

This is due to the old ``Bio.PDB`` module not following PEP8 guidelines that module names are lower case, and classes are title case. Unfortunately here ``PDBParser`` is used for both.

I recommend you remove your new import, and then using the existing import you can refer to the class as ``PDBParser.PDBParser`` instead.

(This is what the  QuantifiedCode check is reporting as *Resolve import naming collision*)

Tests/test_PDB.py
```diff
@@ -1332,6 +1333,24 @@ def test_NACCESS_asa_file(self):
...
+    
+    def test_example(self):
+        p = PDBParser(PERMISSIVE=True)
+        s = p.get_structure("test", sys.argv[1])
+        io = PDBIO()
+        io.set_structure(s)
+        filenumber, filename = tempfile.mkstemp()
+        os.close(filenumber)
+        with open("out2.pdb", "w") as fp:
+            s1 = p.get_structure("test1", sys.argv[1])
```
peterjc: You are still using ``sys.argv[1]`` and ``sys.argv[2]`` for file names. You need to use two automatically named temporary files here (a few lines above you setup a single temporary filename but never used it).

peterjc:
Also the automated style check run via TravisCI is failing,

```
$ flake8 Bio/
Bio/PDB/PDBIO.py:227:1: W391 blank line at end of file
Bio/PDB/PDBIO.py:227:1: W293 blank line contains whitespace
```

You can run flake8 locally after installing it with ``pip install flake8``.

souravsingh:
@peterjc Are there any fixes that are needed?

peterjc:
Yes. There are some simple style issues (extra blank lines and white space - run ``flake8`` to check this as described earlier), and more importantly the tests are failing according to TravisCI:

e.g. from https://travis-ci.org/biopython/biopython/jobs/220635646

```
ERROR: test_example (test_PDB.TestPDBIO)
  File "/home/travis/build/biopython/biopython/Tests/test_PDB.py", line 1389, in test_example
    io.set_structure(s)
NameError: global name 's' is not defined
```

Tests/test_PDB.py
```diff
@@ -1332,6 +1332,23 @@ def test_NACCESS_asa_file(self):
...
         self.assertEqual(len(naccess), 524)
 
+class TestPDBIO(unittest.TestCase):
+    
+    def test_example(self):
+        p = PDBParser(PERMISSIVE=True)
+        filenumber, filename = tempfile.mkstemp()
+        os.close(filenumber)
+        io = PDBIO()
+        io.set_structure(s)
```
peterjc: You have not defined ``s`` yet, the original code did:

```
s = p.get_structure("test", sys.argv[1])
```

However, you need to pick an example PDB file from under ``Tests/PDB/*.pdb``

peterjc:
You also need to replace ``sys.argv[1]`` and ``sys.argv[2]`` with filenames pointing at suitable example PDB files included with Biopython. Or, if none of the existing files work, you would need to try to find two small PDB files we can use here.

souravsingh:
I will resend the PR after fixing the conflicts

PR #1127, opened 2017-03-30
moved self-test from Vector.py

Gasta88:
In reference to issue #820, I've moved the self test part of `Bio/PDB/Vector.py` into `Tests/test_PDB.py`.

### DIFF
--- a/Bio/PDB/Vector.py
+++ b/Bio/PDB/Vector.py
@@ -10,194 +10,8 @@ from __future__ import print_function
 import numpy
 
 
-def m2rotaxis(m):
-    """Return angles, axis pair that corresponds to rotation matrix m."""
-    # Angle always between 0 and pi
-    # Sense of rotation is defined by axis orientation
-    t = 0.5 * (numpy.trace(m) - 1)
-    t = max(-1, t)
-    t = min(1, t)
-    angle = numpy.arccos(t)
-    if angle < 1e-15:
-        # Angle is 0
-        return 0.0, Vector(1, 0, 0)
-    elif angle < numpy.pi:
-        # Angle is smaller than pi
-        x = m[2, 1] - m[1, 2]
-        y = m[0, 2] - m[2, 0]
-        z = m[1, 0] - m[0, 1]
-        axis = Vector(x, y, z)
-        axis.normalize()
-        return angle, axis
-    else:
-        # Angle is pi - special case!
-        m00 = m[0, 0]
-        m11 = m[1, 1]
-        m22 = m[2, 2]
-        if m00 > m11 and m00 > m22:
-            x = numpy.sqrt(m00 - m11 - m22 + 0.5)
-            y = m[0, 1] / (2 * x)
-            z = m[0, 2] / (2 * x)
-        elif m11 > m00 and m11 > m22:
-            y = numpy.sqrt(m11 - m00 - m22 + 0.5)
-            x = m[0, 1] / (2 * y)
-            z = m[1, 2] / (2 * y)
-        else:
-            z = numpy.sqrt(m22 - m00 - m11 + 0.5)
-            x = m[0, 2] / (2 * z)
-            y = m[1, 2] / (2 * z)
-        axis = Vector(x, y, z)
-        axis.normalize()
-        return numpy.pi, axis
-
-
-def vector_to_axis(line, point):
-    """Returns the vector between a point and the closest point on a line
-
-    i.e. the perpendicular projection of the point on the line.
-
-    @type line: L{Vector}
-    @param line: vector defining a line
-
-    @type point: L{Vector}
-    @param point: vector defining the point
-    """
-    line = line.normalized()
-    np = point.norm()
-    angle = line.angle(point)
-    return point - line ** (np * numpy.cos(angle))
-
-
-def rotaxis2m(theta, vector):
-    """Calculate a left multiplying rotation matrix that rotates
-    theta rad around vector.
-
-    Example:
-
-        >>> m=rotaxis(pi, Vector(1, 0, 0))
-        >>> rotated_vector=any_vector.left_multiply(m)
-
-    @type theta: float
-    @param theta: the rotation angle
-
-
-    @type vector: L{Vector}
-    @param vector: the rotation axis
-
-    @return: The rotation matrix, a 3x3 Numeric array.
-    """
-    vector = vector.copy()
-    vector.normalize()
-    c = numpy.cos(theta)
-    s = numpy.sin(theta)
-    t = 1 - c
-    x, y, z = vector.get_array()
-    rot = numpy.zeros((3, 3))
-    # 1st row
-    rot[0, 0] = t * x * x + c
-    rot[0, 1] = t * x * y - s * z
-    rot[0, 2] = t * x * z + s * y
-    # 2nd row
-    rot[1, 0] = t * x * y + s * z
-    rot[1, 1] = t * y * y + c
-    rot[1, 2] = t * y * z - s * x
-    # 3rd row
-    rot[2, 0] = t * x * z - s * y
-    rot[2, 1] = t * y * z + s * x
-    rot[2, 2] = t * z * z + c
-    return rot
-
-
-rotaxis = rotaxis2m
-
-
-def refmat(p, q):
-    """Return a (left multiplying) matrix that mirrors p onto q.
-
-    Example:
-        >>> mirror=refmat(p, q)
-        >>> qq=p.left_multiply(mirror)
-        >>> print(q)
-        >>> print(qq) # q and qq should be the same
-
-    @type p,q: L{Vector}
-    @return: The mirror operation, a 3x3 Numeric array.
-    """
-    p.normalize()
-    q.normalize()
-    if (p - q).norm() < 1e-5:
-        return numpy.identity(3)
-    pq = p - q
-    pq.normalize()
-    b = pq.get_array()
-    b.shape = (3, 1)
-    i = numpy.identity(3)
-    ref = i - 2 * numpy.dot(b, numpy.transpose(b))
-    return ref
-
-
-def rotmat(p, q):
-    """Return a (left multiplying) matrix that rotates p onto q.
-
-    Example:
-        >>> r=rotmat(p, q)
-        >>> print(q)
-        >>> print(p.left_multiply(r))
-
-    @param p: moving vector
-    @type p: L{Vector}
-
-    @param q: fixed vector
-    @type q: L{Vector}
-
-    @return: rotation matrix that rotates p onto q
-    @rtype: 3x3 Numeric array
-    """
-    rot = numpy.dot(refmat(q, -p), refmat(p, -p))
-    return rot
-
-
-def calc_angle(v1, v2, v3):
-    """Calculate the angle between 3 vectors representing 3 connected points.
-
-    @param v1, v2, v3: the tree points that define the angle
-    @type v1, v2, v3: L{Vector}
-
-    @return: angle
-    @rtype: float
-    """
-    v1 = v1 - v2
-    v3 = v3 - v2
-    return v1.angle(v3)
-
-
-def calc_dihedral(v1, v2, v3, v4):
-    """Calculate the dihedral angle between 4 vectors representing 4 connected points.
-
-    The angle is in ]-pi, pi].
-
-    @param v1, v2, v3, v4: the four points that define the dihedral angle
-    @type v1, v2, v3, v4: L{Vector}
-    """
-    ab = v1 - v2
-    cb = v3 - v2
-    db = v4 - v3
-    u = ab ** cb
-    v = db ** cb
-    w = u ** v
-    angle = u.angle(v)
-    # Determine sign of angle
-    try:
-        if cb.angle(w) > 0.001:
-            angle = -angle
-    except ZeroDivisionError:
-        # dihedral=pi
-        pass
-    return angle
-
-
 class Vector(object):
-    """3D vector."""
+    "3D vector"
 
     def __init__(self, x, y=None, z=None):
         if y is None and z is None:
@@ -215,12 +29,12 @@ class Vector(object):
         return "<Vector %.2f, %.2f, %.2f>" % (x, y, z)
 
     def __neg__(self):
-        """Return Vector(-x, -y, -z)."""
+        "Return Vector(-x, -y, -z)"
         a = -self._ar
         return Vector(a)
 
     def __add__(self, other):
-        """Return Vector+other Vector or scalar."""
+        "Return Vector+other Vector or scalar"
         if isinstance(other, Vector):
             a = self._ar + other._ar
         else:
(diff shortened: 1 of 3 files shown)


Bio/PDB/Vector.py
```diff
@@ -10,194 +10,8 @@
...
-            angle = -angle
-    except ZeroDivisionError:
-        # dihedral=pi
-        pass
-    return angle
-
-
 class Vector(object):
-    """3D vector."""
+    "3D vector"
```
peterjc: These should be triple-quotes as per the docstring standard PEP257 https://www.python.org/dev/peps/pep-0257/

Bio/PDB/Vector.py
```diff
@@ -10,194 +10,8 @@
 import numpy
 
 
-def m2rotaxis(m):
```
peterjc: Why are you removing this function (and the others after it) from ``Bio.Vector``? They are part of the module's public API and should stay here.

Bio/PDB/Vector.py
peterjc: This is why the style check is failing under TravisCI, lots of entries like this from the ``pydocstyle`` tool;

```
D300: Use """triple double quotes""" (found '-quotes)
```

You can install this code checking tool with ``pip install pydocstyle``

Bio/PDB/Vector.py
peterjc: This explains why many of the tests are failing on TravisCI, e.g.

```
ERROR: test_DSSP_tool
  File "run_tests.py", line 436, in runTest
  File "/opt/python/2.7.9/lib/python2.7/unittest/loader.py", line 91, in loadTestsFromName
  File "/home/travis/build/biopython/biopython/Tests/test_DSSP_tool.py", line 17, in <module>
    from Bio.PDB import PDBParser
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/__init__.py", line 27, in <module>
    from .Polypeptide import PPBuilder, CaPPBuilder, is_aa, standard_aa_names
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/Polypeptide.py", line 60, in <module>
    from Bio.PDB.Vector import calc_dihedral, calc_angle
ImportError: cannot import name calc_dihedral
```

Gasta88: Apologies, I thought they were testing methods since they were outside the Class definition.

I'll sort them out this weekend.

Gasta88:
One side note, when I test for the division operand I get this error:
```
>>> v1 / 2
TypeError: unsupported operand type(s) for /: 'Vector' and 'int'
```

where `v1` is a Vector object. I'm using Python 3.6

peterjc:
If division works under Python 2, that's likely a bug we should fix under Python 3.

peterjc:
There are still some minor style issue being picked up by TravisCI running flake8,

```
$ flake8 Bio/
Bio/PDB/Vector.py:56:1: W293 blank line contains whitespace
Bio/PDB/Vector.py:75:1: W293 blank line contains whitespace
Bio/PDB/Vector.py:184:1: W293 blank line contains whitespace
```

You should be able to install flake8 locally using ``pip install flake8``.

Bio/PDB/Vector.py
```diff
@@ -69,7 +71,9 @@ def vector_to_axis(line, point):
 
 
 def rotaxis2m(theta, vector):
-    """Calculate a left multiplying rotation matrix that rotates
+    """Rotaxi2m method.
```
peterjc: How about ``Calculate left multiplying rotation matrix`` as the docstring one-line summary?

Bio/PDB/Vector.py
```diff
@@ -107,12 +111,12 @@ def rotaxis2m(theta, vector):
     rot[2, 2] = t * z * z + c
     return rot
 
-
 rotaxis = rotaxis2m
 
 
 def refmat(p, q):
-    """Return a (left multiplying) matrix that mirrors p onto q.
+    """
```
peterjc: The first line of the docstring should be on the triple-quote line, see PEP257 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

Tests/test_PDB.py
```diff
@@ -526,13 +526,6 @@ def test_pdbio_write_truncated(self):
         finally:
             os.remove(filename)
 
-    def test_deepcopy_of_structure_with_disorder(self):
```
peterjc: You seem to have accidentally removed this PDB test for deepcopy?

Tests/test_PDB.py
```diff
@@ -1157,6 +1150,48 @@ def test_transform(self):
...
+        v1 = Vector(0, 0, 1)
+        v2 = Vector(0, 0, 0)
+        v3 = Vector(0, 1, 0)
+        v4 = Vector(1, 1, 0)
+
+        self.assertEqual(calc_angle(v1, v2, v3), 1.5707963267948966)
+        self.assertEqual(calc_dihedral(v1, v2, v3, v4), 1.5707963267948966)
+        ref = refmat(v1, v3)
+        rot = rotmat(v1, v3)
+        self.assertEqual(ref[0].all(), numpy.array([1.0, 0.0, 0.0]).all())
```
peterjc: I think the ``.all()`` bits should be removed, note ``numpy.array([1.0, 0.0, 0.0]).all()`` u
is just ``False``.

Tests/test_PDB.py
```diff
@@ -1157,6 +1150,48 @@ def test_transform(self):
...
+        self.assertEqual((v1 + (1, 2, 3)).get_array().all(), numpy.array([1.0, 2.0, 4.0]).all())
+        self.assertEqual(v1 * v2, 0.0)
+        self.assertEqual((v1 ** v2).get_array().all(), numpy.array([0.0, -0.0, 0.0]).all())
+        self.assertEqual((v1 ** 2).get_array().all(), numpy.array([0.0, 0.0, 2.0]).all())
+        self.assertEqual((v1 ** (1, 2, 3)).get_array().all(), numpy.array([0.0, 0.0, 3.0]).all())
+        self.assertEqual(v1.norm(), 1.0)
+        self.assertEqual(v1.normsq(), 1.0)
+        v1[2] = 10
+        self.assertEqual(v1.__getitem__(2), 10)
+        angle = random() * numpy.pi
```
peterjc: Can this onwards be a separate test method?

peterjc:
TravisCI is failing currently on Python 3.6 with:

```
FAIL: test_Vector (test_PDB.TransformTests)
Test Vector object
  File "/home/travis/build/biopython/biopython/Tests/test_PDB.py", line 1193, in test_Vector
    self.assertEqual((axis - caxis).get_array().all(), numpy.array([0, 0, 0]).all())
AssertionError: True != False
```

As noted earlier, removing the ``.all()`` bits in these tests will likely help here.

Tests/test_PDB.py
Gasta88: Do you mean the part from line 1184 or 1187?

Tests/test_PDB.py
Gasta88: This is a behavior that I don't understand. 
When I launch the unittest of this script from my local machine, the entire test compiles successfully without a warning. 

Is it just this one line or all the others that make a comparison with a `numpy.array`?

Tests/test_PDB.py
peterjc: Lines 1198 to 1193 looked self contained:

```python
   def test_Vector(self):
        ...
        angle = random() * numpy.pi
        axis = Vector(random(3) - random(3))
        axis.normalize()
        m = rotaxis(angle, axis)
        cangle, caxis = m2rotaxis(m)
        self.assertAlmostEqual((angle - cangle), 0.0)
        self.assertEqual((axis - caxis).get_array().all(), numpy.array([0, 0, 0]).all())
```

Could you do:

```python
    def test_Vector(self):
        ...

    def test_angles(self):
        angle = random() * numpy.pi
        axis = Vector(random(3) - random(3))
        axis.normalize()
        m = rotaxis(angle, axis)
        cangle, caxis = m2rotaxis(m)
        self.assertAlmostEqual((angle - cangle), 0.0)
        self.assertEqual((axis - caxis).get_array().all(), numpy.array([0, 0, 0]).all())
```

Tests/test_PDB.py
peterjc: I think you should remove every ``.all()`` in your tests, it just reduces an array to a single boolean value (usually ``True`` but it would be ``False`` for a zero vector).

```
$ python3
Python 3.6.0 (v3.6.0:41df79263a11, Dec 22 2016, 17:23:13) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import numpy
>>> a = numpy.random.random(3)
>>> b = numpy.random.random(3)
>>> a == b#
array([False, False, False], dtype=bool)
>>> a.all() == b.all()
True
>>> a.all()
True
>>> b.all()
True
```

Tests/test_PDB.py
```diff
@@ -41,6 +40,7 @@
 from Bio.PDB import make_dssp_dict
 from Bio.PDB import DSSP
 from Bio.PDB.NACCESS import process_asa_data, process_rsa_data
+from numpy.random import random
```
peterjc: Please move this next to the other numpy imports, e.g. after line 26:

```python
from numpy.linalg import svd, det
```

peterjc:
The style failure looks easy to fix,

```
$ flake8 Bio/
Bio/PDB/Vector.py:56:1: W293 blank line contains whitespace
Bio/PDB/Vector.py:75:1: W293 blank line contains whitespace
Bio/PDB/Vector.py:184:1: W293 blank line contains whitespace
```

You should be able to install this code checking tool locally with ``pep install flake8``.

peterjc:
The line ``from copy import deepcopy`` is still missing from ``test_PDB.py`` which explains this failure:

```
ERROR: test_deepcopy_of_structure_with_disorder (test_PDB.ParseTest)
Test deepcopy of a structure with disordered atoms.
  File "/home/travis/build/biopython/biopython/Tests/test_PDB.py", line 533, in test_deepcopy_of_structure_with_disorder
    _ = deepcopy(self.structure)
NameError: global name 'deepcopy' is not defined
```

peterjc:
I understand now why you were trying to use ``.any()`` earlier, I should have spotted this earlier - one option is to use ``numpy.array_equal`` here:

```python
>>> import numpy
>>> a = numpy.array([1.0, 2.0, 3.0])
>>> b = numpy.array([1.0, 2.0, 3.0])
>>> a == b
array([ True,  True,  True], dtype=bool)
>>> numpy.array_equal(a, b)
True
```

More generally, using ``self.assertAlmostEqual(a, b, places=3)`` should allow for some floating point number variation between platforms.

Gasta88:
Apologies for the error in `Vector.py`, but I realized I've committed only one file change while I left this one unmodified.

The matter with `deepcopy` must have been one clumsy error. Hopefully will be resolved with the next change (famous last words...)

At the moment I'm trying both options to test `numpy.array` objects and see how my version of Python react.

Gasta88:
Before updating this PR, there is one problem. If I run the following commands in the Python console I get the displayed result:
```python
>>> import numpy
>>> from Bio.PDB import Vector
>>> from numpy.random import random
>>> from Bio.PDB import rotaxis,m2rotaxis
>>> angle = random() * numpy.pi
>>> axis = Vector(random(3) - random(3))
>>> axis.normalize()
>>> m = rotaxis(angle, axis)
>>> cangle, caxis = m2rotaxis(m)
>>> axis
<Vector 0.08, -0.10, -0.99>
>>> caxis
<Vector 0.08, -0.10, -0.99>
>>> numpy.array_equal((axis-caxis).get_array(),numpy.array([0,0,0]))
True
```

On the other hand, I get the following if run in Powershell:
```
PS H:\Documents\biopython\Tests> python .\test_PDB.py TransformTests.test_Vector_angles
test_Vector_angles (__main__.TransformTests) ... FAIL

FAIL: test_Vector_angles (__main__.TransformTests)
  File ".\test_PDB.py", line 1201, in test_Vector_angles
    self.assertTrue(numpy.array_equal((axis - caxis).get_array(), numpy.array([0, 0, 0])))

Ran 1 test in 0.107s

FAILED (failures=1)
```

Do you think it depends on the environment?

`test_Vector()` doesn't display any error.

peterjc:
You can give a second argument like ``self.assertTrue(value, msg)``

Try using something like this:

```python
    self.assertTrue(numpy.array_equal((axis - caxis).get_array(), numpy.array([0, 0, 0])),
                    "Expected zero vector, got %r" % (axis - caxis).get_array())

```

I've not checked this but isn't this just trying to compare ``axis`` and ``xaxis``, so would this make sense instead:

```python
    self.assertAlmostEqual(axis, caxis, places=3)
```

peterjc:
Probably what is happening is that ``axis`` and ``caxis`` are very similar but not always equal, and that could depend on the environment (which version of Python and how it was compiled etc).

Gasta88:
I started with the second suggestion since it seems logical, but there is this problem now:
```
TypeError: bad operand type for abs(): 'Vector'
```

It seems that `abs()` doesn't relate well with the `Vector` object, but it's successful with `Vector.get_array()` since it generates a `numpy.array`.

I changed the test to:
```python
self.assertTrue(numpy.array_equal(axis.get_array(), caxis.get_array()),
                    "Expected zero vector, got %r" % (axis - caxis).get_array())
```

but each time the arrays have similar values but not equal. In this case the test will fail more frequently that it should because of a matter of precision.

Does having the `msg` parameter set up helps to maintain testing quality?

peterjc:
If ``self.assertTrue(value)`` fails all you are told is that ``value`` was not True. In this case you don't see what the values given to ``numpy.array_equal`` where, so an explicit failure message is useful.

With the ``self.assertEqual`` and ``assertAlmostEqual`` the default message is very good.

How about something like this (again, untested)?

``` python
self.assertAlmostEqual(axis.get_array(), caxis.get_array(), places=3)
```

Gasta88:
```python
self.assertAlmostEqual(axis.get_array(), caxis.get_array(), places=3)
```
doesn't work either because two `numpy.array` objects are compared and it doesn't like it unless a .all() or .any() is specified in the test.

I'll stick with the message and commit the new files.

peterjc:
How about:
```python
self.assertTrue(numpy.allclose(axis.get_array(), caxis.get_array(),
               "Want %r and %r to be almost equal" % (axis, caxis))
```

The ``numpy.allclose()`` function has optional arguments for the strictness of the check.

peterjc:
Don't worry about the CodeCov numbers - it seems to be a little unreliable at the moment because I made some setting changes on the master branch.

Tests/run_tests.py
```diff
@@ -98,6 +98,7 @@ def is_numpy():
     "Bio.Blast.Applications",
     "Bio.Emboss.Applications",
     "Bio.GenBank",
+    "Bio.Graphics.GenomeDiagram._Colors",
```
peterjc: Please remove this change as it not relevant to the PDB Vector test work.

peterjc:
Other than the extra change to ``run_tests.py``, this looks ready to merge - good job 👍

Gasta88:
`run_tests.py` has been changed as well, thanks for keeping up with this one Peter :)

Gasta88:
`run_tests.py` has been updated. Thanks for keeping up with this one Peter :)

peterjc:
Thank you :)

Issue #1107, opened 2017-03-22
GC_skew () issue


Cashalow:
Hello, 
I found a rather strange issue in the GC_skew() function in Bio.SeqUtils.
If you run this code : 

import Bio
import Bio.SeqUtils
from Bio.Seq import Seq

Bio.SeqUtils.GC_skew(Seq("G"*100 +"A"))

this will fail because GC_skew will try to calculate the skew on the last remaining nucleotide which is not a G/C. Should this function check before trying to divide by zero?

Cheers and thanks for the good work !

peterjc:
You missed out the error, under Python 3.6 I get:

```
>>> import Bio
>>> import Bio.SeqUtils
>>> from Bio.Seq import Seq
>>> Bio.SeqUtils.GC_skew(Seq("G"*100 +"A"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/Bio/SeqUtils/__init__.py", line 106, in GC_skew
    skew = (g - c) / float(g + c)
ZeroDivisionError: float division by zero
```

This is a corner case, due to the windowing scheme the final chunk of sequence can be far smaller than the window size (here one base ``A`` rather than a window of 100bp), and it is therefore more likely to have no ``C`` or ``G`` and thus give a divide by zero exception.

What would you suggest be returned here? ``None`` or the special float value not-a-number (I forget now the best way to define NaN for cross platform and cross version compatibility - something like ``float("NaN")`` though I think).

Cashalow:
Yes indeed but it could also happen with a theoretical window of only A/Ts. I am trying to see how other tools deal with this theoretical case and this one http://genskew.csb.univie.ac.at/ returns "0" when no G/C are found, which as far as I'm concerned makes sense because indeed no GC skew can be observed on such sequences.

peterjc:
You could make a good case for returning zero here too.

peterjc:
Would you like to submit a pull request returning zero in this cases (and explaining this in the docstring)?

Ideally please include a doctest or unittest for this as well.

Cashalow:
Yes sorry, I have prepared the pull request but did not send it yet. I'll try to add a test as well then.

PR #1131, opened 2017-04-04
Small fix to gc skew

Cashalow:
In reference to issue #1107 I have add a check whether G and C are in same quantity and return 0 if so. And a unittest on a window with no G or C.

### DIFF
--- a/Bio/SeqUtils/__init__.py
+++ b/Bio/SeqUtils/__init__.py
@@ -103,7 +103,10 @@ def GC_skew(seq, window=100):
         s = seq[i: i + window]
         g = s.count('G') + s.count('g')
         c = s.count('C') + s.count('c')
-        skew = (g - c) / float(g + c)
+        if g == c:
+            skew = 0
+        else:
+            skew = (g - c) / float(g + c)
         values.append(skew)
     return values
 
--- a/Bio/SeqUtils/__init__.py
+++ b/Bio/SeqUtils/__init__.py
@@ -94,7 +94,9 @@ def GC_skew(seq, window=100):
 
     Returns a list of ratios (floats), controlled by the length of the sequence
     and the size of the window.
-
+    
+    If the number of G and C is equal, returns 0. Fixes the issue when the window is devoid of G/Cs.
+    
     Does NOT look at any ambiguous nucleotides.
     """
     # 8/19/03: Iddo: added lowercase
--- a/Bio/SeqUtils/__init__.py
+++ b/Bio/SeqUtils/__init__.py
@@ -94,9 +94,9 @@ def GC_skew(seq, window=100):
 
     Returns a list of ratios (floats), controlled by the length of the sequence
     and the size of the window.
-    
+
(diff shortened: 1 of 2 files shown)


peterjc:
When ``g == c``, the existing code didn't have any problem - it would already return ``0`.

I was expecting you to check if ``g + c == 0`` to avoid a division by zero error?

peterjc:
One of the things we check with TravisCI is Python style, and that is failing:

```
$ flake8 Bio/
Bio/SeqUtils/__init__.py:97:1: W293 blank line contains whitespace
Bio/SeqUtils/__init__.py:99:1: W293 blank line contains whitespace
```

You can install flake8 using ``pip install flake8``.

peterjc:
You'll need to import ``GC_skew`` at the top of ``test_SeqUtils.py`` in order to fix this error:

``
ERROR: test_GC_skew (test_SeqUtils.SeqUtilsTests)
  File "/home/travis/build/biopython/biopython/Tests/test_SeqUtils.py", line 145, in test_GC_skew
    self.assertEqual(GC_skew(seq)[0], 0)
NameError: global name 'GC_skew' is not defined
```

**Comment by Cashalow** [comment]
Yes indeed sorry I got confused in my branches and I forgot the include in the one I sent.

I'll check the Python styles ASAP.

Regarding my fix, indeed it worked when g == c but I was just thinking that this would save one operation for the cases where G and Cs are present and equal. I could check g + c == 0 instead but I think the count function never returns negative values so my fix should be safe.

**Comment by peterjc** [comment]
Yes, the counts should never be negative.

How about something like this?:

```
    try:
        skew = (g - c) / float(g + c)
    except ZeroDivisionError:
        skew = 0.0
```

**Comment by Cashalow** [comment]
Okay then :) It was also the way this was done for the GC function in the same file, I should have checked it before, sorry.

**Comment by peterjc** [comment]
(The coverage report is a bit odd, but I was tinkering with coverage settings this afternoon)

Other than the docstring this looks good:

```
If the numbers of G and C are equal, returns 0. Fixes the issue when the window is devoid of G/Cs.
```

Something like this?:

```
If a window contains no G or C bases, a skew of zero is used.
```

Cashalow:
Yes I agree to all of this :)

peterjc:
I emailed your GMail account a few days ago: Would you prefer to be acknowledged under your real name, or just as Cashalow?

peterjc:
Thank you - merged as the first pull request for what will be Biopython 1.70.

peterjc:
@sachalau Hi again Sacha. Are you happy to dual licence your contributions under both the Biopython License Agreement AND the BSD 3-Clause License? See #898. Thank you.

sachalau:
yes

Le jeu. 20 sept. 2018 à 16:32, Peter Cock <notifications@github.com> a
écrit :

peterjc:
Thank you - We still need one more author's agreement before we can dual license ``Bio/SeqUtils/__init__.py`` but this is progress.

Issue #1130, opened 2017-04-04
Format change in prosite.dat


francoiscoste:
Hi and thanks for your work on this nice library!

In Biopython 1.68, the new date format of prosite.dat is not supported. Here is the error message:
  
```
/usr/lib64/python3.5/site-packages/Bio/ExPASy/Prosite.py in parse(handle)
     34     handle   - handle to the file."""
     35     while True:
---> 36         record = __read(handle)
     37         if not record:
     38             break

/usr/lib64/python3.5/site-packages/Bio/ExPASy/Prosite.py in __read(handle)
    172                (not dates[1].endswith('(DATA UPDATE)')) or \
    173                (not dates[2].endswith('(INFO UPDATE)')):
--> 174                 raise ValueError("I don't understand date line\n%s" % line)
    175             record.created = dates[0].rstrip(' (CREATED)')
    176             record.data_update = dates[1].rstrip(' (DATA UPDATE)')

ValueError: I don't understand date line
DT   01-APR-1990 CREATED; 01-APR-1990 DATA UPDATE; 01-APR-1990 INFO UPDATE.
```

The format of DT field changed indeed recently according to ftp://ftp.expasy.org/databases/prosite/NEWS to add a day field (and remove parenthesis!) :

New (current) format:
```
DT DD-MMM-YYYY CREATED; DD-MMM-YYYY DATA UPDATE; DD-MMM-YYYY INFO UPDATE.
```
Old format:
```
DT MMM-YYYY (CREATED); MMM-YYYY (DATA UPDATE); MMM-YYYY (INFO UPDATE).
```

Example:
New (current) format:
```
DT 01-APR-1990 CREATED; 01-MAY-2005 DATA UPDATE; 01-OCT-2016 INFO UPDATE.
```
Old format:
```
DT APR-1990 (CREATED); MAY-2005 (DATA UPDATE); OCT-2016 (INFO UPDATE).
```

The release numbers seems to have changed also in January 2017....
Best regard,
François

[Peter: Edited to include triple-back-tick markup of the examples]

peterjc:
Thank you for your well researched report - would you like to try preparing a fix for this (pull request with one new test file) so that it can parse both old and new style ``DT`` lines?

francoiscoste:
Hi again,
I am not familiar with github and biopython, but the fix can be very simple...
I have made a pull request "small fix to deal with new date format in prosite" #1132
I hope it can help...
Best,
François

PR #1132, opened 2017-04-04
small fix to deal with new date format in prosite

francoiscoste:
Following Issue "Format change in prosite.dat" #1130, here is a proposition of simple fix handling both versions.


peterjc:
The TravisCI checks are all working, which is good. I'm not sure why the QuantifiedCode check is failing, but that ought to be fine. CodeCov is noting that we don't have any tests which use the new code

 Can you suggest a suitable (small) example file we can include in the test suite?

francoiscoste:
Here is an except of the new prosite.dat to test the reader:
[prosite2017-tail-head.zip](https://github.com/biopython/biopython/files/893659/prosite2017-tail-head.zip)
I am happy with the licences and being thanked :-)
Best,
François

peterjc:
I just accidentally solved the same thing while fixing an ExPASy bug - see #1215. You've included some additional useful notes about the timing of the change and the more details dates so I'll combine this...

peterjc:
@francoiscoste are you happy with the combined fix on #1215? Thanks!

peterjc:
Closed via #1215.

PR #1144, opened 2017-04-10
Implement SeqRecord .translate method

### LINKED ISSUE #1103: translate method for SeqRecord
We have a ``.reverse_complement(...)`` method in the ``SeqRecord``, using the method of the same name on the ``Seq`` object.

It would be natural to do this for ``.translate(...)`` as well, with similar behaviour for the metadata. i.e. default to dropping it, but make it easy to explicitly retain it (e.g. ``.translate(..., id=True)`` to retain the nucleotide record's ``.id`` on the protein record.

This would have been useful in some recent training course materials.

peterjc:
Intended to close #1103, opening as a pull request for more eyes to look at.

Should we list the translation arguments first, or the annotation arguments?

This does not (yet) have full test coverage (e.g. using the annotation arguments).

### DIFF
--- a/Bio/SeqRecord.py
+++ b/Bio/SeqRecord.py
@@ -1159,6 +1159,98 @@ class SeqRecord(object):
                 answer._per_letter_annotations[key] = value[::-1]
         return answer
 
+    def translate(self,
+                  # Seq translation arguments:
+                  table="Standard", stop_symbol="*", to_stop=False,
+                  cds=False, gap=None,
+                  # SeqRecord annotation arguments:
+                  id=False, name=False, description=False,
+                  features=False, annotations=False,
+                  letter_annotations=False, dbxrefs=False):
+        """Returns new SeqRecord with translated sequence.
+
+        This calls the record's .seq.translate() method (which describes
+        the translation related arguments, like table for the genetic code),
+
+        By default the new record does NOT preserve the sequence identifier,
+        name, description, general annotation or database cross-references -
+        these are unlikely to apply to the translated sequence.
+
+        You can specify the returned record's id, name and description as
+        strings, or True to keep that of the parent, or False for a default.
+
+        You can specify the returned record's features with a list of
+        SeqFeature objects, or False (default) to omit them.
+
+        You can also specify both the returned record's annotations and
+        letter_annotations as dictionaries, True to keep that of the parent
+        (annotations only), or False (default) to omit them.
+
+        e.g. Loading a FASTA gene and translating it,
+
+        >>> from Bio import SeqIO
+        >>> gene_record = SeqIO.read("Fasta/sweetpea.nu", "fasta")
+        >>> print(gene_record.format("fasta"))
+        >gi|3176602|gb|U78617.1|LOU78617 Lathyrus odoratus phytochrome A (PHYA) gene, partial cds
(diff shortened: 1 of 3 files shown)


Bio/SeqRecord.py
```diff
@@ -1159,6 +1159,98 @@ def reverse_complement(self, id=False, name=False, description=False,
...
+            answer.dbxrefs = self.dbxrefs[:]
+        if isinstance(features, list):
+            answer.features = features
+        elif features:
+            # Does not make sense to copy old features as locations wrong
+            raise TypeError("Unexpected features argument %r" % features)
+        if isinstance(annotations, dict):
+            answer.annotations = annotations
+        elif annotations:
+            # Copy the old annotations,
```
cbrueffer: Either stray comma or part of the sentence missing.

cbrueffer:
Looks good to me; I think putting the translation arguments first is the right choice.

peterjc:
@bow you mean use ``**kwargs`` now so that if we have to insert extra arguments later it will not break the implied positional based argument API?

Bio/SeqRecord.py
peterjc: Good spot, fixed - thanks.

bow:
@peterjc , yes. So `SeqRecord.translate`'s signature will be something like:

```
def translate(self,
              # SeqRecord annotation arguments:
              id=False, name=False, description=False,
              features=False, annotations=False,
              letter_annotations=False, dbxrefs=False
              # Seq translation arguments:
              **kwargs):

```

But like I said, I'm fine with the current changes as-is. This is only to make `SeqRecord.translate` less dependent on the actual `Seq.translate` arguments (which I don't imagine will change anytime soon).

peterjc:
I think given so many people will have used ``my_seq.translate(11)`` as shorthand for ``my_seq.translate(table=11)`` from a usability point of view having the translation arguments first is preferable, allowing ``my_record.translate(11)`` to work.

However, there is a chance of the ``Seq`` object's translate method changing (see #1164 and #1224), so perhaps your ``**kwargs`` idea could be used like this to prevent anyone assuming that the ``id=...`` argument etc are at any particular position:

```python
    def translate(self,
                  # Seq translation arguments:
                  table="Standard", stop_symbol="*", to_stop=False,
                  cds=False, gap=None,
                  # SeqRecord annotation arguments:
                  **kwargs):
```

bow:
Aha, I missed the two possible upcoming changes.. The idea was more to gather all arguments that belongs to the less 'stable' method into `**kwargs`. Wrapping `SeqRecord`'s arguments into `**kwargs` is indeed convenient, though it will then require a change on `SeqRecord.translate` when/if `Seq.translate` changes.

Hmm...maybe it is even easier to have a new `SeqRecord` method that accepts a function that transforms the underlying `Seq` object? Then we don't need any hard-coded wrapper methods anymore.

But anyway, maybe I am overthinking this :).

peterjc:
I don't think we can generalise like that - any transformation of the sequence is going to have implications for the annotation, the only safe option would be to drop all of it - at which point there is no need for a ``SeqRecord``, so you might as well use ``my_record.seq.some_method()`` directly.

peterjc:
Rebased with a NEWS entry added. I'm inclined to merge this now that Biopython 1.70 is out, since people seem generally happy with it.

peterjc:
Rebased again with trivial merge conflict in NEWS file fixed.

PR #1147, opened 2017-04-11
remove self-test Bio/PDB/StructureAlignment.py

Gasta88:
In reference to #820 I have:

- Added `Tests/PDB/Alignment_file.fa`
- Removed self-test from `Bio/PDB/StructureAlignment.py`
- Added tests at `Tests/test_PDB.py`


Tests/test_PDB.py
```diff
@@ -31,14 +31,14 @@
     raise MissingPythonDependencyError(
         "Install NumPy if you want to use Bio.PDB.")
 
-from Bio import BiopythonWarning
+from Bio import BiopythonWarning, AlignIO
```
peterjc: Style-wise put that as two import lines please

Tests/test_PDB.py
```diff
@@ -1231,14 +1257,14 @@ def test_entitity_copy(self):
...
 
 def eprint(*args, **kwargs):
-    '''Helper function that prints to stderr.'''
+    """Helper function that prints to stderr."""
     print(*args, file=sys.stderr, **kwargs)
 
 
 def will_it_float(s):
-    '''
+    """
```
peterjc: Thanks for making this docstring (and the others) use the triple-double-quotes style, that's great. Could you also fix this entry to have a blank line after the one-line summary - see PEP257 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

peterjc:
The sequence alignment doesn't look very realistic (not likely to be biologically meaningful), so I assume this is simply in order to use two of the existing PDB files already in the test suite? That makes sense 👍

Tests/test_PDB.py
Gasta88: With your permission, I've added a summary line and then a modified version of the original text separated with a blank line.

`pydocstyle` hasn't complained on my local machine, let's see what TravisCI says.

Gasta88:
Following you request on the main issue #820 , I've decide to pick two PDBs file already present inside the test suite.
If you think I should find a more meaningful couple, I can always add a new test to the script. But in this case I need to add a new PDB file as well :)

Tests/test_PDB.py
peterjc: Maybe we should file an enhancement request with ``pydocstyle`` to check this...

peterjc:
The tests are failing, e.g.

```
ERROR: test_StructAlign (test_PDB.StructureAlignTests)
Tests on module to align two proteins according to a FASTA alignment file.
  File "/home/travis/build/biopython/biopython/Tests/test_PDB.py", line 1223, in test_StructAlign
    al = StructureAlignment(records, m1, m2)
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/StructureAlignment.py", line 45, in __init__
    column = fasta_align.get_column(i)
AttributeError: 'MultipleSeqAlignment' object has no attribute 'get_column'
```

This is due to a change in the alignment objects - we had deprecated ``.get_column`` and it was removed in Biopython 1.69.

This is good news - the new test has found a problem which we can fix :)

Please try replacing that line with ``column = fasta_align[:, i]`` instead.

peterjc:
Excellent, thank you :)

Issue #1152, opened 2017-04-14
Testing Bio/PDB/REsidueDepth.py, missing MSMS software


Gasta88:
Hi,

in the process of testing `Bio/PDB/ResidueDepth.py` I'm finding some problems.

I'm  using one of the PDB in the `Tests/PDB` folder. 
As mentioned in the `Bio/PDB/ResidueDepth.py`, a tool called MSMS is required (source: http://mgltools.scripps.edu/packages/MSMS).

I download and install the application for Windows. This is the result:

```
'\\homes\fgastaldello\Documents\Python Scripts'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
'pdb_to_xyzr' is not recognized as an internal or external command,
operable program or batch file.
'\\homes\fgastaldello\Documents\Python Scripts'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
'msms' is not recognized as an internal or external command,
operable program or batch file.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\fgastaldello\AppData\Local\Continuum\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 685, in runfile
    execfile(filename, namespace)
  File "C:\Users\fgastaldello\AppData\Local\Continuum\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 85, in execfile
    exec(compile(open(filename, 'rb').read(), filename, 'exec'), namespace)
  File "//homes/fgastaldello/Documents/Python Scripts/ResDepth_minitest.py", line 8, in <module>
    rd = ResidueDepth(model, prot_file)
  File "C:\Users\fgastaldello\AppData\Local\Continuum\Anaconda3\lib\site-packages\Bio\PDB\ResidueDepth.py", line 144, in __init__
    surface = get_surface(pdb_file)
  File "C:\Users\fgastaldello\AppData\Local\Continuum\Anaconda3\lib\site-packages\Bio\PDB\ResidueDepth.py", line 90, in get_surface
    "Failed to generate surface file using command:\n%s" % make_surface
AssertionError: Failed to generate surface file using command:
[...]

Issue #1140, opened 2017-04-08
ResidueDepth fails on Windows


JoaoRodrigues:
Related to a question posted on the users mailing list, quoted below:

>AssertionError: Failed to generate surface file using command:msms -probe_radius 1.5 -

Running the code on Windows 10 with the latest available MSMS executable fails because of the code calls a BASH script. This script - `pdb_to_xyzr` - seems to be mapping atom names to radii, so we might be able to add those as a table to the ResidueDepth module and have full Windows compatibility.

peterjc:
I wonder if the Linux subsystem for Windows 10 would fix this for us?

JoaoRodrigues:
The msms executable wouldn't run there then, nor we would have access to
the Windows' Biopython installation. I'm looking at the files, seems to be
a simple fix if I write a parsing function that replaces the BASH script.

underOATH777:
Hi, when is this going to be fixed?

peterjc:
@underOATH777 we don't know. Would you like to help? Do you have access to a Windows machine and any experience with the MSMS tool?

underOATH777:
I definitely would, if I could. My work depends on this. I'm even thinking of installing Ubuntu just to get it done. However not being very familiar with Linux systems is holding me back a little. Do you think I'm better off waiting for a fix, or should I just take the red pill and embark on a Linux adventure?

JoaoRodrigues:
On this today, expect a PR in a couple of hours. I changed machines
recently and had to go through the trouble of getting my compiler chain/git
working correctly before doing any work!

On the other hand, getting a Virtual Machine with Linux wouldn't hurt,
since a lot of programs just bail when it comes to supporting Windows.. Or,
try the new WSL (Bash on Windows). I've had mixed success with compilations
there, but I could get CNS (old FORTRAN code) running there so anything
written after 1995 *might* work :)

underOATH777:
Thanks a lot for your input and efforts. I'm waiting eagerly for the "patch release", if that's what PR means. Cheers. :D

JoaoRodrigues:
@underOATH777 I successfully ran MSMS here on my Win 64 machine and it seems to give proper values for radii (compared to the original BASH script). When the pull gets accepted and merged, can you download the latest version of Biopython (from Git) and test if it works on your end as well?

underOATH777:
Of course, I'd be happy to. :)

underOATH777:
[...]

PR #1158, opened 2017-04-16
Support MSMS calls on Windows

JoaoRodrigues:
Module `ResidueDepth.py` calls a shell script to obtain atomic radii before calling MSMS. I added the radii values used by MSMS and wrote a mapper function to replace the shell script, therefore adding support for Windows systems. Validated against the output of the examples bundled with MSMS.

As a unit test, I added a function in `test_PDB.py` that calls the bit of code that reproduces the `pdb_to_xyzr` script and compare its output to a real output (deposited in the `Tests/PDB` folder). 

Addresses (and hopefully fixes) issue #1140 and should help @Gasta88 in his brilliant self-test quest (issue #1152 ).


JoaoRodrigues:
Also edited the .tex and .lyx files to reflect the changes in the `ResidueDepth` class, namely the fact that it does not need a PDB file anymore (reads directly from the model object). This *will* break compatibility with the previous code, but it should be a simple fix once people read the code..

Doc/biopdb_faq.tex
```diff
@@ -71,7 +71,7 @@
 {\normalsize Institute of Molecular Biology}\\
 {\normalsize University of Copenhagen }\\
 {\normalsize Universitetsparken 15, Bygning 10}\\
-{\normalsize DK-2100 K�benhavn �}\\
+{\normalsize DK-2100 K�benhavn �}\\
```
peterjc: Something weird has happened to the special characters. I've changed this on the master to avoid the problem: 1891973fe3cf5a4ad7e05f6a9ccd8d5d84f48a44

Doc/biopdb_faq.tex
```diff
@@ -1047,27 +1047,27 @@ \subsubsection*{How do I calculate residue depth?}
...
+function), in the form of a Numpy python array with the surface points.
 
 
 \subsubsection*{How do I calculate Half Sphere Exposure?}
 
 Half Sphere Exposure (HSE) is a new, 2D measure of solvent exposure.
 Basically, it counts the number of C$\alpha$ atoms around a residue
 in the direction of its side chain, and in the opposite direction
-(within a radius of 13 �). Despite its simplicity, it outperforms
+(within a radius of 13 �). Despite its simplicity, it outperforms
```
peterjc: Something weird has happened to the special characters. I've changed this on the master to avoid the problem: 1891973fe3cf5a4ad7e05f6a9ccd8d5d84f48a44

peterjc:
Can you replace the new very long ``if/else`` structure in ``Bio/PDB/ResidueDepth.py`` with a loop and a data-structure?

Bio/PDB/ResidueDepth.py
```diff
@@ -44,42 +48,422 @@
...
+            if len(sl) != 9:
                 # skip header
                 continue
             vl = [float(x) for x in sl[0:3]]
             vertex_list.append(vl)
     return numpy.array(vertex_list)
 
 
-def get_surface(pdb_file, PDB_TO_XYZR="pdb_to_xyzr", MSMS="msms"):
+def get_surface(model, MSMS="msms"):
```
peterjc: I suggest you leave ``PDB_TO_XYZR=...`` as the second argument, but as ``PDB_TO_XYZR=None``, and document this as being obsolete.

Then in the code, add something like:

```
if PDB_TO_XYZR is not None:
    # Raise warning or even an error about the API change.
```

That way it will work as is with existing scripts using the default.

JoaoRodrigues:
@peterjc 
>Can you replace the new very long if/else structure in Bio/PDB/ResidueDepth.py with a loop and a data-structure?

That was my first approach, but I couldn't find an easier way than this. There are comparisons between residue names, atom names, partial or full matches.. The other options was to go for regular expressions but I didn't see the added value honestly.

I'll change the header of the function for backwards compatibility then (but add a deprecation warning).

peterjc:
OK, the matching rules do look very fiddly so I can understand using the if/else structure.

Does it make sense to put the atom radii into ``Bio/Data/IUPACData.py`` (assuming they are from IUPAC) or under ``Bio/Data/`` if not? If we do that, keying by atom name would seem more reusable than atomic number? See also ##1119 which is somewhat related.

Finally there is a minor style issue to fix:

```
$ pydocstyle Bio/ BioSQL/ Tests/ Scripts/ Doc/
Bio/PDB/ResidueDepth.py:111 in private function `_get_atom_radius`:
        D202: No blank lines allowed after function docstring (found 1)
Bio/PDB/ResidueDepth.py:451 in public function `get_surface`:
        D202: No blank lines allowed after function docstring (found 1)
```

JoaoRodrigues:
@peterjc Fixed those issues.

Regarding the radii data, these are not IUPAC values, but rather those used by Connolly in his implementation of the surface calculation algorithm. I'd keep these here, since they are tied to MSMS.

peterjc:
``pydocstyle`` tries to enforce PEP257 docstring style guidelines, see also #1170 

Your explanation about the radius values makes sense to me now :)

The CodeCov numbers could be better (patch coverage), but the drop in overall coverage of 0.08% seems to be mostly noise. I'll merge this anyway...

PR #1163, opened 2017-04-17
Stricter alphabet checking in codonalign

peterjc:
What do you think @zruan? I'm wondering if this should instead check for the argument being an instance of class CodonAlphabet?

CC @universvm as he's been trying this module recently.

### DIFF
--- a/Bio/codonalign/codonseq.py
+++ b/Bio/codonalign/codonseq.py
@@ -16,6 +16,8 @@ from math import log
 from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
 from Bio.Alphabet import generic_dna, _ungap
+from Bio.Alphabet import SingleLetterAlphabet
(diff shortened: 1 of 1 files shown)


zruan:
Is `CodonAlphabet` the only alphabet in biopython that is not `SingleLetterAlphabet`? I think a direct checking like `isinstance(codon_alphabet, CodonAlphabet)` is stricker and will avoid ambiguity.

peterjc:
There are some interesting alphabets used in ``test_GAOrganism.py``, and there is the ``ThreeLetterProtein`` - but almost everything else would be based on ``SingleLetterAlphabet``.

OK, let's add a check for ``isinstance(codon_alphabet, CodonAlphabet)`` here instead. Do you want to do this, or should I?

zruan:
Pull request #1180 is made.

peterjc:
Closes in favour of #1180

PR #1166, opened 2017-04-18
Deprecate Bio.PDB.Dice

peterjc:
As discussed on #820, ``Bio.PDB.Dice``was likely intended as an example script using ``PDBIO`` rather than as part of the ``Bio.PDB`` public API.

See mailing list thread:

http://mailman.open-bio.org/pipermail/biopython/2017-April/016132.html
http://mailman.open-bio.org/pipermail/biopython-dev/2017-April/021687.html


peterjc:
I've reverted this in 3b7f121371e2a0f30177a7643acf9ac521996ffd

PR #1154, opened 2017-04-14
moving self_test from Bio/PDB/Superimposer.py

Gasta88:
As mentioned in issue #820, I've moved the self test from `Bio/PDB/Superimposer.py` to `Tests/test_PDB.py`.

The test has been performed on a single structure which has been used twice. Let me know if I need to use two different structures.

PR #1168, opened 2017-04-18
Moved self-test for Bio/PDB/Superimposer.py

Gasta88:
Second attempt to move self-test from `Bio/PDB/Superimposer.py`
Previous PR #1154


peterjc:
Looks good, thanks!

Issue #1136, opened 2017-04-06
SearchIO.BlatIO assertion statements block parsing of blatx (protein query vs dnax database) PSL files


docmanny:
Hello all, 

While working on a program to automate BLAT searches, I stumbled across this error. In short, the mismatch between the query ranges and target ranges that are inherit in blatx searches are triggering the following AssertionError:

```
Traceback (most recent call last):
  File "/usr/lib64/python3.4/site-packages/IPython/core/interactiveshell.py", line 2881, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-34-743ae90aba50>", line 1, in <module>
    qresults = SearchIO.read('/home/manny/blatx_test.psl','blat-psl')
  File "/usr/lib64/python3.4/site-packages/Bio/SearchIO/__init__.py", line 357, in read
    first = next(generator)
  File "/usr/lib64/python3.4/site-packages/Bio/SearchIO/__init__.py", line 314, in parse
    for qresult in generator:
  File "/usr/lib64/python3.4/site-packages/Bio/SearchIO/BlatIO.py", line 401, in __iter__
    for qresult in self._parse_qresult():
  File "/usr/lib64/python3.4/site-packages/Bio/SearchIO/BlatIO.py", line 492, in _parse_qresult
    hsp = _create_hsp(prev_hid, prev_qid, prev)
  File "/usr/lib64/python3.4/site-packages/Bio/SearchIO/BlatIO.py", line 354, in _create_hsp
    assert hsp.hit_end == psl['tend']
AssertionError
```
This bug was replicated by myself using various versions of BLAT (v0.33, a few revisions of v0.34, and the recent v0.35), using various genomes from UCSC (Alligator, Human, and Mouse), and using different protein queries (CDN1B and Insulin in addition to UCHL3). All of the above combinations work if one uses nucleotide queries. 

To replicate this error, run a blat search using `blat [genome.2bit] [protein query] [output.psl] -q=prot -t=dnax -out=psl` to generate the required psl file. ([Here is one such file](https://github.com/biopython/biopython/files/903515/blatx_test.psl.txt), generated by blat-ing [this protein](https://www.ncbi.nlm.nih.gov/protein/48145827?report=fasta) against hg38 with the given command). Then, in python, run:

```
from Bio import SearchIO
qresults = SearchIo.read('blat_test.psl', 'blat-psl')
```

[...]

PR #1175, opened 2017-04-20
Fix BlatIO error caused by discrepancy in query results from proteins

bow:
This is a PR for fixing #1136. The bug was caused by some fields using DNA coordinates while the query is a protein sequence. Since we already detect when a protein query is used, the fix just required using the proper size multiplier for these fields.


peterjc:
Looks good to merge - only waiting to see if we get permission to add the example from #1136 as a test case.

Issue #1170, opened 2017-04-18
Improve pydocstyle (PEP257) compliance


peterjc:
Biopython aspires to follow Python's docstring style guidelines in PEP257 https://www.python.org/dev/peps/pep-0257/ (and use reStructuredText for the docstrings).

We run the automatic tool ``pydocstyle`` on the code as part of the automated TravisCI tests (also used on pull requests). Right now this is configured to ignore many style violations. For example, the new release of pydocstyle v2.0.0 added some extra checks, most of which I just disabled in order that our TravisCI builds pass - see https://github.com/biopython/biopython/commit/cc2ad72f1f237b3ff779b151fdf529abba1f3459

This issue is about (gradually) cleaning up PEP257 compliance issues as flagged by the latest version of the ``pydocstyle`` tool, and removing entries from the ``ignore`` setting in our ``.pydocstyle`` configuration file.

Note that in some cases we as a community might agree to continue to ignore some of these items.

To install/upgrade this tool, use something like this (or just ``pip`` if you are still on Python 2):

```
$ pip3 install --upgrade pydocstyle
```

The following command simples a useful way to see a tally table for picking off the rarest cases to fix them first. The cut is to remove the variable part of the some messages containing text from the offending line.

```
$ pydocstyle Bio/ BioSQL/ Tests/ Scripts/ Doc/ --ignore "" | grep "^        D" | cut -d "(" -f 1 | sort | uniq -c | sort -n
  11         D207: Docstring is under-indented
  12         D104: Missing docstring in public package
  18         D301: Use r""" if any backslashes in a docstring
  24         D214: Section is over-indented 
  33         D405: Section name should be properly capitalized 
  35         D211: No blank lines allowed before class docstring 
  40         D100: Missing docstring in public module
  43         D404: First word of the docstring should not be `This`
  63         D412: No blank lines allowed between a section header and its content 
  85         D103: Missing docstring in public function
 102         D212: Multi-line docstring summary should start at the first line
 104         D403: First word of the first line should be properly capitalized 
 108         D209: Multi-line docstring closing quotes should be on a separate line
 109         D210: No whitespaces allowed surrounding docstring text
 114         D413: Missing blank line after last section 
 118         D101: Missing docstring in public class
 123         D401: First line should be in imperative mood; try rephrasing 
 135         D208: Docstring is over-indented
 140         D406: Section name should end with a newline 
 142         D407: Missing dashed underline after section 
 187         D205: 1 blank line required between summary line and description 
[...]

PR #1176, opened 2017-04-20
Fixes for pydocstyle rule D412

morrme:
Ref. #1170


Bio/Application/__init__.py
```diff
@@ -539,7 +539,6 @@ class _Option(_AbstractParameter):
     take a value, use the _Switch object instead.
 
     Attributes:
-
```
peterjc: As an aside, all these old-style docstrings using lower-case o as a bullet point need updating at some point, e.g. RST bullet points would be simplest.

peterjc:
Many of the example you've changed need other work to adopt RST markup, so in itself this doesn't directly make the documentation any clearer - but I'm happy about bringing it slightly more in line with PEP257 anyway.

peterjc:
Cherry picked and merge conflict fixed, see 77df7d82516f9b75a6932f5e2b0eb920bb0ea18b

Thank you.

Bio/Application/__init__.py
peterjc: e.g. I did some in e8bbaa89b30bd6e55d4ed5a36ab62a17529e76e0

morrme:
@peterjc in accordance with the issue description, I was only focused on one or two rules at a time. I thought that would keep things organized, even if it meant that I'd have to revisit a file later on to make another change.

peterjc:
@morrme good point. I've added this to #1170:

*Update*: You may find some of the modules have loads of issues, and instead it may make sense to work on all the docstring style or reStructuredText problems in that module at once. That's fine too - but again please comment on this issue to avoid duplication of effort.

morrme:
@peterjc Duplication of work is definitely a risk in this approach, as well as an increased number of conflicts. The other challenge with working on multiple problems at once is that there may be some things that are intentional, as you've pointed out, so changing things for one or two related rules at a time allows the reviewer to be able to point these out before too much is overhauled.  I can understand it may be annoying as a reviewer to see one change but then three other things _right there_ that should be fixed, but hopefully once each rule is addressed those issues will all eventually be taken care of.

peterjc:
Eventually :)

Right now its just you and me doing these edits so it should be OK...

PR #1177, opened 2017-04-20
moved self-test from Bio/PDB/Polypeptide.py

Gasta88:
As refer in issue #820, I've moved the self-test from `Bio/PDB/Polypeptide.py` to `Tests/test_PDB.py`

### DIFF
--- a/Bio/PDB/Polypeptide.py
+++ b/Bio/PDB/Polypeptide.py
@@ -451,36 +451,3 @@ class PPBuilder(_PPBuilder):
             return 1
         else:
             return 0
-
-
-if __name__ == "__main__":
-    import sys
-    from Bio.PDB.PDBParser import PDBParser
-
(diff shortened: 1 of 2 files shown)


peterjc:
It looks like you've been caught out by changes to the master branch which added new tests for vectors and ``ResidueDepthTests`` - can you refresh this?

Gasta88:
After I perform the refresh of my master, does GitHub pick up the change automatically or do I have to manually commit something?

Gasta88:
closing PR since it's not synced with Biopython:master.

peterjc:
If you want to continue using the same pull request, you would need to update the same branch. Typically this means doing something like ``git push Gasta88 Gasta88-patch-6 --force``.

However, as using ``--force`` is generally a bad idea as it can delete changes, until you are comfortable with git making a new branch and pull request is safer.

PR #1179, opened 2017-04-21
detele unused get_codon_alphabet method

zruan:
get_codon_alphabet is defined twice. Delete the one that is not used in the code.


peterjc:
Thank you

PR #1182, opened 2017-04-21
KEGG sample data update and small parser fix

kozo2:
I got these data with
```
wget http://rest.kegg.jp/get/C01454+C06876
wget http://rest.kegg.jp/get/C00023+C00017+C00099+C00294+C00298+C00348+C00349+C01386
wget http://rest.kegg.jp/get/ec:6.2.1.25
```


peterjc:
Thank you @kozo2. Are you happy to dual licence this and your past contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

Will this still work on old KEGG files?

Are the updated example files direct from KEGG?

@mdehoon or @widdowquinn could you comment or review this?

kozo2:
@peterjc I am sorry my reply for #898 was late.
https://github.com/biopython/biopython/issues/898#issuecomment-296157924

This will not work on old KEGG COMPOUND files.
But I think that there is no need to support the old format so far.

peterjc:
Thank you @kozo2.

If it is easy to support the old and new formats that would be nice, but not essential. There might be some stubborn people with a copy of the final free KEGG release, but perhaps not anymore?

I'll leave this to @mdehoon or @widdowquinn to decide, as both have used KEGG more than me.

peterjc:
You will need to fix the test failure reported by TravisCI, e.g.

```
test_KEGG
  File "run_tests.py", line 364, in runTest
    outputfile, line_number))
ValueError: 
Output  : 'GENES       RPJ: N234_07060'
Expected: 'GENES       REU: Reut_A1327'
/home/travis/build/biopython/biopython/Tests/output/test_KEGG line 438
```

Use ``python run_test.py -g test_KEGG.py`` and then commit the file ``output/test_kegg``

kozo2:
@peterjc Thank you for the information. I committed `output/test_kegg` (and squashed the commit)  https://github.com/kozo2/biopython/commit/0d4bb3647ed1b8bfbcd78f8128b87c9e1b20e28d#diff-707ad643dcf2d39b73a8e06c0b2c8503

kozo2:
I know the `output/test_kegg`  lacks some field like `EXACT_MASS` or `BRITE`.
I will add code about such fields later.

Bio/KEGG/Compound/__init__.py
```diff
@@ -171,16 +171,9 @@ def parse(handle):
...
-                pathway = (path[:-1], map, name)
-                record.pathway.append(pathway)
-            else:
-                pathway = record.pathway[-1]
-                path, map, name = pathway
-                name = name + " " + data
-                pathway = path, map, name
-                record.pathway[-1] = pathway
+            map, name = data.split("  ")
+            pathway = (map, name)
```
Ghabry: Imo this is bad because it breaks the API. Before the tuple was ("PATH", map, name) now it is (map, name). Any code that relies on these indexing (probably alot) will break.

Better would be pathway = ("PATH", map, name)

joanarcx:
Hi. I am facing an issue that I think is related to this post. Here:

[https://www.biostars.org/p/256408/](url)

I am not a developer, just a user. I'd appreciate if you have any solution.

Regards

peterjc:
@kozo2 do you agree with the change proposed by @Ghabry?

Bio/KEGG/Compound/__init__.py
kozo2: I think it is not good to append "PATH" here.
This field contains only the pathway IDs contrary to the old entry format.

kozo2:
@peterjc I do not think it's good. I would like to hear comments from other people if possible.

peterjc:
I've looked at this and agree with @Ghabry - to preserve the current parser behaviour please keep ``.pathway`` as "A list of 3-tuples: (database, id, pathway)" by using ``PATH`` as the "database".

You could explain in the docstring if you wish that having the "database" here but always "PATH" is a historical artefact.

peterjc:
I wasn't sure that the returned information from the current parser behaviour was being explicitly tested, so I added some minimal testing in https://github.com/biopython/biopython/commit/2538eeb5553f690815f1306cf8cc93965d30640e

ctSkennerton:
I just ran into an issue which will be solved by this PR. 

While I think that it is not perfect to always have "PATH" it is necessary just in case someone wrote code that assumed that the pathway accession is always in the 2nd position of the tuple.

peterjc:
@kozo2 are you familiar with ``git rebase``?

If yes, please rebase this branch against the latest master.

If no, don't worry - I'll take care of this when I merge your work.

Thank you.

kozo2:
@peterjc Thanks for your comment. I think that I rebased well. https://github.com/kozo2/biopython/commits/kegg

kozo2:
@peterjc Shoud I squash the above commits?

peterjc:
If you are happy for me to squash-and-merge this into one commit, this is fine as it is.

I think the reason CodeCov is reporting reduced test coverage is that you've updated all the test files, so the remaining code for the old format never gets tests? Is it worth keeping (some) old test files, perhaps renamed? I will leave this up to you.

peterjc:
Merged, thank you.

PR #1185, opened 2017-04-21
Fixes for pydocstyle rule D403

morrme:
Ref. #1170 

Rule D403 states that 

>  First word of the first line should be properly capitalized 

And thus words like GenBank, HSPFragment, SeqIO, and so on cause errors. This may one to consider keeping as "ignored". 

I did not change any of those terms, but changed others in compliance with this rule.


peterjc:
Thanks. We'd need to look at the remaining cases one-by-one, but I suspect they can be reworded to avoid using module/class names as the first word - this could make the text clearer too.

PR #1188, opened 2017-04-24
Script for git pre-commit hook linting (flake8 and pydocstyle)

### LINKED ISSUE #493: Enforce basic PEP8 style with a pre-commit git hook

peterjc:
Would close #493, and is based on the script proposed by @cbrueffer - cross reference discussion on #883

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -21,6 +21,13 @@ Git Usage
 
 We have a git introduction online at http://biopython.org/wiki/GitUsage
 
+In order to help follow our coding conventions, please install our git
+pre-commit hook as follows::
+
+   $ git clone git@github.com:biopython/biopython.git
+   $ cd biopython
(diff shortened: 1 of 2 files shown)


peterjc:
Just updated to use the ``.flake8`` files (discarding an earlier update also using our ``.pydocstyle`` in the expectation of merging #1199) meaning the same configuration as our TravisCI/Tox style checks.

peterjc:
As of #2074 which moved to a single top level ``.flake8`` file, this bespoke script is no longer needed.

PR #1191, opened 2017-04-25
Add selftest for FragmentMapper

souravsingh:
Fixes a part of #820


Bio/PDB/FragmentMapper.py
```diff
@@ -322,15 +322,5 @@ def __getitem__(self, res):
...
-    p = PDBParser()
-    s = p.get_structure("X", sys.argv[1])
-    m = s[0]
-    fm = FragmentMapper(m, 10, 5, "levitt_data")
-
-    for r in Selection.unfold_entities(m, "R"):
-        print("%s:" % r)
-        if r in fm:
-            print(fm[r])
+    from Bio._utils import run_doctest
```
peterjc: You'd need to expand the existing example in order to make it a self-contained doctest, this alone will fail:

``` python
    >>> model = structure[0]
    >>> fm = FragmentMapper(model, lsize=10, flength=5, dir="fragment_data")
    >>> fragment = fm[residue]
```

Tests/test_PDB.py
```diff
@@ -1576,6 +1576,20 @@ def test_pdb_to_xyzr(self):
...
+    
+    def self_test(self):
+        """ Self test for FragmentMapper module."""
+        p = PDBParser()
+        pdb1= "PDB/1A8O.pdb"
+        s = p.get_structure("X", pdb1)
+        self.assertEqual(m, s[0])
+        fm = FragmentMapper(m, 10, 5, "levitt_data")
+        for r in Selection.unfold_entities(m, "R"):
+            print("%s:" % r)
```
peterjc: Don't use ``print`` in unit-tests, use something like:

```python
self.assertEqual(str(r), "put the expected string here")
```

Tests/test_PDB.py
```diff
@@ -1576,6 +1576,20 @@ def test_pdb_to_xyzr(self):
         assert len(msms_radii) == len(biopy_radii)
         self.assertSequenceEqual(msms_radii, biopy_radii)
 
+class FragmentMapperTests(unittest.TestCase):
+    """ Tests for FragmentMapper module.""""
```
peterjc: This is invalid syntax - there should be three closing double-quote characters.

Also please remove the space before the word 'Tests'.

peterjc:
About the doctest: You put this at the end of file  ``Bio/PDB/FragmentMapper.py``,

```python
if __name__ == "__main__":
    run_doctest(verbose=0)
```

If you tried running the file, you should have seen the current docstring's doctest failed. It is not being complete enough to work as a doctest, https://github.com/biopython/biopython/blob/biopython-169/Bio/PDB/FragmentMapper.py#L28

It would be nice to expand that into a self-contained working doctest but I don't think we have the directory of files ``fragment_data/`` which Thomas referred to.

The simple answer is just remove the entire ``__main__`` if statement from ``Bio/PDB/FragmentMapper.py``.

I don't understand your last commit, adding incomplete example code into the ``__main__`` if statement.

Tests/test_PDB.py
```diff
@@ -1576,6 +1576,20 @@ def test_pdb_to_xyzr(self):
...
+        """Self test for FragmentMapper module."""
+        p = PDBParser()
+        pdb1= "PDB/1A8O.pdb"
+        s = p.get_structure("X", pdb1)
+        self.assertEqual(m, s[0])
+        fm = FragmentMapper(m, 10, 5, "levitt_data")
+        for r in Selection.unfold_entities(m, "R"):
+            self.assertEqual(str(r), "R")
+            if r in fm:
+                print(fm[r])
```
peterjc: You still have a print statement here. Again, can you use something like ``self.assertEqual(fm[r], expected_value)`` which might be tricky given this is in a loop.

souravsingh: Give me a minute, I will have to run the script to check the expected value for the variable.

souravsingh: I am getting the error-
```
File "/home/sourav/miniconda2/lib/python2.7/site-packages/Bio/PDB/FragmentMapper.py", line 67, in _read_fragments
    with open(filename, "r") as fp:
IOError: [Errno 2] No such file or directory: 'levitt_data/lib_10_z_5.txt'
```

peterjc: As explained in the module docstring, these are available from http://csb.stanford.edu/~rachel/fragments/

If the code needs these files all present, we're not going to be able to do this via TravisCI. Rather we'd make the test conditional on the files being pre-downloaded (much like we skip tests if a binary tool is missing).

If the test case only needs on of these files, it may OK to download it as part of the test...

JoaoRodrigues:
Funny coincidence. That machine is down at the moment .. It's my current lab and I've shut it down last week to do some maintenance/upgrades that are taking way longer than I expected. The author (Rachel Kolodny) will visit soon and I can ask her if I can move these fragments someplace else more robust.

peterjc:
Thanks @JoaoRodrigues - strangely I can still see the machine and the text file mentioned. Maybe its still cached. Since you'll be seeing Rachel Kolodny anyway, you could ask if the licensing could be clearer, e.g. for redistribution or modification. I would suggest CC-BY. Currently the files include this in their header:

```

*******************************************************************************
* REFERENCE
*******************************************************************************
*
* If you use this data in research, please cite:
*
* Rachel Kolodny, Patrice Koehl, Leonidas Guibas, Michael Levitt
* Small Libraries of Protein Fragments Accurately Model Native Protein 
* Structures
*******************************************************************************
```

Tests/test_PDB.py
souravsingh: Here is the output I have obtained from the last print statement-
```shell
<Fragment length=5 id=8>
<Fragment length=5 id=1>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=4>
<Fragment length=5 id=7>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=0>
<Fragment length=5 id=6>
<Fragment length=5 id=5>
<Fragment length=5 id=9>
<Fragment length=5 id=8>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=4>
<Fragment length=5 id=7>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=6>
<Fragment length=5 id=0>
<Fragment length=5 id=7>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=4>
<Fragment length=5 id=8>
<Fragment length=5 id=1>
<Fragment length=5 id=2>
<Fragment length=5 id=4>
``

Comment (peterjc) [inline Tests/test_PDB.py]: Then how about:

```python
self.assertTrue(str(fm[r]).startswith("<Fragment length=5 id="))
```

Tests/test_PDB.py
```diff
@@ -1577,6 +1577,22 @@ def test_pdb_to_xyzr(self):
...
 
+class FragmentMapperTests(unittest.TestCase):
+    """Tests for FragmentMapper module."""
+    
+    def self_test(self):
+        """Self test for FragmentMapper module."""
+        p = PDBParser()
+        pdb1 = "PDB/1A8O.pdb"
+        s = p.get_structure("X", pdb1)
+        self.assertEqual(m, s[0])
```
peterjc: You can't compare ``m`` to ``s[0]`` if ``m`` has never been defined.

The original code was ``m = s[0]`` which makes more sense here - defining ``m`` as the first model in the structure (index zero).

peterjc:
In case this helps you: The key line in the original code which we want to test is ``fm = FragmentMapper(m, 10, 5, "levitt_data")``, everything up to then is setting up variables to run the fragment mapping. Everything after that is basically a quick check that it did something sensible - in the original self test this was printed on screen for the user to see, but here we want to do some basic asserts about the data from the mapping.

Tests/test_PDB.py
souravsingh: So we won't need this line, correct?

peterjc: You need ``m = s[0]`` instead of ``self.assertEqual(m, s[0])``

Tests/test_PDB.py
```diff
@@ -1577,6 +1577,22 @@ def test_pdb_to_xyzr(self):
         self.assertSequenceEqual(msms_radii, biopy_radii)
 
 
+class FragmentMapperTests(unittest.TestCase):
+    """Tests for FragmentMapper module."""
+    
+    def self_test(self):
```
peterjc: I was puzzled why the TravisCI tests seemed to be passing given we know the fragment files are not going to be present. One of the things with the ``unittest`` framework is by default you have to name the test methods ``test_...`` for them to be called.

Your method is called ``self_test`` and gets ignored :(

There is no reason (other than the history of where this example came from) to call this a self test. How about something like ``test_fragment_mapper``?

souravsingh: Done

souravsingh:
The test requires only one of the file from the Stanford's server. Would it make sense to host it in the biopython's repository?

peterjc:
Regarding bundling any of the fragment data, let's wait and see what @JoaoRodrigues finds out when he talks to Rachel Kolodny about http://csb.stanford.edu/~rachel/fragments/

There is a copyright question here, and with the current licensing unclear, would it be fair use to bundle one of the files, or should we ask permission? Given we have a personal contact it is simpler to ask first.

JoaoRodrigues:
Hi all,

Rachel Kolodny gives her permission for us to store (any of) the files in Biopython under an open-source license. We should retain the headers as they are though, with the reference and authorship.

peterjc:
Thanks @JoaoRodrigues - that's great news.

If the total file size is negligible, we *could* bundle all of it with Biopython (like we bundle DTD files for parsing NCBI XML files - which has implications for the ``setup.py`` script but we know how to do this)?

However, I think the simpler option of including one fragment file for a unit test is probably more sensible.

What do you think?

JoaoRodrigues:
I'd say bundling the files isn't such a bad idea. They are quite small IIRC (largest is 500KB). On the other hand, how many people do use them .. ?

peterjc:
OK, 500KB for one file is more than I was hoping for - together they might make a sizeable increase to the Biopython tar-ball/zip file .

Let's just bundle one for testing?

peterjc:
Also following #1217, I would prefer putting the new test into its own file, say ``Tests/test_PDB_FragmentMapper.py``, as ``Tests/test_PDB.py`` is getting far too big now.

Note the new test file can be dual licensed from the start (use the same style header as in ``Tests/test_PDB_MMCIF2Dict.py`` for example).

souravsingh:
@peterjc I have made a separate test for FragmentMapper.

peterjc:
Its failing on the style checks,

```
$ flake8 Tests/
Tests/test_PDB_FragmentMapper.py:21:1: E302 expected 2 blank lines, found 1
Tests/test_PDB_FragmentMapper.py:23:1: W293 blank line contains whitespace
Tests/test_PDB_FragmentMapper.py:35:1: W293 blank line contains whitespace
Tests/test_PDB_FragmentMapper.py:36:1: E305 expected 2 blank lines after class or function definition, found 1
Tests/test_PDB_FragmentMapper.py:38:37: W291 trailing whitespace
```

Those are all easy to fix - make sure you have the latest flake8 installed locally as that will be quicker than using TravisCI to check this.

Also, you'll need to add the fragment file to the repository as the tests are failing without it, e.g.

```
ERROR: test_fragment_mapper (test_PDB_FragmentMapper.FragmentMapperTests)
Self test for FragmentMapper module.
  File "/home/travis/build/biopython/biopython/Tests/test_PDB_FragmentMapper.py", line 30, in test_fragment_mapper
    fm = FragmentMapper(m, 10, 5, "levitt_data")
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/FragmentMapper.py", line 259, in __init__
    self.reflist = _read_fragments(lsize, flength, fdir)
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/FragmentMapper.py", line 68, in _read_fragments
    with open(filename, "r") as fp:
IOError: [Errno 2] No such file or directory: 'levitt_data/lib_10_z_5.txt'
```

I'd suggest using folder name ``Tests/PBB/lib_10_z_5.txt`` rather than ``Tests/levitt_data/lib_10_z_5.txt`` as you have now (so change line 30 of the test as well).

souravsingh:
I am not able to access the link to the Fragments data.

peterjc:
Oh dear - right now it seems I can't reach http://csb.stanford.edu/~rachel/fragments/ from home either. I wonder if I have the cached copy of this file on my work computer still...

@JoaoRodrigues was there a known issue with the website?

JoaoRodrigues:
Yes, the machine is actually turned off for a couple of weeks now. I was
*very* surprised it worked :)​ I'll fetch the files tomorrow and put them
on a github repo.

Cheers,

João

peterjc:
I don't seem to have taken a copy of ``lib_10_z_5.txt``, and for anyone wanting to use this code, they'd need the full set. Were you able to get a copy @JoaoRodrigues?

souravsingh:
@peterjc Would it be possible to merge the PR before 1.70?

souravsingh:
@peterjc I have found the file in my machine. Will be making a commit in a few minutes.

souravsingh:
@peterjc I have added the text file, but the self-test is facing an issue.
```python
self.assertEqual(str(r), "<Residue MSE het=H_MSE resseq=151 icode= >")

AssertionError: '<Residue ASP het=  resseq=152 icode= >' != '<Residue MSE het=H_MSE resseq=151 icode= >'

- <Residue ASP het=  resseq=152 icode= >
```

peterjc:
The problem is your test:

```python
        for r in Selection.unfold_entities(m, "R"):
            self.assertEqual(str(r), "<Residue MSE het=H_MSE resseq=151 icode= >")
            if r in fm:
                 self.assertTrue(str(fm[r]).startswith("<Fragment length=5 id="))
```

This loop checks all the residues ``r`` are MSE sequence number 151. The first entry is, but clearly the second entry is ASP sequence number 152, and I would expect the rest to be different as well.

That check is  not directly related to the fragment mapper, so the simple answer is to remove it leaving:

```python
        for r in Selection.unfold_entities(m, "R"):
            if r in fm:
                 self.assertTrue(str(fm[r]).startswith("<Fragment length=5 id="))
```

souravsingh:
@peterjc The PR is now ready for merging

peterjc:
Great - thanks for your patience @souravsingh, and a big thank you to @JoaoRodrigues and Rachel Kolodny regarding using the sample data file.

JoaoRodrigues:
FYI: https://github.com/csblab/fragments

We uploaded the entire fragment library to that repository. Took a while because the server where this was hosted threw quite a fit ..

Thanks @souravsingh and @peterjc for taking care of this.

peterjc:
@JoaoRodrigues could you update the links to this in Biopython please?

JoaoRodrigues:
Done.​

PR #1194, opened 2017-04-27
Fix DSSP error when MMCIF file contains non-standard residues

andrewguy:
When running DSSP on the newer PDBx/mmCIF file format, the following error is observed when the structure contains non-standard residues (e.g. 1AS5.cif structure):

`Bio.PDB.PDBExceptions.PDBException: Structure/DSSP mismatch at <Residue HYP het=  resseq=2 icode= >`

This is due to DSSP labelling any non-standard residues as `X`, whereas line 481 in `DSSP.py` checks against a much larger dictionary of possible residue 3-letter codes (`SCOPData.protein_letters_3_to_1`). For example, the residue `HYP` gets interpreted as `P`, whereas DSSP returns this as `X`.

Using the old PDB file format, these sorts of residues would be labelled as HETATOM, and hence get caught earlier in the `DSSP.py` code, but this isn't the case with mmCIF files.

I'm proposing a simple change where anything that isn't a standard residue gets interpreted as `X`. I've added the relevant unit tests to test DSSP generation on mmCIF files that contain non-standard residues.


JoaoRodrigues:
We should do something about the duplication of these 'amino acids name lists' in the code. I would say the `is_aa` function in `Polypeptide.py` is the better solution, because it supports a `standard` option that checks only against the 20 canonical amino acids. 

@andrewguy would you mind changing from IUPACData to this Polypeptide function and use a try/except (it will raise KeyError on non-standard aas)?

JoaoRodrigues:
Thank you!

andrewguy:
@JoaoRodrigues That makes a lot more sense to do it that way. I've used the `three_to_one` function in `Polypeptide.py`, which also checks against the standard aa.

peterjc:
Thank you both.

@JoaoRodrigues the review thing is fairly new in GitHub, I don't yet really understand it myself.

PR #1196, opened 2017-04-28
PDB HETATM waters should be W not H_HOH (using mmCIF parser)

andrewguy:
The mmCIF parser for PDBx files behaves differently to the PDB parser in the treatment of water molecules. In particular, the mmCIF parser labels waters as 'H_HOH', rather than 'W'. The proposed fix borrows from the code in `PDBParser.py`.


JoaoRodrigues:
Makes sense, good catch 👍

peterjc:
Cross reference #1197

JoaoRodrigues:
@peterjc I did already, it's good IMO.

peterjc:
Thanks both of you, applied. BTW, if you think any of these recent changes deserves direct mention in the ``NEWS.rst`` file, please say so - or include it in a pull request.

PR #1202, opened 2017-05-01
Use flake8-docstrings to run pydocstyle via flake8

### LINKED ISSUE #1199: Adopt flake8-docstrings to run pydocstyle via flake8
Once it has been updated for pydocstyle v2.0.0, I propose we use ``flake8-docstrings`` to run the docstring (PEP257) checks via ``flake8`` instead of calling ``pydocstyle``.

Cross reference https://gitlab.com/pycqa/flake8-docstrings/issues/19

i.e. For both the TravisCI/Tox configuration and the proposed git pre-commit hook.

The advantage of this is we can get ride of our ``.pydocstyle`` configuration file and move those settings into ``.flake8`` and then just need to run a single tool for the style checks.

Cross reference #1170 and #493 and #883

peterjc:
This would close #1199

It replaces the central ``.pydocstyle`` configuration file for ``pydocstyle`` with top-level folder specific ``.flake8`` entries used with ``flake8-docstrings`` via ``flake8`` instead.

Our long term goal should be to unify on a single base-level ``.flake8`` file for the entire project, but right now this setup allows e.g. ``BioSQL/`` and ``Scripts/`` to be more strictly checked until ``Bio/`` and ``Tests/`` catch up.

### DIFF
--- a/Scripts/Restriction/ranacompiler.py
+++ b/Scripts/Restriction/ranacompiler.py
@@ -104,7 +104,7 @@ def regex(site):
 
 
 def is_palindrom(sequence):
-    """Is the sequence (Seq objecT) as a palindrom."""
+    """Check whether the sequence is a palindrome or not."""
     return str(sequence) == str(sequence.reverse_complement())
 
 
--- a/Scripts/Restriction/ranacompiler.py
+++ b/Scripts/Restriction/ranacompiler.py
@@ -104,6 +104,19 @@ def regex(site):
 
 
 def is_palindrom(sequence):
+    """Check whether the sequence is a palindrome or not (DEPRECATED).
+
+    Deprecated alias for is_palindrome (with e at end).
+    """
+    import warnings
+    from Bio import BiopythonDeprecationWarning
+    warnings.warn("is_palindrom is deprecated, please use "
+                  "is_palindrome instead.",
+                  BiopythonDeprecationWarning)
+    return is_palindrome(sequence)
+
+
+def is_palindrome(sequence):
     """Check whether the sequence is a palindrome or not."""
     return str(sequence) == str(sequence.reverse_complement())
 
--- a/Scripts/Restriction/ranacompiler.py
+++ b/Scripts/Restriction/ranacompiler.py
@@ -78,15 +78,17 @@ typedict = {}
 
 class OverhangError(ValueError):
     """Exception for dealing with overhang."""
+
     pass
 
 
 def regex(site):
-    """regex(site) -> string.
+    """Construct a regular expression (string) from a DNA sequence.
 
-    Construct a regular expression from a DNA sequence.
-    i.e.:
+    i.e.::
+
         site = 'ABCGN'   -> 'A[CGT]CG.'
+
     """
     reg_ex = str(site)
     for base in reg_ex:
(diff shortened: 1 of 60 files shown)


peterjc:
Inviting comments, esp from @bow, @MarkusPiotrowski and @morrme

peterjc:
Also CC @cbrueffer as he's worked on the git-commit hook side of things.

peterjc:
There is currently a major catch to watch out for - right now ``flake8`` and/or ``flake8-docstrings`` breaks if there is a trailing comma on the ignore list:

https://gitlab.com/pycqa/flake8/issues/330

peterjc:
Locally I've been using flake8 and pycodestyle installed under Python 3, which means certain differences in the output (like ignoring D302). Right now TravisCI is only running the style checks under Python 2, we'll do both later.

Scripts/Restriction/ranacompiler.py
```diff
@@ -102,20 +104,12 @@ def regex(site):
 
 
 def is_palindrom(sequence):
-    """is_palindrom(sequence) -> bool.
-
-    True is the sequence is a palindrom.
-    sequence is a Seq object.
-    """
+    """Is the sequence (Seq objecT) as a palindrom."""
```
bow: I feel this is better phrased as "Whether the sequence is a palindrome or not." (also there is a typo in the end currently)

Tests/test_MafIO_index.py
```diff
@@ -382,12 +382,12 @@ def test_no_alignment(self):
             self.assertEqual(str(result[0].seq), "N" * 1000)
 
         def test_correct_retrieval_1(self):
-            """
+            """Correct retrievel of Cnksr3 in mouse.
```
bow: This should be "Correct retrieval ..."

Scripts/Restriction/ranacompiler.py
MarkusPiotrowski: In imperative mood: "Check wether the sequence is a palindrome or not."

Tests/test_MafIO_index.py
peterjc: Thank you, fixing.

Scripts/Restriction/ranacompiler.py
peterjc: Thank you. Should we change the method name too (``is_palindrome`` with an e)?

bow: That would be good. Something like:

```
def is_palindrom(...):
    # ... deprecation notice here? I'm not sure if this is overkill since this is a small function ... 
    return is_palindrome(...)

def is_palindrom(...):
    # actual function
```

peterjc: Done, thank you.

peterjc:
Frustratingly I didn't have a pre-commit hook installed on this machine...

peterjc:
I have updated my git pre-commit hook on #1188 to assume just ``.flake8`` configuration.

peterjc:
Thank you @bow & @MarkusPiotrowski for looking at this.

I've just rebased with a slight history clean up; if this passes TravisCI I will merge it...

PR #1204, opened 2017-05-01
Update HSExposure.py

Gasta88:
As mentioned in issue #820, I've removed the self-test from `Bio/PDB/HSExposure.py`.

The self-test was already implemented in `test_PDB.py` from line 848.


peterjc:
Thank you.

PR #1205, opened 2017-05-02
Update MMCIF2dict.py

Gasta88:
As referenced in issue #820, I've moved the self-test from `Bio/PDB/MMCIF2dict.py` into `Tests/test_PDB.py`.


peterjc:
Thank you.

Issue #1203, opened 2017-05-01
Adding missing public docstrings


peterjc:
Using the tool ``pydocstyle`` (installed with ``pip install pydocstyle``), or its flake8 plugin equivalent (installed with ``pip install flake8-docstring``), we can count missing docstrings:

```
$ pydocstyle Bio/ BioSQL/ --select D1 | grep "^        D" | cut -d "(" -f 1 | sort | uniq -c | sort -n
  12         D104: Missing docstring in public package
  21         D100: Missing docstring in public module
  60         D103: Missing docstring in public function
  96         D101: Missing docstring in public class
 271         D105: Missing docstring in magic method
 607         D102: Missing docstring in public method
```

or more simply using flake8:

```
$ flake8 Bio BioSQL --isolated --select D1 --statistics --quiet --quiet
21    D100 Missing docstring in public module
96    D101 Missing docstring in public class
607   D102 Missing docstring in public method
59    D103 Missing docstring in public function
12    D104 Missing docstring in public package
271   D105 Missing docstring in magic method
```

Ideally all the public facing code in Biopython would have docstrings, so it would be great to (gradually) add these where they are missing.

We can list the missed docstrings per file using a command like this:

```
$ flake8 Bio BioSQL --isolated --select D1 | cut -d ":" -f 1 | uniq -c
  20 Bio/bgzf.py
  28 Bio/DocSQL.py
  13 Bio/File.py
   5 Bio/MarkovModel.py
   1 Bio/MaxEntropy.py
   1 Bio/NaiveBayes.py
  12 Bio/pairwise2.py
   7 Bio/ParserSupport.py
  14 Bio/Seq.py
  17 Bio/SeqFeature.py
   6 Bio/SeqRecord.py
   3 Bio/Affy/CelFile.py
   3 Bio/Align/AlignInfo.py
   1 Bio/Align/Applications/_ClustalOmega.py
   1 Bio/Align/Applications/_Clustalw.py
   1 Bio/Align/Applications/_Dialign.py
   1 Bio/Align/Applications/_Mafft.py
   1 Bio/Align/Applications/_MSAProbs.py
   1 Bio/Align/Applications/_Muscle.py
   1 Bio/Align/Applications/_Prank.py
   1 Bio/Align/Applications/_Probcons.py
   1 Bio/Align/Applications/_TCoffee.py
   1 Bio/AlignIO/ClustalIO.py
   3 Bio/AlignIO/EmbossIO.py
[...]

PR #1209, opened 2017-05-04
Adding public docstring to Bio.bgzf

Gasta88:
In reference to issue #1203 , I've added public docstring to the module.

This PR is incomplete because I don't understand what the method `seekable()` does at line 710 and 830.

Is this an implementation of the method `seek()`?

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -404,7 +404,7 @@ def BgzfBlocks(handle):
 
 
 def _load_bgzf_block(handle, text_mode=False):
-    """Load the next BGZF function (INTERNAL PRIVATE)."""
+    """Load the next BGZF function (PRIVATE)."""
     magic = handle.read(4)
     if not magic:
         # End of file
@@ -658,7 +658,7 @@ class BgzfReader(object):
                 return data
 
     def readline(self):
-        """Read single line for the BGZF module."""
+        """Read a single line for the BGZF file."""
         i = self._buffer.find(self._newline, self._within_block_offset)
         # Three cases to consider,
         if i == -1:
@@ -685,7 +685,7 @@ class BgzfReader(object):
             return data
 
     def __next__(self):
-        """Return the next value."""
+        """Return the next line."""
         line = self.readline()
         if not line:
             raise StopIteration
@@ -697,7 +697,7 @@ class BgzfReader(object):
             return self.__next__()
 
     def __iter__(self):
-        """Iterate elements of the BGZF class."""
+        """Iterate over the lines in the BGZF file."""
         return self
 
     def close(self):
@@ -708,6 +708,7 @@ class BgzfReader(object):
         self._buffers = None
 
     def seekable(self):
+        """Return True indicating the BGZF supports random access."""
         return True
 
     def isatty(self):
@@ -748,7 +749,7 @@ class BgzfWriter(object):
         self.compresslevel = compresslevel
 
     def _write_block(self, block):
-        """Write the datablock into the open file."""
+        """Write the datablock into the open file (PRIVATE)."""
         # print("Saving %i bytes" % len(block))
         start_offset = self._handle.tell()
         assert len(block) <= 65536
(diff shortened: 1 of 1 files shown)


Bio/bgzf.py
```diff
@@ -404,7 +404,7 @@ def BgzfBlocks(handle):
 
 
 def _load_bgzf_block(handle, text_mode=False):
-    """Internal function to load the next BGZF function (PRIVATE)."""
+    """Load the next BGZF function (INTERNAL PRIVATE)."""
```
peterjc: We've a convention (which may not be documented anywhere?) using ``(PRIVATE)`` at the end of docstrings, and also ``(DEPRECATED)`` and ``(OBSOLETE)`` as per http://biopython.org/wiki/Deprecation_policy but I don't think we've ever used ``INTERNAL`` like this.

How about:

```python
"""Load the next BGZF function (PRIVATE)."""
```

Bio/bgzf.py
```diff
@@ -656,6 +658,7 @@ def read(self, size=-1):
                 return data
 
     def readline(self):
+        """Read single line for the BGZF module."""
```
peterjc: This isn't for the module, its for the class. How about just:

```python
"""Read a single line from the BGZF file."""
```

Bio/bgzf.py
```diff
@@ -682,6 +685,7 @@ def readline(self):
             return data
 
     def __next__(self):
+        """Return the next value."""
```
peterjc: Next *line* rather than value (this is the convention for how file iterators work).

Bio/bgzf.py
```diff
@@ -693,9 +697,11 @@ def next(self):
             return self.__next__()
 
     def __iter__(self):
+        """Iterate elements of the BGZF class."""
```
peterjc: Change to

```python
"""Iterate over the lines in the BGZF file."""
```

Bio/bgzf.py
```diff
@@ -736,6 +748,7 @@ def __init__(self, filename=None, mode="w", fileobj=None, compresslevel=6):
         self.compresslevel = compresslevel
 
     def _write_block(self, block):
+        """Write the datablock into the open file."""
```
peterjc: Include ``... (PRIVATE).`` at the end please.

peterjc:
Regarding ``.seekable()``, referring to the Python 3 docs https://docs.python.org/2/library/io.html

> Return True if the stream supports random access. If False, seek(), tell() and truncate() will raise IOError.

So how about:

```python
"""Return True indicating the BGZF supports random access."""
```

Bio/bgzf.py
peterjc: Apologies, I missed the word function was being misused here (my own mistake in the original text).

How about:

```python
"""Load the next BGZF block of compressed data (PRIVATE)."""
```

Bio/bgzf.py
```diff
@@ -736,6 +749,7 @@ def __init__(self, filename=None, mode="w", fileobj=None, compresslevel=6):
         self.compresslevel = compresslevel
 
     def _write_block(self, block):
+        """Write the datablock into the open file (PRIVATE)."""
```
peterjc: To be more explicit, I suggest something like:

```python
"""Write provided data to file as a single BGZF compressed block (PRIVATE)."""
```

peterjc:
I've suggested two more minor tweaks - thanks for looking at this file :)

Gasta88:
Thanks, I've applied the changes.

from now on would you prefer to have a separated PR for each edited file or gather few of them into on PR?

peterjc:
It look like my second comment on the first commit got lost.

For ``_load_bgzf_block`` my original text had "function"  rather than "block", so how about:

```python
"""Load the next BGZF block of compressed data (PRIVATE)."""
```

I would suggest one pull request per file or related set of files (e.g. if you updated several entries under ``Bio/SeqIO/`` then one pull request makes sense). This is in part to help with reviewing as I won't always be the module owner.

Gasta88:
Edit with the last comment.

peterjc:
Thank you 👍

PR #1211, opened 2017-05-04
parse PE line, debug sequence version and propagate to record, add te…

FredericBGA:
In order to get an Uniprot fasta header (http://www.uniprot.org/help/fasta-headers), you need to retrieve Protein Existence value and sequence version as well.

The PE tag was not yet parsed.
The value of version seemed to be broken (always retrieved as 0).

Using the patch, you will be able to create PE and SV tags

Tests have been updated according to these changes


peterjc:
This looks good :)

Also, are you happy to be thanked by name (Frederic Sapet?) in the ``NEWS.rst`` and ``CONTRIB.rst`` files?

FredericBGA:
ok :)

peterjc:
Thank you, this will be part of Biopython 1.70

PR #1212, opened 2017-05-04
Update File.py

Gasta88:
As mentioned in issue #1203 ,I've added public docstrings to the module `Bio.File`.

I have a question regarding this module.

Lines from 406 to 431 contain some methods that won't be implemented (they automatically raise a `NotImplementedError` exception).

Would it be good practice to remove them or leave them and document the methods either way?

### DIFF
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -177,7 +177,10 @@ class UndoHandle(object):
         return saved + self._handle.read(size)
 
     def saveline(self, line):
-        """Save the line of the File in memory."""
+        """Store a line in the cache memory for later use.
+
+        This acts to undo a readline, reflecting the name of the class: UndoHandle.
+        """
         if line:
             self._saved = [line] + self._saved
 
@@ -323,7 +326,7 @@ class _IndexedSeqFileDict(_dict_base):
         return key in self._offsets
 
     def __len__(self):
-        """Retun the number of records."""
+        """Return the number of records."""
         return len(self._offsets)
 
     def items(self):
@@ -387,7 +390,7 @@ class _IndexedSeqFileDict(_dict_base):
     def get(self, k, d=None):
         """Return the value in the dictionary.
 
-        If the key (k) is not found, this returns None unless a specified
+        If the key (k) is not found, this returns None unless a
         default (d) is specified.
         """
         try:
@@ -764,7 +767,7 @@ class _SQLiteManySeqFilesDict(_IndexedSeqFileDict):
     def get(self, k, d=None):
         """Return the value in the dictionary.
 
-        If the key (k) is not found, this returns None unless a specified
+        If the key (k) is not found, this returns None unless a
         default (d) is specified.
         """
         try:
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -7,7 +7,6 @@
 
 """Code for more fancy file handles.
 
-
 Classes:
 
     - UndoHandle     File object decorator with support for undo-like operations.
@@ -119,7 +118,9 @@ class UndoHandle(object):
         - peekline    Peek at the next line without consuming it.
 
     """
+
     def __init__(self, handle):
+        """Initialize the class."""
         self._handle = handle
         self._saved = []
         try:
@@ -129,9 +130,11 @@ class UndoHandle(object):
             pass
 
     def __iter__(self):
+        """Iterate over the lines in the File."""
         return self
 
     def __next__(self):
+        """Return the next line."""
         next = self.readline()
         if not next:
             raise StopIteration
@@ -143,11 +146,13 @@ class UndoHandle(object):
             return self.__next__()
 
     def readlines(self, *args, **keywds):
+        """Read lines for the File."""
         lines = self._saved + self._handle.readlines(*args, **keywds)
         self._saved = []
         return lines
 
     def readline(self, *args, **keywds):
+        """Read a single line for the File."""
         if self._saved:
             line = self._saved.pop(0)
         else:
@@ -155,6 +160,7 @@ class UndoHandle(object):
         return line
 
     def read(self, size=-1):
+        """Read the File."""
         if size == -1:
             saved = "".join(self._saved)
             self._saved[:] = []
@@ -171,10 +177,12 @@ class UndoHandle(object):
         return saved + self._handle.read(size)
 
     def saveline(self, line):
+        """Save the line of the File in memory."""
         if line:
             self._saved = [line] + self._saved
 
     def peekline(self):
+        """Look at the first line saved in memory."""
         if self._saved:
             line = self._saved[0]
         else:
@@ -183,19 +191,24 @@ class UndoHandle(object):
         return line
 
     def tell(self):
+        """Return the current position of the file read/write pointer within the File."""
         return self._handle.tell() - sum(len(line) for line in self._saved)
 
     def seek(self, *args):
+        """Set the current position at the offset specified."""
         self._saved = []
         self._handle.seek(*args)
 
     def __getattr__(self, attr):
+        """Return File attribute."""
         return getattr(self._handle, attr)
 
     def __enter__(self):
+        """Open File handle with WITH statement."""
         return self
 
     def __exit__(self, type, value, traceback):
+        """Close File handle with WITH statement."""
         self._handle.close()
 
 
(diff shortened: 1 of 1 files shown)


Bio/File.py
```diff
@@ -143,18 +146,21 @@ def next(self):
             return self.__next__()
 
     def readlines(self, *args, **keywds):
+        """Read lines for the File."""
```
peterjc: ``for`` --> ``from``

Also why ``File`` with a capital F?

How about:

```python
"""Read all the lines from the file as a list of strings."""
```

Bio/File.py
```diff
@@ -143,18 +146,21 @@ def next(self):
...
         lines = self._saved + self._handle.readlines(*args, **keywds)
         self._saved = []
         return lines
 
     def readline(self, *args, **keywds):
+        """Read a single line for the File."""
```
peterjc: Again, ``for`` --> ``from``, and why ``File`` with a capital F?

How about:

```python
"""Read the next line from the file as string."""
```

Bio/File.py
```diff
@@ -171,10 +177,12 @@ def read(self, size=-1):
         return saved + self._handle.read(size)
 
     def saveline(self, line):
+        """Save the line of the File in memory."""
```
peterjc: How about:

```python
"""Undo a read by caching a line to be read back next."""
```

Does that make sense to you reading this?

Bio/File.py
```diff
@@ -171,10 +177,12 @@ def read(self, size=-1):
...
         if line:
             self._saved = [line] + self._saved
 
     def peekline(self):
+        """Look at the first line saved in memory."""
```
peterjc: How about:

```python
"""Return the next line in the file, but do not move forward though the file."""
```

Again, does that make sense to you as someone who is not (yet) familiar with the ``UndoHandle``class?

Bio/File.py
```diff
@@ -183,19 +191,24 @@ def peekline(self):
...
+        """Set the current position at the offset specified."""
         self._saved = []
         self._handle.seek(*args)
 
     def __getattr__(self, attr):
+        """Return File attribute."""
         return getattr(self._handle, attr)
 
     def __enter__(self):
+        """Open File handle with WITH statement."""
```
peterjc: Repetitive use of the word 'with' here and in ``__exit__`` (same in ``bgzf.py`` but I didn't notice till now). How about:

```python
"""Special method called when opening the file using a with-statement."""
```

and

```python
"""Special method called when closing the file using a with-statement."""
```

Bio/File.py
```diff
@@ -292,20 +307,23 @@ def __init__(self, random_access_proxy, key_function,
...
         else:
             return "{}"
 
     def __contains__(self, key):
+        """Return key if contained in the offsets dictionary."""
         return key in self._offsets
 
     def __len__(self):
-        """How many records are there?"""
+        """How many records are there."""
```
peterjc: How about ``Return the number of records.`` here?

Bio/File.py
```diff
@@ -367,7 +385,11 @@ def __getitem__(self, key):
         return record
 
     def get(self, k, d=None):
-        """D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None."""
+        """Return the value in the dictionary.
+        
+        If key not found the second
+        attribute is returned. By default it is None.
+        """
```
peterjc: That's odd line breaking. We could explain the short hand k=key and d=default (which followed the Python built-in dictionary docstring). How about:

```
If the key (k) is not found, this returns None unless a specified
default (d) is specified.
```

Bio/File.py
```diff
@@ -661,7 +684,7 @@ def __contains__(self, key):
                               (key,)).fetchone())
 
     def __len__(self):
-        """How many records are there?"""
+        """Return length of the iterable."""
```
peterjc: I suggest ``Return the number of records indexed.`` here.

Bio/File.py
```diff
@@ -674,12 +697,16 @@ def __iter__(self):
...
+
+            This tries to act like a Python 3 dictionary, and does not return
+            a list of keys due to memory concerns.
+            """
             return [str(row[0]) for row in
                     self._con.execute("SELECT key FROM offset_data;").fetchall()]
 
     def __getitem__(self, key):
-        """x.__getitem__(y) <==> x[y]"""
+        """Return the item when key is specified."""
```
peterjc: Using 'when' seems odd as the key must be given.

I would suggest ``Return record for the specified key.``

Bio/File.py
```diff
@@ -707,7 +734,11 @@ def __getitem__(self, key):
         return record
 
     def get(self, k, d=None):
-        """D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None."""
+        """Return the value in the dictionary.
+        
+        If key not found the second
+        attribute is returned. By default it is None.
+        """
```
peterjc: See earlier comment about the other ``get`` method docstring.

peterjc:
This is interesting and educational for me to read as I wrote much of this file - for some of the other parts of Biopython I'm not going to be able to give such detailed and precise comments ;)

Regarding the code from lines 406 onwards, e.g.

```python
def __setitem__(self, key, value):
         """Would allow setting or replacing records, but not implemented."""
         raise NotImplementedError("An indexed a sequence file is read only.")
```

These are all methods which a Python dictionary would have, and we're trying to mimic a Python dictionary as a familiar object people will know how to use. The docstring looks OK to me, but could be made more explicit, e.g.:

```python
def __setitem__(self, key, value):
         """Would allow setting or replacing records, but not implemented.

         Python dictionaries provide this method for modifying data in the
        dictionary. This class mimics the dictionary interface but is read only.
         """
         raise NotImplementedError("An indexed a sequence file is read only.")
```

Bio/File.py
Gasta88: The last part is a bit intricate. Can I suggest:
```python
"""Store a line in the cache memory for later use.""" 
```

Bio/File.py
Gasta88: Yes, it does.

Bio/File.py
Gasta88: These are fine by me, 
I have some difficulty on choosing imperative mood phrases, but the more I document the more I understand how to use pydocstyle.

Bio/File.py
peterjc: OK. How about this then, trying to make this explicit as the ``saveline`` is the key functionality of the ``UndoHandle`` (along with ``peekline``):

```python
     def saveline(self, line):
        """Store a line in the cache memory for later use.

        This acts to undo a readline, reflecting the name of the class: UndoHandle.
        """
```

Bio/File.py
peterjc: Great, thanks.

Bio/File.py
peterjc: See discussion on #1170, not everyone agrees about strictly enforcing ``D401: First line should be in imperative mood; try rephrasing``, but since we're adding missing docstrings here we might as well try to do this.

Bio/File.py
```diff
@@ -323,7 +323,7 @@ def __contains__(self, key):
         return key in self._offsets
 
     def __len__(self):
-        """How many records are there."""
+        """Retun the number of records."""
```
peterjc: Typo (Return)

Bio/File.py
```diff
@@ -386,9 +386,9 @@ def __getitem__(self, key):
 
     def get(self, k, d=None):
         """Return the value in the dictionary.
-        
-        If key not found the second
-        attribute is returned. By default it is None.
+
+        If the key (k) is not found, this returns None unless a specified
+        default (d) is specified.
```
peterjc: Too many "specified" (might have been my mistake), remove the fist one?

peterjc: Yes, it was my error. Sorry.

Bio/File.py
```diff
@@ -735,9 +763,9 @@ def __getitem__(self, key):
...
```
peterjc: Again, this has my error about an extra "specified", sorry.

Gasta88:
Edited as requested.

peterjc:
CodeCov says the coverage dropped a fraction, but I'm pretty sure that is just noise. We have a couple of stochastic tests meaning coverage does vary run to run.

Merging - thank you!

Issue #1208, opened 2017-05-04
Enhancement: Fix for UPGMA issues


stuckmke:
I've been using the Biopython version 1.66.win32-py2.7.  I noticed two issues when constructing a UPGMA tree:

1. The tree is not ultrametric. This issue was also noted on stack overflow: [http://stackoverflow.com/questions/36461738/biopython-upgma-tree-construction-is-not-giving-ultrametric-tree](url)

2. WPGMA was being used to create the tree as opposed to UPGMA.

I have modified the "upgma" and "_height_of" functions in the TreeConstruction.py source code to fix these issues. Here is a patch:
[TreeConstructionPatch.zip](https://github.com/biopython/biopython/files/975110/TreeConstructionPatch.zip)

peterjc:
Are you able to submit your patch as a GitHub pull request? If you don't know git, as this is just one file you would even be able to do this via the GitHub website (fork the repository, click edit the file, and copy-paste in your new version).

CC @lijax and @etal

stuckmke:
I am new to GitHub, but I believe that I was able to create a pull request.

PR #1213, opened 2017-05-05
Update to UPGMA

stuckmke:
The tree is not ultrametric, and WPGMA was being used to create the tree
as opposed to UPGMA.

This modifies the "upgma" and "_height_of" functions in the
TreeConstruction.py source code to fix these issues.


peterjc:
This is to address #1208.

Note the tests are failing on TravisCI, e.g.

```
FAIL: test_upgma (test_TreeConstruction.DistanceTreeConstructorTest)
  File "/home/travis/build/biopython/biopython/Tests/test_TreeConstruction.py", line 162, in test_upgma
    self.assertTrue(Consensus._equal_topology(tree, ref_tree))
```

stuckmke:
Thanks for letting me know.  I looked more into the test script for the test it is failing (upgma) -- is there any more information on how the reference tree was created?  I see what looks to be the test tree in the docs for Phylo -- is that it?

I originally noticed this issue when I was generating a tree on my research data and Biopython was returning a tree different than MEGA7 was.  I manually calculated the tree and, along with looking at the Biopython source code, found that Biopython was doing _weighted_ pair group method with arithmetic mean, rather than unweighted.  It also was returning a tree that was not ultrametric (which is a necessary assumption for upgma, [source](https://en.wikipedia.org/wiki/UPGMA#Algorithm).  So, since it was performing wpgma instead of upgma, I wonder the reference tree may _also_ be calculated with wpgma (hence a successful Travis CI test in the past) and thus changing the algorithm (i.e. switching methods) would return a different tree and fail the matching test.

I have a three-slide PowerPoint presentation that I presented to my university PI showing the difference in algorithms/results between the two programs that I can provide if you'd like.

peterjc:
@etal? I am out of my depth here with the phylogenetic details.

etal:
Thanks for the contribution, @stuckmke. I haven't tried this code, but the description here makes sense to me. CC @lijax, the original author.

I'm (idly) curious why the new unweighted pair group method involves more code than the original weighted method. Is most of this patch related to enforcing an ultrametric tree, or were there other issues with the original code?

etal:
More to the point, is the original WPGA implementation generally useful enough to be worth keeping as a separate option / method, alongside the new UPGMA?

stuckmke:
The code for the updated UPGMA is longer for two reasons:
1. Enforcing an ultrametric tree (updated lines 730-734), and
2. Adding a second "counting" distance matrix, necessary for the unweighted method (updated lines 581-585, 630-636).

To enforce an ultrametric tree, I modified the `_height_of` function.  The previous code was not actually finding the full path distance from the inner clade to the terminal clade for deep inner clades, thus it was creating branches with inaccurate branch lengths.  In other words, when adding a new branch where the root node would "move back", or up a level, the distance between the previous root node and the new root node would be calculated by the new branch's length minus the singular length of the previous root node's longest terminal branch, rather than the correct calculation of the new branch's length minus the longest path distance to a terminal.  The code as it was written presupposed that the previous root node was _directly_ connected to a terminal branch (and, as such, assumed that the length of the longest terminal branch was equal to the path distance to a terminal -- in non-complex trees with few nodes, such as the test tree, this assumption holds, but breaks down as trees increase in complexity).  However, if the previous root node was only connected to other nodes, and needed to account for multiple branch lengths to find the full path length to a terminal, it could not account for that.  The new code finds the path from the previous root node all the way to the terminal regardless of the number of intermediate nodes, and sums the distances along the path, creating new branches that are ultrametric to the existing branches.  This function would need to be updated for _both_ upgma and a new, separate wpgma function, which I agree is worthwhile to create and keep around.

To calculate an unweighted average, it was necessary to add a second distance matrix (`dm_count`).  In addition to the existing matrix that keeps track of the distances, the second, for counting purposes, keeps track of how many elements contribute to the distances in the distance matrix.  This is necessary for the denominator in the upgma calculation (i.e. it's not always divided by 2 as it is with wpgma).  Each time a new branch is added to the tree, both of the matrices are updated.  The `dm_count` matrix is only necessary for calculating the unweighted average and would not be needed in a separate wpgma function.

As far as I can tell, the existing test tree (upgma.tre) would continue to pass a wpgma Travis CI treeconstruction.py test even after the `_height_of` modifications, but it should probably be more complex (i.e. several nodes deep along all paths) to fully test the `_height_of` function.   I believe it's the second change, which shifts the calculation from weighted to unweighted, that causes the failure, as the returned values from the upgma calculation will be different from the wpgma calculation.  In essence, upgma.tre can probably be renamed wpgma.tre and used to test that function, should you decide to keep the existing code as a separate wpgma function, and new test data for upgma will be needed.

lijax:
I agree with @stuckmke. The original method was not well considered. It's better to have two separate methods and also new test data.

RaumZeit:
Are there any reasons why the proposed changes have not been merged yet?
I just encountered this issue yesterday when I compared my own implementation with the results obtained by Bio.Phylo. From what I can tell after looking at the changes in this PR they indeed fix the wrong implementation of the UPGMA algorithm. However, at least one new test tree should be added to let the CI pass.

etal:
The proposed changes are good to merge. Updating the tests at the same time would be ideal, and if someone is willing to split off the original implementation into a separate function (possibly improved), that would be good too.

Bio/Phylo/TreeConstruction.py
```diff
@@ -1,1076 +1,1145 @@
...
+                clade2.branch_length = min_dist * \
+                    1.0 / 2 - self._height_of(clade2)
+
+            # update node list
+            clades[min_j] = inner_clade
+            del clades[min_i]
+
+            # rebuild distance matrix,
+            # set the distances of new node at the index of min_j
+            for k in range(0, len
```
stuckmke: Lines 646-711 the "old" upgma (which was actually, as in this version, wpgma)

Bio/Phylo/TreeConstruction.py
```diff
@@ -1,1076 +1,1145 @@
...
+                    Clade(branch_length=0.0853101915988, name='Epsilon')
+                    Clade(branch_length=0.136912030623, name='Delta')
+                Clade(branch_length=0.292306275042, name='Alpha')
+            Clade(branch_length=0.0747705106139, name='Beta')
+            Clade(branch_length=0.175229489386, name='Gamma')
+
+
+    """
+
+    methods = ['nj', 'upgma', 'wpgma']
```
stuckmke: Added wpgma as method

Bio/Phylo/TreeConstruction.py
```diff
@@ -1,1076 +1,1145 @@
...
+            raise TypeError("Bad method: " + method +
+                            ". Available methods: " + ", ".join(self.methods))
+
+    def build_tree(self, msa):
+        if self.distance_calculator:
+            dm = self.distance_calculator.get_distance(msa)
+            tree = None
+            if self.method == 'upgma':
+                tree = self.upgma(dm)
+            elif self.method == 'wpgma':
```
stuckmke: Added wpgma as callable method

stuckmke:
Commit 8ae37d7 makes no changes to my updated upgma method, but takes the previous upgma code (that is actually wpgma) and makes it a new callable wpgma method.

Commit 8a29bca renames the old upgma test tree to wpgma, and adds a new corrected upgma tree.

I believe the test_TreeConstruction.py file must also be updated to account for the new wpgma method, but I don't have experience writing Travis test scripts, so I haven't taken this step.  If someone wants to re-write it to incorporate wpgma, please feel free to do so.

peterjc:
There are now merge conflicts in ``Bio/Phylo/TreeConstruction.py``, are you familiar with ``git rebase`` which would probably be the cleanest way to sort this out?

stuckmke:
I'm new with git and not sure how rebase is working (something went haywire above) so I'm going to close this pull request out and try again from scratch.  I'll reference this PR when I open a new one.

peterjc:
Something might have happened with the line endings on the old pull request as GitHub showed all of ``Bio/Phylo/TreeConstruction.py`` as changed. Anyway, the new pull request #1506 looks better, thanks.

PR #1214, opened 2017-05-05
Moved self-test from Bio/PDB/parse_pdb_header.py

Gasta88:
As mentioned in issue #820 , I removed the self test in `Bio/PDB/parse_pdb_header.py`.

The testing is already present in `Tests/test_PDB.py` at line 123.


peterjc:
Thank you

Issue #253, opened 2013-11-02
ExPASy.sprot_search_ful and ExPASy.sprot_search_de do not work


miniocz:
Both will return just pages stating "Please update your links/bookmarks" and eventually correct link at the end of page.
The functions use http://www.expasy.ch/cgi-bin/sprot-search-ful or http://www.expasy.ch/cgi-bin/sprot-search-de eventually, which pages do not exist anymore.

peterjc:
I was reminded of this via #1826 and the tutorial having the (defunct) URLs http://www.expasy.org/cgi-bin/sprot-search-de and http://www.expasy.org/cgi-bin/sprot-search-ful

@bow Should we deprecate these two functions now? What state is the UniProt API wrapper?

bow:
@peterjc Deprecating the functions seems sensible given that the upstream services are not there anymore. To be honest, I also have not touched the API for a very long time.

PR #1215, opened 2017-05-05
ExPASy updates

peterjc:
Who'd like to review this? It has test but they are online only right now.

I suppose adding another prosite file would be good for offline testing of the date line fix.


peterjc:
This does not solve #253, unless someone knows what the old search URL did perhaps we can deprecate or just remove those two functions?

peterjc:
This also closes #1130 (DT line change in Prosite) and now incorporates the documentation changes from @francoiscoste's fix on #1132.

peterjc:
The coverage reports are very slightly worse, because it does not count the new online only tests :(

PR #1218, opened 2017-05-07
Docstring fixes for D200

morrme:
Ref. #1170


peterjc:
Great, thank you - I do hope you automated that, as it would have been a lot of work by hand!

morrme:
@peterjc No, all done by hand (same for D204). I was worried that I'd botch something by doing it any other way.

PR #1227, opened 2017-05-11
Update Residue.py

rodogi:
get_atoms method seems more in touch with the methods used for chain, model and structure.

### DIFF
--- a/Bio/PDB/Residue.py
+++ b/Bio/PDB/Residue.py
@@ -6,6 +6,8 @@
 """Residue class, used by Structure objects."""
 
 # My Stuff
+import warnings
+from Bio import BiopythonDeprecationWarning
 from Bio.PDB.PDBExceptions import PDBConstructionException
 from Bio.PDB.Entity import Entity, DisorderedEntityWrapper
 
@@ -109,6 +111,12 @@ class Residue(Entity):
         return self.segid
 
     def get_atoms(self):
+        for a in self:
+            yield a
+            
+    def get_atom(self):
+        warnings.warn("`get_atom` has been depcreated and we intend to remove it"
+                      " in a future release of Biopython. Please use `get_atoms` instead.")
         for a in self:
             yield a
 
--- a/Bio/PDB/Residue.py
+++ b/Bio/PDB/Residue.py
@@ -108,7 +108,7 @@ class Residue(Entity):
     def get_segid(self):
         return self.segid
 
-    def get_atom(self):
+    def get_atoms(self):
         for a in self:
             yield a
 
--- a/Bio/PDB/Residue.py
+++ b/Bio/PDB/Residue.py
@@ -113,10 +113,11 @@ class Residue(Entity):
     def get_atoms(self):
         for a in self:
             yield a
-            
+
     def get_atom(self):
         warnings.warn("`get_atom` has been depcreated and we intend to remove it"
-                      " in a future release of Biopython. Please use `get_atoms` instead.")
+                      " in a future release of Biopython. Please use `get_atoms` instead.",
+                     BiopythonDeprecationWarning)
         for a in self:
             yield a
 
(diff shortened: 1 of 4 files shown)


peterjc:
You can't just remove an old API like that, it would break any script using the old function name. We'd have to have both in parallel for a while and deprecate the old name.

However, there was talk of deprecating all the PDB module's ``get_*`` methods in preference to using Python properties (e.g. ``.atoms`` instead of ``.get_atoms()`` as so on).

JoaoRodrigues:
While I fully agree with Peter's first sentence, this seems more of a typo
than anything else actually. I *never* noticed this and I've been using
this module for years.. I'd keep an alias to the function and add the
deprecation warning actually.​

peterjc:
So we're agreed :) - rename it but keep a backward compatible alias with a deprecation warning when used. See also http://biopython.org/wiki/Deprecation_policy

rodogi:
I added the new method `get_atoms` while keeping the old one `get_atom` with a deprecation warning.

JoaoRodrigues:
Looks good to me. Thanks @rodogi for spotting & fixing this.

peterjc:
The change looks great 👍 

Does this currently get triggered by any of the unit tests?

rodogi:
Thank you guys for your terrific work in biopython. 

> Does this currently get triggered by any of the unit tests?

I don't really know, I'm still a noob to github.

peterjc:
It looks like the old name does get used in ``test_mmtf.py`` (only):

```
$ grep "\.get_atom(" *.py
test_mmtf.py:            self.mmcif_atoms = [x for x in mmcif_r.get_atom()]
test_mmtf.py:            self.mmtf_atoms = [x for x in mmtf_r.get_atom()]
```

This is confirmed in the test suite output on TravisCI where the new deprecation warning appears:

```
...
test_mmtf ... /home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/PDB/Residue.py:120: BiopythonDeprecationWarning: `get_atom` has been depcreated and we intend to remove it in a future release of Biopython. Please use `get_atoms` instead.
  BiopythonDeprecationWarning)
ok
...
```

So, could you also update ``test_mmtf.py`` to use the new ``.get_atoms()`` version please. With that I'd be happy to merge this pull request.

Separately this shows the old method was not being deliberately tested in ``test_PDB.py``, and it (or rather the new version) ought to be. If you'd like to tackle that as well @rodogi that would be most welcome.

Bio/PDB/Residue.py
```diff
@@ -108,7 +110,14 @@ def get_unpacked_list(self):
     def get_segid(self):
         return self.segid
 
+    def get_atoms(self):
+        for a in self:
+            yield a
+
     def get_atom(self):
+        warnings.warn("`get_atom` has been depcreated and we intend to remove it"
```
peterjc: Typo: ``depcreated`` --> ``deprecated``

peterjc:
@rodogi How would you like to be named in the ``NEWS.rst`` and ``CONTRIB.rst`` files? I'm asking as you don't make this obvious on your GitHub profile, perhaps deliberately. Should you wish to remain semi-anonymous, just ``rodogi`` is fine (there is recent precedent).

You can add yourself to these files as part of the pull request if you like - note the names are sorted alphabetically and for ``CONTRIB.rst`` this is checked automatically as described in the header.

peterjc:
Thank you 👍

peterjc:
@rodogi Are you happy to dual licence your contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_? See #898.

rodogi:
@peterjc Yes I agree to dual licence my contributions under both the Biopython License Agreement AND the BSD 3-Clause License.

peterjc:
@rodogi Thank you 👍

PR #1228, opened 2017-05-15
Biosql doc update

ctSkennerton:
Added in docstrings for public methods for BioSQL

### DIFF
--- a/BioSQL/BioSeq.py
+++ b/BioSQL/BioSeq.py
@@ -135,6 +135,7 @@ class DBSeq(Seq):
         """Add another sequence or string to this sequence.
 
         The sequence is first converted to a Bio.Seq object before the addition.
+        The returned object is a Bio.Seq, not a BioSQL.DBSeq
         """
         # Let the Seq object deal with the alphabet issues etc
         return self.toseq() + other
@@ -143,6 +144,7 @@ class DBSeq(Seq):
         """Add another sequence or string to the left.
 
         The sequence is first converted to a Bio.Seq object before the addition.
+        The returned object is a Bio.Seq, not a BioSQL.DBSeq
         """
         # Let the Seq object deal with the alphabet issues etc
         return other + self.toseq()
--- a/BioSQL/BioSeq.py
+++ b/BioSQL/BioSeq.py
@@ -134,8 +134,8 @@ class DBSeq(Seq):
     def __add__(self, other):
         """Add another sequence or string to this sequence.
 
-        The sequence is first converted to a Bio.Seq object before the addition.
-        The returned object is a Bio.Seq, not a BioSQL.DBSeq
+        The sequence is first converted to a Seq object before the addition.
+        The returned object is a Seq object, not a DBSeq object
         """
         # Let the Seq object deal with the alphabet issues etc
         return self.toseq() + other
@@ -143,8 +143,8 @@ class DBSeq(Seq):
     def __radd__(self, other):
         """Add another sequence or string to the left.
 
-        The sequence is first converted to a Bio.Seq object before the addition.
-        The returned object is a Bio.Seq, not a BioSQL.DBSeq
+        The sequence is first converted to a Seq object before the addition.
+        The returned object is a Seq object, not a DBSeq object
         """
         # Let the Seq object deal with the alphabet issues etc
         return other + self.toseq()
--- a/BioSQL/BioSeq.py
+++ b/BioSQL/BioSeq.py
@@ -40,9 +40,11 @@ class DBSeq(Seq):
         self.start = start
 
     def __len__(self):
+        """Return the length of the sequence."""
         return self._length
 
     def __getitem__(self, index):  # Seq API requirement
+        """Return a subsequence or single letter."""
         # Note since Python 2.0, __getslice__ is deprecated
         # and __getitem__ is used instead.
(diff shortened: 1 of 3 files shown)


BioSQL/BioSeq.py
```diff
@@ -130,10 +132,18 @@ def toseq(self):
         return Seq(str(self), self.alphabet)
 
     def __add__(self, other):
+        """Add another sequence or string to this sequence.
+
+        The sequence is first converted to a Bio.Seq object before the addition.
```
peterjc: I think it is unclear from text which object gets turned into a Seq object. How about something like ``This sequence is first converted to a Seq object before the addition of the other sequence.``? Should we say it will return a ``Seq`` object (not a ``DBSeq`` object)?

peterjc:
Looks good - does pydocstyle complain about over indentation with e.g.

```
        Create a DBSeqRecord object.
 
        Arguments:
            - adaptor - A BioSQL.BioSeqDatabase.Adaptor object
            - primary_id - An internal integer ID used by BioSQL

        You wouldn't normally create a DBSeqRecord object yourself,
        this is done for you when using a BioSeqDatabase object
```

vs:

```
        Create a DBSeqRecord object.
 
        Arguments:

        - adaptor - A BioSQL.BioSeqDatabase.Adaptor object
        - primary_id - An internal integer ID used by BioSQL

        You wouldn't normally create a DBSeqRecord object yourself,
        this is done for you when using a BioSeqDatabase object
```

According to rst.ninjs.org your version (the first version above) produces slightly more compact output compared to the second (how I've been doing recent docstring edits), so would be preferable. Cross reference #1170.

ctSkennerton:
I've updated it to say that those methods return a Bio.Seq object. 

pydocstyle does not warn about the indentation so I guess it is fine.

peterjc:
Note due to our unfortunate historical naming, the ``Seq`` object is defined in ``Bio/Seq.py`` meaning in full it is the ``Bio.Seq.Seq`` object. Therefore I prefer to always do the import as ``from Bio.Seq import Seq`` and refer to it as the ``Seq`` object in user-facing documentation.

Similarly for the ``SeqRecord`` and ``SeqFeature`` objects.

peterjc:
The Python 3.5 test failed but it was a stochastic failure (random points for graph plotting), I've restarted it.

BioSQL/BioSeq.py
```diff
@@ -130,10 +132,20 @@ def toseq(self):
...
+        The returned object is a Bio.Seq, not a BioSQL.DBSeq
+        """
         # Let the Seq object deal with the alphabet issues etc
         return self.toseq() + other
 
     def __radd__(self, other):
+        """Add another sequence or string to the left.
+
+        The sequence is first converted to a Bio.Seq object before the addition.
+        The returned object is a Bio.Seq, not a BioSQL.DBSeq
```
peterjc: See my general comment about what to call the ``Seq`` object. Likewise this is not the ``BioSQL.DBSeq`` object. In full it is the ``BioSQL.BioSeq.DBSeq`` object, so I would recommend just calling it the ``DBSeq`` object.

BioSQL/BioSeq.py
peterjc: (I have expanded on the ``Seq`` object naming in a comment on the main issue)

BioSQL/BioSeq.py
ctSkennerton: Ok, I'll fix up my Language

BioSQL/BioSeqDatabase.py
```diff
@@ -46,7 +46,6 @@ def open_database(driver="MySQLdb", **kwargs):
      - password, passwd - the password to connect with
      - host - the hostname of the database
      - database or db - the name of the database
-
```
peterjc: Please check this is still valid RST. I suspect that we can in fact drop the blank line after "Arguments" provided the bullet point list is indented.

BioSQL/BioSeqDatabase.py
```diff
@@ -389,6 +413,13 @@ def fetch_dbid_by_dbname(self, dbname):
         return rv[0][0]
 
     def fetch_seqid_by_display_id(self, dbid, name):
+        """Return the internal id for a sequence using its display id.
+
+        Argunments:
+            - dbid - the internal id for the sub-database
+            - name - the name of the sequence. Corresponds to the
+            name column of the bioentry table of the SQL schema
```
peterjc: Please check this is valid RST, I suspect the final line should be two spaces more indented as a continuation of the bullet point.

BioSQL/BioSeqDatabase.py
```diff
@@ -403,6 +434,13 @@ def fetch_seqid_by_display_id(self, dbid, name):
         return rv[0][0]
 
     def fetch_seqid_by_accession(self, dbid, name):
+        """Return the internal id for a sequence using its accession.
+
+        Argunments:
+            - dbid - the internal id for the sub-database
+            - name - the accession of the sequence. Corresponds to the
+            accession column of the bioentry table of the SQL schema
```
peterjc: Please check this is valid RST, I suspect the final line should be two spaces more indented as a continuation of the bullet point.

peterjc:
Great, thanks.

PR #1235, opened 2017-05-17
Move self-test from Bio\PDB\PDBParser.py

Gasta88:
As mentioned in issue #820 , I've moved the self test from `Bio\PDB\PDBParser.py` to `Tests\test_PDB.py`.


peterjc:
Thank  you

PR #1237, opened 2017-05-17
PDB.parse_pdb_header can parse headers for missing residues

Bernhard10:
It can now parse certain REMARK 465 headers.

It can certainly detect the presence of this remark and
can parse the content of this remark, if it follows the
template in the pdb-specification.

### DIFF
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -132,13 +132,14 @@ def parse_pdb_header(infile):
     return _parse_pdb_header_list(header)
 
 
-def _parse_remark_465(line, out_dict):
+def _parse_remark_465(line):
     """Parse missing residue remarks.
 
-    Fills two fields in out_dict: has_missing_residues and missing_residues.
-    It specification for REMARK 465 only gives templates, but does not say
-    they have to be followed. So we assume that not all pdb-files with a
-    REMARK 465 can be understood.
+    Returns a dictionary describing the missing residue.
+    The specification for REMARK 465 at
+    http://www.wwpdb.org/documentation/file-format-content/format33/remarks2.html#REMARK%20465
+    only gives templates, but does not say they have to be followed.
+    So we assume that not all pdb-files with a REMARK 465 can be understood.
 
     The dictionary entry "has_missing_residues" will be set to True, if
     at least one REMARK 465 entry with non-empty text exists.
@@ -149,17 +150,19 @@ def _parse_remark_465(line, out_dict):
     empty or incomplete, if the pdb-header is not successfully parsed.
     """
     if line:
-        out_dict["has_missing_residues"] = True
         # Note that line has been stripped.
         assert line[0] != " " and line[-1] not in "\n ", "line has to be stripped"
-    # Optional model number and residue name with 1 (e.g. for RNA) to 3 characters
-    modelnr_and_resname = "(\d+\s[\sA-Z][\sA-Z][A-Z]|[A-Z]?[A-Z]?[A-Z])"
-    chain = "\s([A-Za-z0-9])"
-    # Digit followed by optional insertion code.
-    # Note: Hetero-flags make no sense in contexty with missing residues.
-    ssseq = "\s+(\d+[A-Za-z]?)$"
-    pattern = modelnr_and_resname + chain + ssseq
-    match = re.match(pattern, line)
+    pattern = (r"""
+                (\d+\s[\sA-Z][\sA-Z][A-Z] |   # Either model number + residue name
+                 [A-Z]?[A-Z]?[A-Z])           # Or only residue name with
+                                              # 1 (RNA) to 3 letters
+                \s ([A-Za-z0-9])              # A single character chain
+                \s+(\d+[A-Za-z]?)$            # Residue number: A digit followed
+                                              # by an optional insertion code
+                                              # (Hetero-flags make no sense in
+                                              # context with missing res)
+                """)
+    match = re.match(pattern, line, re.VERBOSE)
     if match is not None:
         residue = {}
         if " " in match.group(1):
@@ -170,12 +173,13 @@ def _parse_remark_465(line, out_dict):
         residue["chain"] = match.group(2)
         try:
             residue["ssseq"] = int(match.group(3))
-        except:
+        except ValueError:
             residue["insertion"] = match.group(3)[-1]
             residue["ssseq"] = int(match.group(3)[:-1])
         else:
             residue["insertion"] = None
-        out_dict["missing_residues"].append(residue)
+        return residue
+    return None
 
 
 def _parse_pdb_header_list(header):
@@ -299,8 +303,11 @@ def _parse_pdb_header_list(header):
                     # print('nonstandard resolution %r' % r)
                     dict['resolution'] = None
             elif hh.startswith("REMARK 465"):
-                # Update the dictionary with content of the remark 465 line (Missing residues)
-                _parse_remark_465(tail, dict)
+                if tail:
+                    dict['has_missing_residues'] = True
+                    missing_res_info = _parse_remark_465(tail)
+                    if missing_res_info:
+                        dict['missing_residues'].append(missing_res_info)
         else:
             # print(key)
             pass
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -192,7 +192,7 @@ def _parse_pdb_header_list(header):
         'author': "",
         'compound': {'1': {'misc': ''}}, 'source': {'1': {'misc': ''}},
         'has_missing_residues': False,
-        'missing_residues' : []}
+        'missing_residues': []}
 
     dict['structure_reference'] = _get_references(header)
     dict['journal_reference'] = _get_journal(header)
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -141,13 +141,8 @@ def _parse_remark_465(line):
     only gives templates, but does not say they have to be followed.
     So we assume that not all pdb-files with a REMARK 465 can be understood.
 
-    The dictionary entry "has_missing_residues" will be set to True, if
-    at least one REMARK 465 entry with non-empty text exists.
-
-    The dictionary entry 'missing_residues' will be filled with those
-    missing residues that can be parsed from the PDB header.
-    WARNING: The list out_dict['missing_residues'] will be
-    empty or incomplete, if the pdb-header is not successfully parsed.
+    Returns a dictionary with the following keys:
+    "model", "res_name", "chain", "ssseq", "insertion"
     """
     if line:
         # Note that line has been stripped.
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -166,7 +166,8 @@ def _parse_remark_465(line):
     if match is not None:
         residue = {}
         if " " in match.group(1):
-            residue["model"], residue["res_name"] = match.group(1).split(" ")
+            model, residue["res_name"] = match.group(1).split(" ")
+            residue["model"] = int(model)
         else:
             residue["model"] = None
             residue["res_name"] = match.group(1)
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -131,6 +131,52 @@ def parse_pdb_header(infile):
                 header.append(l)
     return _parse_pdb_header_list(header)
 
+def _parse_remark_465(line, out_dict):
+    """Parse missing residue remarks.
+
+    Fills two fields in out_dict: has_missing_residues and missing_residues.
+    It specification for REMARK 465 only gives templates, but does not say
+    they have to be followed. So we assume that not all pdb-files with a
+    REMARK 465 can be understood.
+
+    The dictionary entry "has_missing_residues" will be set to True, if
+    at least one REMARK 465 entry with non-empty text exists.
+
+    The dictionary entry 'missing_residues' will be filled with those
+    missing residues that can be parsed from the PDB header.
+    WARNING: The list out_dict['missing_residues'] will be
+    empty or incomplete, if the pdb-header is not successfully parsed.
+    """
+    if line:
+        out_dict["has_missing_residues"] = True
+
+    # Note that line has been stripped.
+    assert  not line or (line[0]!=" " and line[-1] not in "\n "), "line has to be stripped"
+    #Optional model number and residue name with 1 (e.g. for RNA) to 3 characters
+    modelnr_and_resname = "(\d+\s[\sA-Z][\sA-Z][A-Z]|[A-Z]?[A-Z]?[A-Z])"
+    chain="\s([A-Za-z0-9])"
+    #Digit followed by optional insertion code.
+    #Note: Hetero-flags make no sense in contexty with missing residues.
+    ssseq="\s+(\d+[A-Za-z]?)$"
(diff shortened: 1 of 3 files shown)


Bio/PDB/parse_pdb_header.py
```diff
@@ -140,7 +186,10 @@ def _parse_pdb_header_list(header):
         'structure_reference': "unknown",
         'journal_reference': "unknown",
         'author': "",
-        'compound': {'1': {'misc': ''}}, 'source': {'1': {'misc': ''}}}
+        'compound': {'1': {'misc': ''}}, 'source': {'1': {'misc': ''}},
+        'has_missing_residues': False,
+        'missing_residues' : []
```
peterjc: Please include a trailing comma here (ready for when someone adds more entries to the dict)

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -0,0 +1,50 @@
+"""Unit tests for PARTS of the parse_pdb_header module of Bio.PDB"""
```
peterjc: Missing copyright and license header. Assuming you agree with the dual licensing #898, please use:

```python
# Copyright 2017 by Bernhard Thiel.  All rights reserved.
# package.
```

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -0,0 +1,50 @@
...
+
+import unittest
+import warnings
+
+try:
+    import numpy
+    from numpy import dot  # Missing on old PyPy's micronumpy
```
peterjc: Do you use dot, svd, det? If not then just checking ``import numpy` is likely enough.

peterjc:
The style checks are failing,

```
$ flake8 Bio/
Bio/PDB/parse_pdb_header.py:131:1: E302 expected 2 blank lines, found 1
Bio/PDB/parse_pdb_header.py:151:11: E271 multiple spaces after keyword
Bio/PDB/parse_pdb_header.py:151:33: E225 missing whitespace around operator
Bio/PDB/parse_pdb_header.py:152:5: E265 block comment should start with '# '
Bio/PDB/parse_pdb_header.py:154:10: E225 missing whitespace around operator
Bio/PDB/parse_pdb_header.py:155:5: E265 block comment should start with '# '
Bio/PDB/parse_pdb_header.py:156:5: E265 block comment should start with '# '
Bio/PDB/parse_pdb_header.py:157:10: E225 missing whitespace around operator
Bio/PDB/parse_pdb_header.py:191:27: E203 whitespace before ':'
Bio/PDB/parse_pdb_header.py:192:9: E124 closing bracket does not match visual indentation
Bio/PDB/parse_pdb_header.py:300:17: E265 block comment should start with '# '
```

By eye I would expect ``flake8 Tests/`` to fail as well. You can install this command line tool locally use ``pip install flake8``.

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        header = parse_pdb_header("PDB/2XHE.pdb")
+        self.assertTrue(header["has_missing_residues"])
+        self.assertEqual(len(header["missing_residues"]), 142)
+        self.assertIn( {"model":None, "res_name":"GLN", "chain":"B", "ssseq":267, "insertion":None}  , header["missing_residues"])
+        header = parse_pdb_header("PDB/1A8O.pdb")
+        self.assertFalse(header["has_missing_residues"])
+        self.assertEqual(header["missing_residues"], [])
+    def test_parse_remark_465(self):
+        "A UNIT-test for the private function _parse_remark_465"
```
peterjc: Please use tripe double-quotes as per PEP257, and include trailing full stop. e.g.

```
"""Test private function _parse_remark_465 for missing residues."""
```

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -0,0 +1,50 @@
...
+
+from Bio.PDB.parse_pdb_header import parse_pdb_header, _parse_remark_465
+
+
+class ParseReal(unittest.TestCase):
+    """Testing with real PDB file(s)."""
+    def test_parse_pdb_with_remark_465(self):
+        """
+        Tests that parse_pdb_header now can identify some REMARK 465 entries.
+        """
```
peterjc: Please use a single line as per PEP257.

peterjc:
@etal @lennax @JoaoRodrigues Could you comment on the content of this pull request please? I've just made some notes about the style side of things. Thanks!

Tests/test_PDB_parse_pdb_header.py
Bernhard10: I used test_PDB_MMCIFParser.py as a template for my test-file. I'll update the imports.

peterjc: Ah - that probably was probably copied from ``test_PDB.py`` where things like SVD are being used. The ``dot`` import is probably irrelevant given how much PyPy has improved.

I'll check those - thanks!

Bio/PDB/parse_pdb_header.py
```diff
@@ -129,18 +129,67 @@ def parse_pdb_header(infile):
...
+            residue["insertion"] = match.group(3)[-1]
+            residue["ssseq"] = int(match.group(3)[:-1])
+        else:
+            residue["insertion"] = None
+        out_dict["missing_residues"].append(residue)
+
+
 def _parse_pdb_header_list(header):
     # database fields
     dict = {'name': "",
```
lennax: @peterjc what is the policy for resolving namespace clobbering in the midst of content PRs?

peterjc: Could you explain in more detail?

If you mean the existing PDB namespace shadowing caused by the way ``Bio/PDB/__init__.py`` imports things we're largely stuck with that for backward compatibility.

lennax: @peterjc I meant the use of `dict` as a name!

peterjc: Oh yeah - I didn't see it, thanks. Definitely to be avoided for new code, and worth cleaning up where possible in old code (i.e. internal variable names can be changed, argument names cannot). Due to its usage being so common as shorthand for an identifier, trying to avoid built-in ``id`` seems like a lost battle.

In terms of pull requests, I'd prefer that kind of tidying up work be a separate pull request to ensure it gets done in a separate commit even if the pull request gets squashed.

@Bernhard10 if you want to fix that bit of the existing code, could you do it as a second pull request (after this one is merged to avoid any conflicts) please? Thanks!

Bernhard10: Sure, I can fix the variable name in another PR. Except for the variable name: I guess there are lots of things you could do to improve the PDB-parsing code, but I don't know if it is worth the effort considering how inconsistent PDBs can be and considering that the pdb file-format will slowly become outdated.

peterjc:
@lennax Thanks for looking at this - are you happy for this to be merged?

Bio/PDB/parse_pdb_header.py
```diff
@@ -129,18 +129,67 @@ def parse_pdb_header(infile):
     return _parse_pdb_header_list(header)
 
 
+def _parse_remark_465(line, out_dict):
+    """Parse missing residue remarks.
+
+    Fills two fields in out_dict: has_missing_residues and missing_residues.
+    It specification for REMARK 465 only gives templates, but does not say
```
lennax: It would be helpful to provide a link to the specification

Bio/PDB/parse_pdb_header.py
```diff
@@ -247,6 +296,9 @@ def _parse_pdb_header_list(header):
                 except:
                     # print('nonstandard resolution %r' % r)
                     dict['resolution'] = None
+            elif hh.startswith("REMARK 465"):
+                # Update the dictionary with content of the remark 465 line (Missing residues)
+                _parse_remark_465(tail, dict)
```
lennax: Minor: I would personally prefer the syntax `dict.update(_parse_remark_465(tail))` - passing the result data structure doesn't feel very pythonic.

Bio/PDB/parse_pdb_header.py
```diff
@@ -129,18 +129,67 @@ def parse_pdb_header(infile):
...
+    missing residues that can be parsed from the PDB header.
+    WARNING: The list out_dict['missing_residues'] will be
+    empty or incomplete, if the pdb-header is not successfully parsed.
+    """
+    if line:
+        out_dict["has_missing_residues"] = True
+        # Note that line has been stripped.
+        assert line[0] != " " and line[-1] not in "\n ", "line has to be stripped"
+    # Optional model number and residue name with 1 (e.g. for RNA) to 3 characters
+    modelnr_and_resname = "(\d+\s[\sA-Z][\sA-Z][A-Z]|[A-Z]?[A-Z]?[A-Z])"
```
lennax: minor: regex strings should be raw strings to prevent C-style backslash expansion, e.g. `r"\b"`. (not used in this case but it's good practice to avoid confusion)

Bio/PDB/parse_pdb_header.py
```diff
@@ -129,18 +129,67 @@ def parse_pdb_header(infile):
...
+    chain = "\s([A-Za-z0-9])"
+    # Digit followed by optional insertion code.
+    # Note: Hetero-flags make no sense in contexty with missing residues.
+    ssseq = "\s+(\d+[A-Za-z]?)$"
+    pattern = modelnr_and_resname + chain + ssseq
```
lennax: You could tie the comments to the expression and avoid adding them later by using the `re.VERBOSE` flag: https://docs.python.org/2/library/re.html#re.VERBOSE - it would also make it easier to track the whitespace.

Bio/PDB/parse_pdb_header.py
```diff
@@ -129,18 +129,67 @@ def parse_pdb_header(infile):
...
+        residue = {}
+        if " " in match.group(1):
+            residue["model"], residue["res_name"] = match.group(1).split(" ")
+        else:
+            residue["model"] = None
+            residue["res_name"] = match.group(1)
+        residue["chain"] = match.group(2)
+        try:
+            residue["ssseq"] = int(match.group(3))
+        except:
```
lennax: It's never recommended to catch all exceptions because it can catch exits or interrupts. In this case, you are handling errors for calling int, which can throw TypeError or ValueError, so it would be better to use `except (TypeError, ValueError):`. Or at least `except Exception:`

peterjc: Well spotted. I'm surprised ``flake8`` didn't catch that, I will look at ``flake8-blind-except`` as this is an easy automated thing we ought to be checking.

peterjc: See #1253

Bio/PDB/parse_pdb_header.py
Bernhard10: `dict.update` would not work, because there are multiple "REMARK 465" lines. It would have to be something like:

    if tail:
        dict['has_missing_residues'] = True
        missing_res_info = _parse_remark_465(tail)
        if missing_res_info: #None, if we parsed a header-line of REMARK 465
             dict['missing_residues'].append(missing_res_info)

lennax: Ah, good point. I missed the `.append()`.

Bernhard10:
Sorry that it took me so many months before I could come back to this PR.
I made the suggested edits to the code, but I just noticed that the doc-string for `_parse_remark_465` is no longer up to date. I'll fix this Tuesday morning (European time)

JoaoRodrigues:
Will have a look at it tomorrow (Wed, 24th), I guess when all fixes have been made!

Bernhard10:
So from my side I think I have incorporated all needed changes.

peterjc:
Thanks Bernard - over to @JoaoRodrigues who offered to look at this.

JoaoRodrigues:
@Bernhard10 Cool, will have a look then!

Bernhard10:
I have now rebased the code on the newest version of the master branch.
Do you think this is something you will merge soon?

Bio/PDB/parse_pdb_header.py
```diff
@@ -144,7 +190,9 @@ def _parse_pdb_header_list(header):
         'structure_reference': "unknown",
         'journal_reference': "unknown",
         'author': "",
-        'compound': {'1': {'misc': ''}}, 'source': {'1': {'misc': ''}}}
+        'compound': {'1': {'misc': ''}}, 'source': {'1': {'misc': ''}},
+        'has_missing_residues': False,
+        'missing_residues' : []}
```
peterjc: You'll need to remove a space here for the automated PEP8 style checks which are failing on TravisCI,

```
$ flake8 Bio/
Bio/PDB/parse_pdb_header.py:195:27: E203 whitespace before ':'
```

See the ``CONTRIBUTING.rst`` file for more details if you want to run this check locally.

peterjc:
Bernhard has fixed the style issues and taken on board the other feedback - @JoaoRodrigues can you look at this please?

Bio/PDB/parse_pdb_header.py
```diff
@@ -132,6 +132,52 @@ def parse_pdb_header(infile):
...
+
+    Returns a dictionary with the following keys:
+    "model", "res_name", "chain", "ssseq", "insertion"
+    """
+    if line:
+        # Note that line has been stripped.
+        assert line[0] != " " and line[-1] not in "\n ", "line has to be stripped"
+    pattern = (r"""
+                (\d+\s[\sA-Z][\sA-Z][A-Z] |   # Either model number + residue name
+                 [A-Z]?[A-Z]?[A-Z])           # Or only residue name with
```
lennax:
```suggestion
                 [A-Z]{1,3})           # Or only residue name with
```

Bio/PDB/parse_pdb_header.py
```diff
@@ -132,6 +132,52 @@ def parse_pdb_header(infile):
...
+    only gives templates, but does not say they have to be followed.
+    So we assume that not all pdb-files with a REMARK 465 can be understood.
+
+    Returns a dictionary with the following keys:
+    "model", "res_name", "chain", "ssseq", "insertion"
+    """
+    if line:
+        # Note that line has been stripped.
+        assert line[0] != " " and line[-1] not in "\n ", "line has to be stripped"
+    pattern = (r"""
```
lennax: When using `re.VERBOSE` I prefer to do `pattern = re.compile(r"raw pattern", re.VERBOSE)`, then `match = pattern.match(line)` - this prevents trying to use the pattern again and forgetting the verbose flag.

Bio/PDB/parse_pdb_header.py
```diff
@@ -132,6 +132,52 @@ def parse_pdb_header(infile):
...
+                 [A-Z]?[A-Z]?[A-Z])           # Or only residue name with
+                                              # 1 (RNA) to 3 letters
+                \s ([A-Za-z0-9])              # A single character chain
+                \s+(\d+[A-Za-z]?)$            # Residue number: A digit followed
+                                              # by an optional insertion code
+                                              # (Hetero-flags make no sense in
+                                              # context with missing res)
+                """)
+    match = re.match(pattern, line, re.VERBOSE)
+    if match is not None:
```
lennax: I personally prefer early returns: `if match is None: return None`. This allows dedenting the rest of the function.

Bio/PDB/parse_pdb_header.py
peterjc: That is neater :)

peterjc:
Thanks for looking at this @lennax - we should probably merge this now, unless @Bernhard10 wanted to make those stylistic improbvments? I'll probably merge this in the coming week if there is no further comment.

PR #1239, opened 2017-05-18
blat-psl fix for protein queries with negative strand hits

bow:
Continuing the report in #1136 , this is the fix.

If there are no objections, I will merge this when all test pass.


peterjc:
The style checks and tests pass, so please go ahead - thanks.

Issue #1241, opened 2017-05-18
QueryResult object from Bio.SearchIO is not picklable


docmanny:
Hello all,

I first encountered this issue when trying to adapt script running BLAST analysis in parallel to use BLAT and SearchIO:
```
File "/usr/lib64/python3.4/multiprocessing/queues.py", line 247, in _feed
    obj = ForkingPickler.dumps(obj)
  File "/usr/lib64/python3.4/multiprocessing/reduction.py", line 50, in dumps
    cls(buf, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <function QueryResult.<lambda> at 0x7f4bcc2f0bf8>: attribute lookup <lambda> on Bio.SearchIO._model.query failed
```
After confirming that the rest of my code was free of lambdas, I was able to conclude that the error was intrinsic to the QueryResults object, and was able to generate a minimal reproducible example:
```
from Bio import SearchIO
# Note that it doesn't matter what the input file selected is; 
# I tried psl, pslx, and blast8 formats
record = SearchIO.read('./blatx_test.psl', 'blat-psl')
import pickle
with open('./test.pickle', 'w') as testpickle:
    pickle.dump(record, testpickle)
```
```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
_pickle.PicklingError: Can't pickle <function QueryResult.<lambda> at 0x7f1a3667c1e0>: attribute lookup <lambda> on Bio.SearchIO._model.query failed
```
From what I can see in the source code, I suspect that the offending line is in the def__init__() in Bio.SearchIO._model.query: `def __init__(self, hits=(), id=None, hit_key_function=lambda hit: hit.id):`

The inability to pickle QueryResults objects means that SearchIO cannot be easily used in multiproccessing-or-pickling-based scripts. Is this something that could be fixed in Bio.SearchIO, or is the presence of lambda functions in the class something that must be worked around?

peterjc:
Probably we should do:

```python
def __init__(self, hits=(), id=None, hit_key_function=None):
    ...
    if  hit_key_function is None:
       hit_key_function=lambda hit: hit.id
    ...
```

Better yet, define the this as a private function once at module level?

bow:
Ah indeed, the closure makes the object unplickable. @peterjc , I think we should go with not using lambda at all (so private function is an option). How about making it a private staticmethod of the class? Just to avoid cluttering the module namespace, even as a private function.

Anyway, I'll open a PR soon.

[...]

PR #1244, opened 2017-05-18
Remove closure as default argument in QueryResult.__init__

bow:
Continuing from the #1241 , this PR adds a fix so that QueryResult can be pickled.

Things to still consider:
* Should we add pickle (de)serialization tests as well?
* Is making the default argument the class's staticmethod a good idea?


peterjc:
This is certainly an improvement.

Pickle tests are probably a good idea - also for the Seq / SeqFeature / SeqRecord etc.

Issue #1245, opened 2017-05-19
Blat PSL writer in SearchIO errors when handling protein queries


bow:
This is related to #1136 (creating the issue here for book keeping), in this case the cause is the writer class currently ignores the `is_protein` attribute of the input `QueryResult` which causes some coordinate-related sanity checks to fail. I'm making a PR and will close this when all tests pass.

PR #1246, opened 2017-05-19
Fix blat-psl write error when query is protein and add tests

bow:
As reported in #1245 ~ I will merge this after all tests pass.

### DIFF
--- a/Bio/SearchIO/BlatIO.py
+++ b/Bio/SearchIO/BlatIO.py
@@ -642,6 +642,9 @@ class BlatPslWriter(object):
         for hit in qresult:
             for hsp in hit.hsps:
 
+                query_is_protein = getattr(hsp, "query_is_protein", False)
+                blocksize_multiplier = 3 if query_is_protein else 1
+
                 line = []
                 line.append(hsp.match_num)
                 line.append(hsp.mismatch_num)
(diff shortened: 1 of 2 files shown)


Bio/SearchIO/BlatIO.py
```diff
@@ -653,7 +656,8 @@ def _build_row(self, qresult):
                 line.append(hsp.hit_gap_num)
 
                 # check spans
-                assert hsp.query_span_all == hsp.hit_span_all
+                eff_query_spans = [blocksize_multiplier * s for s in hsp.query_span_all]
+                assert hsp.hit_span_all == eff_query_spans
```
peterjc: This should probably throw an exception (``ValueError`` maybe?) rather than use an assert, just in case someone is running Python with asserts turned off.

bow: Good point. Looking at other parts of SearchIO, I think I was quite deliberate in my use of `assert`s. Perhaps it's a good idea to check this in other places as well.

peterjc: There are probably lots of asserts in Biopython which might be better done as explicitly exceptions, don't let this delay merging this pull request.

bow: Ok. I'll update this one in particular to use `ValueError` and keep the other `assert` use in mind for now.

Issue #795, opened 2016-04-03
SearchIO: improve or disable logic handling duplicated IDs in BLAST target database appearing in XML results


mmokrejs:
Hi Bow,
  it happened to me TransDecoder-2.1.0 created duplicate FASTA entries in its output *.mRNA files (see issue https://github.com/TransDecoder/TransDecoder/issues/13). I parsed original accessions from the FASTA files and I do not mind in this case there are duplicates. I see I could probably better extract the blast-ids from the formatted index files which formatdb/makeblastdb created. Can biopython do it for me? I just used the FASTA parser to obtain the IDs but could have used another parser as I still have the target database files. ;-)

```

/usr/lib64/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py:322: BiopythonParserWarning: Adding hit with BLAST-generated ID 'gnl|BL_ORD_ID|521330' since hit ID 'Vespa_velutina__ESTisotig01210' is already present in query 'GOCZ8YT01D9U7U'. Your BLAST database may contain duplicate entries.
  (hit._blast_id, hit.id, query_id), BiopythonParserWarning)
/usr/lib64/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py:322: BiopythonParserWarning: Adding hit with BLAST-generated ID 'gnl|BL_ORD_ID|528349' since hit ID 'Vespa_velutina__ESTisotig27960' is already present in query 'GOCZ8YT01D9U7U'. Your BLAST database may contain duplicate entries.
  (hit._blast_id, hit.id, query_id), BiopythonParserWarning)
/usr/lib64/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py:322: BiopythonParserWarning: Adding hit with BLAST-generated ID 'gnl|BL_ORD_ID|522951' since hit ID 'Vespa_velutina__ESTisotig14083' is already present in query 'GOCZ8YT01DQGJE'. Your BLAST database may contain duplicate entries.
  (hit._blast_id, hit.id, query_id), BiopythonParserWarning)
/usr/lib64/python2.7/site-packages/Bio/SearchIO/BlastIO/blast_xml.py:322: BiopythonParserWarning: Adding hit with BLAST-generated ID 'gnl|BL_ORD_ID|522949' since hit ID 'Vespa_velutina__ESTisotig14081' is already present in query 'GOCZ8YT01DQGJE'. Your BLAST database may contain duplicate entries.
[...]

PR #1250, opened 2017-05-21
Expose 'blast_id' for SearchIO's blast-xml and make ID-mangling adjustable.

bow:
Continuing from #795 , this PR does the following things:

* The raw BLAST IDs of the `QueryResult` and `Hit` objects are now public.
* Adds two keyword arguments to the `blast-xml` parser and writer: `use_raw_query_ids` and `use_raw_hit_ids`. They allow users to disable or enable the ID-mangling that SearchIO is doing (this was previously non-configurable). The default values are set so that no backwards-incompatible changes are produced.

The documentation has been updated accordingly as well.

As usual, I will merge this after the tests pass unless there are any objections.

### DIFF
--- a/Bio/SearchIO/BlastIO/__init__.py
+++ b/Bio/SearchIO/BlastIO/__init__.py
@@ -156,10 +156,18 @@ The blast-xml parser is aware of these modifications and will attempt to extract
 the true sequence IDs out of the descriptions. So when accessing QueryResult or
 Hit objects, you will use the non-BLAST-generated IDs.
 
-Conversely, the blast-xml writer will try to concatenate the true sequence IDs
-with their descriptions and use the BLAST-generated IDs. This enables you to
-write BLAST XML files using SearchIO as if they were written by a real BLAST
-program.
+This behavior on the query IDs can be disabled using the 'use_raw_query_ids'
+parameter while the behavior on the hit IDs can be disabled using the
+'use_raw_hit_ids' parameter. Both are boolean values that can be supplied
+to SearchIO.read or SearchIO.parse, with the default values set to 'False'.
+
+In any case, the raw BLAST IDs can always be accessed using the query or hit
+object's 'blast_id' attribute.
+
+The blast-xml write function also accepts 'use_raw_query_ids' and
+'use_raw_hit_ids' parameters. However, note that the default values for the
+writer are set to 'True'. This is because the writer is meant to mimic native
+BLAST result as much as possible.
 
 
 blast-tab
--- a/Bio/SearchIO/BlastIO/blast_xml.py
+++ b/Bio/SearchIO/BlastIO/blast_xml.py
@@ -220,8 +220,9 @@ def _extract_ids_and_descs(raw_id, raw_desc):
 class BlastXmlParser(object):
     """Parser for the BLAST XML format"""
 
-    def __init__(self, handle):
+    def __init__(self, handle, use_raw_hit_ids=False):
         self.xml_iter = iter(ElementTree.iterparse(handle, events=('start', 'end')))
+        self._use_raw_hit_ids = use_raw_hit_ids
         self._meta, self._fallback = self._parse_preamble()
 
     def __iter__(self):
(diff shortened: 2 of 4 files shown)


Bio/SearchIO/BlastIO/__init__.py
```diff
@@ -156,10 +156,18 @@
...
-program.
+This behavior on the query IDs can be disabled using the 'use_raw_query_ids'
+parameter while the behavior on the hit IDs can be disabled using the
+'use_raw_hit_ids' parameter. Both are boolean values that can be supplied
+to SearchIO.read or SearchIO.parse, with the default values set to 'False'.
+
+In any case, the raw BLAST IDs can always be accessed using the query or hit
+object's 'blast_id' attribute.
+
+The blast-xml can also accept 'use_raw_query_ids' and 'use_raw_hit_ids'
```
peterjc: Missing word - the blast-xml writer?

Bio/SearchIO/BlastIO/blast_xml.py
```diff
@@ -188,26 +188,41 @@
 _RE_ID_DESC_PATTERN = re.compile(" +")
 
 
-def _extract_ids_and_descs(concat_str):
-    # Given a string space-separate string of IDs and descriptions,
-    # return a list of tuples, each tuple containing an ID and
-    # a description string (which may be empty)
+def _extract_ids_and_descs(raw_id, raw_desc):
+    # Given values of the `Hit_id` and `Hit_def` elements, return a tuple
```
peterjc: You might as well turn this informative multi-line comment into a docstring (tagged ```... (PRIVATE).``` at the end of the summary first line).

Bio/SearchIO/BlastIO/__init__.py
bow: Ah yes, good catch. A fix is coming soon.

Bio/SearchIO/BlastIO/blast_xml.py
bow: This is changed now as well ~

Bio/SearchIO/BlastIO/blast_xml.py
```diff
@@ -189,10 +189,14 @@
 
 
 def _extract_ids_and_descs(raw_id, raw_desc):
-    # Given values of the `Hit_id` and `Hit_def` elements, return a tuple
-    # three elements: all IDs, all descriptions, and the BLAST-generated ID.
-    # The BLAST-generated ID is set to `None` if no BLAST-generated IDs are
-    # present.
+    """Extract IDs, descriptions, and raw BLAST ID from the given raw values.
```
peterjc: To be explicit, end the line ``... raw values (PRIVATE).`` (and if you are worried about the length, omit ``the``).

bow: Ah, I missed the `PRIVATE` note. It's added now, and I've also shortened the line.

Bio/SearchIO/BlastIO/__init__.py
```diff
@@ -156,10 +156,18 @@
...
-program.
+This behavior on the query IDs can be disabled using the 'use_raw_query_ids'
+parameter while the behavior on the hit IDs can be disabled using the
+'use_raw_hit_ids' parameter. Both are boolean values that can be supplied
+to SearchIO.read or SearchIO.parse, with the default values set to 'False'.
+
+In any case, the raw BLAST IDs can always be accessed using the query or hit
+object's 'blast_id' attribute.
+
+The blast-xml write also accepts 'use_raw_query_ids' and 'use_raw_hit_ids'
```
peterjc: The English does not flow nicely. How about ``The blast-xml writer also ...`` or ``The blast-xml write function also ...`` here?

peterjc:
Feel free to merge this when you want - I'm just suggesting minor polish.

Bio/SearchIO/BlastIO/__init__.py
bow: Indeed. This was just changed.

bow:
Ok ~ minor polishes should still be included :). I just change parts of the docs as well (a change on NEWS is coming as well since this affects the public-facing API). But before I merge, I think I will squash all documentation commits here into one commit.

Bio/SearchIO/BlastIO/blast_xml.py
```diff
@@ -329,11 +329,11 @@ def _parse_qresult(self):
                     if hit:
                         # need to keep track of hit IDs, since there could be duplicates,
                         if hit.id in key_list:
-                            warnings.warn("Adding hit with BLAST-generated ID "
-                                    "%r since hit ID %r is already present "
-                                    "in query %r. Your BLAST database may contain "
+                            warnings.warn("Renaming hit ID %r to a BLAST-generated ID "
```
peterjc: Really minor change, but I'd remove ``a`` here, and there should only be one BLAST generated ID which could be used?

bow: There is only one BLAST-generated hit for each hit indeed. However this scenario occurs when the same hit ID exists more than once, and for each of those duplicate hit IDs there can be more than one ID that BLAST generates (depending on how many times they occur).

peterjc: Ah - this is subtle isn't it. OK, that's fine as it it - thanks!

PR #1251, opened 2017-05-22
Add public docstring to Bio.Seq

Gasta88:
As mentioned in issue #1203 I've added public docstring to `Bio.Seq` module.

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -5,7 +5,7 @@
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
-"""Provides objects to represent biological sequences with alphabets.
+"""Provide objects to represent biological sequences with alphabets.
 
 See also the Seq_ wiki and the chapter in our tutorial:
     - `HTML Tutorial`_
@@ -14,6 +14,7 @@ See also the Seq_ wiki and the chapter in our tutorial:
 .. _Seq: http://biopython.org/wiki/Seq
 .. _`HTML Tutorial`: http://biopython.org/DIST/docs/tutorial/Tutorial.html
 .. _`PDF Tutorial`: http://biopython.org/DIST/docs/tutorial/Tutorial.pdf
+
 """
 from __future__ import print_function
 
@@ -34,7 +35,7 @@ from Bio.Data import CodonTable
 
 
 def _maketrans(complement_mapping):
-    """Makes a python string translation table (PRIVATE).
+    """Make a python string translation table (PRIVATE).
 
     Arguments:
         - complement_mapping - a dictionary such as ambiguous_dna_complement
@@ -62,7 +63,7 @@ _rna_complement_table = _maketrans(ambiguous_rna_complement)
 
 
 class Seq(object):
-    """A read-only sequence object (essentially a string with an alphabet).
+    """Create a read-only sequence object (essentially a string with an alphabet).
 
     Like normal python strings, our basic sequence object is immutable.
     This prevents you from doing my_seq[5] = "A" for example, but does allow
@@ -130,7 +131,7 @@ class Seq(object):
                                               self.alphabet)
 
     def __str__(self):
-        """Returns the full sequence as a python string, use str(my_seq).
+        """Return the full sequence as a python string, use str(my_seq).
 
         Note that Biopython 1.44 and earlier would give a truncated
         version of repr(my_seq) for str(my_seq).  If you are writing code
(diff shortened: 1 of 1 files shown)


Bio/Seq.py
```diff
@@ -62,7 +63,7 @@ def _maketrans(complement_mapping):
 
 
 class Seq(object):
-    """A read-only sequence object (essentially a string with an alphabet).
+    """Create a read-only sequence object (essentially a string with an alphabet).
```
peterjc: The ``__init__`` could be said to create the object, but that does not make sense for the class docstring. You could just remove the ``A`` and use ``Read-only sequence object (essentially a string with an alphabet).``

Bio/Seq.py
```diff
@@ -198,12 +199,12 @@ def __eq__(self, other):
         return str(self) == str(other)
 
     def __ne__(self, other):
-        """Not equal, see __eq__ documentation."""
+        """Compare sequences with a not-equal operand, see __eq__ documentation."""
```
peterjc: Is this really clearer (likewise for the other special methods)?

Bio/Seq.py
```diff
@@ -513,7 +514,7 @@ def rfind(self, sub, start=0, end=sys.maxsize):
         return str(self).rfind(sub_str, start, end)
 
     def startswith(self, prefix, start=0, end=sys.maxsize):
-        """Does the Seq start with the given prefix?  Returns True/False.
+        """Reurn True if the Seq starts with the given prefix, False otherwise.
```
peterjc: Typo (``Return``)

Bio/Seq.py
```diff
@@ -544,7 +545,7 @@ def startswith(self, prefix, start=0, end=sys.maxsize):
             return str(self).startswith(prefix_str, start, end)
 
     def endswith(self, suffix, start=0, end=sys.maxsize):
-        """Does the Seq end with the given suffix?  Returns True/False.
+        """Return True if the Seq end with the given suffix, False otherwise.
```
peterjc: end --> ends

Bio/Seq.py
```diff
@@ -1139,7 +1140,7 @@ def ungap(self, gap=None):
 
 
 class UnknownSeq(Seq):
-    """A read-only sequence object of known length but unknown contents.
+    """Create a read-only sequence object of known length but unknown contents.
```
peterjc: See comment for the ``Seq`` class docstring, ``Create ...`` is not quite right here.

Bio/Seq.py
```diff
@@ -1391,7 +1394,7 @@ def complement(self):
         return self
 
     def reverse_complement(self):
-        """The reverse complement of an unknown nucleotide equals itself.
+        """Return the reverse complement of an unknown nucleotide equals itself.
```
peterjc: The new version does not make sense to me.

Bio/Seq.py
```diff
@@ -1556,7 +1559,7 @@ def ungap(self, gap=None):
 
 
 class MutableSeq(object):
-    """An editable sequence object (with an alphabet).
+    """Create an editable sequence object (with an alphabet).
```
peterjc: Again, ``Create ...`` is not appropriate.

Bio/Seq.py
```diff
@@ -1986,7 +2001,7 @@ def transcribe(dna):
 
 
 def back_transcribe(rna):
-    """Back-transcribes an RNA sequence into DNA.
+    """Return the RNA back-transcribed sequence into DNA.
```
peterjc: How about ``Return the RNA sequence back-transcribed into DNA.``

Bio/Seq.py
```diff
@@ -2009,7 +2024,7 @@ def back_transcribe(rna):
 
 def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
                    cds=False, pos_stop="X", gap=None):
-    """Helper function to translate a nucleotide string (PRIVATE).
+    """Translate a nucleotide to string (PRIVATE).
```
peterjc: "a nucleotide" implies a single base letter.

How about ``Translate nucleotide string into a protein string (PRIVATE).``

Bio/Seq.py
Gasta88: How about `"""Implement the XYZ operand."""` without any reference to `__eq__`?

peterjc: That seems like an improvement.

Bio/Seq.py
Gasta88: The function itself does not make sense.

How about `Return the reverse complement of an unknown nucleotide,which equals itself.`?

peterjc: If you think the function does not make sense, then clearly the current docstring is not clear enough. How about:

```python
        def reverse_complement(self):
            """Return the reverse complement of an unknown sequence.

            The reverse complement of an unknown nucleotide equals itself:

            >>> from Bio.Seq import UnknownSeq
            >>> from Bio.Alphabet import generic_dna
            >>> example = UnknownSeq(6, generic_dna)
            >>> print(example)
            NNNNNN
            >>> print(example.reverse_complement())
            NNNNNN
            """
```

Gasta88: Much better, thanks.

Gasta88:
File edited.

Bio/Seq.py
```diff
@@ -1391,25 +1394,25 @@ def complement(self):
         return self
 
     def reverse_complement(self):
-        """The reverse complement of an unknown nucleotide equals itself.
+        """Return the reverse complement of an unknown sequence.
 
-        >>> my_nuc = UnknownSeq(10)
-        >>> my_nuc
```
peterjc: Is there a tab here? The display on GitHub looks different for some reason...

peterjc: Well, if it was a tab it is being removed in this change so never mind.

Issue #1128, opened 2017-03-30
Restriction Analysis: Strange behavior of AbaSI in CommOnly


paxelito:
Hi,
I obtained strange results from the restriction analysis based on CommOnly restriction batch for several enzymes, for example focusing on a simple search of AbaSI I find the following results 
```python
    seq = Seq('AAGGTCGGGCAGGA', IUPACAmbiguousDNA())
    rest.AbaSI.search(seq) -> [3, 4]
    rest.AbaSI.characteristic() -> (12, 9, None, None, 'C')
    rest.AbaSI.elucidate() -> CNNNNNNNNN_NN^N
```
The enzyme cuts forward at 3 and 4. If the "^" is placed 12 bases after the recognized sequence, how the first cut can be 3? 

Really thx

peterjc:
Why do you say the enzyme cuts at 3 and 4?

(Also beware that due to a what I consider to be an unfortunate design choice, ``Bio.Restriction`` counts from one rather than zero like Python and most of Biopython)

paxelito:
Hi,
in accordance with the documentation 
``` python
>>> ecoseq = my_seq + Seq(EcoRI.site, amb) + my_seq
>>> ecoseq
Seq('AAAAAAAAAAAAAAGAATTCAAAAAAAAAAAAAA', IUPACAmbiguousDNA())
>>> EcoRI.search(ecoseq)
[16]
 ```
> The position returned by the method search is the first base of the downstream segment produced by a restriction (i.e. the first base after the position where the enzyme will cut)

and

> A more detailled view of the restriction site can be produced using the elucidate() method. The ^ refers to the position of the cut in the sense strand of the sequence,  _ to the cut on the antisense or complementary strand. ^_ means blunt.

as in 
``` python
>>> EcoRI.elucidate()
'G^AATT_C'
```
thus, if I understand correctly, 
``` python
>>> AbaSI.elucidate()
'CNNNNNNNNN_NN^N'
```
means that the `AbaSI` enzyme cuts in the sense strand **12 bases after** the recognized site `C` as also documented in 
``` python
>>> AbaSI.characteristic()
(12, 9, None, None, 'C')
```
thus, looking at my sequence, if search() returns 3 it means that the recognized sequence should be 12 bases before that is **not possible**. 

Moreover, looking at this [AbaSI](https://www.neb.com/products/r0665-abasi) description 

[...]

PR #1256, opened 2017-05-25
Restriction: Bugfixes (CommOnly,  non-palindromic enzymes), more tests

MarkusPiotrowski:
-  Solved a bug from a previous commit (addition of pre-defined RestrictionBatches, #1128)
-  Solved a bug that returned  a positive restriction when the enzyme cutted just in front of the sequence
-  Solved a bug for non-palindromic enzymes: a combined ```re``` expression with the ```|``` operator didn't give the expected results; now do two searches for seq and rev-comp seq.
-  Added tests to ```test_Restriction.py```

### DIFF
--- a/Tests/test_Restriction.py
+++ b/Tests/test_Restriction.py
@@ -7,9 +7,10 @@
 
 from Bio.Restriction import Analysis, Restriction, RestrictionBatch
 from Bio.Restriction import CommOnly, NonComm, AllEnzymes
-from Bio.Restriction import (Acc65I, Asp718I, EcoRI, EcoRV, KpnI, SmaI,
-                             MluCI, McrI, SacI, AanI)
-from Bio.Seq import Seq
+from Bio.Restriction import (Acc65I, Asp718I, BamHI, EcoRI, EcoRV, KpnI, SmaI,
+                             MluCI, McrI, NdeI, BsmBI, AanI, EarI, SnaI)
+from Bio.Restriction import FormattedSeq
+from Bio.Seq import Seq, MutableSeq
 from Bio.Alphabet.IUPAC import IUPACAmbiguousDNA
 from Bio import BiopythonWarning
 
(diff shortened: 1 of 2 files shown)


MarkusPiotrowski:
Anyone an idea what's the problem in pypy3?
```bash
ERROR: test_change (test_Restriction.RestrictionBatchPrintTest)

Test that change() changes something.

  File "/home/travis/build/biopython/biopython/Tests/test_Restriction.py", line 389, in test_change

    analysis.change(rb=batch)

  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.5/Bio/Restriction/Restriction.py", line 2352, in change

    raise AttributeError('Analysis has no attribute %s' % k)

AttributeError: Analysis has no attribute rb
```
The test runs fine in all other versions (Python 2.7, Python 3.x, pypy).

peterjc:
My guess is this is a subtly in object equality and string optimisation (it is a Python implementation detail that two strings "rb" and "rb" are often but not always the same object in memory):

```
$ python
Python 2.7.10 (default, Feb  7 2017, 00:08:15) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
>>> id("rd")
4352818632
>>> id("rd")
4352818592
>>> id("rd")
4352818552
>>> a = "rd"
>>> id(a)
4352818472
>>> id("rd")
4352818472
>>> quit()
```

I think this is being triggered by the strange use of ``is`` rather than ``==`` here:

```python
            elif k is 'rb':
                self = Analysis.__init__(self, v, self.sequence, self.linear)
            elif k is 'linear':
                setattr(self, 'linear', v)
                self.search(self.sequence, v)
```

https://github.com/MarkusPiotrowski/biopython/blob/8471fa88a59e034aa3d745ddc6849b7a6832a887/Bio/Restriction/Restriction.py#L2341

MarkusPiotrowski:
@peterjc Thanks! That was the problem. I replaced the ```is``` with ```==``` and now it passed.

Tests/test_Restriction.py
```diff
@@ -303,8 +519,75 @@ def test_analysis_restrictions(self):
...
+                         "RestrictionBatch(['Asp718I', 'EcoRI', 'KpnI', "
+                         "'SmaI', 'SnaI'])")
 
 
 if __name__ == "__main__":
     runner = unittest.TextTestRunner(verbosity=2)
     unittest.main(testRunner=runner)
+    batch = EcoRI + SmaI + KpnI
+    my_seq = Seq('GAATTCCCGGGATGCGTA')
+    analysis = Analysis(batch, my_seq)
```
peterjc: Can you remove this little self test? It would not be used via ``run_tests.py``?

peterjc:
Great - other than the tiny ``__main__`` change, this looks good to me 👍

Tests/test_Restriction.py
MarkusPiotrowski: Upps, obviously the result of misplaced drag'n'drop...

peterjc:
Assuming TravisCI is happy, I think can be merged.

Is it worth adding anything to the ``NEWS.rst`` file for the overall updates to the restriction code since the last release?

PR #1263, opened 2017-05-29
add docstring to SeqFeature.py

Gasta88:
As mentioned in issue #1203, I've added public docstring to `Bio.SeqFeature` module.

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -164,9 +164,11 @@ class SeqFeature(object):
             self.ref_db = ref_db
 
     def _get_strand(self):
+        """Get function for the strand property (PRIVATE)."""
         return self.location.strand
 
     def _set_strand(self, value):
+        """Set function for the strand property (PRIVATE)."""
         try:
             self.location.strand = value
         except AttributeError:
@@ -183,12 +185,14 @@ class SeqFeature(object):
                           """)
 
     def _get_ref(self):
+        """Get function for the reference property (PRIVATE)."""
         try:
             return self.location.ref
         except AttributeError:
             return None
 
     def _set_ref(self, value):
+        """Set function for the reference property (PRIVATE)."""
         try:
             self.location.ref = value
         except AttributeError:
@@ -204,12 +208,14 @@ class SeqFeature(object):
                        """)
 
     def _get_ref_db(self):
+        """Get function for the database reference property (PRIVATE)."""
         try:
             return self.location.ref_db
         except AttributeError:
             return None
 
     def _set_ref_db(self, value):
+        """Set function for the database reference property (PRIVATE)."""
         self.location.ref_db = value
     ref_db = property(fget=_get_ref_db, fset=_set_ref_db,
                       doc="""Feature location reference's database.
@@ -218,12 +224,14 @@ class SeqFeature(object):
                           """)
 
     def _get_location_operator(self):
+        """Get function for the location operator property (PRIVATE)."""
         try:
             return self.location.operator
         except AttributeError:
             return None
 
     def _set_location_operator(self, value):
+        """Set function for the location operator property."""
         if value:
             if isinstance(self.location, CompoundLocation):
                 self.location.operator = value
@@ -237,7 +245,7 @@ class SeqFeature(object):
                                  doc="Location operator for compound locations (e.g. join).")
 
     def __repr__(self):
-        """A string representation of the record for debugging."""
+        """Represent the feature as a string for debugging."""
         answer = "%s(%s" % (self.__class__.__name__, repr(self.location))
         if self.type:
             answer += ", type=%s" % repr(self.type)
@@ -253,7 +261,7 @@ class SeqFeature(object):
         return answer
 
     def __str__(self):
-        """A readable summary of the feature intended to be printed to screen."""
+        """Return the full feature as a python string."""
         out = "type: %s\n" % self.type
         out += "location: %s\n" % self.location
         if self.id and self.id != "<unknown id>":
@@ -265,7 +273,7 @@ class SeqFeature(object):
         return out
 
     def _shift(self, offset):
-        """Returns a copy of the feature with its location shifted (PRIVATE).
+        """Return a copy of the feature with its location shifted (PRIVATE).
 
         The annotation qaulifiers are copied.
         """
@@ -276,7 +284,7 @@ class SeqFeature(object):
                           qualifiers=OrderedDict(self.qualifiers.items()))
 
     def _flip(self, length):
-        """Returns a copy of the feature with its location flipped (PRIVATE).
+        """Return a copy of the feature with its location flipped (PRIVATE).
 
         The argument length gives the length of the parent sequence. For
         example a location 0..20 (+1 strand) with parent length 30 becomes
@@ -292,7 +300,7 @@ class SeqFeature(object):
                           qualifiers=OrderedDict(self.qualifiers.items()))
 
     def extract(self, parent_sequence):
-        """Extract feature sequence from the supplied parent sequence.
+        """Extract the feature's sequence from supplied parent sequence.
 
         The parent_sequence can be a Seq like object or a string, and will
         generally return an object of the same type. The exception to this is
@@ -349,7 +357,7 @@ class SeqFeature(object):
     __nonzero__ = __bool__
 
     def __len__(self):
-        """Returns the length of the region described by a feature.
+        """Return the length of the region where the feature is located.
 
         >>> from Bio.Seq import Seq
         >>> from Bio.Alphabet import generic_protein
@@ -478,6 +486,7 @@ class Reference(object):
     """
 
     def __init__(self):
+        """Initialize the class."""
         self.location = []
         self.authors = ''
         self.consrtm = ''
@@ -488,7 +497,7 @@ class Reference(object):
         self.comment = ''
 
     def __str__(self):
-        """Output an informative string for debugging."""
+        """Return the full Reference object as a python string."""
         out = ""
         for single_location in self.location:
             out += "location: %s\n" % single_location
@@ -503,12 +512,13 @@ class Reference(object):
         return out
 
     def __repr__(self):
+        """Represent the Reference object as a string for debugging."""
         # TODO - Update this is __init__ later accpets values
         return "%s(title=%s, ...)" % (self.__class__.__name__,
                                       repr(self.title))
 
     def __eq__(self, other):
-        """Check if two Reference objects should be considered equal
+        """Check if two Reference objects should be considered equal.
 
         Note that the location is not compared, as __eq__ for the
         FeatureLocation class is not defined.
@@ -589,7 +599,7 @@ class FeatureLocation(object):
     """
 
     def __init__(self, start, end, strand=None, ref=None, ref_db=None):
-        """Specify the start, end, strand etc of a sequence feature.
+        """Initialize the class.
 
         start and end arguments specify the values where the feature begins
         and ends. These can either by any of the ``*Position`` objects that
@@ -661,9 +671,11 @@ class FeatureLocation(object):
         self.ref_db = ref_db
 
     def _get_strand(self):
+        """Get function for the strand property (PRIVATE)."""
         return self._strand
 
     def _set_strand(self, value):
+        """Set function for the strand property (PRIVATE)."""
         if value not in [+1, -1, 0, None]:
             raise ValueError("Strand should be +1, -1, 0 or None, not %r"
                              % value)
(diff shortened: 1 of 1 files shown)


Bio/SeqFeature.py
```diff
@@ -164,9 +164,11 @@ def __init__(self, location=None, type='', location_operator='',
             self.ref_db = ref_db
 
     def _get_strand(self):
+        """Return the strand where the sequence feature is located (PRIVATE)."""
```
peterjc: Here ``_get_strand`` and ``_set_strand`` are used via the ``property`` line later to define the ``strand`` property which does have a docstring. So in a sense they are documented...

Do you agree something like ``Get function for the strand property (PRIVATE).`` is better? The key point I would want to get across is the link to the ``.strand`` property.

(This same point applies to the other get/set property methods)

Bio/SeqFeature.py
```diff
@@ -292,7 +300,7 @@ def _flip(self, length):
                           qualifiers=OrderedDict(self.qualifiers.items()))
 
     def extract(self, parent_sequence):
-        """Extract feature sequence from the supplied parent sequence.
+        """Extract the sequence feature from the supplied parent sequence.
```
peterjc: Rather than ``sequence feature``, I would keep ``feature sequence`` or perhaps use ``feature's sequence``. The key point is this method returns a sequence (not a feature object).

Bio/SeqFeature.py
```diff
@@ -870,7 +883,7 @@ def _flip(self, length):
 
     @property
     def parts(self):
-        """Read only list of parts (always one, the Feature Location).
+        """Read only listof sections (always one, the FeatureLocation object).
```
peterjc: missing space in ``listof``.

Bio/SeqFeature.py
```diff
@@ -927,7 +940,7 @@ def nofuzzy_end(self):
             raise
 
     def extract(self, parent_sequence):
-        """Extract feature sequence from the supplied parent sequence."""
+        """Extract the feature location from the supplied parent sequence."""
```
peterjc: See comment earlier about the other ``.extract`` method.

Bio/SeqFeature.py
```diff
@@ -1153,20 +1168,22 @@ def __nonzero__(self):
...
 
     def __iter__(self):
+        """Iterate over the parent positions within the CompoundLocation object."""
         for loc in self.parts:
             for pos in loc:
                 yield pos
 
     def _shift(self, offset):
-        """Returns a copy of the location shifted by the offset (PRIVATE)."""
+        """Return a copy of the locations shifted by an offset (PRIVATE)."""
```
peterjc: CompoundLocation to match some of your other changes?

Bio/SeqFeature.py
```diff
@@ -1359,30 +1376,33 @@ class ExactPosition(int, AbstractPosition):
...
 
     def __repr__(self):
-        """String representation of the ExactPosition location for debugging."""
+        """Represent the ExactPosition object as a string for debugging."""
         return "%s(%i)" % (self.__class__.__name__, int(self))
 
     @property
     def position(self):
-        """Legacy attribute to get position as integer (OBSOLETE)."""
+        """Legacy attribute to get location as integer (OBSOLETE)."""
```
peterjc: I think this should be position (consistent with how position and location are used here and in the Tutorial).

Bio/SeqFeature.py
```diff
@@ -1537,15 +1562,17 @@ def position(self):
 
     @property
     def extension(self):
-        """Legacy attribute to get extension (from left to right) as an integer (OBSOLETE)."""
+        """Legacy attribute to get the dimension (from left to right) as an integer (OBSOLETE)."""
```
peterjc: We've not used dimension elsewhere, I would stick with extension here.

Bio/SeqFeature.py
```diff
@@ -1537,15 +1562,17 @@ def position(self):
...
         return self._right - self._left
 
     def _shift(self, offset):
+        """Legacy attribute to get extension (zero) as integer (PRIVATE)."""
```
peterjc: Copy and paste error? See the other ``_shift`` methods where you had a sensible docstring, e.g.

``Return a copy of the position object with its location shifted (PRIVATE).``

Bio/SeqFeature.py
```diff
@@ -1640,15 +1669,17 @@ def position(self):
 
     @property
     def extension(self):
-        """Legacy attribute to get extension (from left to right) as an integer (OBSOLETE)."""
+        """Legacy attribute to get the dimension (from left to right) as an integer (OBSOLETE)."""
```
peterjc: Again, I would avoid introducing dimension as a new term.

Bio/SeqFeature.py
```diff
@@ -1640,15 +1669,17 @@ def position(self):
...
         return self._right - self._left
 
     def _shift(self, offset):
+        """Legacy attribute to get extension (zero) as integer (PRIVATE)."""
```
peterjc: Copy and paste error? See the other ``_shift`` methods where you had a sensible docstring, e.g.

``Return a copy of the position object with its location shifted (PRIVATE).``

Bio/SeqFeature.py
```diff
@@ -1701,20 +1733,23 @@ def position(self):
 
     @property
     def extension(self):
-        """Legacy attribute to get extension (zero) as integer (OBSOLETE)."""
+        """Legacy attribute to get zero as integer (OBSOLETE)."""
```
peterjc: I think the older version was clearer in this case.

Bio/SeqFeature.py
```diff
@@ -1772,20 +1808,23 @@ def position(self):
...
```
peterjc: Again, keep the old ``.extension`` docstring here.

Bio/SeqFeature.py
```diff
@@ -1772,20 +1808,23 @@ def position(self):
...
-        """A string representation of the location for debugging."""
+        """Represent the location as a string for debugging."""
         return "%s(%i)" % (self.__class__.__name__, int(self))
 
     def __str__(self):
+        """Return a representation of the AfterPosition object (with python counting)."""
         return ">%s" % self.position
 
     def _shift(self, offset):
+        """Legacy attribute to get extension (zero) as integer (PRIVATE)."""
```
peterjc: Copy and paste error? See the other ``_shift`` methods where you had a sensible docstring, e.g.

``Return a copy of the position object with its location shifted (PRIVATE).``

Bio/SeqFeature.py
```diff
@@ -1701,20 +1733,23 @@ def position(self):
...
-        """A string representation of the location for debugging."""
+        """Represent the locationas as a string for debugging."""
         return "%s(%i)" % (self.__class__.__name__, int(self))
 
     def __str__(self):
+        """Return a representation of the BeforePosition object (with python counting)."""
         return "<%s" % self.position
 
     def _shift(self, offset):
+        """Legacy attribute to get extension (zero) as integer (PRIVATE)."""
```
peterjc: Copy and paste error? See the other ``_shift`` methods where you had a sensible docstring, e.g.

``Return a copy of the position object with its location shifted (PRIVATE).``

Bio/SeqFeature.py
```diff
@@ -1879,10 +1919,12 @@ def __str__(self):
         return out
 
     def _shift(self, offset):
+        """Legacy attribute to get extension (zero) as integer (PRIVATE)."""
```
peterjc: Copy and paste error? See the other ``_shift`` methods where you had a sensible docstring, e.g.

``Return a copy of the position object with its location shifted (PRIVATE).``

Bio/SeqFeature.py
Gasta88: I can't due to flake8. It doesn't like when the signature is used inside the summary docstring.

Bio/SeqFeature.py
Gasta88: Sorry, I can't find the other comment about `.extract` method.

Bio/SeqFeature.py
Gasta88: Sorry, can you elaborate more on this comment?

Bio/SeqFeature.py
peterjc: We're not actually enforcing that bit of flake8's checks yet are we? Right now the automated tests should be ignoring ``D402	First line should not be the function’s "signature"``

https://github.com/biopython/biopython/blob/master/Bio/.flake8

(This may be a false positive worth reporting to ``flake8``)

Bio/SeqFeature.py
peterjc: I was wondering about using:

```
"""Return a copy of the CompoundLocation shifted by an offset (PRIVATE)."""
```

peterjc:
The key point is the ``.extract`` method uses the location information to return a sequence (of letters), not a location. How about this for the ``SeqFeature`` object's ``.extract`` method:

```
"""Extract feature's sequence from supplied parent sequence.

Uses the feature's location to pull out the region described from
the given parent sequence, applying compound operations like join,
or any reverse complements, if required.

The given sequence can be a string or a Biopython sequence object.
"""
```

(From memory the parent sequence can be a string or a Seq-like object)

Maybe it needs a doctest as an example?

Bio/SeqFeature.py
```diff
@@ -237,7 +245,7 @@ def _set_location_operator(self, value):
                                  doc="Location operator for compound locations (e.g. join).")
 
     def __repr__(self):
-        """A string representation of the record for debugging."""
+        """Represent the feature's sequence as a string for debugging."""
```
peterjc: No, it does not give the feature's sequence as a string (which might be ``ACGT...``). In the original ``record`` was misleading, would ``feature`` be better?

Bio/SeqFeature.py
```diff
@@ -253,7 +261,7 @@ def __repr__(self):
         return answer
 
     def __str__(self):
-        """A readable summary of the feature intended to be printed to screen."""
+        """Return the full feature's sequence as a python string."""
```
peterjc: No, it does not give the feature's sequence as a string (which might be ``ACGT...``). Was the original unclear?

Bio/SeqFeature.py
```diff
@@ -1705,16 +1737,19 @@ def extension(self):
         return 0
 
     def __repr__(self):
-        """A string representation of the location for debugging."""
+        """Represent the locationas as a string for debugging."""
```
peterjc: Typo, ``locationas`` --> ``location``

Bio/SeqFeature.py
```diff
@@ -1760,6 +1795,7 @@ class AfterPosition(int, AbstractPosition):
 
     # Subclasses int so can't use __init__
     def __new__(cls, position, extension=0):
+        """Create a new instance in AfterPosition object."""
```
peterjc: Odd phrasing, how about: ``Create a new instance of the AfterPosition object.`` here?

Bio/SeqFeature.py
Gasta88: I wasn't imperative, what about `Return a summary of the feature intended to be printed to screen.`?

peterjc: That's good.

Gasta88:
Apologies, I'm getting confused a bit with the `.extract` method.

I've edited the one for the `SeqFeature` object, and I've also changed the one inside `FeatureLocation` and `CompoundLocation` object.

I've also provided doctest examples to the new ones.

Bio/SeqFeature.py
```diff
@@ -927,7 +940,21 @@ def nofuzzy_end(self):
...
+
+        The parent_sequence can be a Seq like object or a string, and will
+        generally return an object of the same type. The exception to this is
+        a MutableSeq as the parent sequence will return a Seq object.
+
+        >>> from Bio.Seq import Seq
+        >>> from Bio.Alphabet import generic_protein
+        >>> from Bio.SeqFeature import FeatureLocation
+        >>> seq = Seq("MKQHKAMIVALIVICITAVVAAL", generic_protein)
+        >>> fL = FeatureLocation(8, 15)
```
peterjc: That should be ``fl`` (lower case) to match style guidelines and the line below. This is why TravisCI is failing on the doctests.

peterjc:
Other than the doctest, this looks ready to merge. Good job 👍

Gasta88:
Thanks for fixing the doctest, apology but I was off on holiday for the past two weeks.

How should we proceed on this PR?

peterjc:
I would have merged it already but waited for the tests, thanks for the reminder...

peterjc:
I squashed this down into one commit (all under your name) to keep the git history reasonably clear - thank you!

PR #1264, opened 2017-06-01
Updated fragment library link to new URL.

JoaoRodrigues:
Referring to #1191 , updated url. Thanks for pointing this out @peterjc

### DIFF
--- a/Bio/PDB/FragmentMapper.py
+++ b/Bio/PDB/FragmentMapper.py
@@ -18,7 +18,7 @@ J Mol Biol. 2002 323(2):297-307.
 
 The definition files of the fragments can be obtained from:
 
-U{http://github.com/csblab/fragments/}
+http://github.com/csblab/fragments/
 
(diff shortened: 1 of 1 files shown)


Bio/PDB/FragmentMapper.py
```diff
@@ -18,7 +18,7 @@
 
 The definition files of the fragments can be obtained from:
 
-U{http://csb.stanford.edu/~rachel/fragments/}
+U{http://github.com/csblab/fragments/}
```
peterjc: That weird ``U{...}`` markup is not valid RST, and we should fix that while you're working on this. The simplest fix is to make it a plain URL.

I find http://rst.ninjs.org/ useful for RST checking.

JoaoRodrigues: Never worked with this before. Thanks for the website. Changed it.

peterjc:
Cheers.

PR #1269, opened 2017-06-06
Add "testseq.py" to Scripts

Adil-Iqbal:
Add "testseq" function.

### DIFF
--- a/Bio/GA/__init__.py
+++ b/Bio/GA/__init__.py
@@ -1,2 +1,14 @@
-# make files in this directory importable
-"""A selection of genetic algorithm code."""
+# Copyright 2001 by Brad Chapman.  All rights reserved.
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
+"""Genetic Algorithm library (DEPRECATED)."""
+
+import warnings
+from Bio import BiopythonDeprecationWarning
+warnings.warn("Bio.GA has been deprecated, and we intend to remove"
+              " it in a future release of Biopython. Please consider using"
+              " DEAP instead.  If you would like to"
+              " continue using Bio.SomeModule, please contact the Biopython"
+              " developers via the mailing list or GitHub.",
+              BiopythonDeprecationWarning)
--- a/DEPRECATED.rst
+++ b/DEPRECATED.rst
@@ -30,8 +30,23 @@ installation simply aborts with a error message.
 
 Python 3.3
 ==========
-Still suported but deprecated as of Release 1.67, triggering a warning
-at installation time.
+No longer supported as of Release 1.70, having triggered a warning with
+release 1.67 onwards.
+
+Jython
+======
+Biopython is mostly working under Jython 2.7.0, but support for Jython
+is deprecated as of Release 1.70.
+
+Bio.GA
+======
+This was deprecated in Biopython 1.70. Please consider using a dedicated
+genetic algorithm library like DEAP instead.
+
+Bio.NeuralNetwork
+=================
+This was deprecated in Biopython 1.70. Please consider using a dedicated
+machine learning library like scikit-learn or TensorFlow instead.
 
 Bio.Phylo.CDAOIO.CDAOError
 ==========================
--- a/Tests/test_GACrossover.py
+++ b/Tests/test_GACrossover.py
@@ -9,13 +9,17 @@ from Bio.Seq import MutableSeq
 from Bio.Alphabet import SingleLetterAlphabet
 
 # local stuff
-from Bio.GA.Organism import Organism
-from Bio.GA.Crossover.General import SafeFitnessCrossover
-from Bio.GA.Crossover.GeneralPoint import GeneralPointCrossover
-from Bio.GA.Crossover.GeneralPoint import InterleaveCrossover
-from Bio.GA.Crossover.TwoPoint import TwoPointCrossover
-from Bio.GA.Crossover.Point import SinglePointCrossover
-from Bio.GA.Crossover.Uniform import UniformCrossover
+import warnings
+from Bio import BiopythonDeprecationWarning
+with warnings.catch_warnings():
+    warnings.simplefilter('ignore', BiopythonDeprecationWarning)
+    from Bio.GA.Organism import Organism
+    from Bio.GA.Crossover.General import SafeFitnessCrossover
+    from Bio.GA.Crossover.GeneralPoint import GeneralPointCrossover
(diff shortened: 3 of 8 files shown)


peterjc:
Rather than assuming ``/`` as a path separator for splitting (which will break on Windows), use ``os.path.split(...)`` and ``os.path.join(...)`` for the path manipulations. Or just assume that it will be run from the ``Tests/`` folder and hard-code the relative path ``../Scripts/`` (which should work on Windows which will accept the other slashes as input)?

andrewguy:
I'm still concerned that the random number generation is a little unconventional, and more confusing than it needs to be.

I understand your concern that the code needs to be intuitive for those unfamiliar with coding in general, but an equally valid concern here is code readability for anyone who reads the source code. Random number generation should be straightforward, and there are fairly well established ways of doing it.

As written, you're seeding a random number generator with the output from _another_ random number generator, every time you make a call to `random()`.

Can we maybe get another set of eyes on this? This is what I imagine it should look like, but happy for people to disagree - https://github.com/Adil-Iqbal/Personal-Projects/pull/1/files

Adil-Iqbal:
That's a very good point. 

No sense trying to reinvent the wheel. Besides, if this code is to be used in an educational capacity, no sense trying to reinforce concepts that would have to be unlearned.   

I'll see about reverting the seeding method.

Adil-Iqbal:
The following was copy/pasted from the mailing list:

**TL:DR - Game plan: (1) Resolve unittest importing issue. (2) Revert testseq seeding. (3) Revisit testseq doctests. (4) Update unittest for reverted seeding. (2) Thank you for your patience.**

Hello all, 

I'd like to let everyone know where 'testseq' is in development. You can find the code for it here: https://github.com/Adil-Iqbal/Personal-Projects/tree/master/Test%20Sequence

The biggest issue (unexpectedly) was importing the 'testseq' module into the 'test_testseq' unit test. Absolute and relative importing did not work since the 'Scripts' folder is not a part of the 'Bio' package. After trying out several solutions, the one that looks the most promising is using the "importlib" module. The issue is not yet resolved but I am confident that it is just a matter of working out the details.

I've gotten the chance to tackle part of another big issue, which is python version and operating system compatibility. Since the 'testseq' function uses randomization, its output can be different across versions and platforms. This poses a problem for code testing. Luckily there are a few work-arounds that I've implemented in the unit test which should work fine. I will also be revisiting the 'testseq' doctests to ensure they are also compatible, though I have not yet run into any issues with that on Travis CI. (Peter warned me about this. ^_^)

Today, an argument was presented for reverting the seeding implementation, which was that user intuition should preferably be developed with respect to best practices. Unless anyone has a reasonable rebuttal, I will probably revert the seeding to an earlier state. Though, that will have to wait until after I've resolved the importing issue.

Also, I'd like to thank you for being patient with me. I would've completed this sooner, but I have other obligations as well. 

Warmest regards,
Adil

Tests/run_tests.py
```diff
@@ -173,7 +170,6 @@ def is_numpy():
 # Skip Bio.bgzf doctest for broken gzip, see http://bugs.python.org/issue17666
 def _have_bug17666():
     """Debug function to check if Python's gzip is broken (PRIVATE).
-
```
peterjc: Please restore this blank line which is a deliberate part of the docstring for private function ``_have_bug17666`` - see https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

peterjc:
Travis is failing on the style checks before ever running the tests:

```
$ flake8 Tests/
Tests/run_tests.py:1:1: D205 1 blank line required between summary line and description
Tests/run_tests.py:171:1: D205 1 blank line required between summary line and description
Tests/run_tests.py:323:1: D205 1 blank line required between summary line and description
```

You can install flake8 locally, see https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

andrewguy:
Your latest build is failing with Python 2.7 because you are doing division, and haven't added `from __future__ import division`. Python 2.7 division defaults to floor division, whereas Python 3+ is floating point division.

I would really recommend testing what you can on your local machine before pushing changes to github.

You can use either pip's virtualenv, or conda's env to set up self-contained distributions for testing (if you haven't already done so). You can set up a Python 2.7 environment this way, and make sure your build passes on both 2.7 and 3+. 

Peter's suggestion of running Flake 8 locally is also a very good one.

Adil-Iqbal:
Thanks a lot guys. I really appreciate it. I've got flake8 up and running locally. I'll try to get virtualenv going too.

Tests/run_tests.py
```diff
@@ -7,9 +7,7 @@
 This will find all modules whose name is "test_*.py" in the test
 directory, and run them.  Various command line options provide
 additional facilities.
-
```
peterjc: You have accidentally remove this (and two more) blank line from the module's docstring.

peterjc:
I wonder if the complicated import tricks you are trying in order to be able to do ``from Scripts import testseq`` or similar could be avoided by including ``Scripts/`` in ``sys.path`` in the same way that ``run_tests.py`` ensures our complied C modules can be imported from the ``build/`` subdirectory?

Tests/run_tests.py
Adil-Iqbal: Reverted. Will include changes in next commit.

Adil-Iqbal:
That method will certainly succeed in importing the module. The issue is that once the importing is complete, the doctests are run. And the doctest *also* contain an import statement. That import statement is subject to the same importing limitations which I will describe below:

Prior to 3.3, heirarchical importing (importing with "dot" notation) required a directory to be initialized with a "__init__.py" file. ([Source](https://docs.python.org/2/tutorial/modules.html#packages)) This created a bunch of confusion, so after 3.3, heirarchical importing was generalized to all directories, regardless of the directory being initialized. 

Since we can't add a "__init__.py" file to folders carelessly, its best to remove non-Bio modules from the "run_tests.py" file prior to version 3.3.  That's why the most recent build passed the Travis CI test. Thankfully, the unit test for 'testseq' runs all of the same tests that are in the doctests, so we can be confident that 'testseq' is compatible with all versions of python. 

All versions of python after 3.3 don't require an extra "__init__.py" file, so the doctests will run fine on 3.3, 3.4, 3.5, 3.6, and PyPy3.

peterjc:
I'm pretty sure you can remove these lines from ``testseq.py`` and the doctest will still work as the classes etc are defined in the current file so already present in the local scope:

```python
>>> from Scripts.testseq import testseq
```

Adil-Iqbal:
Closing pull request.

PR #722, opened 2015-12-09
Account for the 4-letter KEGG organism codes.

allista:
Otherwise these were getting into the previouse gene record as gene idetifiers.
Example: 
[http://rest.kegg.jp/get/ec:5.4.2.2](http://rest.kegg.jp/get/ec:5.4.2.2)

> BACU: 103004993(PGM1) 103005357(PGM2)

Issue #888, opened 2016-07-22
Handling of bootstrap support values in Bio.Phylo Newick parser


twrightsman:
Migrated from https://redmine.open-bio.org/issues/3407

@bendmorris said:

> This was reported to me by Arlin Stoltzfus (quote):
>> There is a description of Newick here:
>> 
>> http://evolution.genetics.washington.edu/phylip/newicktree.html
>> and a BNF here:
>> 
>> http://evolution.genetics.washington.edu/phylip/newick_doc.html
>> Note that this allows square-bracketed comments.
>> 
>> Bootstrap values commonly are represented in 2 ways, one of which is wrong. The wrong way to represent bootstrap values is to present them as internal node labels. Labels for internal nodes are given as follows:
>> 
>> ```
>> ((( human: 0.1, chimp:0.1 ) primates: 0.2, (rat:0.1, mouse:0.1) rodents:0.2), cat:0.3 )
>> ```
>> where "primates" and "rodents" are internal node labels. They go between the right paren and the (optional) colon and distance. If you put numbers in the label position, a graphic renderer may place them on the nodes, which is why some people represent bootstrap values this way.
>> 
>> However, the preferred way to represent bootstrap values is to make them syntactic comments (enclosed in square brackets) placed after all other node information, i.e., after the optional colon & branch length. Both examples are shown here:
>> 
>> ```
>> ((raccoon:19.19959,bear:6.80041):0.84600[50],((sea_lion:11.99700, seal:12.00300):7.52973[100],
>> ((monkey:100.85930,cat:47.14069):20.59201[80], weasel:18.87953):2.09460[75]):3.87382[50],dog:25.46154);
>> ```
>> 
>> I recommend that you only support the second version, and treat the first version as a case of internal node labels.
>> 
>> Arlin
>> -------
>> Arlin Stoltzfus (arlin@umd.edu)
>> Fellow, IBBR; Adj. Assoc. Prof., UMCP; Research Biologist, NIST
>> IBBR, 9600 Gudelsky Drive, Rockville, MD, 20850
>> tel: 240 314 6208; web: www.molevol.org"

twrightsman:
@peterjc replied:

> Where does the statement "the preferred way to represent bootstrap values is to make them syntactic comments (enclosed in square brackets) placed after all other node information" come from?
> 
[...]

PR #1276, opened 2017-06-10
Fixed newick format writer: node labels should not be separated by colon.

allista:
As per @peterjc [request](https://github.com/biopython/biopython/pull/722#issuecomment-299436002), opening another PR for newick fix.

**Additional info:** 
In newick, confidence is stored as a label of an internal node.
But:

> For example, interior nodes can have names in that standard. These names follow the right parenthesis for that interior node, as in this example:
> (B:6.0,(A:5.0,C:3.0,E:4.0)Ancestor1:5.0,D:11.0);

[from here](http://evolution.genetics.washington.edu/phylip/newicktree.html)

So no need for a colon before a label or a confidence value.

### DIFF
--- a/Bio/Phylo/NewickIO.py
+++ b/Bio/Phylo/NewickIO.py
@@ -333,7 +333,7 @@ class Writer(object):
                     return (':' + format_branch_length
                             ) % (clade.branch_length or 0.0) + _get_comment(clade)
                 else:
-                    return (':' + format_confidence + ':' + format_branch_length
+                    return (format_confidence + ':' + format_branch_length
                             ) % (clade.confidence, clade.branch_length or 0.0) + _get_comment(clade)
 
         return make_info_string


peterjc:
Good job - this is what I was hoping for, a nice clean pull request with this once change. Thanks!

I'm going to ask someone else for a review, but can you think of a test case which could be added for this?

allista:
Sorry for my ignorance ^_^ but what is a test case in this context?

I can provide a tree file formatted as described.
Or you can take any tree generated by, e.g. MrBayes or fasttree. I suppose most ML tools will do too.

peterjc:
The test itself could be done as a doctest, but a unittest method would probably be simpler. There is a chapter in our Tutorial about our test framework.

I'd like some bit of code which calls this changed code in ``_info_factory``, so that we can confirm the new version has the colons in the right place. This could be as simple as producing a string from a simple tree like your initial question like ``(B:6.0,(A:5.0,C:3.0,E:4.0)Ancestor1:5.0,D:11.0);``

allista:
Will do!

etal:
I think we had the colon there so that the Newick parser knows that the strings following the internal node represent support values, with an empty string as the internal node label. We may have gone back and forth on this.

etal:
Per #1315 and #888, it sounds like we should merge this pull request.

cymon:
Hi Folks,

Not sure if this will help, but my understanding is that the value after
the colon at a node is typically the branch length, anything before the
colon is a label. The label can be any information: bootstrap value, node
age, clade name etc... The way bootstrap values are typically handled is to
make them into labels, if that is what you want to do - if you don't want
to label your nodes with bootstrap values then information should be stored
in brackets after the branch length as a comment (I'm not sure how many
nexus parsers in software handle comments correctly - few I would guess).

I'd suggest doing what PAUP does and have an option to write the bootstrap
values as labels when writing the tree:

SaveBootP    No|Brlens|NodeLabels|Both            Brlens

Note that here the default is to write the bootstrap values in place of the
branch lengths rather than as labels - not sure why.

Cheers, C.

peterjc:
Is there an obvious way to make this an option in our NEXUS and Newick writers?

peterjc:
@etal do you want to merge this to deal with #888 and #1315?

etal:
Yes, and let's default to the way Cymon suggests.

peterjc:
So you'd like a change with an option along the lines of of @cymon's comment last summer?

etal:
Is appveyor stalled for an unexpected reason? That's the only check we're waiting on.

etal:
Yes, I'll defer to anyone who is actively using these other programs. I'm not familiar with the current conventions for Newick and Nexus.

peterjc:
I was assuming there is no AppVeyor report (and nothing to click on to restart a stalled build) because this pull request pre-dates our AppVeyor setup. We can in principle go ahead and rebase-and-merge anyway (or at least some of us can, that may be restricted to those of us with repository admin rights?).

If you want me to do the merge, I can do that.

I'd do this now, but I'm a little unclear if any further tweaking was required as per Cymon's suggestion, or if it is fine as is.

etal:
Since the tests passed on Travis and there's nothing platform-specific here, I'm happy for you to merge this as it is. If another user can identify a need to make this behavior optional, I'd prefer to let them provide the code and test case, since I'm not in a position to do so myself.

peterjc:
OK, merging - thank you everyone. Let's see if further polish is needed, and if the associated issues can be closed now.

@allista do you want to be acknowledged by name in the ``NEWS.rst`` file for the next release? You are already in the ``CONTRIB.rst`` file.

allista:
Thanks! Considering the size of the fix I would say that CONTRIB.rst is more than enough.

peterjc:
OK, thank you! On a related point, if you are happy to dual license your contributions, could you agree over on #898 please.

allista:
Done.

peterjc:
Thank you

PR #1280, opened 2017-06-12
Note lack of support for minimal MEME format

j-andrews7:
I agree to my contributions being dual licensed under both our original "Biopython License Agreement" and the more widely used "3-Clause BSD License".

Doubt that statement's necessary for a doc change, but whatever.

### DIFF
--- a/Bio/motifs/meme.py
+++ b/Bio/motifs/meme.py
@@ -145,7 +145,9 @@ def __read_datafile(record, handle):
         if line.startswith('TRAINING SET'):
             break
     else:
-        raise ValueError("Unexpected end of stream: 'TRAINING SET' not found.")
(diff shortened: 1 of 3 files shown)


peterjc:
How about making the exception message more explicit instead? e.g. Adding ``This can happen with minimal MEME files from XXX which are not supported yet.``

j-andrews7:
@peterjc Probably a much better idea.

peterjc:
Thanks for agreeing to the dual-licensing. Are you happy to be thanked by name in the ``NEWS.rst`` and ``CONTRIB.rst`` files?

j-andrews7:
Sure.

peterjc:
Thank you!

PR #1283, opened 2017-06-14
Change Enzyme __str__ to use new-lines not spaces.

peterjc:
By using new-lines, the string output closely resembles the raw file format.

Includes test for the modified __str__ method.

(I spotted this will looking at gaps in our test coverage)


cbrueffer:
Looks fine to me.

peterjc:
Thank you for looking at this.

PR #1284, opened 2017-06-14
Do not append additional periods to GenBank DEFINITION field

biologyguy:
Currently, if converting from FASTA format where a period already exists
 in the def line, an additional period is appended.

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -834,7 +834,7 @@ class GenBankWriter(_InsdcWriter):
         # see ftp://ftp.ncbi.nih.gov/genbank/gbrel.txt [3.4.5]
         # and discussion https://github.com/biopython/biopython/pull/616
         # So let's add a period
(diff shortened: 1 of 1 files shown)


peterjc:
Rather than:

```python
descr += '.' if descr and descr[-1] != "." else ""
```

how about:

```python
if not descr.endswith("."):
    descr += '.'
```

In any case, some might argue this is working as intended? Did you read the discussion on https://github.com/biopython/biopython/pull/616 yet?

biologyguy:
Hi Peter,
Ya, you're solution would be better for sure. Having just read through your discussion with Bertrand, however, I see you both spent some time really thinking about this; I'll kill my pull request. The change blew up unit tests in my own software, but I'll certainly conform to the will of the people :)

peterjc:
Just because you're the first person to report being "bitten" by the change doesn't mean we made the best possible choice. Personally I've rarely noticed trailing dots in FASTA files. If you have real world examples - for example a public database using this - that would sway me. Thanks!

PR #1291, opened 2017-06-22
Python 3 compatible sorting of atoms in PDB Residues

### LINKED ISSUE #1189: Bio.PDB sorting not tested and Python 2 only
The ``cmp`` function is Python 2 only, sorting using a key function is the preferred idiom for Python 2 and 3.

``Bio/PDB/Chain.py`` class ``Chain`` uses ``cmp`` in never used private method ``._sort```, could remove this?

``Bio/PDB/Structure.py`` class ``Structure``, same.

``Bio/PDB/Model.py`` class ``Model``, same.

``Bio/PDB/Residue.py`` class ``Residue`` defines ``._sort`` and uses in public ``.sort`` but not tested.

Spotted via flake8 tests (installed and running under Python 3),

```
$ flake8 --config=/dev/null --select F821 Bio/PDB/
Bio/PDB/Chain.py:34:20: F821 undefined name 'cmp'
Bio/PDB/Chain.py:36:20: F821 undefined name 'cmp'
Bio/PDB/Chain.py:37:16: F821 undefined name 'cmp'
Bio/PDB/Model.py:52:16: F821 undefined name 'cmp'
Bio/PDB/Residue.py:52:20: F821 undefined name 'cmp'
Bio/PDB/Residue.py:62:20: F821 undefined name 'cmp'
Bio/PDB/Residue.py:67:16: F821 undefined name 'cmp'
Bio/PDB/Structure.py:36:16: F821 undefined name 'cmp'
```

peterjc:
This should close #1189, Bio.PDB sorting not tested and Python 2 only.

Note this removes the (apparently unused) private ``._sort()`` methods of all the objects, and updates the only public ``.sort()`` method which is on the ``Residue`` object - for which I have added a test.

We could define (Python 3 compatible) public ``.sort()`` methods for the ``Chain``, ``Model`` and ``Structure`` based on the existing Python 2 only private sort methods - but are they useful?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,14 @@ cache: pip
 matrix:
   include:
     - stage: basics
+      python: 2.7
+      env: TOXENV=style
+      addons:
+        apt:
+          packages:
+      before_install: echo "Going to run basic checks"
+    - stage: basics
+      python: 3.6
       env: TOXENV=style
       addons:
         apt:
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -13,30 +13,6 @@ class Chain(Entity):
         self.level = "C"
         Entity.__init__(self, id)
 
-    # Private methods
-
-    def _sort(self, r1, r2):
-        """Sort function for residues in a chain (PRIVATE).
-
-        Residues are first sorted according to their hetatm records.
-        Protein and nucleic acid residues first, hetatm residues next,
-        and waters last. Within each group, the residues are sorted according
-        to their resseq's (sequence identifiers). Finally, residues with the
-        same resseq's are sorted according to icode.
-
-        Arguments:
-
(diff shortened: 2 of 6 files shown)


peterjc:
@etal @JoaoRodrigues what do you think about this PDB sort change?

etal:
These changes all look good and safe to me.

It would be reasonable for Chain, Model, and Structure be able to sort their child entities by ID. I don't recall needed to do that back when I was working with protein structures, though, since the files from PDB normally have that done already.

peterjc:
I agree - but the PDB model details are not fresh enough in my mind to quickly implement sorting at the other levels.

peterjc:
I've rebased this to deal with the merge conflict... waiting for tests...

peterjc:
Rebased again, with the missing colon in the TravisCI file fixed.

PR #1293, opened 2017-06-22
Use setuptools in place of distutils

### LINKED ISSUE #1183: Updating setup.py to assume and use setuptools
RE: http://mailman.open-bio.org/pipermail/biopython-dev/2017-April/021690.html

> Hello all,
>
> If you are not interested in the packaging details for how we release Biopython, you can stop reading now.
>
> According to https://packaging.python.org/current/ we should recommend using pip for installation (which we now do), and use setuptools to define projects and create Source Distributions (which we no not take advantage of fully).
>
> According to https://packaging.python.org/installing/ we can assume both pip and setuptools are available on Python 2.7.9 and 3.4 onwards (possibly after invoking ensurepip). Note we've deprecated and are likely to soon drop Python 3.3 support [*].
>
> With Biopython 1.69 we included a minimal requirments.txt file which makes it easy to install our numpy dependency but we do not yet specify numpy via install_requires in setup.py - in part because we do not yet assume setuptools will be present. Here's a nice explanation of these two mechanisms and the differing usecases: https://caremad.io/posts/2013/07/setup-vs-requirement/
>
> In short I would like to require setuptools be present for building Biopython, which will let us simply our setup.py file and take advantage of many of the meta-data fields which we currently omit (e.g. classifiers for which versions of Python supported, which gets advertised on PyPI etc).
> 
> Peter
> 
> [*] http://mailman.open-bio.org/pipermail/biopython-dev/2017-April/021674.html

peterjc:
This would close #1183.

It does add classifiers to ``setup.py``, but otherwise does not attempt to make any changes to the installation process. I am deliberately being cautious here for Biopython 1.70, and if that throws up no issues we can move forward with further modernisations to the ``setup.py`` file.

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -116,7 +116,6 @@ commands =
 skip_install = True
 deps =
 commands =
-    python setup.py sdist --manifest-only
     python setup.py sdist --formats=gztar,zip
 
 [testenv:bdist_wheel]
--- a/setup.py
+++ b/setup.py
@@ -499,6 +499,28 @@ setup_args = {
     "description": 'Freely available tools for computational molecular biology.',
     "long_description": readme_rst,
     "download_url": 'http://biopython.org/DIST/',
+    "classifiers": [
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Developers',
+        'Intended Audience :: Science/Research',
+        'License :: Freely Distributable',
+        # Technically the "Biopython License Agreement" is not OSI approved,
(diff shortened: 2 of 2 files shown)


chapmanb:
This looks great to me Peter. Thanks so much for modernizing the build.

peterjc:
OK, fingers crossed - hopefully we'll get quick feedback if this does cause issues downstream.

peterjc:
Cross reference #1307 where I am proposing further changes to ``setup.py`` using ``setuptools`` for our NumPy dependency

Issue #443, opened 2014-12-10
Add wrapper for EBI's dbfetch


peterjc:
See http://www.ebi.ac.uk/Tools/dbfetch/

This makes sense as a European alternative to the NCBI Entrez efetch, or TogoWS's entry API calls.

Gasta88:
Hi @peterjc , this issue is quite old but if people still find some interest in it, I would like to develop a wrapper for it.

I have few questions:
1) I would like to use other wrapper in Biopython as a reference (ex: those one present in `Bio.Align.Applications`. Is it correct to use the `Bio.Application` module to develop them?

2) On the website there is a reference to their web service: [http://www.ebi.ac.uk/Tools/webservices/services/dbfetch](url). Is it the case to use this instead of the direct access as mentioned originally?

peterjc:
Our framework using ``Bio.Application`` is for wrapping command line tools, so no.

This is for wrapping an online resource, which is why ``Bio.Entrez`` or ``Bio.TogoWS`` would be a better example. Or ``Bio.KEGG.REST``.

Maybe under ``Bio.EBI`` would be a good namespace, although it does not follow PEP8 and use lower case...

http://www.ebi.ac.uk/Tools/webservices/services/dbfetch appears to be is a SOAP based service, which I would avoid (at least for now), and focus on the simple REST style API described at http://www.ebi.ac.uk/Tools/dbfetch/

Gasta88:
Hi @peterjc,

I've made some progress and created a simple module for dbfetch. I have few questions on how to test it and integrate it with Biopython:

I'll place it in `Bio\EBI`. The module will be name `__init__.py`. This wrapper offer two methods only:

> One method to retrieve meta-info regarding all the database accessible via dbfetch. A handle is returned.
> One method to create the query to the selected database via four parameters: database, ids, format and syntax. A handle is returned.

For the testing part I'm creating a script that will be located in `Tests\test_EBI_dbfetch.py`. 

I'm not sure if I should test each single database via the `EBI.dbfetch()` method or not. After all only the parameters of the method change, but not the object type returned.
Moreover I don't thinks it's safe to check the content of any handle because today an identifier might be connected to an entity, while in the future it can change with a new version of the database. Do you agree with this?

Would it be easier to upload what I've done in a PR and take it from there?

peterjc:
[...]

PR #1298, opened 2017-06-26
Adding Bio.EBI.dbfetch

Gasta88:
I'll open this PR and then carry on with the editing for module location and other issues.

This refer issue #443


setup.py
```diff
@@ -332,6 +332,7 @@ def is_Numpy_installed():
     'Bio.Compass',
     'Bio.Crystal',
     'Bio.Data',
+	'Bio.EBI.dbfetch'
```
peterjc: The indentation looks odd here.

Gasta88:
It seems that Travis is not able to load the module.

How do you make new modules available for testing?

peterjc:
I think you need to add ``Bio.EBI`` in ``setup.py``, not just ``Bio.EBI.dbfetch``

setup.py
```diff
@@ -332,7 +332,7 @@ def is_Numpy_installed():
     'Bio.Compass',
     'Bio.Crystal',
     'Bio.Data',
-    'Bio.EBI.dbfetch',
+    'Bio.EBI',
```
peterjc: Given you are using ``Bio/EBI/dbfetch/__init__.py`` then I think you need both ``Bio.EBI`` and ``Bio.EBI.dbfetch`` here in ``setup.py``

Alternatively and perhaps more simply, I think you could use ``Bio/EBI/dbfetch.py`` instead and just ``Bio.EBI``in ``setup.py``?

Gasta88: Apologies for the late reply, but I was on holiday.

I've followed your second suggestion.

peterjc:
You will still need to have a ``Bio/EBI/__init__.py`` file for Python 2.7 to work (or anything before Python 3.3), see https://www.python.org/dev/peps/pep-0420/

Gasta88:
Sorry, but i don't understand.

I've renamed the module from `dbfetch.py` to `__init__.py` so that the module import command would be shorter and clearer.

The file itself contains the two methods necessary to use the service and TravisCI is not raising any error for any version of Python.

Does the folder system need to look something like this?
```
|\Bio
|-----\EBI
|---------\dbfetch.py
|--------- \__init__.py
```
where the new `__init__.py` file should import `dbfetch.py` only?
Other implementations (ex: Entrez or TofgoWS) only have a `__init__.py` file with the module methods in it.

peterjc:
The fact that many bits of Biopython put code directly into the ``__init__.py`` files is not always the best approach. Here your code is specific to the EBI's online resource ``dbfetch``. From a name space point of view having the functions under ``Bio.EBI.dbfetch`` seemed best, which means ``Bio/EBI/__init__.py`` can literally be an empty file, ideally something like https://github.com/biopython/biopython/blob/biopython-169/Bio/Blast/__init__.py

So yes,

* ``Bio/EBI/__init__.py`` (no real code)
* ``Bio/EBI/dbdetch.py`` (your code)

And in ``setup.py`` include ``Bio.EBI`` in the package list.

Gasta88:
Thanks! This is my first time contributing to a new module, hopefully I'll get better with the next ones :)

peterjc:
You still have this in your ``setup.py`` file,

```python
    'Bio.EBI.dbfetch',
    'Bio.EBI',
```

Since you've got rid of the directory ``Bio/EBI/dbfetch/`` and just have ``Bio/EBI/`` all you need in ``setup.py`` for the pacakge list is:

```python
    'Bio.EBI',
```

peterjc:
Can you add this to ``Tests/test_EBI_dbfetch.py`` please so that it will get skipped when run via ``run_tests.py --offline``:

```python
import requires_internet
requires_internet.check()
```

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    return _q("dbfetch.databases?style=json")
+
+
+def dbfetch(db, ids, format=None, style=None):
+    """Create a URL to use the tool via RESTful method.
+
+    db = database to query.
+    ids = single id or list of ids separated by commas. Can be provvided as
+          list or string object.
```
Gasta88: TravisCI is throwing a `flake8` error that I can't reproduce locally. 
I've updated the package but when I run it everything is ok.

peterjc:
Ah, this one:

```
$ flake8 Bio/
Bio/EBI/dbfetch.py:71:1: RST301 Unexpected indentation.
Bio/EBI/dbfetch.py:72:1: RST201 Block quote ends without a blank line; unexpected unindent.
```

You'll also need to do ``pip install flake8-rst-docstrings`` for those checks which are via my flake8 plugin using docutils - https://github.com/peterjc/flake8-rst-docstrings

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+    here http://www.ebi.ac.uk/Tools/dbfetch/dbfetch/dbfetch.databases
+
+    """
+    return _q("dbfetch.databases?style=json")
+
+
+def dbfetch(db, ids, format=None, style=None):
+    """Create a URL to use the tool via RESTful method.
+
+    db = database to query.
```
peterjc: This does not look like valid RST, I would suggest a bullet point list like the example here:
https://github.com/biopython/biopython/blob/biopython-170/Bio/SeqRecord.py#L160

peterjc:
On the mailing list Jun Aruga volunteered to look at the new module this weekend:
http://mailman.open-bio.org/pipermail/biopython-dev/2017-August/021858.html

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import print_function
+
+from Bio._py3k import urlopen as _urlopen
+from Bio._py3k import _binary_to_string_handle
+
+import requires_internet
+requires_internet.check()
+
+
+def _q(arg1, arg2=None, arg3=None, arg4=None):
```
junaruga: The argument names: arg1, arg2, arg3, arg4 is hard to understand the behavior.
The reason you used those may be that you wanted to call `_q` from both by `def dbfetch` and `def db_info`.
Is this case, replacing `def _q` to `def _query_by_db`, below sample looks better, isn't it?

```
_BASE_URL = "http://www.ebi.ac.uk/Tools/dbfetch/dbfetch%s"


def _query_by_url(url):
    resp = _urlopen(_BASE_URL % url)
    return _binary_to_string_handle(resp)


def db_info():
    return _query_by_url("dbfetch.databases?style=json")


def dbfetch(db, ids, format=None, style=None):
    ...

    return _query_by_db(db, ids, format, style)


def _query_by_db(db, ids=None, format=None, style=None):
    url = '/%s' % db
    if ids:
        url += '/%s' % ids
    if format:
        url += '/%s' % format
    if style:
        url += '/%s' % style
    resp = _urlopen(_BASE_URL % url)
    return _binary_to_string_handle(resp)
```

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+    URL = "http://www.ebi.ac.uk/Tools/dbfetch/dbfetch%s"
```
junaruga: I am not sure using constant in the function is correct way or not.

As a reference, TogoWS (`__init__.py`) is using `_BASE_URL` at outside of the function.
If you want to use it inside of function, other API is using `url` as a variable.

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+    if arg3 and arg4:
```
junaruga: This string appending logic is a little bit duplicated logic such as `"/%s..." % (arg1 ..)`.
I showed another logic's sample for that in above part.

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
+
+"""Provides code to access the EBI dbfetch tool.
+
+This module aims to make dbfetch easy to use. See:
+http://www.ebi.ac.uk/Tools/dbfetch/
+
+The dbfetch tool from EBI provvides access to different databases via a simple
```
junaruga: Typo: provvide => provide (There are 3 parts in the file)

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+URL that the module constructs. Errors are handled regarding databse access and
+HTPP level.
+
+The functionality is somewhat similar to Biopython's Bio.TogoWS, Bio.KEGG.REST
+and Bio.Entrez modules.
+
+EBI provvides extensive guidelines for the databases, their format and the
+syntax necessary to use the tool. See:
+http://www.ebi.ac.uk/Tools/dbfetch/syntax.jsp
+
```
junaruga: Do we need to write that we do not support the SOAP services (  http://www.ebi.ac.uk/Tools/webservices/services/dbfetch ) in this comment?
That is written in the comment of TogoWS `__init__.py`.

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,82 @@
...
+    URL = "http://www.ebi.ac.uk/Tools/dbfetch/dbfetch%s"
+    if arg3 and arg4:
+        args = "/%s/%s/%s?%s" % (arg1, arg2, arg3, arg4)
+    elif arg3:
+        args = "/%s/%s/%s" % (arg1, arg2, arg3)
+    elif arg2:
+        args = "/%s/%s" % (arg1, arg2)
+    else:
+        args = "/%s" % (arg1)
+    resp = _urlopen(URL % (args))
```
junaruga: The URL`http://www.ebi.ac.uk/Tools/dbfetch/dbfetch/{db}/{id}/{format}?{style}` that is used in this code, is popular than the URL that is mentioned in below page?

> http://www.ebi.ac.uk/Tools/dbfetch/syntax.jsp
> The general format of the dbfetch URL is:
> dbfetch?db={db}&format={format}&style={style}&id={id}&pageHtml={pageHtml}

Why do you prefer it?

junaruga:
I commented!

Right now there is 11 commits in this PR.
After the reviewing is done, I expect the commits will be squash to proper units.

Gasta88:
Thanks for the review @junaruga! This is my first attempt in building a module and I'm not surprised it looks rough from a more expert view.

I've implemented the changes proposed and explain my reason where asked.

peterjc:
Thanks @junaruga - you've been very though.

TravisCI is failing due to a problem on the master branch, which I have hopefully fixed now... in any case as this is an online test we'd run it locally anyway since TravisCI and AppVeyor are setup to run only the offline tests.

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,91 @@
...
+http://www.ebi.ac.uk/Tools/dbfetch/
+
+The dbfetch tool from EBI provides access to different databases via a simple
+URL that the module constructs. Errors are handled regarding databse access and
+HTTP level.
+
+The functionality is somewhat similar to Biopython's Bio.TogoWS, Bio.KEGG.REST
+and Bio.Entrez modules.
+
+EBI provvides extensive guidelines for the databases, their format and the
```
junaruga: Typo: "provvides" => "provides"

Bio/EBI/dbfetch.py
```diff
@@ -0,0 +1,91 @@
...
+
+    """
+    return _query_by_url("dbfetch.databases?style=json")
+
+
+def dbfetch(db, ids, format=None, style=None):
+    """Create a URL to use the tool via RESTful method.
+
+    - db    - database to query.
+    - ids   - single id or list of ids separated by commas. Can be provvided as
```
junaruga: Typo: "provvided" => "provided"

junaruga:
@Gasta88 Okay, I commented about the typo now. and this PR tests are failed for the style check on Travis.

@peterjc Okay!

peterjc:
I *think* that if you make another commit to fix those typos, then TravisCI should test with the latest master branch where I fixed that style issue.

Gasta88:
Since the failure is coming from codecov, should I do something?

peterjc:
Don't worry about CodeCov here. I'd like to tweak the CodeCov settings - in this case your work has tests but as they are online, they do not currently get run on TravisCI/AppVeyor - so don't appear to have much test coverage.

I'd have to manually run the tests locally before merging.

I think you've addressed all @junaruga's point now, right?

Gasta88:
Yes I have.

junaruga:
@Gasta88 @peterjc It looks good to me. Thanks.

peterjc:
I'm OK with merging this, unless as per discussion on #443 directly people to https://github.com/bebatut/ebisearch is deemed a better bet? Let's discuss that on #443.

Gasta88:
Since I'm trying to integrate EBIsearch, can I close this PR and place the editing into another branch/PR?

peterjc:
Sure, a new branch and pull request is fine if that's easier.

(I suggest you don't delete this branch immediately - it might still be of interest for comparing the alternative code)

PR #1299, opened 2017-06-26
Ignore leading spaces in value

ecederstrand:
Fixes parsing of quoted, multi-line qualifier values in GenBank feature tables. The following would previously raise `ValueError: Problem with 'CDS' feature:[...]` because the quote in the `/property` qualifier was not properly detected:

```
FH   Key             Location/Qualifiers
FT   CDS             1..756
FT                   /*tag=  a
FT                   /product= "Lactobacillus kefir T76I/V95M/S96L/E145A/F147L
FT                   /V148I/T152A/L153M/Y190G/A202F/M206C/Y249F mutant
FT                   ketoreductase (KRED) protein"
FT                   /partial
FT                   /note= "No stop codon is shown"
```

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -299,7 +299,10 @@ class InsdcScanner(object):
                     # New qualifier
                     i = line.find("=")
                     key = line[1:i]  # does not work if i==-1
-                    value = line[i + 1:].lstrip()  # we ignore 'value' if i==-1
+                    value = line[i + 1:]  # we ignore 'value' if i==-1
+                    if i and value.startswith(' ') and value.lstrip().startswith('"'):
+                        warnings.warn("White space after equals in qualifier", BiopythonParserWarning)
+                        value = value.lstrip()
                     if i == -1:
                         # Qualifier with no key, e.g. /pseudo
                         key = line[1:]
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -299,7 +299,7 @@ class InsdcScanner(object):
                     # New qualifier
                     i = line.find("=")
                     key = line[1:i]  # does not work if i==-1
-                    value = line[i + 1:]  # we ignore 'value' if i==-1
+                    value = line[i + 1:].lstrip()  # we ignore 'value' if i==-1
                     if i == -1:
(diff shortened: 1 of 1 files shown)


peterjc:
Where did the problematic file come from? I'm assuming not from the NCBI.

In this case I agree the error looks easy to cope with - but I would prefer to issue a parser warning as the file is invalid. e.g.

```python
warnings.warn("White space after equals in qualifier on line: %r" % line,
              BiopythonParserWarning)
```

or, if the file has so many of these that you think a single warning is enough use a fixed message. perhaps:

```python
warnings.warn("White space after equals in qualifier", BiopythonParserWarning)
```

ecederstrand:
This came from the GENESEQN database, more specifically record AZF43916. I'll add a warning in a separate commit.

peterjc:
Could you contact GENESEQN to ask they fix their formatting?

Are you familiar with Python's ``unittest`` and testing for warnings?

ecederstrand:
The thing with GENESEQ is that it's not a strict EMBL format, just EMBL-like (because why use a standard when you can make your own...). There's no restriction on spaces in the features table in the specification, as far as I can see. See http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.179.2297&rep=rep1&type=pdf (see Appendix I, page 24).

We have implemented a full GENESEQ parser that we may eventually contribute back. It builds on the EMBL parser from Biopython, and this patch is one of the things that was easy to upstream until then.

I'm happy to contribute under both licenses and being mentioned as contributor.

peterjc:
I think GENESEQ ought to be a little stricter here. See also the INSDC table format http://www.insdc.org/files/feature_table.html as used in GenBank and EMBL flat files.

In terms of extending the EMBL parser, see also the IMGT parser:
https://github.com/biopython/biopython/blob/biopython-169/Bio/SeqIO/InsdcIO.py#L143
https://github.com/biopython/biopython/blob/biopython-169/Bio/GenBank/Scanner.py#L924

Thank you for agreeing to the licensing, and to being named.

Does adding a test for the warning look easy for you to do? Its a bit fiddly compared to checking for an exception.

peterjc:
If you want to come add a unit test for this later, that'd be great - but I've merged this as is - thank you.

PR #1301, opened 2017-06-28
Adding '.count_overlap()' method to Seq.py

chris-rands:
Following [issue 1294](https://github.com/biopython/biopython/issues/1294), I have added a `.count_overlap()` method to the `Seq`, `UnknownSeq` and `MutableSeq` classes in `Bio/Seq.py`. I have also added a line in each `.count()` method docstring referring to the new method.

Unresolved issues:
- The implementations for the 3 classes are basically the same and all rely on conversion of `self` to strings so that `str.find()` can be used. There might be better implementations for the `UnknownSeq` and `MutableSeq` classes. 
- Doctests have been added but no other tests yet. In the Issue, it was suggested that tests should be added to `Tests/test_Seq_objs.py` but I was not clear how best to implement this. 

I would appreciate feedback, suggestions and any contributions, thank you!

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -431,7 +431,7 @@ class Seq(object):
         # If it has one, check the alphabet:
         sub_str = self._get_seq_str_and_check_alphabet(sub)
         return str(self).count(sub_str, start, end)
-    
+
     def count_overlap(self, sub, start=0, end=sys.maxsize):
         """Return an overlapping count.
 
@@ -1467,7 +1467,7 @@ class UnknownSeq(Seq):
         count() method is much for efficient.
         """
         # The implementation is currently identical to that of
-        # Seq.count_overlap() 
+        # Seq.count_overlap()
         sub_str = self._get_seq_str_and_check_alphabet(sub)
         self = str(self)
         overlap_count = 0
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -1461,6 +1461,14 @@ class UnknownSeq(Seq):
         3
         >>> UnknownSeq(4, character="N").count_overlap("NNN")
         2
+        >>> UnknownSeq(7, character="N").count_overlap("NN")
+        6
+        >>> UnknownSeq(7, character="N").count_overlap("NNN")
+        5
+        >>> UnknownSeq(7, character="N").count_overlap("NN", 1, 5)
+        3
+        >>> UnknownSeq(7, character="N").count_overlap("NN", -5, -1)
+        3
 
         Where substrings do not overlap, should behave the same as
         the count() method:
@@ -1481,17 +1489,32 @@ class UnknownSeq(Seq):
         HOWEVER, do not use this method for such cases because the
         count() method is much for efficient.
         """
-        # The implementation is currently identical to that of
-        # Seq.count_overlap()
         sub_str = self._get_seq_str_and_check_alphabet(sub)
-        self = str(self)
-        overlap_count = 0
-        while True:
-            start = self.find(sub_str, start, end) + 1
-            if start != 0:
-                overlap_count += 1
+        if len(sub_str) == 1:
+            if str(sub_str) == self._character:
+                if start == 0 and end >= self._length:
+                    return self._length
+                else:
+                    # This could be done more cleverly...
+                    return str(self).count(sub_str, start, end)
             else:
-                return overlap_count
+                return 0
+        else:
+            if set(sub_str) == set(self._character):
+                if start == 0 and end >= self._length:
+                    return self._length - len(sub_str) + 1
+                else:
+                    # This could be done more cleverly...
+                    self = str(self)
+                    overlap_count = 0
+                    while True:
+                        start = self.find(sub_str, start, end) + 1
+                        if start != 0:
+                            overlap_count += 1
+                        else:
+                            return overlap_count
+            else:
+                return 0
 
     def complement(self):
         """Return the complement of an unknown nucleotide equals itself.
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -484,10 +484,10 @@ class Seq(object):
         count() method is much for efficient.
         """
         sub_str = self._get_seq_str_and_check_alphabet(sub)
-        self = str(self)
+        self_str = str(self)
         overlap_count = 0
         while True:
-            start = self.find(sub_str, start, end) + 1
+            start = self_str.find(sub_str, start, end) + 1
             if start != 0:
                 overlap_count += 1
             else:
@@ -1485,9 +1485,6 @@ class UnknownSeq(Seq):
         0
         >>> UnknownSeq(4, character="N").count_overlap("AA") == UnknownSeq(4, character="N").count("AA")
         True
-
-        HOWEVER, do not use this method for such cases because the
-        count() method is much for efficient.
         """
         sub_str = self._get_seq_str_and_check_alphabet(sub)
         if len(sub_str) == 1:
@@ -1505,10 +1502,10 @@ class UnknownSeq(Seq):
                     return self._length - len(sub_str) + 1
                 else:
                     # This could be done more cleverly...
-                    self = str(self)
+                    self_str = str(self)
                     overlap_count = 0
                     while True:
-                        start = self.find(sub_str, start, end) + 1
+                        start = self_str.find(sub_str, start, end) + 1
                         if start != 0:
                             overlap_count += 1
                         else:
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -1457,7 +1457,7 @@ class UnknownSeq(Seq):
 
         e.g.
 
-        >>> from Bio.Seq import UnknownSeq 
+        >>> from Bio.Seq import UnknownSeq
         >>> UnknownSeq(4, character="N").count_overlap("NN")
         3
         >>> UnknownSeq(4, character="N").count_overlap("NNN")
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -392,6 +392,8 @@ class Seq(object):
         This behaves like the python string method of the same name,
         which does a non-overlapping count!
 
+        For an overlapping search use the newer count_overlap() method.
+
         Returns an integer, the number of occurrences of substring
         argument sub in the (sub)sequence given by [start:end].
         Optional arguments start and end are interpreted as in slice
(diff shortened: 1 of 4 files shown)


peterjc:
For the ``Seq`` object I would suggest you edit the existing docstring section:

```python
         """...

         HOWEVER, please note because python strings and Seq objects (and
         MutableSeq objects) do a non-overlapping search, this may not give
         the answer you expect:
 
         >>> "AAAA".count("AA")
         2
         >>> print(Seq("AAAA").count("AA"))
         2
 
         An overlapping search would give the answer as three!
         """
```

Bio/Seq.py
```diff
@@ -430,6 +432,58 @@ def count(self, sub, start=0, end=sys.maxsize):
...
+        >>> from Bio.Seq import Seq
+        >>> print(Seq("AAAA").count_overlap("AA"))
+        3
+        >>> print(Seq("ATATATATA").count_overlap("ATA"))
+        4
+        >>> print(Seq("ATATATATA").count_overlap("ATA", 3, -1))
+        1
+
+        Where substrings do not overlap, should behave the same as
+        the count() method:
```
peterjc: I would show this in the doctest that follows - call both ``.count()`` and ``.count_overlap()`` one after the other?

peterjc:
I would think this should work for the ``UnknowSeq``  object, basically repeating the ``.count()`` method code but with a different calculation:

```python
        sub_str = self._get_seq_str_and_check_alphabet(sub)
        if len(sub_str) == 1:
            if str(sub_str) == self._character:
                if start == 0 and end >= self._length:
                    return self._length
                else:
                    # This could be done more cleverly...
                    return str(self).count(sub_str, start, end)
            else:
                return 0
        else:
            if set(sub_str) == set(self._character):
                if start == 0 and end >= self._length:
                    return end - start + 1  # TODO, check this is not off by one!
                else:
                    # This could be done more cleverly...
                    return str(self).count(sub_str, start, end)
            else:
                return 0
```

Both here and in the original ``.count()`` method, ideally we really want to avoid ``str(self)`` as we could be creating for example a 5 million character string if this was from say a GenBank assembly file which does not provide the actual sequence.

peterjc:
For the test, perhaps something like this?

```python
    def test_str_count_overlap_GG(self):
        """Check our count_overlap method using GG."""

        expected = [3, 3, 3, 3, 1, ...]  # FILL THIS IN!

        for seq, exp in zip(self._examples, expected):
            # Using search term GG as a string
            self.assertEqual(seq.count_overlap("GG"), exp)
            # Using search term GG as a Seq with generic alphabet
            self.assertEqual(seq.count_overlap(Seq("GG")), exp)

    def test_str_count_overlap_NN(self):
        """Check our count_overlap method using GG."""

        expected = [0, 0, ..., 11, 0, 0, 0, ...]  # FILL THIS IN!

        for seq, exp in zip(self._examples, expected):
            # Using search term NN as a string
            self.assertEqual(seq.count_overlap("NN"), exp)
            # Using search term GG as a Seq with generic alphabet
            self.assertEqual(seq.count_overlap(Seq("NN")), exp)

```

This could use ``self._start_end_values`` as well for start/end values. Or more simply you could construct explicit examples, e.g.:

```python
    def test_str_count_overlap_NN(self):
        self.assertEqual(Seq("ACGTTTACG").count_overlap("TT"), 2)
        self.assertEqual(Seq("ACGTTTACG", generic_dna).count_overlap("TT"), 2)
        self.assertEqual(Seq("ACGTTTACG", generic_dna).count_overlap("TT", 3, 50), 2)
        ...
```

chris-rands:
Thank you Peter; I will look at this properly next week and try to carry out all your suggestions

By the way, why does `Seq.py` include all these `end=sys.maxsize` default arguments? Why not use `end=None` instead?

peterjc:
Well ``sys.maxsize`` works under both Python 2 and 3, previously it was ``sys.maxint`` - see https://github.com/biopython/biopython/commit/9ab622b8bdabe51760201c5e7b19c479cd5ffc2f

But to answer your real question, it was to follow the python ``UserString`` defaults, see https://github.com/biopython/biopython/commit/890e1d5f8db228e15495f62a29b3e9271d861370

Adil-Iqbal:
Just out of curiosity, what is the rationale for having a separate method for this functionality? Would it not be simpler to include a Boolean argument called "overlap" within the existing count method?

my_seq.count_overlap ("ATG")

versus

my_seq.count ("ATG", overlap=True)

peterjc:
Adil: See discussion on https://github.com/biopython/biopython/issues/1294
about new method vs extra argument.

chris-rands:
Hi Peter, I have tried to address all your comments above including adding more to the docstrings, changing the implementation of `UnknownSeq.count_overlap()` and adding tests to `Tests/test_Seq_objs.py`.

NOTE, I did not implement `UnknownSeq.count_overlap()` the way you suggested, because this did not work I think. Please check that the implementation I went with looks correct.

Let me know what you think and if there is anything else to do, thanks!

Bio/Seq.py
```diff
@@ -1870,6 +2013,68 @@ def count(self, sub, start=0, end=sys.maxsize):
...
+        >>> my_mseq.count_overlap("AT", 2, -1) == my_mseq.count("AT", 2, -1)
+        True
+
+        HOWEVER, do not use this method for such cases because the
+        count() method is much for efficient.
+        """
+        # The implementation is currently identical to that of
+        # Seq.count_overlap() apart from the definition of sub_str
+        sub_str = str(sub)
+        self = str(self)
```
peterjc: Redefining ``self`` seems like bad style, how about defining ``self_str = str(self)`` instead?

Bio/Seq.py
```diff
@@ -1375,6 +1440,82 @@ def count(self, sub, start=0, end=sys.maxsize):
...
+        >>> UnknownSeq(4, character="N").count_overlap("A")
+        0
+        >>> UnknownSeq(4, character="N").count_overlap("A") == UnknownSeq(4, character="N").count("A")
+        True
+        >>> UnknownSeq(4, character="N").count_overlap("AA")
+        0
+        >>> UnknownSeq(4, character="N").count_overlap("AA") == UnknownSeq(4, character="N").count("AA")
+        True
+
+        HOWEVER, do not use this method for such cases because the
```
peterjc: I would have thought ``.count("AA")`` vs ``.count_overlap("AA")`` does not actually make much difference for the ``UnknownSeq``? i.e. Should the ``HOWEVER...`` bit be removed?

Bio/Seq.py
```diff
@@ -1375,6 +1440,82 @@ def count(self, sub, start=0, end=sys.maxsize):
...
+                    return str(self).count(sub_str, start, end)
+            else:
+                return 0
+        else:
+            if set(sub_str) == set(self._character):
+                if start == 0 and end >= self._length:
+                    return self._length - len(sub_str) + 1
+                else:
+                    # This could be done more cleverly...
+                    self = str(self)
```
peterjc: Please avoid redefining ``self``; see also separate comment about how to implement this efficiently

peterjc:
Yes, you're right - my outline for ``UnknownSeq.count_overlap()`` was faulty. This should be a trivial integer calculation using start and end (for the search window) and the length of the target sequence.

i.e. Avoid calling ``str(self)`` which would pointlessly create a large string from the ``UnknownSeq`` object.

chris-rands:
Thanks Peter, I'll use `self_str = str(self)` of course, apologies.

Indeed, I should have thought of that, I'll make an implementation using just the the start, end and substring size. I think the same logic could apply for `UnknownSeq.count()` (which currently includes `str(self).count()` calls)?

Otherwise, do the `Tests/test_Seq_objs.py` tests look ok?

peterjc:
Yes, looking at it now with fresh eyes we could improve ``UnknownSeq.count`` as well.

I've not examine the coverage but the tests look good in principle.

I'd want to make sure we have pathological examples like a search string longer than the main string:

```python
>>> UnknownSeq(3, character="N").count("NNN")
1
>>> UnknownSeq(3, character="N").count("NNNN")
0
>>> UnknownSeq(3, character="N").count_overlap("NNN")
1
>>> UnknownSeq(3, character="N").count_overlap("NNNN")
0
```

chris-rands:
Ok, thanks, shall I look to change `UnknownSeq.count()` here too or should that be a separate issue/pull request?

I will add some pathological tests along with the other changes.

peterjc:
What ever is easier for you - they overlap so much and I'd be reviewing both changes anyway.

chris-rands:
I've just committed a modified version of `Unknown.count_overlap()` that no longer uses `str(self)` and so avoids the memory burden. I'd be grateful for a careful review!

The maths is simple, but I still found it quite confusing honestly. The edge case you mention where the substring is larger than `self` is handled and I also thought about negative `start` and/or `end` values. Also values greater than the length of `self` or less than the -length of `self` needed to be truncated I thought. Finally `None` inputs now default to the `0` for `start` and `sys.maxsize` for `end`

Tests/test_Seq_objs.py
```diff
@@ -181,6 +181,88 @@ def test_str_count(self):
...
+
+        assert len(self._examples) == len(expected)
+
+        for seq, exp in zip(self._examples, expected):
+            # Using search term GG as a string
+            self.assertEqual(seq.count_overlap("GG"), exp)
+            # Using search term GG as a Seq with generic alphabet
+            self.assertEqual(seq.count_overlap(Seq("GG")), exp)
+
+        # Testing Seq() and MutableSeq() with variable start and end arguments
```
peterjc: Maybe split this into a separate test method, e.g. insert

```python
        def test_count_overlap_start_end(self):
```

Bio/Seq.py
```diff
@@ -1464,6 +1464,8 @@ def count_overlap(self, sub, start=0, end=sys.maxsize):
         2
         >>> UnknownSeq(4, character="N").count_overlap("NNNNN")
         0
+        >>> UnknownSeq(4,character="N").count_overlap('NNN',1,2)
```
peterjc: The docstring isn't following PEP8 in terms of spaces after commas - but I think this and most of the other doctests here would be better off in the unit tests as they do little in terms of documentation.

peterjc:
This looks about ready - other than moving some of the new doctests into the main test suite I didn't see anything worth delaying this, so it could well make it into Biopython 1.70 which I hope to do this week.

chris-rands:
Thank you Peter, that sounds great, I'll try and move most of the new doctests to the main suite today

Very happy on both counts

chris-rands:
Just committed the changes, let me know if there is any more to do, thanks!

peterjc:
You can add yourself to the ``NEWS.rst`` and ``CONTRIB.rst`` files if you like (sorted alphabetically), otherwise I'd do that after merging.

I'm just waiting for the tests to finish now, thanks.

chris-rands:
Thanks, I've added myself to the files

peterjc:
Can you push that final commit with the name addition?

chris-rands:
Done. Something on the `count_overlap()` method could be added to `NEWS.rst` if you think it's appropriate, up to you of course

peterjc:
Lovely, thank you.

chris-rands:
Thank you Peter!

PR #1306, opened 2017-07-03
Add 'testseq' function

Adil-Iqbal:
1. Reverted Seeding.
2. Updated doctests for reverted seeding.
3. Reviewed unit test to ensure that it works with reverted seeding.
4. Allowed table argument to accept names and objects.
5. Updated unit test to attempt importing normally before using the "manually_import" function.

Everything in this pull request has been discussed earlier with the exception of the very last item.  I have wrapped the 'manually_import' function in a try/except block. My reason for doing this is below:

Python has a built-in import utility module that gives access to import internals. Over the last few versions of python, this built-in module has gone through several re-works. My concern is that it will be changed further in the future. If that happens, the 'unit test will cause an import error. 

Since importing works fine in Python 3.3 and beyond, I have wrapped the call to the "manually_import" function in a try/except block. The unit test will now attempt to import the module normally before it tries to use the "manually_import" function.  

This change will allow the unit test to continue working regardless of whether the import utility module is re-worked in subsequent versions of python.


Tests/run_tests.py
```diff
@@ -165,6 +173,12 @@ def is_numpy():
     DOCTEST_MODULES.remove("Bio.SeqIO")
     DOCTEST_MODULES.remove("Bio.SearchIO")
 
+# Skip doctest for all modules outside of 'Bio' package prior to Python 3.3
+if float(sys.version[:3]) < 3.3:
+    for i, name in enumerate(DOCTEST_MODULES):
+        if is_outside_bio(name):
```
peterjc: I don't think the extra function is needed, how about ``if not name.split(".", 1)[0].startswith("Bio"):`` here?

Tests/run_tests.py
```diff
@@ -70,6 +70,13 @@ def is_numpy():
         return False
 
 
+def is_outside_bio(name):
+    name = name.split(".")
+    if "Bio" in name[0]:
```
peterjc: I'd use ``name[0].startswith("Bio")`` or maybe even ``name[0] in ["Bio", "BioSQL"]`` to be explicit?

Tests/run_tests.py
Adil-Iqbal: Sounds good. Will implement.

Tests/run_tests.py
peterjc: See later comment on where this function was being used.

Tests/run_tests.py
peterjc: Hang on, on reflection this would be clearer: ``if not name.startswith(("Bio.", "BioSQL.")):``

Note this is passing a tuple of two elements as the argument to the starts-with method.

Adil-Iqbal: Fewer operations in fewer keystrokes. Awesome! 
I will wait until the CI tests are finished before I commit anything further.

Adil-Iqbal:
Alright! Done. Unless anyone has other concerns?

I ended up removing a lot of the code from the "manually_import" function since it just isn't necessary after python version 3.3.

Tests/run_tests.py
```diff
@@ -165,6 +166,12 @@ def is_numpy():
     DOCTEST_MODULES.remove("Bio.SeqIO")
     DOCTEST_MODULES.remove("Bio.SearchIO")
 
+# Skip doctests for all modules outside Bio & BioSQL prior to Python 3.3
+if float(sys.version[:3]) < 3.3:
```
peterjc: Sorry, I should have spotted this earlier - you should not assume you can break up the version string like that.

Instead use ``sys.version_info`` which is a (named) tuple, i.e. something like this should work?

```python
if sys.version_info < (3, 3):
```

Adil-Iqbal: Not a problem.

peterjc:
The style checks failed on Travis, but it was not your fault:

```
$ flake8 Bio/
Bio/Phylo/TreeConstruction.py:485:25: W503 line break before binary operator
```

Its been fixed in c7598291083c8c186ebe7267518c047f9d2a746c

If you are comfortable with the ``git rebase`` command, that would be the ideal way to update your branch. However, as the previous commit passed all our tests, I am not worried about it.

Because the the namespace and import issues, I'm going to ask for a second opinion before merging this...

Tests/test_testseq.py
```diff
@@ -0,0 +1,109 @@
...
+    module_path = os.path.split(module_path)
+    if module_path[1] == "Tests":
+        module_path = os.path.join(module_path[0], name.pop(0))
+    else:
+        raise ImportError("Must call 'manually_import' from inside 'Tests' folder.")
+    for i, step in enumerate(name):
+        module_path = os.path.join(module_path, step)
+    name = name[-1][:-3]
+    import imp
+    return imp.load_source(name, module_path)
```
picrin: Although the method is hacky, I can see it's utility. As more code is added to the "Scripts" directory will we need to call this method from other tests in the "Tests" directory? Would it be more suitable to move this method to a generic "utils" module?

I just see that this can become a common issue with code added to Scripts directory.

peterjc: If we do end up with more code in ``Scripts/`` being tested like this, then that seems like a good idea. Right now having the code here seems fine to me.

peterjc:
Adam Kurkiewicz on the mailing list http://mailman.open-bio.org/pipermail/biopython-dev/2017-July/021813.html reports a stochastic failure:

> I can get the test cases to pass OK 90% of the time inside Tests by
running
>
> ``python test_testseq.py``
>
> However, these test cases are still non-deterministic, in particular
test_codon_tables fails about 10% of the time. Is this intended
behaviour?
>
> That's the stack trace:
>
> ```
> picrin@lamport:~/programming/biopython/adil/Tests$ python
> test_testseq.py
> test_alphabet (__main__.TestTestseq)
> Testing 'alphabet' argument... ... ok
> test_codon_tables (__main__.TestTestseq)
> Testing codon sets... ... FAIL
> test_gc_target (__main__.TestTestseq)
> Testing 'gc_target' argument... ... ok
> test_messenger (__main__.TestTestseq)
> Testing 'messenger' argument... ... ok
> test_seeding (__main__.TestTestseq)
> Testing 'rand_seed' argument... ... ok
> test_size (__main__.TestTestseq)
> Testing 'size' and 'truncate' arguments... ... ok
> 
> ======================================================================
> FAIL: test_codon_tables (__main__.TestTestseq)
> Testing codon sets...
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "test_testseq.py", line 87, in test_codon_tables
>     self.assertFalse(seq[0] == "M")
> AssertionError: True is not false
> 
> ----------------------------------------------------------------------
> Ran 6 tests in 0.008s
> 
> FAILED (failures=1)
> ```

peterjc:
I'm going to close this as stale, although the hint of a stochastic failure in the Codon Tables is slightly concerning.

PR #1309, opened 2017-07-05
Feature location equality

kblin:
As discussed with @peterjc on Twitter, it would be nice to have a strict* equality check for `FeatureLocation` and `CompoundLocation` objects. This allows to directly compare locations using `==` and `!=` instead of having to go via `list(loc_a) == list(loc_b)` or similar.

* This is strict in as much as a `BeforePosition(x)` or `AfterPosition(x)` is assumed to be equal to an `ExactPosition(x)`, but this matches our other non-string conversions of fuzzy features. This can lead to a situation where `loc_a == loc_b` is `True` but `str(loc_a) == str(loc_b)` is `False`. It does match the behaviour for comparisons via a `list` conversion, though.

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -1198,6 +1198,24 @@ class CompoundLocation(object):
             for pos in loc:
                 yield pos
 
+    def __eq__(self, other):
+        """Check if all parts of CompoundLocation are equal to all parts of other CompoundLocation"""
+        if not isinstance(other, CompoundLocation):
+            return False
+        if not len(self.parts) == len(other.parts):
+            return False
+        if not self.operator == other.operator:
+            return False
+        identical = True
+        for i, self_part in enumerate(self.parts):
+            other_part = other.parts[i]
+            identical = identical and self_part == other_part
+        return identical
+
+    def __ne__(self, other):
+        """Needed for py2, not needed for py3"""
+        return not (self == other)
+
     def _shift(self, offset):
         """Return a copy of the CompoundLocation shifted by an offset (PRIVATE)."""
         return CompoundLocation([loc._shift(offset) for loc in self.parts],
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -851,6 +851,16 @@ class FeatureLocation(object):
             for i in range(self._start, self._end):
                 yield i
 
+    def __eq__(self, other):
+        """Check if start, end and strand of this location are the same as in the other location"""
+        if not isinstance(other, FeatureLocation):
+            return False
+        return self._start == other.start and self._end == other.end and self._strand == other.strand
+
+    def __ne__(self, other):
+        """Needed for py2, not needed for py3"""
+        return not (self == other)
+
     def _shift(self, offset):
         """Return a copy of the FeatureLocation shifted by an offset (PRIVATE)."""
         # TODO - What if offset is a fuzzy position?
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -852,13 +852,18 @@ class FeatureLocation(object):
                 yield i
 
     def __eq__(self, other):
-        """Check if start, end and strand of this location are the same as in the other location"""
+        """Implement equality by comparing all the location attributes."""
         if not isinstance(other, FeatureLocation):
             return False
-        return self._start == other.start and self._end == other.end and self._strand == other.strand
+        return self._start == other.start and \
+            self._end == other.end and \
+            self._strand == other.strand and \
+            self.ref == other.ref and \
+            self.ref_db == other.ref_db
 
     def __ne__(self, other):
-        """Needed for py2, not needed for py3"""
+        """Implement the not-equal operand."""
+        # This is needed for py2, but not for py3.
         return not (self == other)
 
     def _shift(self, offset):
@@ -1199,21 +1204,21 @@ class CompoundLocation(object):
                 yield pos
 
     def __eq__(self, other):
-        """Check if all parts of CompoundLocation are equal to all parts of other CompoundLocation"""
+        """Check if all parts of CompoundLocation are equal to all parts of other CompoundLocation."""
         if not isinstance(other, CompoundLocation):
             return False
-        if not len(self.parts) == len(other.parts):
+        if len(self.parts) != len(other.parts):
             return False
-        if not self.operator == other.operator:
+        if self.operator != other.operator:
             return False
         identical = True
-        for i, self_part in enumerate(self.parts):
-            other_part = other.parts[i]
+        for self_part, other_part in zip(self.parts, other.parts):
             identical = identical and self_part == other_part
         return identical
 
     def __ne__(self, other):
-        """Needed for py2, not needed for py3"""
+        """Implement the not-equal operand."""
+        # This is needed for py2, but not for py3.
         return not (self == other)
 
     def _shift(self, offset):
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -864,7 +864,7 @@ class FeatureLocation(object):
     def __ne__(self, other):
         """Implement the not-equal operand."""
         # This is needed for py2, but not for py3.
-        return not (self == other)
+        return not self == other
 
     def _shift(self, offset):
         """Return a copy of the FeatureLocation shifted by an offset (PRIVATE)."""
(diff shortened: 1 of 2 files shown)


Bio/SeqFeature.py
```diff
@@ -851,6 +851,16 @@ def __iter__(self):
             for i in range(self._start, self._end):
                 yield i
 
+    def __eq__(self, other):
+        """Check if start, end and strand of this location are the same as in the other location"""
```
peterjc: Can you include a trailing full stop on the docstrings as per PEP257 please (likewise for the rest of the docstrings, see also #1170).

Bio/SeqFeature.py
```diff
@@ -851,6 +851,16 @@ def __iter__(self):
...
+        if not isinstance(other, FeatureLocation):
+            return False
+        return self._start == other.start and self._end == other.end and self._strand == other.strand
```
peterjc: We need to check ``.ref`` and ``.ref_db`` as well, which come up in things like assembly contig entries or trans-splicing. That would need a test too.

After which the docstring can be simplified to something like:

```python
"""Implement equality by comparing all the location attributes."""
```

Bio/SeqFeature.py
```diff
@@ -851,6 +851,16 @@ def __iter__(self):
...
+
+    def __ne__(self, other):
+        """Needed for py2, not needed for py3"""
```
peterjc: I'd prefer that as a comment, with something simple like this as the docstring:

```python
        """Implement the not-equal operand."""
```

Bio/SeqFeature.py
```diff
@@ -1188,6 +1198,24 @@ def __iter__(self):
...
+    def __eq__(self, other):
+        """Check if all parts of CompoundLocation are equal to all parts of other CompoundLocation"""
+        if not isinstance(other, CompoundLocation):
+            return False
+        if not len(self.parts) == len(other.parts):
+            return False
+        if not self.operator == other.operator:
+            return False
+        identical = True
+        for i, self_part in enumerate(self.parts):
```
peterjc: Would ``for self_part, other_part in zip(self.parts, other.parts):`` be nicer here given you've confirmed the have the same number of parts above? e.g.

```python
for self_part, other_part in zip(self.parts, other.parts):
    if not self_part == other_part:
        return False
return True
```

Bio/SeqFeature.py
```diff
@@ -1188,6 +1198,24 @@ def __iter__(self):
             for pos in loc:
                 yield pos
 
+    def __eq__(self, other):
+        """Check if all parts of CompoundLocation are equal to all parts of other CompoundLocation"""
+        if not isinstance(other, CompoundLocation):
+            return False
+        if not len(self.parts) == len(other.parts):
```
peterjc: Was this deliberate over using ``if len(self.parts) != len(other.parts):`` (and likewise for the rest of the method)?

Bio/SeqFeature.py
kblin: I personally prefer the more explicit form, as I always get confused by `zip()`. But if you prefer `zip`, I can certainly use that.

Bio/SeqFeature.py
kblin: Nah, that was just me being daft. :)

Bio/SeqFeature.py
kblin: sure.

Bio/SeqFeature.py
kblin: I actually didn't look at `FeatureLocation` as I initially started to compare `CompoundLocation`s, where both `ref` and `ref_db` are hardcoded to `None`. Didn't occur to me to look at `FeatureLocation`.

kblin: What data types are `.ref` and `.ref_db`? I don't see them used in any other tests.

Bio/SeqFeature.py
peterjc: I'd prefer the zip option here. Note just use ``zip`` as in this context it will work fine under Python 2 (it not be an iterator but the list memory overhead is not important).

Bio/SeqFeature.py
peterjc: They are strings or None IIRC. Do you fancy adding a test for those too while your looking at this? e.g. in ``Tests/one_of.gb`` we have locations like this:

```
     CDS             join(2201..2479,U18267.1:120..246,U18268.1:130..288,
                     U18270.1:4691..4788,U18269.1:82..>128)
                     /gene="TMPO"
```

kblin: What would you like having tested in this context?

Bio/SeqFeature.py
```diff
@@ -851,6 +851,21 @@ def __iter__(self):
...
+        return self._start == other.start and \
+            self._end == other.end and \
+            self._strand == other.strand and \
+            self.ref == other.ref and \
+            self.ref_db == other.ref_db
+
+    def __ne__(self, other):
+        """Implement the not-equal operand."""
+        # This is needed for py2, but not for py3.
+        return not (self == other)
```
peterjc: Your C training is showing perhaps - you don't need those brackets.

kblin: I've used the same style the `Seq` object `_ne_` function used that you linked me to on Twitter.

Bio/SeqFeature.py
```diff
@@ -1188,6 +1203,24 @@ def __iter__(self):
...
+        if not isinstance(other, CompoundLocation):
+            return False
+        if len(self.parts) != len(other.parts):
+            return False
+        if self.operator != other.operator:
+            return False
+        identical = True
+        for self_part, other_part in zip(self.parts, other.parts):
+            identical = identical and self_part == other_part
+        return identical
```
peterjc: Do you find this form with the local variable ``identical`` clearer than my suggestion, which would short-cut comparing all the parts as soon as one difference is found?

```
for self_part, other_part in zip(self.parts, other.parts):
    if self_part != other_part:
        return False
return True
```

Bio/SeqFeature.py
peterjc: I'd just check the attributes of the parts of the ``CompoundLocation`` for that CDS (or a similar feature) as created by the GenBank parser. For example the ``2201..2479`` bit should have ``.ref`` and ``.ref_db`` equal to ``None``, while the ``U18267.1:120..246`` bit would I think have ``.ref`` equal to ``U18267.1`` and ``.ref_db`` still ``None``.

(This bit of ``Bio.SeqFeature`` seems to have been designed with a bit more scope than is currently used in the GenBank/EMBL parsers)

Bio/SeqFeature.py
peterjc: Sheepish face - So you did https://github.com/biopython/biopython/blob/biopython-169/Bio/Seq.py#L198 - and it was me that wrote that line three years ago: 06c6c9d99fb5230b822be6000cc6bc404044cc38 - I'll fix that...

Bio/SeqFeature.py
kblin: Ah, I missed that part. I'll implement that as well.

Bio/SeqFeature.py
kblin: Seeing how I'll do another patch anyway, I'll also remove the brackets.

peterjc:
You can remove the ``identical`` variable now.

You are already in the ``CONTRIB.rst`` file, but we would add your name the ``NEWS.rst`` as contributing to Biopython 1.70.

I'm happy to come back to this later today, merge it, and do that final tidying myself - but if you want to update the pull request now, please go ahead. Thanks.

kblin:
Thanks, you beat me to it.

peterjc:
I've noticed that we *may* now want to update the ``Reference`` object's ``__eq__`` to also check for equality of the associated locations (e.g. as used in GenBank files). Do you have any objection to this plan? It is possible that the current behaviour is more useful if comparing references from multiple GenBank records where the genomes would be of different lengths and thus the reference locations would not be expected to match...

kblin:
Seeing how both the comment and the commit message strongly imply that the only reason the location isn't checked is due to `__eq__` not being defined on location objects, I don't see why not.

That said, I think my use case was to e.g. test that two records are identical, so I'm definitely not bothered by different lengths for the references. This might differ for other use cases.

peterjc:
OK, done in 8546b9c5bb22234b0c39ba427d6800b4e3266f3f

kblin:
:+1:

PR #1310, opened 2017-07-05
Rewrite Installation in RST

souravsingh:
A Work in progress fix for #1295 

I have also enabled editing for maintainers in case any of the maintainer wishes to push fixes to the documentation.


Doc/install/INSTALL.rst
```diff
@@ -0,0 +1,95 @@
...
+
+1) Unix/Linux
+-------------
+
+GCC is recommende as the C compiler, which is usually available as part of the standard set of packages on any Unix or Linux system.
+
+2) Mac OS X
+------------
+
+Install Apple’s XCode suite from the App Store, and then from the Xcode options also install the optional command line utilities.
```
peterjc: I believe you now do not need the full XCode suite, the command line tools alone is possible.

Doc/install/INSTALL.rst
```diff
@@ -0,0 +1,95 @@
...
+
+  $ python --version
+  
+This command will show the version of python2 present in your system. To check the exact version of python3 interpreter, type::
+
+  $ python3 --version
+    
+This command will show the version of python3 present in your system. If you aren't able to find the Python software available 
+in your Linux system, you can download Python from the site here- `https://www.python.org/downloads/source/ <https://www.python.org/downloads/source/>`_ and select the source gzipped tarball for the version you would like to install.
+For example, If you want to install Python 3.6 in your system, Look for the latest stable Python 3.6 release and Click on the "Download Gzipped source tarball".
```
peterjc: Do we really want to document how to install Python from source? Better pointing elsewhere?

souravsingh: I don't mind linking the python.org for instructions, but wouldn't it be an issue if we tell users to go to some link everytime for some software or for a package( like we can tell users to go to NumPy docs link for instead of explaining the installation ourselves).

peterjc: Likewise I don't want us to say any more than absolutely necessary about how to install NumPy.

(This is going to be much easier on Biopython 1.70 with pip automatically pulling in numpy if missing, see #1302 which is now merged)

souravsingh: So should we point people to this for installation of Python- https://wiki.python.org/moin/BeginnersGuide

souravsingh:
@peterjc I noticed the issue with Windows conda packages just now. In that case, should we maybe release a conda package for Biopython using conda-forge. The recipe creation process should be trivial, but I am worried about the conflict with the ``bioconda`` channel.

souravsingh:
I also suggest putting the conda instructions as an alternative step, since I believe not everyone uses conda.

peterjc:
Yes, already noted as a good idea on https://github.com/biopython/biopython.github.io/issues/118 but it applies to the main repository too.

souravsingh:
@peterjc The Biopython conda recipe is now ready here- https://github.com/conda-forge/biopython-feedstock

Doc/install/INSTALL.rst
```diff
@@ -159,9 +159,9 @@ This will install the biopython package in your system. For checking the install
 Installing Biopython using Anaconda
 ===================================
 
-For Anaconda users, the Biopython source files are available in the `bioconda <https://bioconda.github.io/>`_ channel. To install Biopython, just do::
+For Anaconda users, the Biopython source files are available in the `conda-forge <https://conda-forge.org/>`_ channel. To install Biopython, just do::
 
-  $ conda install -c bioconda biopython
+  $ conda install -c conda-forge biopython
```
peterjc: Isn't conda-forge a default channel? Can we just recommend: ``conda install biopython`` instead?

souravsingh: I checked the command ``conda install biopython`` in Windows, It still fetches the 1.68 binaries uploaded to default channel. 

``conda install -c conda-forge biopython`` works really well as it fetches the latest binaries for Windows which is uploaded to the ``conda-forge`` channel.

peterjc: Ah - this is likely due to the default channel priority setting. That's fine then, thanks for checking.

Doc/install/INSTALL.rst
souravsingh: I realised now that the page for Python Beginner's Guide is outdated and doesn't give exact instructions. Should we use the Django Girls page for Python installation?

souravsingh:
Do we need to make any more updates to the doc? I would like a review from someone in Biopython developer community.

Doc/install/INSTALL.rst
peterjc: See the XCode instructions we give on http://biopython.org/wiki/Download about running ``xcode-select --install``

Doc/install/INSTALL.rst
```diff
@@ -0,0 +1,195 @@
...
+
+2) Mac OS X
+------------
+
+Install Apple’s XCode suite from the App Store, and then from the Xcode options also install the optional command line utilities.
+
+3) Windows
+-----------
+
+We recommend you install Biopython and its dependencies using the provided pre-compiled Windows Installers. i.e. You don’t need a C compiler.
```
peterjc: This is now out of date, we provide wheel files but not installers.

Doc/install/INSTALL.rst
```diff
@@ -111,7 +111,7 @@ UNIX/Linux
 For installing Biopython, you will require a software package called ``pip`` to do so. If you are using the latest 3.x version of Python, you will have the package preinstalled in your UNIX/Linux system. You can check the version of the software by doing::
 
   $ pip --version
-  pip 9.0.1 from /usr/lib/python3.6/site-packages (python 3.6)
+  pip 9.0.1 from /usr/lib/python3.6/site-packages (python 3.5)
```
peterjc: This looks wrong - there is a mix of Python 3.6 and 3.5 here now (the old version was all Python 3.6)

Doc/install/INSTALL.rst
```diff
@@ -0,0 +1,232 @@
...
+  $ make
+
+Finally, you’ll need to have root permissions on the system and then install everything::
+  
+  $ make install
+  
+If there were no errors and everything worked correctly, you should now be able to type ``python`` at a command prompt and enter into the python interpreter::
+
+  $ python
+  Python 2.7.5 (...)
```
peterjc: Given the example above was installing Python 3.6.1, this is not consistent.

peterjc:
I still don't think we should describe how to install Python in our documentation, but just point at the official python.org information for that (and other popular options like conda).

souravsingh:
I didn't notice the consistency in Python version across the docs, that should be fixed by a quick correction.

As for pointing users for Python installation, I remember the docs being outdated on the installation ( the docs cover Python 3.3, plus there are many inconsistencies in the docs across platforms).

I suggest pointing users to Anaconda( at least that is what Software Carpentry does) for the latest Python distribution.

peterjc:
Yes, cleaning up all the inconsistencies in the current install documentation is part of #1295.

As part of this, the less extra and not important information we include, the less documentation have to maintain.

souravsingh:
Added link to Hitchhiker's Guide to Python for the installation of Python.

Doc/install/INSTALL.rst
```diff
@@ -19,7 +19,7 @@ GCC is recommende as the C compiler, which is usually available as part of the s
 2) Mac OS X
 ------------
 
-Install Apple’s XCode suite from the App Store, and then from the Xcode options also install the optional command line utilities.
+Install Apple’s XCode suite from the App Store, and then install the commandline tools using command- ``xcode-select --install`` on the terminal.
```
peterjc: You don't need the XCode suite at all anymore. I think the wording on http://biopython.org/wiki/Download is still accurate (I don't have a clean Mac to double check this on, those I have access to already have the command line tools).

Also missing  a space in "command line" and there is an extra minus/hyphen before the the command.

souravsingh: I wasn't sure earlier about this. I confirmed this right now on a friend's Mac. Running the command itself is enough for command-line tools and installing XCode also ensures that you get the command-line tools.

peterjc: Thanks for doing that - good to know what we're recommending on http://biopython.org/wiki/Download is still current for getting the compilers on macOS 👍

souravsingh:
Closing the PR since it is going nowhere

PR #1311, opened 2017-07-05
TravisCI: Build LaTeX documentation (Tutorial etc)

peterjc:
Rather than using Tox, this just uses the TravisCI config.

Build times seem quicker than our style checks, so having this as another basic check in the first stage seems best (rather than as part of the main test stage).

This doesn't (yet) do anything with the compiled PDF and HTML files, but ensures we don't have any undetected LaTeX syntax errors lurking until next time someone tries to build things (e.g. the next release).

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -78,12 +78,12 @@ deps =
 commands =
     #The bash call is a work around for special characters
     #The /dev/null is to hide the verbose output but leave warnings
-    bash -c \'python setup.py install > /dev/null\'
+    bash -c 'python setup.py install > /dev/null'
     #The bash call is a work around for the cd command
-    nocov: bash -c \'cd Tests && python run_tests.py --offline\'
+    nocov: bash -c 'cd Tests && python run_tests.py --offline'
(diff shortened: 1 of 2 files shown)


peterjc:
Note this depends on the Debian LaTeX and hevea packages being white-listed on the Travis setup, which they are for precise - but not yet on trusty - see https://github.com/travis-ci/apt-package-whitelist/issues/4199

peterjc:
Huh, didn't remember opening the pull request but just resurrected the branch. Current issue is hevea failing, most likely as the version with Ubuntu Xenial is too old.

It is my intension once this is working to add automatic deployment to the website, at least when the version tag is not a dev version (similar effect to only deploying on tagged commits, but more flexible). See the API deployment work.

Caiofcas:
Is this still a problem now that the repo uses github actions mainly?

peterjc:
This would need re-doing to work on GitHub Actions - the fundamental problem about how to quickly get a virtual machine with LaTeX remains a potential bottleneck.

PR #1314, opened 2017-07-11
self no longer re-defined in MutableSeq.count_overlap method

chris-rands:
I missed one example of re-defining `self` previously (see [this pull request](https://github.com/biopython/biopython/pull/1301)), apologies

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2054,10 +2054,10 @@ class MutableSeq(object):
         # The implementation is currently identical to that of
         # Seq.count_overlap() apart from the definition of sub_str
         sub_str = str(sub)
-        self = str(self)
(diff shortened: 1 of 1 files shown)


peterjc:
Thanks - Given we've just released Biopython 1.70, I was worried for a moment that the test coverage missed something, but this is just a style issue. Phew.

chris-rands:
Thanks, yes, I should have been clearer, it's purely a code style point (and also to mirror [the implementation](https://github.com/biopython/biopython/blob/master/Bio/Seq.py#L487) in `Seq.count_overlap`), won't impact any results or tests, sorry I missed it though

peterjc:
I missed this too during the previous pull request's review. Thanks for following up on this 👍

Issue #1317, opened 2017-07-12
Bio.Affy.CelFile doctest fails on Windows 64-bit Python 2.7


peterjc:
Reported in https://github.com/biopython/DIST/pull/3 by @benfulton, I could reproduce the doctest failure under AppVeyor as part of testing the wheel process:

https://ci.appveyor.com/project/biopython/biopython-wheels/build/1.0.10/job/ow4ualic5sha1fcy

```
======================================================================
FAIL: read_v4 (Bio.Affy.CelFile)
Doctest: Bio.Affy.CelFile.read_v4
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Miniconda-x64\envs\test_env\lib\doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for Bio.Affy.CelFile.read_v4
  File "C:\projects\biopython-wheels\biopython\build\lib.win-amd64-2.7\Bio\Affy\CelFile.py", line 137, in read_v4
----------------------------------------------------------------------
File "C:\projects\biopython-wheels\biopython\build\lib.win-amd64-2.7\Bio\Affy\CelFile.py", line 154, in Bio.Affy.CelFile.read_v4
Failed example:
    print(c.intensities.shape)
Expected:
    (5, 5)
Got:
    (5L, 5L)
----------------------------------------------------------------------
Ran 274 tests in 134.514 seconds
```

This was using 64-bit Python 2.7 via miniconda.

peterjc:
I am testing a simple change to the doctest to avoid this.

peterjc:
Fixed in #1318, specifically by commit https://github.com/biopython/biopython/commit/a1d14584f3fa8002b42ebcf9b79ff26fe82235d7

PR #1318, opened 2017-07-12
Include 64-bit Python 2 on Appveyor, fix doctest

peterjc:
This expands the AppVeyor testing to include 64-bit Python 2 which would have caught bug #1317 (doctest int/long issue), and then fixes that test.

It also addresses some minor docstring style issues in ``Bio.Affy.CelFile``


peterjc:
CodeCov reported a tiny drop in coverage in ``Bio/GA/Evolver.py``, but that's a stochastic bit of code so almost to be expected.

Issue #1320, opened 2017-07-13
test_Phylo.py UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3


peterjc:
Spin out from #855 which was specifically for ``test_NCBIXML.py`` but has the same root cause.

Some of the test XML files contain a non-ASCII accented character:

```
$ grep $'\xc3\xbc' PhyloXML/*.xml
PhyloXML/distribution.xml:            <desc>ETH Zürich</desc>
PhyloXML/phyloxml_examples.xml:                  <desc>ETH Zürich</desc>
```

Note while ``PhyloXML/distribution.xml`` fails to do so, ``PhyloXML/phyloxml_examples.xml`` does define an encoding,

```xml
<?xml version="1.0" encoding="UTF-8"?>
```

Testing with Biopython 1.70 with my default locale, everything is fine as a UTF8 encoding is the default. However, under some systems (including the ``multibuild`` systems for compiling wheels), you can get a default encoding of ascii.

The failure can be recreated under Python 3 as follows, here on Mac OS X using Python 3.6:

```
$ LANG=C python3 test_Phylo.py
test_convert (__main__.IOTests)
Convert a tree between all supported formats. ... ok
test_convert_phyloxml_binary (__main__.IOTests)
Try writing phyloxml to a binary handle; fail on Py3. ... ERROR
test_convert_phyloxml_filename (__main__.IOTests)
Write phyloxml to a given filename. ... ERROR
test_convert_phyloxml_text (__main__.IOTests)
Write phyloxml to a text handle. ... ERROR
test_format_branch_length (__main__.IOTests)
Custom format string for Newick branch length serialization. ... ok
test_int_labels (__main__.IOTests)
Read newick formatted tree with numeric labels. ... ok
test_newick_read_multiple (__main__.IOTests)
Parse a Nexus file with multiple trees. ... ok
test_newick_read_scinot (__main__.IOTests)
Parse Newick branch lengths in scientific notation. ... ok
test_newick_read_single1 (__main__.IOTests)
Read first Newick file with one tree. ... ok
test_newick_read_single2 (__main__.IOTests)
Read second Newick file with one tree. ... ok
test_newick_read_single3 (__main__.IOTests)
Read Nexus file with one tree. ... ERROR
test_newick_write (__main__.IOTests)
Parse a Nexus file with multiple trees. ... ok
test_phylo_read_extra (__main__.IOTests)
Additional tests to check correct parsing ... ok
[...]

Issue #1319, opened 2017-07-13
test_SeqIO_SeqXML.py UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3


peterjc:
Spin out from #855 which was specifically for ``test_NCBIXML.py`` but has the same root cause.

One of the test XML files contains non-ASCII accented characters (etc):

```
SeqXML/rna_example.xml:        <description>åÅüöÖßøä¢£$€香肠</description>
```

Note that this XML file does define an encoding:

```xml
<?xml version="1.0" encoding="UTF-8"?>
```

Testing with Biopython 1.70 with my default locale, everything is fine as a UTF8 encoding is the default. However, under some systems (including the ``multibuild`` systems for compiling wheels), you can get a default encoding of ``ascii``.

The failure can be recreated under Python 3 as follows, here on Mac OS X using Python 3.6:

```
$ LANG=C python3 test_SeqIO_SeqXML.py
test_duplicated_dbxref (__main__.TestDetailedRead)
Read multiple cross references to a single source ... ERROR
test_duplicated_property (__main__.TestDetailedRead)
Read property with multiple values ... ERROR
test_empty_description (__main__.TestDetailedRead)
Check empty description. ... ERROR
test_full_characters_set_read (__main__.TestDetailedRead)
Read full characters set for each type ... ERROR
test_global_species (__main__.TestDetailedRead)
Check global species. ... ERROR
test_local_source_definition (__main__.TestDetailedRead)
Check local source. ... ERROR
test_local_species (__main__.TestDetailedRead)
Check local species. ... ERROR
test_read_minimal_required (__main__.TestDetailedRead)
Check minimal record. ... ERROR
test_special_characters_desc (__main__.TestDetailedRead)
Read special XML characters in description. ... ERROR
test_unicode_characters_desc (__main__.TestDetailedRead)
Test special unicode characters in the description. ... ERROR
test_read_write_dna (__main__.TestReadAndWrite)
Read and write DNA. ... ok
test_read_write_globalSpecies (__main__.TestReadAndWrite)
Read and write global species. ... ok
test_read_write_protein (__main__.TestReadAndWrite)
Read and write protein. ... ok
test_read_write_rna (__main__.TestReadAndWrite)
Read and write RNA. ... ERROR
test_write_species (__main__.TestReadAndWrite)
Test writing species from annotation tags. ... ok
test_for_errors (__main__.TestReadCorruptFiles)
Handling of corrupt files. ... ok
[...]

PR #1322, opened 2017-07-13
Parse XML files in bytes mode to avoid encoding error

### LINKED ISSUE #855: test_NCBIXML.py UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3
Hi all,

`test_NCBIXML.py` fails with these errors (Python-3.5.1):

`UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 313: ordinal not in range(128)`

Full log: https://kojipkgs.fedoraproject.org//work/tasks/4147/14504147/build.log

peterjc:
This would resolve #855 and #1319 

A similar approach might work for #1320 and #1321 

For testing you need to (temporarily) change the locale settings, e.g. with ``LANG=C``, see detailed notes on #855.

### DIFF
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -579,9 +579,17 @@ def parse(handle, debug=0):
     BLOCK = 1024
     MARGIN = 10  # must be at least length of newline + XML start
     XML_START = "<?xml"
+    NEW_LINE = "\n"
+    NULL = ""
 
-    text = handle.read(BLOCK)
     pending = ""
+    text = handle.read(BLOCK)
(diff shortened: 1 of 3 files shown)


peterjc:
CodeCov is getting quite fussy (they have a moving goal post by default, the more we improve our base coverage, the stricter this check becomes).

It thinks we dropped two lines of coverage dealing with the ``SAXParseException``...

peterjc:
@smsaladi who helped debug this on #855 says he likes the binary mode idea: https://github.com/biopython/biopython/issues/855#issuecomment-315184239

peterjc:
Thanks Bow

PR #1327, opened 2017-07-20
Don't raise StopIteration from generators; return does this explicitl…

### LINKED ISSUE #967: Update use of StopIteration for PEP479
I'm using `biopython v1.68` in `anaconda 3.5.1` and am using a structure to read through a fastq file like so:

```
handle = gzip.open(path, "rt")
for record in SeqIO.parse(handle, "fastq"):
    do_stuff()
handle.close()
```

I keep getting this error and can't find anything about it in the source. Does anyone know what is going on here? Thanks!

> /Bio/SeqIO/QualityIO.py:1031: PendingDeprecationWarning: generator 'FastqGeneralIterator' raised StopIteration
> for title_line, seq_string, quality_string in FastqGeneralIterator(handle):

cbrueffer:
…y; fixes #967

### DIFF
--- a/Bio/AlignIO/NexusIO.py
+++ b/Bio/AlignIO/NexusIO.py
@@ -42,7 +42,7 @@ def NexusIterator(handle, seq_count=None):
     n = Nexus.Nexus(handle)
     if not n.matrix:
         # No alignment found
-        raise StopIteration
+        return
 
     # Bio.Nexus deals with duplicated names by adding a '.copy' suffix.
     # The original names and the modified names are kept in these two lists:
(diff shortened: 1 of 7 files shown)


peterjc:
Looks good - just waiting for AppVeyor to finish, just in case...

peterjc:
Note there are still lots more cases to review

bow:
@cbrueffer indeed this looks ok ~ I do also still notice that we still use the `raise` from other places (e.g. in `Bio/AlignIO/MauveIO.py` and `Bio/AlignIO/PhylipIO.py`).

cbrueffer:
I've had a look at all "raise StopIteration" cases in the codebase; I think the remaining one are needed (I could be missing something).

peterjc:
Thank you

cbrueffer:
Expanding on this, the cases @bow mentioned don't use yield, so I'm not sure whether this change applies there?

peterjc:
Good point. Best run with the ``PendingDeprecationWarning`` visible to find out for sure if Python thinks they need fixing...

Issue #563, opened 2015-06-18
GenomeDiagram example Proux_et_al_2002_Figure_6.py needs updating


TheChymera:
[This example file](https://github.com/biopython/biopython/blob/82272b1f36e3b7dd28296bde518b248db946b590/Doc/examples/Proux_et_al_2002_Figure_6.py) fails with:

```
Traceback (most recent call last):
  File "test.py", line 127, in <module>
    assert record.name not in feature_sets
AssertionError
```

peterjc:
Are you running the example unmodified? How did you get the sample GenBank files?

TheChymera:
Yes, unmodified. Got them from the GenBank website.

peterjc:
I can't yet explain the error message you have (unless it is a simple mistake like saving a record with the wrong filename?).

Note for ``NC_003212`` via the website you must ask for the FULL record in GenBank format, otherwise you get a summary record with no genes and a ``CONTIG`` entry instead of the sequence.

The annotation of this bacteria has been revised and the gene names in this script are now recorded under ``old_locus_tag``. However, even with the locus tag detection tweaked https://github.com/biopython/biopython/commit/8457989f090df0e304e83cabff6d8cdc1a61a88e , there is something different about the number of features (and thus the colours pre-assigned to them). I suspect something on the annotation has changed subtly...

sbliven:
This should be fixed as of 94c9d7e right?

peterjc:
Yep, thank you!

PR #1329, opened 2017-07-20
GenomeDiagram bug fixes

sbliven:
Fixes two bugs in GenomeDiagram.

1. Short arrows fill the full track width. If a feature was small enough not to have an arrow shaft, the display got messed up. This could be a bug with reportlab's handling of polygons with degenerate edges, but it is easy to work around.

arrow_bug2017-07-20-16 43 44

2. It was not possible to set 0 margins (using Diagram's x & y parameters). I also implement a slight change to the default fragment_size, so that if only one fragment is used (meaning the gene takes the full width of the image without wrapping) then it doesn't add extra padding.

```
gd_diagram.draw(format="linear",orientation="landscape", pagesize=(800,800), fragments=1,start=0,end=150,x=0,y=0,fragment_size=None )
```

Screenshot comparison (with random data):

margins


peterjc:
I thought short arrows are draw just as triangles (head only), but a test case would help be understand what is wrong here (doesn't have to be added to the ``Tests/`` - just an example to try out while reviewing this). e.g. Does this break on both circular and linear diagrams.

peterjc:
Updating the example for Proux et al  was filed as #563 - something on the annotation seemed to have changed.

peterjc:
@widdowquinn What do you think about the margin changes? I like the function for picking the first non-``None`` value. Would you prefer the default stays as ``0.9`` regardless of the number of fragments? That seems safer for backward compatibility.

widdowquinn:
I'm not as fond of the function solution as you are. Am I just being crotchety to think that:

`x or y`

is clearer and more concise, and less likely to introduce accidental new behaviour, than:

```
def _first_defined(*args):
    """Return the first non-null argument"""
    for arg in args:
        if arg is not None:
            return arg
    return None

_first_defined(x, y)
```

especially when there are 14 calls for each instantiation?

The neater solution I was thinking of at the time would probably be something like a `DrawerParameters()` object, so we'd have (pseudocode ahead):

```
class DrawerParameters():
  def __init__(self, p1=p1_default, p2=p2_default):
     self.p1 = p1
     self.p2 = p2

class Diagram():
  def __init__(self, **kwargs):
    self.params = DrawerParameters()
    for kwd, val in kwargs:
      # maybe also a sanity test here for whether the attribute is defined?
      setattr(self.params, kwd, val)

  def draw(self, **kwargs):
    for kwd, val in kwargs:
      # maybe also a sanity test here for whether the attribute is defined?
      setattr(self.params, kwd, val)
      if format == 'linear':
        # change linear specific arguments
      else:
        # change circular specific arguments
```

widdowquinn:
Aesthetically, I don't like that there's no margin/padding between the vertical zero marker and the image edge.

Since this is modifiable by the user (so if the user wants no margin, they can already have it) and, as you point out, backwards compatibility (of a trivial sort) breaks, I don't see an advantage from changing it.

peterjc:
The reason I dislike ``x or y`` is this idiom will skip any value of ``x`` which evaluates as false, which includes zero (sometimes a useful value - like a margin) as well as None (makes sense as a marker for no value specified).

widdowquinn:
I see your point, re: arguments equal to zero. `_first_defined()` isn't very neat IMO, but at least it avoids that.

I do think a separate `parameters` object/dictionary (and the corresponding design change) is a better option.

peterjc:
@widdowquinn pending further work to change the parameters model, does this pull request look OK to merge?

@sbliven did you have a test case for the short arrows issue? I want to check this on both linear and circular diagrams before merging.

peterjc:
I've cherry-picked the Proux *et al* (2002) Figure 6 fix as 94c9d7e5736b3072924a0ae77a72069b412a6d23

sbliven:
Apologies for not following up on this. When I got notification for @peterjc's comment I realized I had never posted the test.

Here is a test example for the arrow issue, both with the current pypi version and my branch:

[Test (biopython 1.70)](https://gist.github.com/sbliven/f0a549d3bb0f43cbebe36a057b109ed2)
[Test (genomediagram branch)](https://gist.github.com/sbliven/05455af038db957b61c15fea2a73e724)

Thanks for cherry-picking the Proux figure. I've merged master into this branch to reflect that (or would a rebase be preferred?).

@widdowquinn While zero margin does look funny, it was needed for my application because I was already adding padding around the element externally. Thus I really did need to differentiate `x=0` from the default value. I agree that most users would set a non-zero `x` parameter. I believe that without this patch it is not possible to have zero margin.

As to the `fragment_size` parameter, I can't think of any reason you would not want this set to 1 in the case of a single fragment (since there is no need for a divider between fragments, why would you add extra padding?). I think the change of default reflects this, while maintaining backwards compatibility with any code that had specifically tweaked the margins.

peterjc:
Thanks for the test case - so the short arrow head bug is specific to the linear diagrams. I forget now how different the code is here (but the circular arrows are more complicated as the arrow head edges are not actually straight lines).

A rebase would be preferable but as I'd likely rebase-and-merge this anyway, or squash-merge, it isn't pressing.

Regarding margins, for horizontal figures the labels can be at an angle thus spill into the margin - so a non-zero default margin is sensible. I think leaving the defaults as they are would be best - while supporting zero margins as a new feature.

(It was a long time ago, but I remember wanting to use zero-margins for GenomeDiagram in one usecase...)

sbliven:
@peterjc This patch does not set zero-margins for any default. The default `x` and `y` margins remain at 5%. It just removes the additional 10% vertical padding that was reserved as spacing between fragments (but included also before the first fragment and after the last one) if only a single fragment is present.

peterjc:
Thanks for the clarification. I know @widdowquinn is away right now, so unfortunately we can't expect an immediate reply on that side of things. Time permitting I will look at the arrow head on my own though.

sbliven:
I tried to get travis to pass, but I can't reproduce the error. `tox -c .travis-tox.ini -e style` succeeds on my machine, and the line travis mentions is a comment (Bio/Graphics/GenomeDiagram/_LinearDrawer.py:115:1: RST203 Definition list ends without a blank line; unexpected unindent.)

peterjc:
Ah, the ``flake8`` errors with the ``RST`` prefix are from (my) RST docstring plugin, you'd need ``pip install flake8-rst-docstrings``:

https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

This error can "solved" with a blank line at the end of the docstring, but I do wonder if this is a false positive: https://github.com/peterjc/flake8-rst-docstrings/issues/1

peterjc:
Confirmed - that was a false positive about the docstring wanting a blank line at the end - https://github.com/peterjc/flake8-rst-docstrings/issues/1 fixed by flake8-rst-docstrings v0.0.7

peterjc:
Mixed news. We had a test case for short sigils but it was only being used for the ``ARROW`` sigil until 68c5cdaccfc2426a5bcd7ac89ff3dcf5ed984a26 where I generalised this to include the ``BIGARROW``, ``JAGGY`` and ``OCTO`` as well.

The red sigils are so small arrows can become just a triangle for the linear diagram (this special case is currently only dealt with in the circular drawer):

gd_sigil_short_arrow

In this test case however, the ``BIGARROW`` still looks fine (it is meant to cover the grey strand specific boxes and the blue boxes on the opposite strand), so there is something important about the aspect ratio or something in your example this is mising:

gd_sigil_short_bigarrow

For the jaggy sigil it just changes the angles to squeeze into the space:

gd_sigil_short_jaggy

However, this shows similar corner case with linear diagrams, tiny features and the ``OCTO`` sigil (the red ones spill out over the grey box):

gd_sigil_short_octo

Unfortunately with your fix my ``BIGARROW`` test breaks for negative strand features (where ``headbase`` is expected to be negative):

gd_sigil_short_arrow

How about something like this as the condition?

```python
if abs(headlength) < 0.99 * abs(boxwidth):
```

peterjc:
In 3b642ddaba7cdeecb14432af4b104386ca18b2f8 I have hopefully fixed the ``OCTO`` sigil to match the existing behaviour on circular diagrams for small features,

gd_sigil_short_octo

sbliven:
Nice additional tests, @peterjc.

How about this?
```
    if (headbase if orientation == "right" else -headbase) > 0:
```

I'm still seeing a problem with the PNG output if `arrowshaft_height=1.0`. Maybe a better solution would be to just de-duplicate the polygon vertices.

sbliven:
I have a fixed & squashed version #1385

PR #1332, opened 2017-07-21
Save a PDB file to a string

jgreener64:
It would be nice to be able to save a PDB file to a string as well as a file using the same function. This is currently possible on Python 2:

```python
from Bio.PDB import PDBParser, PDBIO
from StringIO import StringIO

parser = PDBParser()
structure = parser.get_structure("1AKE", "1AKE.pdb")

io_pdb = PDBIO()
io_pdb.set_structure(structure)
io_str = StringIO()
io_pdb.save(io_str)

io_str.getvalue()
```

However `StringIO` was updated and in Python3 expects unicode strings, so the above doesn't work. Converting the strings to unicode in `save` as in this PR makes the above work (the import changes to `from io import StringIO`). Let me know if this is the right way to get round the problem.

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -194,7 +194,7 @@ class PDBIO(object):
             if not preserve_atom_numbering:
                 atom_number = 1
             if model_flag:
-                fp.write("MODEL      %s\n" % model.serial_num)
+                fp.write(u"MODEL      %s\n" % model.serial_num)
             for chain in model.get_list():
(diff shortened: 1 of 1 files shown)


jgreener64:
Just spoken to @peterjc and sorted this.

peterjc:
Within the Biopython codebase we use ``from Bio._py3k import StringIO`` to deal with this Python 2 vs 3 difference. The external library ``six`` can also be useful in this kind of situation.

PR #1333, opened 2017-07-21
We will drop support for Python 2.7 no later than 2020

peterjc:
See http://www.python3statement.org/ and https://www.python.org/dev/peps/pep-0373/#update

Proposed on the mailing list, see http://mailman.open-bio.org/pipermail/biopython-dev/2017-June/021794.html


cbrueffer:
:+1:

peterjc:
Typo: ``end-of-list`` --> ``end-of-life``

PR #1334, opened 2017-07-21
Small changes to Bio.PDB

jgreener64:
On looking through the Bio.PDB module to get the hang of it I corrected a few style things that could be merged, though they are not essential:

- Correct some spelling
- Remove unused variables/imports
- Remove redundant brackets

### DIFF
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -31,8 +31,7 @@ from Bio.PDB.parse_pdb_header import _parse_pdb_header_list
 class PDBParser(object):
     """Parse a PDB file and return a Structure object."""
 
-    def __init__(self, PERMISSIVE=True, get_header=False,
-                 structure_builder=None, QUIET=False):
+    def __init__(self, PERMISSIVE=True, structure_builder=None, QUIET=False):
         """Create a PDBParser object.
 
         The PDB parser call a number of standard methods in an aggregated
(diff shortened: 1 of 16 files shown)


Bio/PDB/PDBParser.py
```diff
@@ -31,8 +31,7 @@
 class PDBParser(object):
     """Parse a PDB file and return a Structure object."""
 
-    def __init__(self, PERMISSIVE=True, get_header=False,
-                 structure_builder=None, QUIET=False):
+    def __init__(self, PERMISSIVE=True, structure_builder=None, QUIET=False):
```
peterjc: That looks like an API change (removing the ``get_header`` option) which is not backwards compatible.

Bio/PDB/QCPSuperimposer/__init__.py
```diff
@@ -87,7 +87,7 @@ def set(self, reference_coords, coords):
         self.coords = coords
         n = reference_coords.shape
         m = coords.shape
-        if n != m or not(n[1] == m[1] == 3):
+        if n != m or not n[1] == m[1] == 3:
```
peterjc: Is this definitely equivalent? The comparison of three terms is in itself perhaps not the best style...

peterjc:
With this second and larger contribution, are you happy to be thanked by name in the NEWS and CONTRIB file? Thanks!

jgreener64:
Yes sounds good! Thanks for the merge.

PR #1335, opened 2017-07-21
Move self-test for Genbank to unittest

souravsingh:
The PR fixes a part of #820 by moving the ``Bio.GenBank.Scanner`` self-test to a unitest.


Bio/GenBank/Scanner.py
```diff
@@ -1653,374 +1653,5 @@ def _feed_misc_lines(self, consumer, lines):
...
-
-    print("")
-    print("EMBL Iteration")
-    print("==============")
-    e = EmblScanner()
-    for record in e.parse_records(StringIO(embl_example), do_features=True):
-        print("%s %s %s" % (record.id, record.name, record.description))
-        print(record.seq)
+    from Bio._utils import run_doctest
+    run_doctest(verbose=0)
```
peterjc: Are there any doctests in this file? If not, there is no reason to having these three lines of code.

souravsingh: There was a self test in the file, which I have migrated as it's own unittest.

peterjc: If there are no doctests, we don't need this:

```python
if __name__ == "__main__":
    run_doctest(verbose=0)
```

souravsingh:
@peterjc There are print statements in the unittest. Can you guide me through it?

peterjc:
Basically replace the print lines with a unit test method checking something similar, e.g.

```python
        print("%s %s %s" % (record.id, record.name, record.description))
```

becomes:

```python
        self.assertEqual(record.id, "FILL IN THIS STRING")
        self.assertEqual(record.name, "FILL IN THIS STRING")
        self.assertEqual(record.description, "FILL IN THIS STRING")
```

while for:

```python
        print(record.seq)
```

you might just do something like this:

```python
        self.assertEqual(len(record.seq), 12345)
```

peterjc:
You'll also need to make sure ``flake8`` is happy:

```
$ flake8 Bio/
Bio/GenBank/Scanner.py:1657:1: W293 blank line contains whitespace
Bio/GenBank/Scanner.py:1657:1: W391 blank line at end of file
```

PR #1339, opened 2017-07-24
Also clean aux files created in Doc/Tutorial

smoe:
The include instructions in Doc/tutorial.tex each spawn a corresponding aux file.

Just something I noticed for repeated builds of the Debian package.

Cheers,

Steffen

### DIFF
--- a/Doc/Makefile
+++ b/Doc/Makefile
@@ -33,6 +33,7 @@ clean: clean-subdirs
 	rm -f biopdb_faq.aux
 	rm -f biopdb_faq.log
 	rm -f biopdb_faq.out
+	rm -f Tutorial/*.aux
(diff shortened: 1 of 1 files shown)


peterjc:
Good point - we missed this when breaking up the Tutorial into per-chapter files.

smoe:
Happy with the dual licence. It was not expected but I'd be lucky and happy should you decide to add my name to any such honorable list over that one line patch. Maybe just wait for another patch of mine down the road.

peterjc:
Thanks Steffen - if you prefer, sure, we can wait for a more substantial patch before adding your name explicitly as a contributor. Thank you for agreeing to the licensing.

Issue #754, opened 2016-01-18
Tutorial use of urlopen, bytes and Python 3


peterjc:
As noted on #751, there are a couple of uses of `urlopen` in the tutorial to get a handle to feed to a text based parser. These need at least an updated import, e.g.

```
>>> import urllib
>>> handle = urllib.urlopen("http://www.somelocation.org/data/someswissprotfile.dat")
```

to (*fixed error python 2 example, could also use urllib2*):

```
>>> from urllib import urlopen  # Python 2
>>> from urllib.request import urlopen  # Python 3
>>> handle = urlopen("http://www.somelocation.org/data/someswissprotfile.dat")
```

However, the subtlety here is it will be a binary handle giving bytes not (unicode) strings. Would we also need to wrap the handle in `io.TextIOWrapper` (or our `EvilHandleHack` workaround for Python 3.3, see #723) as done internally by the `Bio.Entrez` and `Bio.TogoWS` interfaces?

i.e. something like this for Python 3:

```
>>> from urllib.request import urlopen
>>> from io import TextIOWrapper
>>> handle = TextIOWrapper(urlopen("http://www.somelocation.org/data/someswissprotfile.dat"))

```

peterjc:
We've dropped Python 3.3 support, so don't need to worry about that special case.

Gasta88:
I'll be happy to have a look at it. 

What tutorial docs are we referring to? the [biopython cookbook?](http://biopython.org/DIST/docs/tutorial/Tutorial.html)

peterjc:
Yes, source to which is these per-chapter LaTeX files:

https://github.com/biopython/biopython/tree/master/Doc/Tutorial

You can compile it with the ``Makefile`` in the main ``Doc`` directory, i.e.

```
cd Doc
make
```

Note that would need ``pdflatex`` (for the PDF) and ``hevea`` (for the HTML) installed.

Note we have an ad hoc doctest system used for testing many of the examples, see ``Tests/test_Tutorial.py`` for details.

Gasta88:
Thanks.
I'm on Ubuntu 17.04 and I've installed `texlive`, `pdflatex` and `hevea`.

Following your instructions I get several warnings when I run `make` but it still compile the files and I get the `.pdf` and `.html` ones. Is it normal?

peterjc:
[...]

Issue #751, opened 2016-01-13
Tutorial problems - uniprot


tiagoantao:
Problems on uniprot

1. There is no myswissprotfile.dat.gz (much code is hypothetical)

2. urllib.urlopen does not work on Python 3

3. myprositefile.dat does not exist

4. prosite.dat does not exist

5. mysingleprositerecord.dat does not exit

6. prosite.doc ...

7. enzyme.dat

8. myprositerecord.html

Again, the chapter seems more hypothetical that usable.

Gasta88:
Hi @tiagoantao, this is a long shot but I've recently had a look at this chapter and came to these conclusions:
1) Given that we want to maintain at the minimum the amount of files in `Doc/examples` I would suggest to leave those files that are not further manipulated in the doctests. In particular I'm referring to `myswissprotfile.dat.gz`, `mysingleprositerecord.dat` and `myprositerecord.html`.

2) In order to make `urlopen` work, I've uploaded a `.dat`. file. This can be used for other doctests where the files are further manipulated. Other small files that can be uploaded should be `enzyme.dat` and `prosite.doc`.

With other users consent I'll be happy to have a look at it.

peterjc:
@Gasta88 What do you mean for getting ``urlopen`` to work with a sample ``.dat`` file in the repository?

Regarding Prosite, we have potentially suitable test files under ``Tests/Prosite/`` and ``Tests/Prosite/Doc/``

PR #1340, opened 2017-07-25
add new example for Uniprot Tutorial

Gasta88:
This new example is for the part of the tutorial where `urlopen()` is used.
This PR is in reference to issue #754 and #751.


peterjc:
So far this pull request adds a new SwissProt plain text file ``Doc/examples/Q9BZS1.dat`` but as far as I can see nothing refers to the new file anywhere.

You are planning to update existing examples which refer to generic files to use this specific example?

peterjc:
The  TravisCI failure is unrelated, http://www.atgc-montpellier.fr/download/binaries/phyml/PhyML-3.1.zip was down (link works for me).

Gasta88:
You are right, I decided to upload the file first and then use the URL in the `chapter_uniprot.tex` urlopen example. This will be another PR.

I did this because in this way I can safely test the URL for the .dat file now incorporate the `master`.

Do you prefer to have everything into one single PR?

peterjc:
One PR adding the example file and using it in the Tutorial would make sense to me. Can you add the tutorial changes to this PR?

Doc/Tutorial/chapter_uniprot.tex
```diff
@@ -21,10 +21,15 @@ \subsection{Parsing Swiss-Prot records}
...
 \begin{verbatim}
->>> import urllib
->>> handle = urllib.urlopen("http://www.somelocation.org/data/someswissprotfile.dat")
+>>> from future.standard_library import install_aliases
+>>> install_aliases()
+>>> 
+>>> from urllib.request import urlopen
+>>> from Bio import SwissProt
+>>> 
+>>> handle=urlopen("https://raw.githubusercontent.com/biopython/biopython/master/Doc/examples/Q9BZS1.dat")
```
peterjc: Will this actually give you text mode on Python 3?

peterjc:
I was thinking more something like this:

```latex
..., or
% TODO: The urlopen example needs to be in text mode for Python 3..                                                                 
\begin{verbatim}                                                                                                                    
>>> import urllib # Python standard library                                                                                         
>>> handle = urllib.urlopen("http://somewhere.org/mydatafile.txt")                                                                  
\end{verbatim}
to open a file stored on the Internet before calling \verb|read|.
```

becomes:

```latex
..., or
\begin{verbatim}                                                                                                                    
>>> from urllib.request import urlopen  # Python 3 only
>>> from io import TextIOWrapper
>>> handle = TextIOWrapper(urlopen("http://somewhere.org/mydatafile.txt"))
\end{verbatim}
to open a file stored on the Internet before calling \verb|read|.

Note for Python 2 this can be done more concisely using:
\begin{verbatim}
>>> from urllib2 import urlopen  # Python 2                                                                                                                    
>>> handle = urlopen("http://somewhere.org/mydatafile.txt")                                                                  
\end{verbatim}
```

Note that is untested, we might need to use ``TextIOWrapper(BufferedReaderurlopen("http://somewhere.org/mydatafile.txt")))`` as in https://github.com/biopython/biopython/blob/biopython-170/Bio/_py3k/__init__.py#L101

peterjc:
See also https://github.com/biopython/biopython/commit/fa080ea34c98cb7d773de919b5956937ca599d1f

Gasta88:
I see, by using `http://example.org/<FILE>` there won't be any need to upload a `.dat`. file to make the example work.

This clarify my comment in issue #751 as well.

peterjc:
We may not have understood each other. The domain ``example.org`` is a special domain for use in documentation. The point of https://github.com/biopython/biopython/commit/fa080ea34c98cb7d773de919b5956937ca599d1f was not to use some random group's real website, but one of the special example domain names.

The example itself won't literally work as written - the ``example.org`` domain will not return a useful file for our code to parser.

Gasta88:
Apologies, I thought that the main reason for this issue and 751 was to make the chapter less hypothetical by using real examples/ files.
If the scope is to have examples that won't run as doctests, I'll stick with changing the syntax according to the Python version.

peterjc:
The point of @tiagoantao's #751 is indeed to replace generic examples with specific examples.

I was mentioning commit fa080ea as it also altered one of the examples in question.

Gasta88:
Would it be OK to use `Tests/SwissProt/sp001` to have a real example for the urlopen example?

peterjc:
How about using one of the more naturally named examples in that folder like ``F2CXE6.txt`` (for which we also have the UniProt XML file which could be useful for comparison)?

i.e. using something like ``urlopen("https://raw.githubusercontent.com/biopython/biopython/master/Tests/SwissProt/F2CXE6.txt")``

Doc/Tutorial/chapter_uniprot.tex
```diff
@@ -25,15 +25,15 @@ \subsection{Parsing Swiss-Prot records}
 >>> from urllib.request import urlopen # Python 3 only
 >>> from io import TextIOWrapper
 >>> 
->>> handle = TextIOWrapper(urlopen("http://example.org/myprositefile.dat"))
+>>> handle = TextIOWrapper(urlopen("https://raw.githubusercontent.com/biopython/biopython/master/Tests/SwissProt/F2CXE6.txt"))
```
peterjc: That's a SwissProt text file, for a Prosite file maybe use ``Tests/Prosite/ps00107.txt``

Doc/Tutorial/chapter_uniprot.tex
```diff
@@ -25,15 +25,15 @@ \subsection{Parsing Swiss-Prot records}
...
 \end{verbatim}
 to open the file stored on the Internet before calling \verb|read|.
 
 Note for Python 2 this can be done by using:
 \begin{verbatim}
 >>> from urllib import urlopen #Python 2 only
 >>>
->>> handle = urlopen("http://example.org/myprositefile.dat")
+>>> handle = urlopen("https://raw.githubusercontent.com/biopython/biopython/master/Tests/SwissProt/F2CXE6.txt")
```
peterjc: Again, that's a SwissProt text file, this should be a Prosite file

peterjc: Never mind, reading the text it ought to have been a SwissProt file in the first place.

Doc/Tutorial/chapter_uniprot.tex
peterjc: Sorry, never mind, reading the text about this example ought to have been a SwissProt text file in the first place.

Doc/Tutorial/chapter_uniprot.tex
```diff
@@ -57,16 +57,20 @@ \subsection{Parsing Swiss-Prot records}
 %TODO - Check the single quotes when printing record.description here:
 \begin{verbatim}
 >>> print(record.description)
-'RecName: Full=Chalcone synthase 3; EC=2.3.1.74; AltName: Full=Naringenin-chalcone synthase 3;'
```
peterjc: What was the ``myaccessionnumber`` value used earlier for this:

```
 >>> from Bio import ExPASy
 >>> handle = ExPASy.get_sprot_raw(myaccessionnumber)
```

```
 >>> from Bio import SwissProt
 >>> record = SwissProt.read(handle)
```

```
>>> print(record.description)
```

Can we make this into an online Tutorial doctest using the special comment line ``%doctest . internet``?

Doc/Tutorial/chapter_uniprot.tex
Gasta88: There is also another issue with the file that you mentioned (since they were my first choice):

```python
handle = TextIOWrapper(urlopen("https://raw.githubusercontent.com/biopython/biopython/master/Tests/Prosite/ps00107.txt"))
>>> record = SwissProt.read(handle)
  File "/home/gasta/.local/lib/python3.5/site-packages/Bio/SwissProt/__init__.py", line 145, in read
    record = _read(handle)
  File "/home/gasta/.local/lib/python3.5/site-packages/Bio/SwissProt/__init__.py", line 182, in _read
    _read_id(record, line)
  File "/home/gasta/.local/lib/python3.5/site-packages/Bio/SwissProt/__init__.py", line 294, in _read_id
    raise ValueError("ID line has unrecognised format:\n" + line)
ValueError: ID line has unrecognised format:
ID   PROTEIN_KINASE_ATP; PATTERN.
```
Probably a different module from `Bio.SwissProt` must be used.

Doc/examples/Q9BZS1.dat
```diff
@@ -0,0 +1,739 @@
+ID   FOXP3_HUMAN             Reviewed;         431 AA.
```
peterjc: Why are you adding this file?

Doc/Tutorial/chapter_uniprot.tex
Gasta88: there was no `myaccessionnumber` specified before.

I've refrained to make the test online since I remember you mentioned a problem with the CI on Windows.

Doc/Tutorial/chapter_uniprot.tex
peterjc: Yes, the text and example code did not match up with the old example URL - see ``Bio.ExPASy.Prosite`` for dealing with ``Tests/Prosite/ps00107.txt``.

I was confused too.

Doc/examples/Q9BZS1.dat
Gasta88: It is a mistake from my part, but I've removed it from my branch weeks ago and it shouldn't be here.

I'll look into removing this file from the PR.

peterjc: Thanks

Doc/Tutorial/chapter_uniprot.tex
peterjc: Right now we don't run the online tests under TravisCI or AppVeyor, so it should be fine as long as it works on your machine.

Gasta88:
I've place the tag `%doctest . internet` for the `urlopen` example in Python 2 and 3.
Running `test_Tutorial` in each version will always throw an error for the opposite one.

Even if I silence one of the two, Travis will throw an error for the testing in the opposite version of Python.  
Any idea on how can I solve this?

peterjc:
@Gasta88 options:

1. add yet another special case to ``test_Tutorial.py`` for declaring Python version requirements
2. use a try/except or if statement as part of the doctest (which perhaps makes the example more confusing)
3. comment out the ``%doctest ...`` line until we drop Python 2

PR #1341, opened 2017-07-26
Improving UnknownSeq.count() when it is used with variable start and end args

chris-rands:
In pull request #1301 it was mentioned that `UnknownSeq.count()` could be improved, because when variable start and/or end argument are used, the current implementation converts the entire sequence to a string. This will have a substantial memory burden for long sequences. The comments in the current code also hint at this: `# This could be done more cleverly...` (for example [here](https://github.com/chris-rands/biopython/blob/master/Bio/Seq.py#L1429)).

Since Peter and I implemented `UnknownSeq.count_overlap()` in [97709cc](https://github.com/biopython/biopython/commit/97709cc7a4b8591e794b442796d41e4f7b855a0f), I knew a similar approach should work for `UnknownSeq.count()` and this is what is implemented here.

The new implementation passes the original doctests and the tests in `Tests/test_Seq_objs.py`; no further tests have been added. Someone should carefully check the logic and ensure that I haven't done anything stupid.

Note, I haven't done any formal benchmarks of speed or memory usage of the old vs. suggested new implementations, I don't know if this is necessary or indeed handled by Biopython's own tests.

Please advise how to proceed, thank you!

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -1421,24 +1421,29 @@ class UnknownSeq(Seq):
         1
         """
         sub_str = self._get_seq_str_and_check_alphabet(sub)
-        if len(sub_str) == 1:
-            if str(sub_str) == self._character:
-                if start == 0 and end >= self._length:
-                    return self._length
-                else:
-                    # This could be done more cleverly...
-                    return str(self).count(sub_str, start, end)
-            else:
-                return 0
-        else:
-            if set(sub_str) == set(self._character):
-                if start == 0 and end >= self._length:
-                    return self._length // len(sub_str)
-                else:
-                    # This could be done more cleverly...
-                    return str(self).count(sub_str, start, end)
-            else:
-                return 0
+        len_self, len_sub_str = self._length, len(sub_str)
+        # Handling case where substring not in self
+        if set(sub_str) != set(self._character):
+            return 0
+        # Setting None to the default arguments
+        if start is None:
+            start = 0
+        if end is None:
+            end = sys.maxsize
+        # Truncating start and end to max of self._length and min of -self._length
+        start = max(min(start, len_self), -len_self)
+        end = max(min(end, len_self), -len_self)
+        # Convert start and ends to positive indexes
+        if start < 0:
+            start += len_self
+        if end < 0:
+            end += len_self
+        # Handle case where end <= start (no negative step argument here)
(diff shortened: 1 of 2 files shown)


peterjc:
That looks good - once the tests have run we should be able to check the resulting code coverage report, and make sure all branches of the changed code are exercised by the existing test suite, and if so this looks ready to merge 👍

chris-rands:
Thanks Peter, I note the coverage of `Bio/Seq.py` has decreased slightly; I think because the tests never explicitly feed `start` or `end` values equal to `None`, but I'm not sure this is an issue, or I could add tests? Otherwise, how do you interpret the checks? Why did the Travis CI build fail?

peterjc:
The TravisCI build failed on Python 3.6 due to a download issue, I restarted that.

You're right, we don't explicitly test using ``start=None`` or ``end=None``:

https://codecov.io/gh/biopython/biopython/compare/789b58cfccf411b98a7f20b5544f4f13ffc496cd...244f3167a2f8835a53e9c8b462086a342f03b5b8/src/Bio/Seq.py#L1429

It would be good to add those to the unit test.

chris-rands:
Thanks, these tests should apply just for  [`test_str_count`](https://github.com/biopython/biopython/blob/master/Tests/test_Seq_objs.py#L180) or more broadly? The `.count_overlap()` methods already have unit tests including `None` for `start` or `end` values, but otherwise I don't think it is tested in `Tests/test_Seq_objs.py`.

My initial thought was just to add `None` to [` _start_end_values`](https://github.com/biopython/biopython/blob/master/Tests/test_Seq_objs.py#L100). However, [`test_str_getitem`](https://github.com/biopython/biopython/blob/master/Tests/test_Seq_objs.py#L488) isn't currently set-up to handle such tests with `None`, so that would need some re-factoring if we take that approach I guess?

peterjc:
Have a play, what ever you think works best should be fine - even as simple as adding a couple of test methods with constant arguments just for these corner cases (without trying to be too clever with lists of alternative arguments).

chris-rands:
Thanks, in fact I think I found a simple solution (just committed), let me know what you think

peterjc:
Good job - that looks nice. Let's wait for the tests to finish and confirm the code coverage...

peterjc:
Thank you!

PR #1342, opened 2017-07-26
removed selftest from Bio/KDTree/KDTree.py

Gasta88:
In reference to issue #820, I've removed the self test from this module. The test has been developed 6 months ago and sits correctly in `Tests/test_KDTree.py`


peterjc:
Did you mean to include additions to ``Tests/test_KDTree.py`` as part of this pull request?

(I am not sure why AppVeyor is confused - GitHub says this branch can be merged?)

Gasta88:
Yes, GitHub didn't display any problem.

I didn't add `Tests/test_KDTree.py` because it's already in place.

peterjc:
The current ``Tests/test_KDTree.py`` does not seem to cover all the functionality used in the old self test. e.g. ``get_radii``

This is confirmed based on the current coverage reports - see https://codecov.io/gh/biopython/biopython/src/master/Bio/KDTree/KDTree.py

Gasta88:
Thanks for pointing that out, I have one more query for this unittest.

for this module the creator wrote the testing methods inside the module itself, calling them later in `Tests/test_KDTree.py`.

Can I follow the same example or would it be best to move all the methods into the testing script?

peterjc:
I would suggest moving those private test functions into the test suite (which can be done later, or as part of this pull request if you prefer).

Tests/test_KDTree.py
```diff
@@ -12,31 +12,142 @@
     raise MissingExternalDependencyError(
         "Install NumPy if you want to use Bio.KDTree.")
 
-try:
```
peterjc: Please lease the try/except - this is still used on Jython to skip this test.

Gasta88:
The method `test_all_search` might raise few questions.

I've decided to test the `all_get_radii` and `all_get_indices` methods, but the computation is taking quite a toll on my machine. Therefore I've taken into account a subset of 100 iterations to test only.

The `test_KDTree_exceptions` is classic fuzzy testing to try few exceptions on the class.

peterjc:
It seems the tests are timing out under PyPy (Python 2 and 3), can you try that locally? Perhaps even fewer iterations are needed.

Gasta88:
Took me 20 seconds on my machine, what's the timeout limit for TravisCI?

peterjc:
Its the time out for ``run_tests.py --offline`` we need to worry about.

Gasta88:
Travis CI seems to have timed out for Python 3. 
I've checked with both versions (2.7 and 3) on my machine with the `run_tests.py --offline` and everything looks ok.

peterjc:
I've had a play with the ``travis_wait`` command, e.g. https://travis-ci.org/peterjc/biopython/jobs/261067283 - not yet working as I'd like

It may be more practical to ask TravisCI nicely if they could increase the timeout for the biopython repository's builds...

peterjc:
The ``travis_wait`` change in now included on the master branch via #1359 so that should help here with timeouts.

peterjc:
It looks like there is a Python 3 specific problem:

```
ERROR: test_all_search (test_KDTree.KDTreeTest)
  File "/home/travis/build/biopython/biopython/Tests/test_KDTree.py", line 199, in test_all_search
    self.assertTrue(test_all_search((nr_points / 10), dim, bucket_size, query_radius))
  File "/home/travis/build/biopython/biopython/Tests/test_KDTree.py", line 159, in test_all_search
    coords = random.random((nr_points, dim))
  File "mtrand.pyx", line 856, in mtrand.RandomState.random_sample (numpy/random/mtrand/mtrand.c:15541)
  File "mtrand.pyx", line 167, in mtrand.cont0_array (numpy/random/mtrand/mtrand.c:6141)
TypeError: 'float' object cannot be interpreted as an integer
```

Tests/test_KDTree.py
```diff
@@ -195,7 +195,8 @@ def test_KDTree(self):
             self.assertTrue(test(nr_points, dim, bucket_size, radius))
 
     def test_all_search(self):
-        self.assertTrue(test_all_search(nr_points, dim, bucket_size, query_radius))
+        for i in range(0, 5):
+            self.assertTrue(test_all_search((nr_points / 10), dim, bucket_size, query_radius))
```
peterjc: Use explicit integer division here (``//`` not ``/``) in order to work as expected on Python 2 and 3

peterjc:
Python 3.6 on Travis failed:

```
FAIL: test_all_search (test_KDTree.KDTreeTest)
  File "/home/travis/build/biopython/biopython/Tests/test_KDTree.py", line 199, in test_all_search
    self.assertTrue(test_all_search((nr_points // 10), dim, bucket_size, query_radius))
```

Is this stochastic?

Gasta88:
It appears so, I didn't notice that the value was floating between `nr_points` and `nr_points`-1.

I'll use a different test.

peterjc:
Thank you!

PR #1348, opened 2017-08-01
edit chapter_cluster.tex urlopen doctest

Gasta88:
As mentioned in issue #754 , I've added the doctest for Python 2 and 3. The doctest remain offline but uses an example from `Tests/Cluster/cyano.txt`.


peterjc:
Looks good - tested by hand against Python 2.7.10 and 3.6.0 adding:

```python
from Bio import Cluster
record = Cluster.read(handle)
print(record.data)
```

Thanks

PR #1305, opened 2017-07-03
Simplify Pypy2/3 Travis setup

rht:
- [x] This is contingent on https://github.com/travis-ci/travis-cookbooks/pull/861.

PR #1351, opened 2017-08-01
Simplify Pypy2/3 Travis setup.

junaruga:
This PR is what I did cherry-pick from https://github.com/biopython/biopython/pull/1305 keeping the commits, rebasing on the latest master branch.
Only I changed only 3rd commit hash, because I thought that was better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
       python: 3.6
       env: TOXENV=py36-cover
     - stage: test
-      python: pypy
+      python: pypy-5.7.1
       env: TOXENV=pypy-nocov
     - stage: test
       python: pypy3
@@ -76,9 +76,6 @@ addons:
 #
 # There is no GenePop Ubuntu pacakge, although it is in BioConda.
 #
-# There are TravisCI provided versions of PyPy and PyPy3, but currently too old.
-# We therefore deactivate that, and download and unzip portable PyPy binaries.
-#
 # We also need DSSP for testing but it is not available in the repositories.
 # Try to download the binary for Linux and place it in $HOME/bin
 #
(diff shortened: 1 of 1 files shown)


junaruga:
Done :)

peterjc:
Personally I prefer ``mkdir -p bin`` but this is a matter of taste. Thanks for looking at this.

peterjc:
Thank you for testing this - resolved via #1305

junaruga:
> Personally I prefer mkdir -p bin but this is a matter of taste. Thanks for looking at this.

Why do you like it?

When I saw `mkdir -p`, I could not understand why it was used for just 1 depth directory.
In my understanding `mkdir -p` is used to create more than 2 depth's directories like this.

```
$ mkdir -p foo/bar
```

After running tests on Travis without the code, later I found the intention.

In my opinion, `mkdir -p` does not tell people "we are considering the case `bin/` directory has already existed" enough good.

So, I think below pattern is better for people to guess the intention from the code.
But as you already merged, I am fine for that.

```
- |
  [ bin ! -d ] && mkdir bin
```

```
- mkdir bin || true
```

```
- |
  if [ ! -d bin ]; then
    mkdir bin
  fi
```

peterjc:
The GNU ``mkdir`` used on Linux has very short clear help text for ``-p``:

```
$ man mkdir
...
       -p, --parents
              no error if existing, make parent directories as needed
...
```

The help text for the macOS version of ``mkdir`` is not as clear, but still says _"with this option specified, no error will be reported if a directory given as an operand already exists"_:

```
$ man mkdir
...
     -p      Create intermediate directories as required.  If this option is not specified, the full path prefix of each operand must
             already exist.  On the other hand, with this option specified, no error will be reported if a directory given as an operand
             already exists.  Intermediate directories are created with permission bits of rwxrwxrwx (0777) as modified by the current
             umask, plus write and search permission for the owner.
...
```

junaruga:
OK, there is the help text in the manual.

But my point is which is more readable. That is if people can read without stress.

1. People recognize `mkdir -p bin`. => People wonder why it exists? => People recognize `mkdir -p` returns exist status: 0 regardless the directory exists from `man mkdir`.

2. People recognize `[ ! -d bin ] && mkdir bin`. => People recognize that means "if bin does not exist, create bin directory" from the code.

The flow 2 is better, isn't it?

peterjc:
Some people will not be be familiar with the ``mkdir`` command options, but equally ``[ ! -d bin ] && mkdir bin`` or ``test -d bin || mkdir bin`` are using quite advanced features and combine two commands. To me at least, the ``mkdir`` approach is simpler and more understandable.

junaruga:
Thank you for the explanation. I could understand the situation.
That makes sense.

PR #1352, opened 2017-08-02
Show matches vs mismatches in pairwise2 printing

peterjc:
This would help with situations like https://www.biostars.org/p/265338/#265624 by making the matches and mismatches of the aligned region easier to see.

Note space is already used to mean an unaligned region, thus I picked ``:`` to go with ``|``.

@MarkusPiotrowski and @bow: Could you review this please (eg is there any convention in ``Bio.SearchIO`` we could follow here)?

### DIFF
--- a/Bio/pairwise2.py
+++ b/Bio/pairwise2.py
@@ -214,7 +214,7 @@ To see a description of the parameters for a function, please look at
 the docstring for the function via the help function, e.g.
 type ``help(pairwise2.align.localds``) at the Python prompt.
 
-"""
+"""  # noqa: W291
 from __future__ import print_function
 
 import warnings
--- a/Bio/pairwise2.py
+++ b/Bio/pairwise2.py
@@ -56,7 +56,7 @@ printout, use the ``format_alignment`` method of the module:
     >>> from Bio.pairwise2 import format_alignment
     >>> print(format_alignment(*alignments[0]))
     ACCGT
-    |||||
+    |.||.
     A-CG-
       Score=3
     <BLANKLINE>
@@ -103,12 +103,12 @@ Some examples:
     >>> for a in pairwise2.align.globalxx("ACCGT", "ACG"):
     ...     print(format_alignment(*a))
     ACCGT
-    |||||
+    |.||.
     A-CG-
       Score=3
     <BLANKLINE>
     ACCGT
-    |||||
+    ||.|.
     AC-G-
       Score=3
     <BLANKLINE>
@@ -118,12 +118,12 @@ Some examples:
     >>> for a in pairwise2.align.localxx("ACCGT", "ACG"):
     ...     print(format_alignment(*a))
     ACCGT
-    ||||
+    |.||
     A-CG-
       Score=3
     <BLANKLINE>
     ACCGT
-    ||||
+    ||.|
     AC-G-
       Score=3
     <BLANKLINE>
@@ -134,12 +134,12 @@ Some examples:
     >>> for a in pairwise2.align.globalmx("ACCGT", "ACG", 2, -1):
     ...     print(format_alignment(*a))
     ACCGT
-    |||||
+    |.||.
     A-CG-
       Score=6
     <BLANKLINE>
     ACCGT
-    |||||
+    ||.|.
     AC-G-
       Score=6
     <BLANKLINE>
@@ -150,12 +150,12 @@ Some examples:
     >>> for a in pairwise2.align.globalms("ACCGT", "ACG", 2, -1, -.5, -.1):
     ...     print(format_alignment(*a))
     ACCGT
-    |||||
+    |.||.
     A-CG-
       Score=5
     <BLANKLINE>
     ACCGT
-    |||||
+    ||.|.
     AC-G-
       Score=5
     <BLANKLINE>
@@ -167,14 +167,14 @@ Some examples:
     >>> for a in pairwise2.align.globalms("A", "T", 5, -4, -1, -.1):
     ...     print(format_alignment(*a))
     A-
-    ||
+    ..
     -T
       Score=-2
     <BLANKLINE>
     >>> for a in pairwise2.align.globalms("A", "T", 5, -4, -3, -.1):
     ...	    print(format_alignment(*a))
     A
-    |
+    .
     T
       Score=-4
     <BLANKLINE>
@@ -187,7 +187,7 @@ Some examples:
     >>> for a in pairwise2.align.globaldx("KEVLA", "EVL", matrix):
     ...     print(format_alignment(*a))
     KEVLA
-    |||||
+    .|||.
     -EVL-
       Score=13
     <BLANKLINE>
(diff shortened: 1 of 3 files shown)


MarkusPiotrowski:
@peterjc I have thought about similar things since I'm working on ```pairwise2``` but I came to the view to let things as they are. While a simple match/mismatch in the match line is fine for DNA sequences, you may want something more elaborate when using protein sequences, e.g. something indicating similarity. And this may depend on the matrix which was used during the alignment. And also for DNA sequences you may want to tag gaps different to mismatches. 
So in my view there is little gain in the proposed change but a risk of breaking existing code from the ```pairwise2``` users.

peterjc:
I agree this could be made much more complicated by considering the scoring, there are conventions in EMBOSS and ClustalW which might make sense to follow here.

It looks like I missed updating the doctests in the pull request; I will leave that extra work for now since you are not keen.

Any thoughts @bow? Should we close this?

bow:
@peterjc, @MarkusPiotrowski 

As far as visualizations aid go, I'm almost ok with the changes actually (but I don't feel strongly about accepting/rejecting it either). The only changes I would make is indeed to distinguish between gaps and mismatches. But even then, these are only visualization aids and not meant to be something that end users can work with (e.g. so that they can query the what the second mismatch of the alignment).

If we are making the alignments interpretable, I think a more suitable solution would be having something like CIGAR strings (but different, since CIGARs do not distinguish matches and mismatches).

peterjc:
FYI, CIGAR has ``M`` for a match or mismatch but also supports ``=`` and ``X`` for an explicit match and mismatch. The resulting CIGAR string is over course more verbose and takes more space to store.

peterjc:
Given the only change is to the "human readable" output, I doubt this would break existing scripts.

Regarding showing gaps more clearly, any suggestion for the symbol to use (dot?), and how to implement that cleanly?

bow:
Ah, indeed there are the `X` the `=` cigar strings. Could we not just use `-` for the gaps as well? I think `:` is ok for the mismatch symbol.

peterjc:
Current releases local alignment example from ``test_pairwise2.py`` using ``|`` for the aligned region:

```
-AxBx
 ||||
zA-Bz
  Score=1.9
```

Current suggestion on pull request using ``|`` for a match and ``:`` for a mismatch or gap:

```
-AxBx
 |:|:
zA-Bz
  Score=1.9
```

Bow's suggestion using ``-`` for gaps:

```
-AxBx
 |-|:
zA-Bz
  Score=1.9
```

That looks better than I expected. Worth trying?

MarkusPiotrowski:
EMBOSS uses a ```:``` for similar amino acids and ```.``` for mismatches. Gaps are shown as blanks:
```
EMBOSS_001       251 LFTDWDNNKEAD-AIVSQGGSVIISPLGKILAGPNFESEGLITADLDLGD    299
                     ||:  ||.:||| .||:.|||||||||||||||||||||.:|||||||.|
EMBOSS_001       251 LFS--DNFREADPTIVTAGGSVIISPLGKILAGPNFESETIITADLDLSD    298
```
```-``` was also my first idea, but I think it looks ugly. How about using just a blank? I know, in local alignments blanks indicate the non-aligned parts, but I think you cannot misinterpret the result. 
What about ```X``` for mismatch?

**Comment by peterjc** [comment]
You've convinced me ``:`` was a bad idea. Using ``.`` for mismatches is easy and would match EMBOSS, while ``X`` would be unique but clear.

Given this code uses blanks for unaligned (for local alignments), using it for gaps seems a little bit confusing - but again, it would match other tools like EMBOSS.

**Comment by peterjc** [comment]
As long as you are aware if it was a local or global alignment, then yes, the leading/trailing spaces on the middle line are not ambiguous.

**Comment by MarkusPiotrowski** [comment]
The problem with ```format_aligment``` for local alignments is that it shows the whole sequences and not only the aligned part. However, if we only show the aligned part then we should add the starting position at the beginning:

```
10 FTDWDNNKEAD-AI
   |X  ||XX||| X|
 5 FS--DNFREADPTI
```
And you can infer from the alignment result if it was global or local from the end and start positions: 
```
from Bio import pairwise2 as pw

seq1 = 'AAGCTT'
seq2 = 'GCTT'

alignments = pw.align.globalxx(seq1, seq2)
print(alignments)

alignments = pw.align.localxx(seq1, seq2)
print(alignments)
```
```
[('AAGCTT', '--GCTT', 4.0, 0, 6)]
[('AAGCTT', '--GCTT', 4.0, 2, 6)]
```

**Comment by peterjc** [comment]
@MarkusPiotrowski I like your idea to solve the space issue by changing the display of local alignments. Would you like to implement that on top of this branch?

**Comment by MarkusPiotrowski** [comment]
Yes, I will try this locally (after this branch has been merged) and then make a separate PR.

**Comment by peterjc** [comment]
The newly introduced trailing whitespace has its downsides for linting with flake8.

Over to you @MarkusPiotrowski - merge as is and then work on changing how local alignments are shown?

**Comment by MarkusPiotrowski** [comment]
Yes, please.

**Comment by peterjc** [comment]
Thank you both for looking at this.

**Comment by MarkusPiotrowski** [comment]
@peterjc 
I'm working on implementing the ```print_format``` function for local alignments, so that only the aligned part of the sequence is shown, together with its starting position like this:
```
# Global alignment
--TTTTTGGGAATTCCC-------
  .....||||.|||||       
CCCCCCCGGGATTTCCCGGGGGGG
  Score=-15

# Local alignment
5 GGGAATTCCC
  ||||.|||||
7 GGGATTTCCC
  Score=9
```

My question is: zero-based or one-based positions? Since this is a pretty print I would prefer the one-based output, which is what EMBOSS' Water is doing:
```
EMBOSS_001         6 GGGAATTCCC     15
                     ||||.|||||
EMBOSS_001         8 GGGATTTCCC     17
```

**Comment by peterjc** [comment]
Ping @bow - how do SearchIO or the BLAST alignment objects present coordinates in their alignment pretty prints?

Otherwise given this is pretty printing, I have no particular stance here. As long as it explicit in the docstring ;)

**Comment by bow** [comment]
SearchIO displays a 0-based coordinate. Here's an example parsing one of our test cases:

```
>>> qress = list(SearchIO.parse("xml_2226_blastn_001.xml", "blast-xml"))
>>> print(qress[1].hits[0].hsps[0])
      Query: gi|356995852:1-490 Mus musculus POU domain, class 5, transcripti...
        Hit: gi|356995852|ref|NM_013633.3| Mus musculus POU domain, class 5, ...
Query range: [0:490] (1)
  Hit range: [0:490] (1)
Quick stats: evalue 0; bitscore 905.98
  Fragments: 1 (490 columns)
     Query - GAGGTGAAACCGTCCCTAGGTGAGCCGTCTTTCCACCAGGCCCCCGGCTCGGGGTGCCC~~~GAACA
             |||||||||||||||||||||||||||||||||||||||||||||||||||||||||||~~~|||||
       Hit - GAGGTGAAACCGTCCCTAGGTGAGCCGTCTTTCCACCAGGCCCCCGGCTCGGGGTGCCC~~~GAACA
>>> qress[1].hits[0].hsps[0].hit_range
(0, 490)
>>> qress[1].hits[0].hsps[0].hit_start
0
>>> qress[1].hits[0].hsps[0].hit_end
490
>>> 
```

My reason for doing so was simply because it's part of SearchIO's 'contract' with users that coordinates are all converted to 0-based, half open (per the documentation).

For pretty-printing alignment formats, since there are no specific object model per-se (and different alignment tool may do it differently), I would probably go with 1-based. Regardless of the indexing base, I would personally find it helpful if the output includes a mention of the coordinate base (i.e. just output `0-based coordinate` or `1-based coordinate`). Not enough tools / formats do this IMO (maybe SearchIO should also do this).

JoaoRodrigues:
As someone who often writes this kind of little helper function for alignments, I'd go with 1-based as well as it's the natural way of counting amino acids/nucleotides in a sequence.

PR #1355, opened 2017-08-02
Reduce a number of virtuanl machine in Travis CI to save the resource.

junaruga:
I know reducing a number of virtual machine is effective to save Travis CI's resource in my experience.
I think we can summarize `sdist`, `bdist_wheel` test on default Python (2.7) for that.

The result message is also helpful for us to know the result.

```
TOXENV='style,sdist,bdist_wheel' tox -c .travis-tox.ini
...
  style: commands succeeded
  sdist: commands succeeded
  bdist_wheel: commands succeeded
  congratulations :)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,13 +33,7 @@ matrix:
           packages:
       before_install: echo "Going to run basic checks"
     - stage: basics
-      env: TOXENV=sdist
-      addons:
-        apt:
-          packages:
-      before_install: echo "Going to run basic checks"
-    - stage: basics
-      env: TOXENV=bdist_wheel
+      env: TOXENV=sdist,bdist_wheel
       addons:
         apt:
           packages:


junaruga:
Hum.
A PR before this PR: https://travis-ci.org/biopython/biopython/builds/260375302
This PR: https://travis-ci.org/biopython/biopython/builds/260392394

junaruga:
We may reduce the total running time, if we set Limit concurrent jobs: 6 in Travis settings page.

Ref: https://docs.travis-ci.com/user/customizing-the-build/#Limiting-Concurrent-Builds
Settings -> General -> "Limit concurrent jobs": 6 (= number of virtual machines on stage: test)

junaruga:
Hum, this PR might not be effective.

junaruga:
How do you think this idea?

peterjc:
Lately the Python 3.6 test has timed out sometimes :(

Right now this repeats ``sdist`` and ``bdist_wheel`` under both Python 2 and 3, which seems to be of limited value.

Timewise, combining ``sdist`` and ``bdist_wheel`` makes more sense, so that the first stage is then three targets of similar run time (that and ``style`` under Python 2 and 3).

junaruga:
> Timewise, combining sdist and bdist_wheel makes more sense, so that the first stage is then three targets of similar run time (that and style under Python 2 and 3).

That's good idea.
I rebased and building again now.

peterjc:
Thank you.

Issue #1356, opened 2017-08-07
Python3 PDB NeighborSearch: TypeError: '<' not supported between instances of 'Residue' and 'Residue'


harryjubb:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print platform.python_implementation(); print(platform.platform())
import Bio; print(Bio.__version__)
```

```
>>> import sys; print(sys.version)
3.6.2 (default, Jul 26 2017, 17:09:31)
[GCC 5.4.0 20160609]

>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-4.4.0-87-generic-x86_64-with-debian-stretch-sid

>>> import Bio; print(Bio.__version__)
1.70
```

### Expected behaviour

`ns.search_all(2.0, level='R')` to return all pairs of residues in the NeighborSearch object that are within 2 Angstroms of each other.

### Actual behaviour

```
Traceback (most recent call last):
  File "/home/ubuntu/repos/pipeline/venv362/lib/python3.6/site-packages/doit/action.py", line 403, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/home/ubuntu/repos/pipeline/tasks/pdb_analysis.py", line 687, in fix_ligand_overlaps
    clashing_waters = ns.search_all(2.0, level='R')
  File "/home/ubuntu/repos/pipeline/venv362/lib/python3.6/site-packages/Bio/PDB/NeighborSearch.py", line 128, in search_all
    next_level_pair_list = self._get_unique_parent_pairs(next_level_pair_list)
  File "/home/ubuntu/repos/pipeline/venv362/lib/python3.6/site-packages/Bio/PDB/NeighborSearch.py", line 65, in _get_unique_parent_pairs
    elif p1 < p2:
TypeError: '<' not supported between instances of 'Residue' and 'Residue'
```

### Steps to reproduce

Minimal script to reproduce the problem:

```python
import urllib.request
from Bio.PDB import PDBParser
from Bio.PDB import NeighborSearch

urllib.request.urlretrieve('https://files.rcsb.org/download/10GS.pdb', '10gs.pdb')

pdbparser = PDBParser()
structure = pdbparser.get_structure('', '10gs.pdb')

ns = NeighborSearch(list(structure.get_atoms()))

clashes = ns.search_all(2.0, level='R')
```

peterjc:
[...]

PR #1360, opened 2017-08-08
Implements Py3-compatible rich comparison for all Bio.PDB objects

JoaoRodrigues:
This pull request addresses issue #1356 and refactors #1189 and PR #1291. It implements rich comparisons for every object of the SMCRA hierarchy, including disordered wrappers. This allows direct comparison of any two objects and allows calling the `sorted` method directly, unifying the API.

The test suite was expanded and refactored to test all these possibilities. The `a_structure.pdb` test PDB was slightly tweaked to avoid adding yet another file. I had to fix a number of other tests to accomodate these changes. All tests pass under Python 2.7 and Python 3.

Thanks @harryjubb for raising the problem that led to this fix, and @peterjc for the earlier work and usual leadership 👍 

``CONTRIB.rst`` files.


Bio/PDB/Residue.py
```diff
@@ -54,19 +96,8 @@ def sort(self):
...
-            """
-            try:
-                i = ["N", "CA", "C", "O"].index(atom.name)
-            except ValueError:
-                i = 4
-            return (i, atom.name, atom.altloc)
-
-        self.child_list.sort(key=sort_index)
+
+        self.child_list.sort
```
peterjc: Should that not be ``self.child_list.sort()`` for an in-place sorting of the list?

peterjc:
Is it worth adding public ``.sort()`` methods to each level of objects? i.e. Where there used to be a private ``._sort()`` method until #1189

Bio/PDB/Residue.py
JoaoRodrigues: Well spotted!

peterjc:
Are you happy with running ``flake8`` locally with the plugins as described in the CONTRIBUTING file? The TravisCI style checks are failing right now:

```
$ flake8 Bio/
Bio/PDB/Residue.py:92:1: D202 No blank lines allowed after function docstring
```

After that it will run ``flake8 Tests/`` as well.

JoaoRodrigues:
> Is it worth adding public .sort() methods to each level of objects? i.e. Where there used to be a private ._sort() method until #1189

I would actually propose removing them altogether and have people use the `sorted` method instead. This is already supported anyway.

peterjc:
OK - should we deprecate the single public ``.sort()`` method then?

For the git pre-commit hook, see issue #493 and possible implementation on #1188

Bio/PDB/Residue.py
```diff
@@ -54,19 +96,8 @@ def sort(self):
...
-            The first integer is 0, 1, 2, 3, 4 for atoms N, CA, C, O, other.
-            """
-            try:
-                i = ["N", "CA", "C", "O"].index(atom.name)
-            except ValueError:
-                i = 4
-            return (i, atom.name, atom.altloc)
-
-        self.child_list.sort(key=sort_index)
+
```
peterjc: You'll need to remove this blank link to make ``flake8`` happy:

```
$ flake8 Bio/
Bio/PDB/Residue.py:92:1: D202 No blank lines allowed after function docstring
```

JoaoRodrigues:
>OK - should we deprecate the single public .sort() method then?

I would vote for that.

JoaoRodrigues:
@etal @peterjc Some tests are failing on Py 3.4 and made me realize I might have misunderstood how to write these methods. For example, for an `__eq__` method for equality, should I explicitly write what I expect to compare in the _other_ object - i.e. `return self.id == other.id` - or should I leave it to the language and implement it on both objects like `return self.id == other`? It seems that the latter is the best choice because it tries to figure out how to compare the objects first and if it cannot, then it raises the normal error 'Cannot compare objects x and y' instead of a missing attribute in case the _other_ object is say, a list.

Bio/PDB/Residue.py
```diff
@@ -108,6 +108,10 @@ def sort(self):
         by name, with any alternative location specifier for disordered
         atoms (altloc) as a tie-breaker.
         """
+        warnings.warn("The custom sort() method will be deprecated in the "
```
peterjc: change "deprecated" to "removed" (or change the tense to "... has been deprecated in favour of...")

JoaoRodrigues: Done.

JoaoRodrigues:
The coverage is decreasing because of all the rich comparison methods, which have a few conditionals in there, but the testing is actually more extensive than before. If others agree with these changes, we should merge this.

peterjc:
Seems fine to me - @etal, @lennax, etc any comments before we merge this?

Issue #1368, opened 2017-08-18
Conversion from embl to fasta fails


ofajardo:
### Setup

I am reporting a problem with Biopython version 1.68 and 1.70, Python version 2.7.11, and operating
system windows 7 and linux Red Hat as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```
(On linux)
2.7.11 |Anaconda 2.5.0 (64-bit)| (default, Dec  6 2015, 18:08:32)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
CPython
Linux-2.6.32-431.20.3.el6.x86_64-x86_64-with-redhat-6.3-Santiago
1.68

(*Please copy and run the above in your Python, and copy-and-paste the output*)

### Expected behaviour

Using SeqIO SeqIO.convert to convert from a protein embl file to fasta file should give a good output. In Biopython 1.66 it is still working, but broken in 1.68 and 1.70.

### Actual behaviour

fasta file produced is full of ????? instead of aminoacids.

(*Please fill this in, and provide any exception message in full*)

### Steps to reproduce

Python code:

````python
from Bio import SeqIO

file_handler_emblFile = open("101ma.entret")
file_handler_fastaFile = open("101ma.fasta", "w")
count = SeqIO.convert(file_handler_emblFile, "embl", file_handler_fastaFile, "fasta")
````

embl input file 101ma.entret:

```
ID   101MA                             Unannotated; unspecified; UNC;                154 AA.
DT   1997-12-13 (PDB date in HEADER line)
DE   MOLECULE:MYOGLOBIN; MISC:; ENGINEERED:YES; CHAIN:A; MUTATION:YES;
OS   PHYSETER CATODON
KW   LIGAND BINDING, OXYGEN STORAGE, OXYGEN BINDING, HEME, OXYGEN TRANSPORT
SQ   Sequence 154 AA;
     MVLSEGEWQLVLHVWAKVEADVAGHGQDILIRLFKSHPETLEKFDRVKHLKTEAE
     MKASEDLKKHGVTVLTALGAILKKKGHHEAELKPLAQSHATKHKIPIKYLEFISE
     AIIHVLHSRHPGNFGADAQGAMNKALELFRKDIAAKYKELGYQG
//
```

fasta output in biopython 1.66 (correct)

```
[...]

PR #1369, opened 2017-08-18
Handle EMBL SQ lines with no coordinates

peterjc:
This should fix issue #1368, but does not yet have a test case.

Also do we have a test case explaining the 0 nt situation https://github.com/biopython/biopython/commit/81a2cee65da4198af3021df1d4ef1160e3812dfc was intended to fix?


peterjc:
Before the fix:

```
$ python3 -c "from Bio import SeqIO; print(repr(SeqIO.read('/tmp/101ma.entret', 'embl').seq)); import Bio; print(Bio.__version__)"
UnknownSeq(154, alphabet = Alphabet(), character = '?')
1.70

```

With the fix:

```
$ python3 -c "from Bio import SeqIO; print(repr(SeqIO.read('/tmp/101ma.entret', 'embl').seq)); import Bio; print(Bio.__version__)"
/Users/pc40583/repositories/biopython/Bio/GenBank/Scanner.py:632: BiopythonParserWarning: EMBL sequence line missing coordinates
  BiopythonParserWarning)
Seq('MVLSEGEWQLVLHVWAKVEADVAGHGQDILIRLFKSHPETLEKFDRVKHLKTEA...YQG', Alphabet())
1.71.dev0
```

Note the new warning message (which appears once only by default, not once per bad line).

peterjc:
We do have a relevant test case ``EMBL/embl_with_0_line.embl`` created in 2518ec1ddeec590cc92c54a4bb760c3fe7c0493a and my change breaks it, this needs more work...

peterjc:
Updated with test case from @ofajardo based on public PDB data - with his permission, see https://github.com/biopython/biopython/issues/1368#issuecomment-323514779

PR #1384, opened 2017-09-03
Add deprecation warning for PSEA

souravsingh:
Related to the discussion here in the mailing list- http://lists.open-bio.org/pipermail/biopython-dev/2017-July/021850.html

I am not able to communicate to the dev and the user mailing list, so I am putting the discussion about the Deprecation here.

### DIFF
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -19,6 +19,8 @@ import os
 from Bio.PDB.Polypeptide import is_aa
 from Bio import BiopythonDeprecationWarning
 
+import warnings
+
 warnings.warn("Bio.PDB.PSEA has been deprecated, and we intend to remove"
               " it in a future release of Biopython. If you would like to"
               " continue using Bio.PDB.PSEA, please contact the Biopython"
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -17,6 +17,13 @@ ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
 import os
 
 from Bio.PDB.Polypeptide import is_aa
+from Bio import BiopythonDeprecationWarning
+
+warnings.warn("Bio.PDB.PSEA has been deprecated, and we intend to remove"
+              " it in a future release of Biopython. If you would like to"
+              " continue using Bio.PDB.PSEA, please contact the Biopython"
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
I am sorry, but I seriously do not understand the reason to deprecate this module besides 'the software it supports is old'. Again, NACCESS is old(er) and DSSP is old(er) and we still support those modules.

Unless this particular module is being a pain to maintain (our Python code, not the tool itself), I am *strongly* against this deprecation. Unless we decide to re-code it ourselves and keep the functionality.

souravsingh:
Currently as it stands, ``PSEA`` works fine for Mac, but doesn't compile for Windows and Linux, which is mentioned by @peterjc in the discussion here- http://lists.open-bio.org/pipermail/biopython-dev/2017-July/021851.html

So as far as it goes, We have two approaches-

1) Get the software to work on Linux, which I have tried on three different distros and doesn't work.

2) Deprecate the module for now and mention that the functionality is being re-coded as a separate package.

JoaoRodrigues:
I got the code from [here](http://abcis.cbs.cnrs.fr/PSEA/DOWNLOAD/) and compiled it on a BASH shell on Windows (WSL). You need to add an include to `stdlib.h` and compile with `gcc -o p-sea -psea.c -lm` to link with the math library. 

Regardless, this is *not* our software so it doesn't matter if it compiles or not. What matters is that *our* code works. Deprecating a code just because it wraps old software is a bad idea in my opinion. The only decent alternative I see is to replace it with a better alternative ...

peterjc:
It sounds like Joao's approach would work to compile this for Linux too. If so, we should probably mention this somewhere like the PSEA module docstring?

It would also be fairly easy to define a BioConda package which we could use for TravisCI testing.

peterjc:
Note Joao did comment on the mailing list earlier voicing his objection:
http://lists.open-bio.org/pipermail/biopython-dev/2017-July/021854.html

@souravsingh why can't you post on the mailing lists anymore?

souravsingh:
The mail rebounds back saying the sending is failed due to "List not found".

peterjc:
@souravsingh please forward the rebound email to my GMail, and I'll pass it on to our mailing list team. Thanks!

peterjc:
It looks like @souravsingh got stuck trying to do a BioConda package,
https://github.com/bioconda/bioconda-recipes/pull/5948

PR #1385, opened 2017-09-04
Genomediagram Bug Fixes (squashed)

sbliven:
This is a cleaner version of #1329.

It squashes commits down to two main issues.

1. Allows GenomeDiagrams to be created with zero margins. Also reduces unneeded extra y padding in the case of single fragments (but this is still generated if needed).

2. Work around a bug in reportlab's PNG output of polygons with degenerate vertices. 84079c6 simplifies the approach used in #1329, so it should work with all polygons regardless of the strand or source of degeneracy (short arrow shafts, thick arrow shafts, octogons, etc).

Here are some results from running Tests/test_GenomeDiagrams.py:

BIGARROW:
gd_sigil_bigarrow_shafts
gd_sigil_short_bigarrow

ARROW:
gd_sigil_short_arrow

OCTO:
gd_sigil_short_octo

### DIFF
--- a/Bio/Graphics/GenomeDiagram/_Diagram.py
+++ b/Bio/Graphics/GenomeDiagram/_Diagram.py
@@ -36,7 +36,7 @@ from Bio.Graphics import _write
 
 
 def _first_defined(*args):
-    """Return the first non-null argument"""
+    """Return the first non-null argument (PRIVATE)."""
     for arg in args:
         if arg is not None:
             return arg
--- a/Bio/Graphics/GenomeDiagram/_Diagram.py
+++ b/Bio/Graphics/GenomeDiagram/_Diagram.py
@@ -35,6 +35,14 @@ from ._Track import Track
 from Bio.Graphics import _write
 
 
+def _first_defined(*args):
+    """Return the first non-null argument"""
+    for arg in args:
+        if arg is not None:
+            return arg
+    return None
+
+
 class Diagram(object):
     """Diagram container.
 
@@ -79,7 +87,7 @@ class Diagram(object):
     def __init__(self, name=None, format='circular', pagesize='A3',
                  orientation='landscape', x=0.05, y=0.05, xl=None,
                  xr=None, yt=None, yb=None, start=None, end=None,
-                 tracklines=False, fragments=10, fragment_size=0.9,
+                 tracklines=False, fragments=10, fragment_size=None,
                  track_size=0.75, circular=True, circle_core=0.0):
         """Initialize.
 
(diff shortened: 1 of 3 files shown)


peterjc:
@widdowquinn what do you think of the glyph simplification code to remove near-identical points? Would you worry about side effects at high magnification? On the plus side it should make the output images a little simpler (which is a good thing as the complex PDFs can take a while to render).

My preference is to explicitly switch to a triangle for when an arrow is just the arrow-head.

sbliven:
@peterjc Although there is support for merging close points in the code, I'm currently only merging those which are identical. So there's no risk of artifacts at high magnification from deduplicating. I guess you could worry that straight lines can be represented by several segments, but this seem like a very minor issue.

If you want, I can add back the code for switching to a triangle, thereby saving two vertices per triangle.

Note that the #1329 version of the code contained a bug not in this one. With `shaft_width=1.0` the outer corners of the arrow were degenerate. This showed up, for instance, in the second row of GD_sigil_arrow_shafts.png extending off the page to the right. It seems easier to keep the de-duplication code rather than thinking through all the edge cases.

peterjc:
If ``_deduplicate`` didn't (initially) have the delta option, so it was clearly just removing duplicated points I'd be happier. And perhaps all the glyph drawing code should use it (at least in the linear drawer)?

sbliven:
@peterjc Done.

Tests/test_GenomeDiagram.py
```diff
@@ -1070,6 +1073,25 @@ def test_diagram_via_object_pdf(self):
...
 
+def _check_prereqs():
+    """Check that the input data is available"""
+    if not os.path.exists("Graphics"):
+        os.mkdir("Graphics")
+    if not os.path.exists("GenBank"):
+        os.mkdir("GenBank")
+
+    gbfilename = os.path.join("GenBank", "NC_005816.gb")
+    if not os.path.exists(gbfilename):
```
peterjc: Unless the test script is being run from the wrong directory, this will exist - https://github.com/biopython/biopython/blob/master/Tests/GenBank/NC_005816.gb

(I may be missing something, but at face value I would revert this commit)

sbliven: Ah, I've apparently been running tests the wrong way. Reverted.

peterjc:
I've asked TravisCI to rebuild that timed out build.

I'm hoping to have Leighton's input, but he's been out of the office a lot recently, so in the short term I might just pull out the glyph changes (since I did most of the fancy shapes) and apply those directly.

peterjc:
Rebased onto the current master, hopefully without breaking anything, leaving just the margin changes.

sbliven:
@peterjc Thanks for the rebase. Is anything else required to merge this?

peterjc:
Just waiting on comments / approval from @widdowquinn (and we'd need to add your name to the ``NEWS.rst`` entry again as we're into a new release now).

widdowquinn:
This all looks good, to me. Apologies for the delay!

peterjc:
Updated NEWS and rebased...

peterjc:
Thank you @widdowquinn, and thank you again and also for your patience @sbliven - merged!

Issue #1288, opened 2017-06-20
Improving test coverage


peterjc:
We have for some time run the continuous integration tests on TravisCI via the Python tool ``coverage`` and uploaded the metrics to https://codecov.io/github/biopython/biopython/

For Biopython 1.69 the test coverage reported was about 80%, as we approach Biopython 1.70 we are now up to 84%.

This has been a combination of removing some deprecated code, and adding more tests - including efforts like #820 for moving the remaining self-tests blocks under ``Bio/`` into our test framework under ``Tests/`` where those tests will be run regularly via continuous integration.

Note that currently our TravisCI configuration deliberately DOES NOT include the online tests (we are using ``python run_tests.py --offline``), and due to installation time overheads DOES NOT have all the optional dependencies installed. This means the numbers on CodeCov.io are in a sense under-reporting our potential true test coverage.

I would like us to improve the coverage further - the sunburst image from CodeCov.io which takes into account the lines of code and percentage coverages highlights particular targets including ``Bio.Nexus``,  ``Bio.codonalign``, ``Bio.KEGG``, ``Bio.PopGen``.

Gasta88:
I'm particularly interested in `Bio.KEGG.REST` thanks to the recent development with the dbfetch module.

From CodeCov.io it seems that the private method `_q()` is never covered. Likewise, most exceptions are not tested.

I could try to expand the test script `Tests\test_KEGG_online.py` to verify the content of the handles and perform some fuzzy testing. Does it sound OK?

peterjc:
@Gasta88 Right now (deliberately) the TravisCI tests are all run in offline mode, so none of the network only code in Biopython will be reported as covered. Please run ``coverage`` locally before trying to improve the online test coverage, as otherwise you might be wasting some effort.

Regarding ``Bio.KEGG``, apart from the (online) ``REST.py`` file there is poor coverage in ``Bio/KEGG/__init__.py`` (looks like helper functions for output), ``Bio/KEGG/Enzyme/__init__.py`` and ``Bio/KEGG/Compound/__init__.py``. The KGML folder looks better percentage-wise.

peterjc:
Good news: I've enabled AppVeyor for continuous integration under Windows:

https://github.com/biopython/biopython/commit/09ba3065d191d204e860d67954d7f1ef7bb4a1b0
https://ci.appveyor.com/project/biopython/biopython/history

[...]

PR #1386, opened 2017-09-07
added test for Nexus/Tree module

Gasta88:

This is for `Bio.Nexus.Tree` module.

### DIFF
--- a/Tests/test_Nexus.py
+++ b/Tests/test_Nexus.py
@@ -409,7 +409,7 @@ Root:  16
         self.assertEqual(output, expected)
         self.assertEqual(t3.is_compatible(t2, threshold=0.3), [])
 
-    def test_internal_node_labels(self):
+    def test_TreeTest2(self):
         """Handle text labels on internal nodes.
         """
         ts1b = "(Cephalotaxus:125.000000,(Taxus:100.000000,Torreya:100.000000)"\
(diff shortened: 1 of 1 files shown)


peterjc:
What do you think @etal ?

Gasta88:
Small side note: I tried to test `merge_with_support` and `consensus` methods, but failed on my attempts.

I was successful in python 2.x, but line 832 in `Bio.Nexus.Tree` has a `key` function using `lambda` and two arguments.
```python
consensus_ids.sort(lambda x, y: len(consensus.node(x).data.taxon) - len(consensus.node(y).data.taxon))
```
 According to the docs, even in 2.x it should get only one argument but it seems more flexible that python 3.x.

Every attempts in python 3.x failed, therefore I'm more than happy to hear any suggestions.

peterjc:
This line assumes Python 2 sorting with a ``cmp=...`` argument:

https://github.com/biopython/biopython/blob/biopython-170/Bio/Nexus/Trees.py#L832

```python
consensus_ids.sort(lambda x, y: len(consensus.node(x).data.taxon) - len(consensus.node(y).data.taxon))
```

Likely fix:

```python
consensus_ids.sort(key = lambda x: len(consensus.node(x).data.taxon))
```

etal:
Looks good to me (once cmp vs. key is sorted out). Great to see better tests for this sub-package.

Gasta88:
Thanks @peterjc , I've used the your suggestion before but the test seems to succeed some times and fail others. I'll try few more attempts and expand further the testing coverage.

If I had to edit `Bio.Nexus.Tree`, would it be preferable to propose a separate PR before moving forward with this one, or collect everything in here?

peterjc:
Hard to say - if the test work is linked to a fix in ``Bio.Nexus.Tree`` then one pull request seems logical.

PR #1393, opened 2017-09-21
mmCIF writer

jgreener64:
It would be good for Biopython to have an mmCIF file writer, so I implemented one based on the `PDBIO` class. It lets you do:

```python
p = MMCIFParser()
s = p.get_structure("1fat", "1fat.cif")
io = MMCIFIO()
io.set_structure(s)
io.save("out.cif")
```

Selector classes work in the same way as `PDBIO`, e.g. `io.save("out.cif", CASelector())`. It also lets you write out an mmCIF dictionary:

```python
d = MMCIF2Dict("1fat.cif")
io = MMCIFIO()
io.set_dict(d)
io.save("out.cif")
```

I have tested the writer on some random PDB IDs and it runs without errors and produces files that look correct. I also checked it writes a large file (1HTQ, 10 models of 97,000 atoms) correctly and it does so (it takes ~1 minute).

Let me know what you think and if there are any suggestions.

``CONTRIB.rst`` files.


peterjc:
@etal, @JoaoRodrigues, @lennax - who'd be best to review this?

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
+"""Write an mmCIF file.
```
peterjc: Please leave a blank line after the module docstring title, see PEP257 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
...
+                inds = []
+                for i in key_list:
+                    try:
+                        inds.append(mmcif_order[key].index(i))
+                    # Unrecognised key - add at end
+                    except ValueError:
+                        inds.append(len(mmcif_order[key]))
+                z = list(zip(inds, key_list))
+                z.sort()
+                key_lists[key] = [k for _, k in z]
```
peterjc: Would this not be the same?

```python
key_lists[key] = [z[k] for k in sorted(z)]
```

(See also separate comment about replacing the above two lines with ``z = sorted(zip(inds, key_list))``)

In fact, I wonder if you can avoid building the ``inds`` list in the first place by sorting the keys directly?

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
...
+
+        # Write out top data_ line
+        if data_val:
+            out_file.write("data_" + data_val + "\n")
+            out_file.write("#\n")
```
peterjc: If disk IO is going to be slow with mmCIF files, you might as well combine these two lines into a single call to write:

```python
out_file.write("data_" + data_val + "\n#\n")
```

If performance is not critical, then your version is much clearer.

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
...
+
+        for key, key_list in key_lists.items():
+            # Pick a sample mmCIF value, which can be a list or a single value
+            sample_val = self.dic[key + "." + key_list[0]]
+            val_type = type(sample_val)
```
peterjc: Can you use ``isinstance(...)`` rather than comparing types in this loop?

Bio/PDB/MMCIFParser.py
```diff
@@ -75,7 +75,6 @@ def _build_structure(self, structure_id):
             element_list = mmcif_dict["_atom_site.type_symbol"]
         except KeyError:
             element_list = None
-        seq_id_list = mmcif_dict["_atom_site.label_seq_id"]
```
peterjc: Removing this (and again later) puzzles me, but I am not familiar enough with the mmCIF format to understand the intent.

Bio/PDB/__init__.py
```diff
@@ -31,6 +31,8 @@
 # IO of PDB files (including flexible selective output)
 from .PDBIO import PDBIO, Select
 
+from .MMCIFIO import MMCIFIO
```
peterjc: Is this needed? The ``Bio.PDB`` module already does too many automatic imports (and name shadowing) so I'd prefer to avoid it.

Tests/test_PDB.py
```diff
@@ -913,6 +916,83 @@ def test_pdbio_missing_occupancy(self):
...
+            io.save(filename)
+            struct2 = self.mmcif_parser.get_structure("1a8o", filename)
+            nresidues = len(list(struct2.get_residues()))
+            self.assertEqual(len(struct2), 1)
+            self.assertEqual(nresidues, 158)
+        finally:
+            os.remove(filename)
+
+    def test_mmcifio_write_residue(self):
+        """Write a single residue using MMCIFIO"""
```
peterjc: Missing full stop at the end of the docstring.

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
...
+        "auth_seq_id",
+        "auth_comp_id",
+        "auth_asym_id",
+        "auth_atom_id",
+        "pdbx_PDB_model_num",
+    ]
+}
+
+
+class MMCIFIO(object):
```
peterjc: You have a module called ``MMCIFIO`` and a class called ``MMCIFIO`` which is a very bad idea as they clash and will be a source of confusion depending on how they are imported.

This is one reason why PEP8 says use lower case for module names and title case for classes https://www.python.org/dev/peps/pep-0008/#naming-conventions

I appreciate that ``Bio.PDB`` breaks those rules (as do other bits of Biopython), but it is historic code and we're stuck with it for backwards compatibility.

I would nevertheless suggest considering using ``Bio/PDB/mmcifio.py`` for the filename (but am interested to hear what other people would recommend here).

JoaoRodrigues:
I can have a look at this. Much needed addition, thanks @jgreener64 !

jgreener64:
Thanks for the comments @peterjc. I will address these and wait for @JoaoRodrigues to have a look.

Bio/PDB/MMCIFParser.py
jgreener64: In line 109-112 `seq_id_list` is set depending on which of `_atom_site.auth_seq_id` and `_atom_site.label_seq_id` are present. This line is therefore not used and causes an error if `_atom_site.label_seq_id` is not defined in the mmCIF file.

Technically this is separate to this PR but I put it here as I need this for the tests to pass (the mmCIF writer writes out `_atom_site.auth_seq_id` but not `_atom_site.label_seq_id` and hence reading the written files back in was erroring).

Bio/PDB/MMCIFIO.py
jgreener64: Yes I see the issue. Changing the filename to `mmcifio.py` seems sensible and would be in line with PEP8 conventions, and this is probably worth inconsistency within the PDB module.

Bio/PDB/__init__.py
jgreener64: Yes I will remove this.

Bio/PDB/MMCIFIO.py
```diff
@@ -0,0 +1,301 @@
...
+                        except Exception:
+                            pass
+                        sb.structure[0]['A'].add(pdb_object)
+                    else:
+                        # Atom
+                        sb.init_residue('DUM', ' ', 1, ' ')
+                        try:
+                            parent_id = pdb_object.parent.parent.id
+                            sb.structure[0]['A'].id = parent_id
+                        except Exception:
```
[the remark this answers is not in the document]
jgreener64: Looking at the id setter code, `ValueError` seems the correct one.

peterjc:
Mental note: The case renaming can complicate the git history (and cause trouble if used on Windows or a Mac with default case insensitive file system), so we should squash this branch to ensure that case change is not in the master branch history.

(I'd probably have squashed the branch to a single commit anyway, given the number of small tweak)

jgreener64:
Yes, it has already confused me (the initial rename didn't push up, hence the delete and re-add).

peterjc:
Sorry, I've been hit by this issue before with git, and could have warned you.

jgreener64:
No it's one of those things that always gets forgotten until it happens haha.

Thanks to @StefansM for testing this and pointing out that there are three `atom_site` fields marked as essential that are not written here:

`label_seq_id` - I think this can start from 1 and increment for each residue, not writing out anything for HETATMS.
`label_asym_id` - I think this can start from A and increment with each molecule change.
`label_entity_id` - I think this can start from 1 and increment with each molecule change.

These are required by the mmCIF format and so should be written out, however they may be written out differently to the original file as they are not parsed in - I think this is fine. I will add these.

peterjc:
@JoaoRodrigues we're hoping you could take a look at this

JoaoRodrigues:
Hi @jgreener64 @peterjc 

Sorry, been quite busy lately .. I tried combining a multi-model/multi-chain viral capsid PDB file (360 chains) into a single CIF file with your code and it worked perfectly.

Regarding changes, based on my previous experience with PDB to CIF conversions/writing, I would add a test where you read a multi-chain/multi-model PDB (it can be only 4 atoms), write a CIF with it, and check the contents match. Same thing for reading a CIF/writing a CIF.

Super nice work!

jgreener64:
Thanks @JoaoRodrigues will implement your suggestions!

jgreener64:
I have implemented the suggested changes, let me know if anything else needs doing.

jgreener64:
I also added a small section to the docs, let me know if this needs to change.

JoaoRodrigues:
Thanks @jgreener64 

I'll have a look asap once it passes the tests.

JoaoRodrigues:
This is all good by me as well @peterjc 

Awesome contribution @jgreener64 , thank you!

peterjc:
Thank you both. I've merged it (squashed into one commit, in part to avoid issues with Windows and the file rename), but then realised there is nothing about this in the ``NEWS.rst`` file for the next release, and this deserves to me mentioned. @jgreener64 Would you mind doing a followup pull request for that?

JoaoRodrigues:
Thanks @peterjc 

This *must* be there, it's really a major contribution for the `Bio.PDB` module. If @jgreener64 cannot do it, I'll gladly write up the news in this one.

jgreener64:
Thanks both, I'll make another PR for `NEWS.rst`.

Issue #705, opened 2015-12-04
Updating Installation Docs


souravsingh:
Since Biopython is also available in PyPI, It would be good if we add this info in the installation docs.We can add this as an alternative to the normal method for installation of Biopython

peterjc:
What did you have in mind? There is more than one way to use PyPI...

See also *[Biopython-dev] Should we recommend "pip install biopython"?*
http://lists.open-bio.org/pipermail/biopython-dev/2015-October/021105.html

And this thread on the NumPy list, *[Numpy-discussion] Proposal: stop supporting 'setup.py install'; start requiring 'pip install .' instead*
https://mail.scipy.org/pipermail/numpy-discussion/2015-October/074086.html

souravsingh:
I believe we should support ``setup.py install`` since a few have issues with pip installation.Moreover, to upgrade to the latest package, one can do so by using the ``--upgrade`` option with ``pip install``.So mentioning that could be good

peterjc:
Belatedly closing this issue - we added ``pip install biopython`` to the README last year in https://github.com/biopython/biopython/commit/8a16a61109f40ed353cac3970a9965e3e7505916 (with the traditional ``python setup.py build``, ``test``, and ``install`` also covered).

PR #1394, opened 2017-09-22
pip upgrade, uninstall; re-order compilation information.

peterjc:
Cross reference #1295 for overhauling the installation documentation.

This pull request addresses some additional points relating to issue #705 by describing how to update and uninstall using pip.

It also re-orders the dependency information to have a single block about what you need for compiling Biopython (including adding the fact you will need the Python header files).

It also list pypy before jython (since we've deprecated our support for Jython).

I hereby agree to dual licence this and any previous contributions under both the _Biopython License Agreement_ AND the _BSD 3-Clause License_.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,9 +60,11 @@ For the impatient
 
 Python 2.7.9 onwards, and Python 3.4 onwards, include the package management
 system "pip" which should allow you to install Biopython (and its dependency
-NumPy if needed) with just::
+NumPy if needed), upgrade or uninstall with just one terminal command::
 
     pip install biopython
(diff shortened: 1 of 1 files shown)


peterjc:
https://github.com/biopython/biopython.github.io/commit/fb4318d118846b321598540743a824586f4da917 added ``pip uninstalled biopython`` to the website instructions.

peterjc:
https://github.com/biopython/biopython.github.io/commit/e92a805a0d9d8b5c680390c672b25fb7e9817966 mentions the header files on the website instructions.

(I don't like this duplication, as part of #1295 we should address this)

PR #1398, opened 2017-09-26
EBIsearch integration

Gasta88:
This pull request addresses issue #443 

The module has been located in `Bio\EBI\ebisearch.py`.
A testing script has been allocated in `Tests`.

### DIFF
--- a/Bio/EBI/__init__.py
+++ b/Bio/EBI/__init__.py
@@ -1,0 +1,5 @@
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
+
+"""Provide code to access the EBI tools."""
--- a/Bio/EBI/ebisearch.py
+++ b/Bio/EBI/ebisearch.py
@@ -1,0 +1,691 @@
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
+
+"""Provide code to access the EBI Search API.
+
+This module aims to make EBI Search API easy to use. See:
+https://www.ebi.ac.uk/ebisearch/overview.ebi
+
+The EBI Search API is a text search engine to access biological data resources
+hosted at the European Bioinformatics Institute (EMBL-EBI).
+
+The following code was developed by Berenice Batut (berenice.batut@gmail.com)
+
+References:
+Park Y.M., Squizzato S., Buso N., Gur T., Lopez R. (2017)
+The EBI search engine: EBI search as a service making biological data accessible
+for all.;Nucleic Acids Research, May 2, 2017; DOI: 10.1093/nar/gkx359
+
+"""
+
+import requests
+
+import requires_internet
+requires_internet.check()
+
+baseUrl = 'http://www.ebi.ac.uk/ebisearch/ws/rest'
+sizeLimit = 100
+startLimit = 250000
+
+
+def get_domain_details(domain):
+    """Return a dictionary with the details of a given domain in EBI.
+
+    >>> get_domain_details("allebi")
+
+    domain: domain id in EBI
+    """
+    url = baseUrl + '/' + domain
+    r = requests.get(
+        url,
+        headers={"accept": "application/json"})
+    r.raise_for_status()
+    return r.json()
+
+
+def get_details_and_subdomains(domain, level, verbose=False):
+    """Return the details (id and name) for the domain and its subdomains.
+
+    domain: domain id in EBI
+    verbose: Boolean to define the printing info
+    """
+    domain_details = {domain['id']: domain['name']}
+    if verbose:
+        print("\t" * level + "%s -- %s" % (domain['id'], domain['name']))
+
+    if "subdomains" not in domain:
+        return domain_details
+    else:
+        for subdomain in domain["subdomains"]:
+            domain_details.update(
+                get_details_and_subdomains(subdomain, level + 1, verbose))
+    return domain_details
+
+
+def get_domains(verbose=False):
+    """Return the list of domains in EBI as a dictionary.
+
+    The key being the domain id and the value the domain name.
+
+    verbose: boolean to define the printing info
+    """
+    allebi = get_domain_details("allebi")
+    domain_details = {}
+    for domain in allebi["domains"]:
+        domain_details.update(get_details_and_subdomains(domain, 0, verbose))
+    return domain_details
+
+
+def print_domain_hierarchy():
+    """Print the hierarchy of the domains."""
+    get_domains(verbose=True)
+
+
+def check_domain(domain):
+    """Check if a domain exist in EBI.
+
+    domain: id of a domain to check
+    """
+    if domain not in get_domains(verbose=False):
+        err_str = "The domain does not correspond to the id of a known domain"
+        err_str += " in EBI. "
+        err_str += "The list of EBI domains and their id can be "
+        err_str += "accessed with get_domains"
+        raise ValueError(err_str)
+
+
+def get_number_of_results(domain, query):
+    """Return the number of results for a query on a specific domain in EBI.
+
+    domain: domain id in EBI
+    query: query for EBI
+    """
+    check_domain(domain)
+    url = baseUrl + '/' + domain + '?query=' + query + '&size=0'
+    r = requests.get(
+        url,
+        headers={"accept": "application/json"})
+    r.raise_for_status()
+    return r.json()['hitCount']
+
+
+def get_subdomain_fields(domain):
+    """Return the fields of a domain and its subdomains.
+
+    domain: domain id in EBI
+    """
+    fields = {
+        "searchable": {},
+        "retrievable": {},
+        "sortable": {},
+        "facet": {},
+        "topterms": {}
+    }
+    if "subdomains" not in domain:
+        for field in domain["fieldInfos"]:
+            field_desc = field["description"]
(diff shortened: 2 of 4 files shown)


Gasta88:
I've updated to the latest version of `flake8` and `pydocstyle`. When I run them on the new scripts I don't get any error, but TravisCI seems to think differently.

I'm updating the modules via this command: `pip install <moduleName> --update`

peterjc:
Note we have separate flake8 setting per top level folder, but this is the failure on TravisCI:

```
$ flake8 Bio/
Bio/EBI/ebisearch.py:590:1: D300 Use """triple double quotes"""
```

We install several flake8 plugins - see https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst - which include docstring checks - in this case it looks like ``flake8-docstrings`` is failing.

Bio/EBI/__init__.py
```diff
@@ -0,0 +1,5 @@
+# This code is part of the Biopython distribution and governed by its
+# license.  Please see the LICENSE file that should have been included
+# as part of this package.
```
peterjc: Can we have all of ``Bio.EBI`` dual licenced from the start please? See #898 and our current wording as used in ``Bio/SeqRecord.py``,

```python
# package.
```

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,691 @@
...
+for all.;Nucleic Acids Research, May 2, 2017; DOI: 10.1093/nar/gkx359
+
+"""
+
+import requests
+
+import requires_internet
+requires_internet.check()
+
+baseUrl = 'http://www.ebi.ac.uk/ebisearch/ws/rest'
```
peterjc: Please avoid "camel case", I would use ``base_url`` here, or ``BASE_URL`` if this is regarded as a constant - see https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,691 @@
...
+sizeLimit = 100
+startLimit = 250000
+
+
+def get_domain_details(domain):
+    """Return a dictionary with the details of a given domain in EBI.
+
+    >>> get_domain_details("allebi")
```
peterjc: If this is a intended to be a doctest, the output is missing. If this is just documentation, can you include a snippet of the output perhaps?

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,691 @@
...
+
+    domain: domain id in EBI
```
peterjc: It is not explicit that this is one of the arguments (but we have not settled on a common preferred reStructuredText style for this). I would suggest using the bullet point list style in ``Bio/Seq.py`` or the very explicit ``:param domain: domain ID in EBI`` style used in ``Bio/Phylo``.

Tests/test_EBI_ebisearch.py
```diff
@@ -0,0 +1,126 @@
+# This code is part of the Biopython distribution and governed by its
```
peterjc: Missing copyright statement with author names.

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,691 @@
+# This code is part of the Biopython distribution and governed by its
```
peterjc: Missing copyright statement with author names (and the dual licensing point raised separately).

Gasta88: Sorry, this part is not clear. What years should I place for the copyright?
If I understood right it should be something like:
```
# Copyright 20xx-20xx Berenice Batut.  All rights reserved.
```

peterjc: See e.g. ``Bio/SeqRecord.py`` for our convention, the first line(s) are the copyright bits (Bernice as original author, and then revisions by you), followed by the license information.

Gasta88:
Should we give consent to be cited on the `NEWS.rst` and `CONTRIB.rst` at each PR? I think I gave it ages ago in previous PRs.

Gasta88:
`requests` module is not supported in Biopython; I'll take care to use `from Bio._py3k.urlopen` instead.

Gasta88:
Sorry @peterjc, but I need your insight on one problem here.
Originally the module return `json` format output because the `requests` module allow the user to attached headers info along with the request to the server.

`urlopen` does not allow it and the output is returned by default in XML format.
Implementing the module with `Bio._py3k.urlopen`, I probably need to use a module that can handle XML format. Does Biopython have a preferred one?

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,751 @@
...
+References:
+Park Y.M., Squizzato S., Buso N., Gur T., Lopez R. (2017)
+The EBI search engine: EBI search as a service making biological data accessible
+for all.;Nucleic Acids Research, May 2, 2017; DOI: 10.1093/nar/gkx359
+
+"""
+
+import requests
+
+import requires_internet
```
peterjc: We only use this in our tests framework, so you can remove it from the main module.

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,751 @@
...
+requires_internet.check()
+
+base_url = 'http://www.ebi.ac.uk/ebisearch/ws/rest'
+sizeLimit = 100
```
peterjc: Camel case as in line above

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,751 @@
...
+startLimit = 250000
```
peterjc: Camel case as in the lines above

peterjc:
Ah - you mean the ``import requests`` bit, referencing the 3rd party library https://pypi.python.org/pypi/requests

If you can rework the calls to avoid this extra dependency that would be preferable.

Note we're using ``from urllib2 import urlopen, Request`` on Python 2 but ``from urllib.request import urlopen, Request, ...`` under Python 3.

Are you sure you can't do this with the ``headers=`` argument to ``urlopen``? See e.g. https://stackoverflow.com/questions/385262/how-do-i-send-a-custom-header-with-urllib2-in-a-http-request

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,798 @@
...
+    Arguments:
+     - domain - domain id in EBI.
+
+    Example:
+
+    >>> dom_dict = ebisearch.get_domain_details(domain="allebi")
+    >>> print(str(dom_dict['domains'][0])[:150])
+    >>> {u'description': u'Results from all the EBI resources', u'subdomains':
+        [{u'description': u'', u'subdomains': [{u'fieldInfos': [{u'description'
+        : u'Luce
```
peterjc: The indentation is wrong, if this is to be a working doctest the output of the print statement it should be exact - and therefore truncated at 75 might be better than 150. Right now there also looks to be a comma or something missing before the "Luce" entry.

Bio/EBI/ebisearch.py
```diff
@@ -0,0 +1,798 @@
...
+
+def check_domain(domain):
+    """Check if a domain exist in EBI.
+
+    Arguments:
+
+     - domain - id of a domain to check.
+    """
+    if domain not in get_domains(verbose=False):
+        err_str = "The domain does not correspond to the id of a known domain"
```
peterjc: You could write this as a multi-line string. If you did that there is no need to create the temporary variable ``err_str``.

Bio/EBI/ebisearch.py
Gasta88: Phaving a mixture of doctest and unittest is not a good idea. 
I could remove this part here and test the function in `Tests/test_EBI_ebisearch.py`?

Gasta88:
A style error seems to be causing the issue, but I've updated `flake8` to the latest version and I can't catch anything on my local machine.

Gasta88:
Hi @peterjc , I've checked again if the code had some style errors but nothing is flagged by `flake8` or `pydocstyle`. 
Is there a way that I can trigger TravisCI to check again without using a dummy commit?

MarkusPiotrowski:
@Gasta88 It looks like a "DOI" comment is responsible for the problem.

The log says:
```
BioSQL Scripts Doc ; then echo 'Please use https://doi.org/... not the doi: or DOI: style.'; false; fi
```
and in ```ebisearch.py```, line 22 you use ```DOI: 10.1093/nar/gkx359```  Try to replace this with ```https://doi.org/..```.

Gasta88:
Thanks, I did see that!

peterjc:
Sorry, the DOI thing is recent - we have a few ``grep`` commands run via TravisCI/Tox for misc style issues not covered in ``flake8``.

peterjc:
@bebatut as the original author of https://github.com/bebatut/ebisearch could you please state here you are happy to contribute your code to Biopython, and dual licence your contributions under both the Biopython License Agreement AND the BSD 3-Clause License? See #898.

Also, are you happy to be thanked by name in the NEWS.rst and CONTRIB.rst files?

(@Gasta88 if we have agreement from Bérénice on our public mailing list archive, we cite that rather than waiting for repeating that here on GitHub.)

bebatut:
@peterjc sorry for the late answer!

I would be happy:
- to contribute my code to Biopython
- to dual license my contributions under both license
- to be thanked in the `NEWS.rst` and `CONTRIB.rst`

:smile:

Issue #1397, opened 2017-09-25
PDB/mmCIF parsers spawn multiple threads (thanks to numpy)


StefansM:
If the underlying BLAS library allows it, numpy will spawn a thread for every core in the machine as soon as it is imported. As far as I can tell, these are POSIX threads rather than python threads, so the overhead is not huge, but this can still cause problems when a program is expected to run in a single thread. If you try to run 64 jobs on a 64 core machine, you will end up with 4096 threads, which will almost certainly breach the resource limits on a shared-use machine, causing your jobs to fail and your system administrator to be angry.

This behaviour can be disabled by setting [some environment variables](https://stackoverflow.com/a/31622299), but this is not particular obvious, especially if biopython was installed by an administrator and the underlying dependencies aren't known.

It's not clear to me what the best solution, if any, is to this. I see a few options:

1. Ignore it. This is the usual behaviour of numpy and it is up to the user to configure their system correctly.

2. As above, but add a warning to the FAQ.

3. Override the environment variables that define the number of threads that are used before numpy is imported. This has the disadvantage of changing the environment variable in the user's code as well as the library code.

4. Whenever numpy is to be used, fork off a subprocess with the correct environment variables. This has all the overhead associated with forking and interprocess communication, and would probably just be a huge hassle.

Personally, I found this non-obvious enough that it caused me some real problems with system resource limits (the default on the machines I'm using is 1024 threads or processes) and cost me some time debugging. I didn't expect a PDB parser to be spawning 64 threads, so I spent time combing through my own code first.

### Example
I am using CPython on `Linux-2.6.32-504.el6.x86_64-x86_64-with-redhat-6.6-Santiago` and biopython `1.71.dev0`. In one terminal, I open a python interpreter:

```
$ python
Python 3.6.1 (default, Jun 28 2017, 11:45:40) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getpid()
27015
>>> 
```
I keep that interpreter open, and in a bash prompt on the same machine I check the number of threads that it is using:
```
$ ps --no-header -L -p 27015 | wc -l
1
```
As expected. Now, in the python interpreter I import numpy and then check the number of threads in use:
```
$ ps --no-header -L -p 27015 | wc -l
64
```
The same thing happens when I import `Bio.PDB` or any other module that uses numpy.

### Numpy configuration
[...]

PR #1401, opened 2017-09-26
Add an FAQ entry for limiting NumPy threads

StefansM:
This pull request addresses issue #1397 by adding an entry to the FAQ giving some workarounds to the problem.

I won't be offended if you ask me to rewrite it: I tend to be a bit too verbose when it comes to this sort of thing.

---

``CONTRIB.rst`` files.


Doc/Tutorial/chapter_introduction.tex
```diff
@@ -312,6 +312,24 @@ \section{Frequently Asked Questions (FAQ)}
...
 
+  \item \emph{Why does Biopython start so many threads?} \\
+  Some parts of Biopython make use of \href{http://www.numpy.org}{NumPy}, which in turn makes use of a system-dependent BLAS (Basic Linear Algebra Subprograms) library. Some BLAS libraries automatically spawn one thread for each available CPU core. On systems where this is the case, simply importing NumPy (or any Biopython module relying on NumPy) will spawn several threads. The number of threads can be changed by setting an environment variable before NumPy is loaded: if your version of NumPy is linked to the Intel MKL, set \verb|MKL_NUM_THREADS| to the desired number; for OpenBLAS, set \verb|OMP_NUM_THREADS|; otherwise, consult the documentation for your BLAS library.
+
+  To do this in Python, you could put this at the beginning of your script before loading Biopython:
+  \begin{verbatim}
+  import os
+  try:
+    os.environ["OMP_NUM_THREADS"] = "1"
```
peterjc: Four space indentation please

Doc/Tutorial/chapter_introduction.tex
```diff
@@ -312,6 +312,24 @@ \section{Frequently Asked Questions (FAQ)}
   \item \emph{Why doesn't} \verb|Bio.Fasta| \emph{work?} \\
   We deprecated the \verb|Bio.Fasta| module in Biopython 1.51 (August 2009) and removed it in Biopython 1.55 (August 2010). There is a brief example showing how to convert old code to use \verb|Bio.SeqIO| instead in the \href{https://github.com/biopython/biopython/blob/master/DEPRECATED.rst}{DEPRECATED.rst} file.
 
+  \item \emph{Why does Biopython start so many threads?} \\
+  Some parts of Biopython make use of \href{http://www.numpy.org}{NumPy}, which in turn makes use of a system-dependent BLAS (Basic Linear Algebra Subprograms) library. Some BLAS libraries automatically spawn one thread for each available CPU core. On systems where this is the case, simply importing NumPy (or any Biopython module relying on NumPy) will spawn several threads. The number of threads can be changed by setting an environment variable before NumPy is loaded: if your version of NumPy is linked to the Intel MKL, set \verb|MKL_NUM_THREADS| to the desired number; for OpenBLAS, set \verb|OMP_NUM_THREADS|; otherwise, consult the documentation for your BLAS library.
```
peterjc: Should we say *"... which in turn may make use of ..."* here (ie *may*)? After all, NumPy might be built with ATLAS instead of BLAS?

peterjc:
@StefansM Other than those very minor tweaks, this looks good. I've double checked the LaTeX rendering locally (since the verbatim section can be tricky given all the special characters), and that's perfect.

@mdehoon you do a lot more with NumPy than me - any comments on this?

StefansM:
@peterjc I've fixed the indentation of the python sample. I should know better than to trust autoindent for code embedded in LaTeX documents.

I also reworded the text slightly following your suggestion. The Intel MKL isn't just a BLAS library, so I changed that to "mathematics library".

peterjc:
Looks good. If there are no objections or further suggestions, I'll merge this later this week.

peterjc:
Apparently it can be ``$OPENBLAS_NUM_THREADS`` which needs setting, see https://mail.python.org/pipermail/numpy-discussion/2017-September/077224.html

StefansM:
It looks like it can be `OPENBLAS_NUM_THREADS`, `GOTO_NUM_THREADS` or `OMP_NUM_THREADS`, depending on how OpenBLAS was linked: https://github.com/xianyi/OpenBLAS#set-the-number-of-threads-with-environment-variables

I think these are the most common libraries that NumPy is linked with:

* **Netlib LAPACK/BLAS**: Single-threaded only, nothing to set.
* **ATLAS**: [Set at compile time](http://math-atlas.sourceforge.net/faq.html#tnum), nothing that can be done.
* **BLIS**: Set [`BLIS_NUM_THREADS`](https://github.com/flame/blis/wiki/Multithreading#the-automatic-way)
* **Intel MKL**: Set [`MKL_NUM_THREADS`](https://software.intel.com/en-us/articles/recommended-settings-for-calling-intel-mkl-routines-from-multi-threaded-applications)
* [**OpenBLAS**](https://github.com/xianyi/OpenBLAS#set-the-number-of-threads-with-environment-variables):
  + Compiled with OpenMP: Set `OMP_NUM_THREADS`
  + Otherwise, set `OPENBLAS_NUM_THREADS` or `GOTO_NUM_THREADS`.
  + Practically speaking, it's probably best to set both `OPENBLAS_NUM_THREADS` and `OMP_NUM_THREADS`.

I've not actually tested whether this problem occurs with these other libraries, but it seems likely. I'm happy to add these to the FAQ entry if you think it wouldn't bulk it out too much.

peterjc:
Let's give it a day or two in case that thread on the NumPy discussion list has a more authoritative answer (e.g. in some cases is it function call time rather than import time which matters?)

peterjc:
We never did merge this - and surprisingly I don't recall any queries in the years since about threads via numpy. I'm going to just close this, sorry, rebasing is non-trivial.

Note the tutorial has since moved from LaTeX to reStructuredText and the FAQ is now here:
https://github.com/biopython/biopython/blob/biopython-184/Doc/Tutorial/chapter_introduction.rst

PR #1415, opened 2017-10-20
Fixing the D104 and removing it from the ignore list in Bio/.flake8

azneto:
This pull request addresses issue #1203 

I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ AND the _BSD 3-Clause License_.

I am not happy to be thanked by name in the ``NEWS.rst`` and
``CONTRIB.rst`` files.

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

PR #1414, opened 2017-10-19
adding docstring to __init__ functions

azneto:
Adding the docstring 'Initialize the class.' to every def __init__ lacking docstring

This pull request addresses issue #1203

I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ AND the _BSD 3-Clause License_.

I am not happy to be thanked by name in the ``NEWS.rst`` and
``CONTRIB.rst`` files.

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

PR #1413, opened 2017-10-19
adding a few examples to Bio/Seq.py

azneto:
This pull request addresses issue #1203

I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ AND the _BSD 3-Clause License_.

I am happy to be thanked by name in the ``NEWS.rst`` and
``CONTRIB.rst`` files.

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

PR #1418, opened 2017-10-20
Thank Adhemar for recent docstring work

peterjc:
See recent contributions in #1413, #1414, #1415 from @azneto


azneto:
:+1:  thanks

Issue #825, opened 2016-05-17
BioSQL tests should not require CREATE DATABASE permissions


peterjc:
Ideally from a security point of view, we'd do something like this (e.g. with PostgreSQL):

``` bash
$ psql -U postgres -c "CREATE USER biosql WITH PASSWORD 'testing'"
CREATE ROLE
$ psql -U postgres -c "CREATE DATABASE biosql_test"
CREATE DATABASE
$ psql -U postgres -c "GRANT ALL PRIVILEGES ON DATABASE biosql_test TO biosql"
GRANT
```

And allow local access to this specific database via the `pg_hba.conf` file:

```
# Access from local host for running BioSQL tests                                                                                       
local   "biosql_testing" biosql                            md5                                                                          
host    "biosql_testing" biosql      127.0.0.1/32          md5                                                                          
host    "biosql_testing" biosql      ::1/128               md5 
```

Here I'm creating the empty database `biosql_test` and an associated locked down user account `biosql` with a password `testing` (settings which would then go into the `test_BioSQL_psycopg2.py` file, and be mimicked for TravisCI). I do _not_ want to have to grant the `biosql` user permission to create new databases:

```
$ psql -U postgres -c "CREATE USER biosql WITH PASSWORD 'testing' CREATEDB"
CREATE ROLE
```

Currently (aside from SQLite), `common_BioSQL.py` function `_do_db_create()` essentially runs `DROP DATABASE biosql_test` and then `CREATE DATABASE biosql_test` before (re)importing the schema. In theory we could drop any tables and then (re)import the schema instead?

emily-bradley:
@peterjc I would like to work on this issue.

peterjc:
I am the de facto maintainer for the Biopython BioSQL code, but I am by no means a database expert. So, if you know that side of things this is a nice choice to work on, albeit a relatively low priority issue.

Note the way we configure the usernames and passwords for the BioSQL tests has changed since I opened this issue - see file ``Tests/biosql.ini.sample`` and how it is used in ``.travis.yml`` for our continuous integration tests.

https://github.com/biopython/biopython/blob/master/Tests/biosql.ini.sample

However, the general principle remains true, and would apply in cases like wanting to run the tests on a relatively locked down machine.

peterjc:
Implemented for PostgreSQL in #1420.

PR #1420, opened 2017-10-21
Changes not require postgresql CREATE DATABASE permissions.

azneto:
This pull request addresses issue #825

Changes not require postgresql CREATE DATABASE permissions.

- biosql.ini.sample and biosql.ini.appveyor
Modification to use a regular user rather than postgres

- .travis.yml and .appveyor.yml
Modification to create a database and a regular user

- .travis-tox.ini
Adding psycopg2 to every postgres version. Without this, tests are skipped.

- common_BioSQL.py
1. moving do_db_create do do_db_cleanup
2. check for DBDRIVER psycopg2 rather than pgdb
3. postgresql: delete everything from TESTEDB owned by DBUSER
4. mysql: mantain previous behaviour
   
- test_BioSQL_*
Modifications to remove the create database command and fixing a few comments

I (*am/am not*) happy to be thanked by name in the ``NEWS.rst`` and
``CONTRIB.rst`` files.


peterjc:
@ctSkennerton Conner, you've done a fair amount of work on Biopython's BioSQL support - are you able to review this pull request? Thanks!

ctSkennerton:
I'll take a look at it this week

ctSkennerton:
This looks good to me. Tests run fine on my local machine too.

Tests/common_BioSQL.py
```diff
@@ -130,48 +130,58 @@ def check_config(dbdriver, dbtype, dbhost, dbuser, dbpasswd, testdb):
...
+    """Cleanup everything from TESTDB.
 
     Relevant for MySQL and PostgreSQL.
     """
-    # first open a connection to create the database
-    server = BioSeqDatabase.open_database(driver=DBDRIVER, host=DBHOST,
-                                          user=DBUSER, passwd=DBPASSWD)
 
-    if DBDRIVER == "pgdb":
+    if DBDRIVER == "psycopg2":
```
peterjc: Should this perhaps be ``if driver in ["psycopg2", "pgdb"]:`` here?

Tests/biosql.ini.appveyor
```diff
@@ -22,6 +22,6 @@ testdb=biosql_test
 [pg]
 # Covers DBDRIVER="psycopg2" etc
 dbhost=localhost
-dbuser=postgres
-dbpasswd=Password12!
+dbuser=biosql_user
+dbpasswd=biosql_pass
```
peterjc: These changes are to match the account created in ``.appveyor.yml`` rather than the default account provided by AppVeyor, so the comment earlier is now out of date:

```python
# Values match https://www.appveyor.com/docs/services-databases/
```

I suggest moving that comment to the ``[mysql]`` section, and adding a new comment to the ``[pg]`` section.

Tests/biosql.ini.sample
```diff
@@ -20,6 +20,6 @@ testdb=biosql_test
 [pg]
 # Covers DBDRIVER="psycopg2" etc
 dbhost=localhost
-dbuser=postgres
-dbpasswd=
+dbuser=biosql_user
+dbpasswd=biosql_pass
```
peterjc: This change means that the settings no long match those provided by TravisCI, so a note that the account is created in ``.travis.yml`` would be helpful.

peterjc:
I've noted some minor issues after which I am happy to merge this.

Going further, are you familiar enough with MySQL to say if we can or should do the same there?

Tests/common_BioSQL.py
azneto: I actually had considered this. I'll add pgdb there, just in case.

Tests/biosql.ini.appveyor
azneto: OK

Tests/biosql.ini.sample
azneto: OK

azneto:
I have experience with mysql but I don't use it anymore for years. The problem is that mysql doesn't work with schemas like postgresql, so we can't use the same code to clean the database. 

It's possible to create a mysql regular user inside .travis.yml and .appveyor.yml and set the .ini files to use it. This user is able to drop and create databases the same way the root user does, so the cleanup code for mysql will remain the same. 

The only difference really is that the user will have to create a database and a user and the testing code will no longer get admin privileges.

peterjc:
OK, let's leave testing MySQL as it is. You're already in the NEWS and CONTRIB file, so other than fixing a typo, this is ready to merge. Thank you for working on this.

azneto:
Ok

peterjc:
Merged, thank you :)

PR #1429, opened 2017-10-26
Look at the charset of the content type of the handle to determine the encoding for io.TextIoWrapper

BioGeek:
This pull request addresses issue #1402.

I think the root cause is as follows: 

-  the handle that `Entrez.efetch` returns under Python 3 is a `io.TextIOWrapper` object.
- `io.TextIOWrapper`  has [an optional argument `encoding` ](https://docs.python.org/3/library/io.html#io.TextIOWrapper). It defaults to [`locale.getpreferredencoding(False)`](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding).
 - the encoding by NCBI of the record is in `utf-8` (we presume).
- If the `locale.getpreferredencoding(False)` of the user is not equal to `utf-8`, we get `UnicodeDecodeErrors`.

In the case that the record returned by NCBI is in XML format, then we can determine the encoding from the first line of the XML. But how do we determine the encoding of a record returned in text format?

Some digging in the code revealed that, in the function [`_binary_to_string_handle(handle)`](https://github.com/biopython/biopython/blob/1d22a21e238dd6b82e597325acb256a5a60f8def/Bio/_py3k/__init__.py#L99) , the `handle` is a [`http.client.HTTPResponse`](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse) object. This has a `headers` attribute which is a [`HTTPMessage`](https://docs.python.org/3/library/email.compat32-message.html#email.message.Message) object.  In our case, it has the following Content-Type:

`('Content-Type', 'text/plain; charset=UTF-8')`

So, this pull request looks at the charset of the content type of the handle to determine the encoding. If no `headers`attribute is found, it falls back to the default behaviour and uses  `locale.getpreferredencoding(False)`.

----

``CONTRIB.rst`` files.


BioGeek:
Note: all tests pass, but the file where I added an extra test, [`test_Entrez_online.py`, is skipped because internet is not available](https://ci.appveyor.com/project/biopython/biopython/build/0.0.448/job/j9bb69gmu2ru47rm#L880).

And apologies for pushing from the master branch. Next time I'll make a proper branch.

Bio/_py3k/__init__.py
```diff
@@ -94,11 +94,16 @@ def _is_int_or_long(i):
...
 
     import io
+    import locale
 
     # Python 3.4 onwards, the standard library wrappers should work:
     def _binary_to_string_handle(handle):
         """Treat a binary (bytes) handle like a text (unicode) handle."""
-        wrapped = io.TextIOWrapper(io.BufferedReader(handle))
+        try:
+            encoding = handle.headers.get_content_charset()
```
peterjc: I would add a comment here that you are checking the HTML headers which would be present if this was a network handle. Without context I find this is hard to understand.

peterjc:
This looks good - the test coverage has dropped ever so slightly because you've added code which does not get tested on TravisCI/AppVeyor where we deliberately do not run the online tests.

peterjc:
Looks good in local testing on the mac, here with the old code your new test fails in a non-UTF8 locale like ``C`` (ASCII only):

```
$ LANG=C python3 test_Entrez_online.py
test_ecitmatch (test_Entrez_online.EntrezOnlineCase)
Test Entrez.ecitmatch to search for a citation ... ok
test_efetch_biosystems_xml (test_Entrez_online.EntrezOnlineCase)
Test Entrez parser with XML from biosystems ... ok
test_efetch_gds_utf8 (test_Entrez_online.EntrezOnlineCase)
Test correct handling of encodings in Entrez.efetch ... ERROR
test_efetch_taxonomy_xml (test_Entrez_online.EntrezOnlineCase)
Test Entrez using a integer id - like a taxon id ... ok
test_egquery (test_Entrez_online.EntrezOnlineCase)
Test Entrez.egquery which searches in all Entrez databases for a single text query ... ok
test_elink (test_Entrez_online.EntrezOnlineCase)
Test Entrez.elink with multiple ids, both comma separated and as list ... ok
test_epost (test_Entrez_online.EntrezOnlineCase)
Test Entrez.epost with multiple ids, both comma separated and as list ... ok
test_espell (test_Entrez_online.EntrezOnlineCase)
Test misspellings with Entrez.espell ... ok
test_medline_from_url (test_Entrez_online.EntrezOnlineCase)
Test Entrez into Medline.read from URL ... ok
test_parse_from_url (test_Entrez_online.EntrezOnlineCase)
Test Entrez.parse from URL ... ok
test_read_from_url (test_Entrez_online.EntrezOnlineCase)
Test Entrez.read from URL ... ok
test_seqio_from_url (test_Entrez_online.EntrezOnlineCase)
Test Entrez into SeqIO.read from URL ... ok
test_webenv_search (test_Entrez_online.EntrezOnlineCase)
Test Entrez.search from link webenv history ... ok
Entrez (Bio)
Doctest: Bio.Entrez ... ok
ecitmatch (Bio.Entrez)
Doctest: Bio.Entrez.ecitmatch ... ok
efetch (Bio.Entrez)
Doctest: Bio.Entrez.efetch ... ok
egquery (Bio.Entrez)
Doctest: Bio.Entrez.egquery ... ok
einfo (Bio.Entrez)
Doctest: Bio.Entrez.einfo ... ok
elink (Bio.Entrez)
Doctest: Bio.Entrez.elink ... ok
esearch (Bio.Entrez)
Doctest: Bio.Entrez.esearch ... ok
espell (Bio.Entrez)
Doctest: Bio.Entrez.espell ... ok
esummary (Bio.Entrez)
Doctest: Bio.Entrez.esummary ... ok

ERROR: test_efetch_gds_utf8 (test_Entrez_online.EntrezOnlineCase)
Test correct handling of encodings in Entrez.efetch
  File "/Users/pc40583/repositories/biopython/Tests/test_Entrez_online.py", line 221, in test_efetch_gds_utf8
    result = handle.read()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 342: ordinal not in range(128)

Ran 22 tests in 18.583s

FAILED (errors=1)
```

Then updating to your fix, the new test passes :)

I wonder if for this test method we should explicitly temporarily change the locale to ``C``?

peterjc:
I did wondering if a better fix would be to add an optional encoding argument to ``_binary_to_string_handle``, but this private function is currently only used for network handles (in ``Bio.Entrez``, ``Bio.ExPASy``, ``Bio.KEGG`` and ``Bio.TogoWS``), so as written seems fine. We might perhaps rename this function to indicate it is primarily for network (urllib) handles?

BioGeek:
I have also thought about  temporarily changing the locale inside the test to make more explicit what we are testing, but I wasn't sure of  a proper way to do this.

Changing the name of `_binary_to_string_handle` is fine for me.

peterjc:
Merged, thank you.

If you can find a good what to cleanly change the locale for this one test, that would be worth adding as a new pull request.

Issue #1431, opened 2017-10-30
BWA test failure


pgarland:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

3.6.3 (default, Oct  3 2017, 21:16:13) 
[GCC 7.2.0]
>>> CPython
Linux-4.13.0-1-amd64-x86_64-with-debian-buster-sid
>>> 1.71.dev0

### Expected behaviour

BWA test succeeds and shows the output of running the BWA index command.

### Actual behaviour
```bash
(biopython-3.6) pgarland@rimbaud:~/Hacking/biopython/Tests$ python run_tests.py test_BWA_tool.py
Python version: 3.6.3 (default, Oct  3 2017, 21:16:13) 
[GCC 7.2.0]
Operating system: posix linux
test_BWA_tool ... FAIL
======================================================================
FAIL: test_index (test_BWA_tool.BwaTestCase)
Test for creating index files for the reference genome fasta file
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pgarland/Hacking/biopython/Tests/test_BWA_tool.py", line 102, in test_index
    % (cmdline, stdout))
AssertionError: False is not true : FASTA indexing failed:
bwa index -a bwtsw BWA/human_g1k_v37_truncated.fasta
Stdout:

----------------------------------------------------------------------
Ran 1 test in 0.493 seconds

FAILED (failures = 1)
```

### Steps to reproduce

Running the biopython test suite.

peterjc:
It would appear the test has caught a major behaviour change in BWA, which is worth knowing about. A little investigation would be needed, but assuming that to be the case and that BWA is well behaved, checking for a non-zero return makes sense. CC @saketkc (original wrapper author).

peterjc:
Cross reference Debian issue: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=880236

saketkc:
Thank you @pgarland and @peterjc! It seems to break because of the change upstream:
https://github.com/lh3/bwa/commit/298284d754aa946b24a9d40d12cd911c57e8eacd
See #1437

pgarland:
I tried the bwa banch from the repo of @saketkc and running the BWA_tool test now passes on my system.

peterjc:
Good detective work @saketkc, that BWA commit explains the stderr/stdout change.

PR #1437, opened 2017-11-01
Fix BWA tests

saketkc:
Upstream changes to stdout and stderr are flaky and hence
returncodes should be checked instead.

This pull request addresses issue #1431

``CONTRIB.rst`` files.

### DIFF
--- a/Tests/test_BWA_tool.py
+++ b/Tests/test_BWA_tool.py
@@ -91,20 +91,15 @@ class BwaTestCase(unittest.TestCase):
         cmdline = BwaIndexCommandline(bwa_exe)
         cmdline.set_parameter("infile", self.reference_file)
         cmdline.set_parameter("algorithm", "bwtsw")
-        try:
-            stdout, stderr = cmdline()
-        except ApplicationError as err:
-            self.assertTrue(err.returncode == 0,
-                            "FASTA indexing failed:\n%s\nStdout:%s\nStderr:%s\n"
-                            % (cmdline, stdout, stderr))
-
-
         stdout, stderr = cmdline()
         for extension in self.reference_extensions:
             index_file = self.reference_file + "." + extension
             self.assertTrue(os.path.exists(index_file),
                             "Index File %s not found"
                             % (index_file))
+        self.assertTrue('Finished constructing BWT' in str(stdout) + str(stderr),
+                        "FASTA indexing failed:\n%s\nStdout:%s\nStderr:%s\n"
+                        % (cmdline, stdout, stderr))
 
     def do_aln(self, in_file, out_file):
         """Test for generating sai files given the reference and read file"""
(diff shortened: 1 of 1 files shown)


Tests/test_BWA_tool.py
```diff
@@ -91,27 +91,33 @@ def test_index(self):
         cmdline = BwaIndexCommandline(bwa_exe)
         cmdline.set_parameter("infile", self.reference_file)
         cmdline.set_parameter("algorithm", "bwtsw")
+        try:
+            stdout, stderr = cmdline()
+        except ApplicationError as err:
+            self.assertTrue(err.returncode == 0,
```
peterjc: Style wise, how about ``self.assertEqual(err.returncode, 0, ...)`` here?

However, including this assert puzzles me. If we get an ``ApplicationError``, surely we expect the return code to be non-zero?

peterjc:
Is it useful to replace ``"fail to locate the index" not in stderr`` with ``"fail to locate the index" not in stdout + stderr`` given where that string appears will depend on the version of BWA?

saketkc:
@peterjc Does this look good now?

Tests/test_BWA_tool.py
```diff
@@ -97,9 +97,9 @@ def test_index(self):
             self.assertTrue(os.path.exists(index_file),
                             "Index File %s not found"
                             % (index_file))
-        self.assertTrue(stdout.startswith("[bwt_gen]"),
-                        "FASTA indexing failed:\n%s\nStdout:%s"
-                        % (cmdline, stdout))
+        self.assertTrue('Finished constructing BWT' in str(stdout) + str(stderr),
```
peterjc: I'm pretty sure ``stdout`` and ``stderr`` are both strings, so you don't need the ``str(...)`` here.

Tests/test_BWA_tool.py
```diff
@@ -108,13 +108,12 @@ def do_aln(self, in_file, out_file):
         cmdline.read_file = in_file
         self.assertTrue(os.path.isfile(in_file))
         stdout, stderr = cmdline(stdout=out_file)
-
-        self.assertTrue("fail to locate the index" not in stderr,
-                        "Error aligning sequence to reference:\n%s\nStderr:%s"
-                        % (cmdline, stderr))
+        self.assertTrue("fail to locate the index" not in str(stderr) + str(stdout),
```
peterjc: Again, I don't think you need the str(...) here.

Tests/test_BWA_tool.py
saketkc: This is required because either can be none.

```
$ python test_BWA_tool.py 
test_index (__main__.BwaTestCase)
Test for creating index files for the reference genome fasta file ... ok
test_sampe (__main__.BwaTestCase)
Test for generating samfile by paired end sequencing. ... ERROR
test_samse (__main__.BwaTestCase)
Test for single end sequencing. ... ERROR

ERROR: test_sampe (__main__.BwaTestCase)
Test for generating samfile by paired end sequencing.
  File "test_BWA_tool.py", line 150, in test_sampe
    self.do_aln(self.infile1, self.saifile1)
  File "test_BWA_tool.py", line 111, in do_aln
    self.assertTrue("fail to locate the index" not in stderr + stdout,
TypeError: cannot concatenate 'str' and 'NoneType' objects

ERROR: test_samse (__main__.BwaTestCase)
Test for single end sequencing.
  File "test_BWA_tool.py", line 132, in test_samse
    self.do_aln(self.infile1, self.saifile1)
  File "test_BWA_tool.py", line 111, in do_aln
    self.assertTrue("fail to locate the index" not in stderr + stdout,
TypeError: cannot concatenate 'str' and 'NoneType' objects

```

peterjc: Ah. That stems from where we use ``subprocess`` and the ``.communicate(...)`` method, which says it will return ``None`` if capturing that stream was not requested. This could be clearer in our ``Bio.Application`` documentation.

However, here we do request both streams (the default in the ``Bio.Application`` wrapper's ``__call__`` method), so maybe the subprocess docs are also incomplete? I am puzzled, but understand why you used ``str(...)`` now.

peterjc:
Thank you.

peterjc:
You are not yet in the NEWS file for this release, I've filed #1446 to try to avoid this in future. Comments on that are most welcome.

In the past I would just follow this up with a manual commit directly to the master branch, but now we've protected the master branch, so that all changes must go in as a pull request.

Unless you've got another pull request pending, could you open a new pull request to add yourself to the release notes for Biopython 1.71 in the ``NEWS.rst`` file please?

PR #1406, opened 2017-10-12
Add lookahead statement to regex patterns to catch overlapping cut sites

biologyguy:
Without this, any additional cut sites are consumed by re.finditer() if
 they are partially contained in another cut site, meaning they get
 missed.

## Previous behaviour (wrapped in a seqbuddy call):
```
>>> seqbuddy test.fa --find_restriction_sites SphI

# ### Restriction Sites (indexed at cut-site) ### #
Seq1
SphI	13
# ############################################### #

LOCUS       Seq1                     131 bp    DNA              UNK 01-JAN-1980
DEFINITION  Seq1.
ACCESSION   Seq1
VERSION     Seq1
KEYWORDS    .
SOURCE      .
  ORGANISM  .
            .
FEATURES             Location/Qualifiers
     SphI            8..21
ORIGIN
        1 catgcacgca tgcatgcacg caggcaaggg actcttcaaa aagaaacagg aaaaaacttt
       61 gcaagattga gaagggattg atagccaaac tgagcgcttc cctttactac aatacaaagt
      121 aaaaattttt g
//
```

## New behaviour:
```
>>> seqbuddy test.fa --find_restriction_sites SphI

# ### Restriction Sites (indexed at cut-site) ### #
Seq1
SphI	13, 17
# ############################################### #

LOCUS       Seq1                     131 bp    DNA              UNK 01-JAN-1980
DEFINITION  Seq1.
ACCESSION   Seq1
VERSION     Seq1
KEYWORDS    .
SOURCE      .
  ORGANISM  .
            .
FEATURES             Location/Qualifiers
     SphI            8..21
     SphI            12..25
ORIGIN
        1 catgcacgca tgcatgcacg caggcaaggg actcttcaaa aagaaacagg aaaaaacttt
       61 gcaagattga gaagggattg atagccaaac tgagcgcttc cctttactac aatacaaagt
      121 aaaaattttt g
//
```

I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ AND the _BSD 3-Clause License_.

I am happy to be thanked by name in the ``NEWS.rst`` and
``CONTRIB.rst`` files.

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

PR #1439, opened 2017-11-04
Restriction.py: Add lookahead statement to regex to catch overlapping site

MarkusPiotrowski:
This pull request superseeds PR #1406 from @biologyguy.

In addition to the changes in PR #1406 I did some smaller changes addressing problems:

- with ftp downloads in ```rebase_update.py``` (workaround for Python 2 and explaining why an IOError may occur for Python 3)
- with Python 2/3 differences regarding the type of ```freq``` in ```Restricion_Dictionary``` (the problem was caused by using floor division in ```...4 / ...``` in ```ranacompiler.py```) 
- I reversed an earlier bugfix which is now resolved by the lookahead statement

The commits from @biologyguy were cherry-picked and rebased against the current master. I already edited ```NEWS.rst```.


MarkusPiotrowski:
Rebasing to current master for easy committing

Scripts/Restriction/rebase_update.py
```diff
@@ -64,7 +64,21 @@ def getfiles(self, *files):
...
+            try:
+                self.retrieve(file, filename)
+                # The following line is a workaround for an urllib bug in
+                # Python 2.7.11 - 2.7.xx (?). It does not seem to work on
+                # Python 3.xx. Try to remove the line in new Python versions.
+                urlcleanup()
+            except IOError as e:
+                print(e)
+                print('This error is probably due to a non-solved ftp bug in '
+                      'recent Phyton versions. Please download the emboss '
```
peterjc: Typo (Python)

Bio/Restriction/Restriction.py
```diff
@@ -938,22 +938,20 @@ def _search(cls):
...
         revmodif = cls._rev_modify
         s = str(cls)
         cls.on_minus = []
 
-        for start, group in iterator_for:
+        for start, group in iterator:
             if group(s):
                 cls.results += [r for r in modif(start)]
+            else:
+                cls.on_minus += [r for r in revmodif(start)]
```
peterjc: Does this alter the order in which motifs on the reverse strand are reported? I am assuming everything is fine as the tests pass, but wanted to ask just in case.

peterjc:
Other than a typo and a minor query, this looks good. It would be good to add an entry to the NEWS file like:

```
The restriction enzyme list in Bio.Restriction has been updated to the
November 2017 release of REBASE.
```

Bio/Restriction/Restriction.py
MarkusPiotrowski: No, ```cls.results``` and ```cls.on_minus``` are combined afterwards and then sorted. Actually, this is just a reversion of an earlier commit of mine, where I separated the forward and reverse search, because I realized that a 'hit' on one strand prevented a 'hit' on the other strand. At that time 'lookahead search' was not in my vocabulary and indeed the test case was an overlapping recognition sequence

peterjc: Thanks for the clarification.

MarkusPiotrowski:
Again rebased to current master for easy commiting

biologyguy:
Thanks for taking over on this @MarkusPiotrowski 
Is there anything else you need me to do?

peterjc:
I think we're good - Markus include the NEWS entry I suggested, along with both your names. Merging...

Issue #1441, opened 2017-11-06
mmCIF issue parsing multine text and line-end comments


StefansM:
I've just encountered a possible issue using `Bio.PDB.MMCIF2Dict` to parse a more-complex-than-usual CIF file. This was tested on the most recent development version of biopython.

### Multiline text strings don't preserve newlines

The CIF format uses a pair of semicolons in column 1 to delimit multiline blocks of text. The current parser merges each line in the block into one:

```python
from Bio.PDB.MMCIF2Dict import MMCIF2Dict
>>> MMCIF2Dict(io.StringIO("""\
  2 data_test
  3 _test
  4 ;
  5 This should be
  6     parsed as
  7 multiline text.
  8 ;
  9 """))["_test"]
'This should beparsed asmultiline text.'
```

According to the [CIF spec](https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax) (section 17 and 20), this should be parsed as
```
\n
This should be\n
    parsed as\n
multiline text
```
with no trailing `\n`.

I think the cause of this issue is [this block of code](https://github.com/biopython/biopython/blob/e7eebb34187e55f716a30a96c5dd432ffcdb770d/Bio/PDB/MMCIF2Dict.py#L101-L105):
```python
                for line in handle:
                    line = line.strip()
                    if line == ';':
                        break
                    token += line
```

I would change that to something like this:
```python
quote_buffer = []
for line in handle:
    line = line.rstrip()
    if line == ';':
        token = "\n".join(quote_buffer)
        break
    quote_buffer.append(line)
```

This would also fix a similar problem encountered when a semicolon appears on a line of its own with an indent. That is,
```python
>>> MMCIF2Dict(io.StringIO("""\
  2 data_
  3 loop_
  4 _test
  5 ;
  6     ;
  7     Foo
  8 ;
  9 """))["_test"]
['', 'Foo', '']
```
should give the result ``["\n    ;\n    Foo"]``.

### Parsing of line-end comments

The current mmCIF parser only understands comments at the  beginning of lines:
```python
>>> MMCIF2Dict(io.StringIO("""\
  2 data_
  3 loop_
  4 _test
  5 foo # Ignore this
  6 bar
  7 baz
  8 """))["_test"]
['foo', '#', 'Ignore', 'this', 'bar', 'baz']
```
[...]

PR #1445, opened 2017-11-07
Fixes for mid-line comments and multiline text blocks in mmCIF files

StefansM:
This pull request addresses issue #1441.

The changes to lines 36&ndash;41 of `test_PDB_MMCIF2Dict.py` were necessary to correct the test. The field `_struct_ref.pdbx_seq_one_letter_code` should include the newlines, but the test previously accepted the concatenated result as correct.

The test could do with some cleanup (line 35 is over 6,000 characters long), but that doesn't fall within the scope of this pull request.

----

``CONTRIB.rst`` files.


Tests/test_PDB_MMCIF2Dict.py
```diff
@@ -21,6 +21,9 @@
 
 from Bio.PDB.MMCIF2Dict import MMCIF2Dict
 
+import io
```
peterjc: Using ``io.StringIO`` makes Python 2 and 3 at once tricky, which would explain why you needed the explicit unicode strings later.

Why we normally use ``from Bio._py3k import StringIO`` which does ``from StringIO import StringIO`` under Python 2 or ``from io import StringIO`` under Python 3.

Either was seems fine (and the difference only matters under Python 2 anyway which will eventually go away).

peterjc:
I'd like @jgreener64 to take a look at the code, but from a technical perspective it looks nicely done with tests. Good job.

Would you like to add yourself to the ``NEWS.rst`` and ``CONTRIB.rst`` files? That lets you decide exactly how to list your name (which you do not currently openly disclose on your GitHub profile).

Note the names are sorted alphabetically (see comment at top of the contributors file, this is checked by TravisCI). Alternatively I can do that by adding a commit to your branch.

Tests/test_PDB_MMCIF2Dict.py
StefansM: Ah-ha! I thought there might be something like that available, but missed `Bio._py3k`. I don't develop on Python 2, but the build bot did its job nicely and let me know there was a problem. I don't mind switching to `Bio._py3k` if you consider that more idiomatic for biopython.

peterjc: Sure - my guess is you can use ``from Bio._py3k import StringIO`` and remove the ``u`` prefixes.

Note that in general I would recommend using the third-party library ``six`` for this kind of compatibility shim, but we wanted to keep the Biopython dependencies to a minimum and have a few other more project specific hacks under ``Bio._py3k`` anyway.

peterjc:
Ah - I missed you were already in the ``CONTRIB.rst`` file. Thanks for your further contribution 👍

StefansM:
Thanks Peter. I'm already listed correctly in `CONTRIB.rst`, so I've just added myself to `NEWS.rst`.

jgreener64:
Having looked at the code I am happy with it, nice one @StefansM.

peterjc:
Thank you both :)

Issue #666, opened 2015-11-09
Allow easier access to fasta line wrapping 


ColinAnthony:
For many applications, line-wrapped fasta files pose some complications.
It would be great to have the Bio.SeqIO.FastaIO.FastaWriter: "wrap=None" option exposed to SeqIO, to allow SeqIO.write(data, handle, wrap=None, format)
or
set "wrap=none" as the default.

peterjc:
Hi Colin. SeqIO is deliberately simple and avoids this. What would you expect the wrap argument to do with FASTQ (line wrapping is possible but strongly discouraged), GenBank (can sometimes be useful to disable line wrapping in breach of the spec), etc? I fear this is the thin end of the wedge...

[Out of interest, why do you need FASTA line wrapping anyway? The default has rarely if ever caused me problems.]

digfish:
The problem is, it's impossible to use a FASTA file with fixed wrapping like that, because the function parse() in the same module Bio.SeqIO is not even able to read that file ! 
Try it for yourself !
If I want wrapping I just turn on the wrapping on the text editor !
Let's not complicate matters: a FASTA file have two lines for every sequence: the first one with the identifiers and the second one with the proper sequence !

ColinAnthony:
The problem I was having was that I was getting line wrapping as a default
with SeqIO.write . I don't want line wrapping.
I want a two line per record fasta file.
From your reply, it seems that this shouldn't be the case, so perhaps I am
missing something.
I recall opening files in gedit and seeing the lines were wrapped (also if
I 'head' the file in bash , the lines are wrapped).
I will double check my code and the output. Perhaps something else is going
on here?

peterjc:
It seems @digfish has found a separate problem with the FASTA parser - it can usually read the FASTA files written by Biopython: Please file a separate bug for that.

In any case, the convention is very clear: FASTA sequence lines can be wrapped, and in most cases *are* wrapped. Remember the sequence can be millions (even billions) or letters long in the case of chromosomes - and singles lines of that length do not work well at the command line terminal nor in most editors.

However, as @ColinAnthony notes, there are times (when dealing with shorter sequences) that no-line-wrapping FASTA files with exactly two lines per record have advantages.

digfish:
I'm working with reads with approximately 255 bp. I understand the motivation for having wrapping for reads with more than > 1000 bp. 
[...]

PR #1450, opened 2017-11-09
add fasta-nowrap to SeqIO.write API

Zsailer:
This pull request addresses issue #666. pinging @peterjc for review. 

It extends the `SeqIO.write` API to allow writing to fasta format without line wrapping by adding a `fasta-nowrap` option to the format list. 

To implement this, I subclassed `FastaWriter` to `FastaNoWrapWriter` and set `wrap=None` in super. Not sure if this is the best way to do this. I'm happy to change it.

This doesn't update the Docs. I wanted to make sure the logic works before changing the docs. 

**Contributing stuff**

``CONTRIB.rst`` files.


Zsailer:
Oops, should have ran tests before submitting. Sorry. 

Looks like the tests are written such that `SeqIO.parse` must be symmetric to `SeqIO.write`. That means the parsing formats must also have a "fasta-nowrap" option. Should I just use the standard fasta parser as a place holder to start?

peterjc:
The place holder makes sense - up until now all the output formats have had a parser, so the tests reflect that. You could instead try adding a special case to the tests?

peterjc:
I think to fix the test, you'd need to update this line

https://github.com/biopython/biopython/blob/biopython-170/Tests/test_SeqIO.py#L433

```python
    elif format == "fasta":
```

to:

```python
    elif format in ["fasta", "fasta-nowrap"]:
```

Zsailer:
That helped fix some of the test. 

I'm breaking some new tests. Looks like I need to add a bunch of "expected output" lines to `output.test_SeqIO`.

Zsailer:
Ok, all tests are passing now.

peterjc:
I hope you spotted ``run_test.py -g test_SeqIO.py`` (see the testing chapter in the tutorial), but anyway - you got there :)

Would you like to add yourself to the ``NEWS.rst`` and ``CONTRIB.rst`` files too (sorted alphabetically)?

Zsailer:
Oh wow, I totally missed that. :man_facepalming: 

Thanks for the review! I've added myself to `NEWS.rst` and `CONTRIB.rst`. If you need anything else from me, let me know. :)

peterjc:
(The TravisCI Python 2.7 test failed in the KD Tree code, looks like a stochastic failure and I have re-run that)

Zsailer:
Thanks @peterjc. Looks like the tests passed this time!

peterjc:
See https://github.com/biopython/biopython/issues/666#issuecomment-350701713 where perhaps ``fasta-2line`` would be better than ``fasta-nowrap`` (important for parsing).

peterjc:
(Hang on, I've had a rebase glitch while adding the parser on top of your branch...)

peterjc:
I had to do a ``git rebase`` and ``git push --force`` to sort out my initial glitch which pulled some unrelated commits onto this pull request.

Anyway, how does this look @Zsailer?

There is scope for also implementing SeqIO indexing for the new FASTA variant as well.

peterjc:
Updated again since master branch now enforces ``pydocstyle`` check D301:

```
$ flake8 Bio/SeqIO/FastaIO.py
Bio/SeqIO/FastaIO.py:76:1: D301 Use r""" if any backslashes in a docstring
```

Zsailer:
@peterjc, this looks good! Thanks for pushing this forward!

peterjc:
Great, thanks. Once the tests have passed on this, I'll apply it.

(The stricter docstring checking is fairly recent, it will make life harder for a few of the already open pull requests)

peterjc:
@bow Does this look good to you? We could add more on the tests side, and perhaps implement direct indexing of two-line-per-record FASTA as well - but this seems a useful unit of work now.

bow:
@peterjc this looks good ~ and @Zsailer thanks for the PR! My only small question is about the writer: is there a reason it is called `FastaNoWrap...` while the iterator is `FastaTwoLine..` and the proposed format name is `fasta-2line`?

peterjc:
Good catch on the write class name - I agree with you. I'll fix that before merging this.

PR #1453, opened 2017-11-14
Clusterbuster support

j-andrews7:
This pull request addresses issue #909.

This pull request contains support for additional motif formats originally written by @ghuls with a few updates and additional tests to fix a few small issues. All relevant tests and flake8 conformance pass in my environment, let me know if there are any issues.

``CONTRIB.rst`` files.


peterjc:
There was a weird merge commit at the start of your branch so I rebased this to the master, and also added your names to the ``NEWS.rst`` file (you are both already in the ``CONTRIB.rst`` file).

peterjc:
You'd also accidentally checked in a bunch of compiled files, which I have fixed and updated the ``.gitignore`` file which ought to have prevented this:

```
Bio/Cluster/cluster.cpython-36m-x86_64-linux-gnu.so
Bio/KDTree/_CKDTree.cpython-36m-x86_64-linux-gnu.so
Bio/Nexus/cnexus.cpython-36m-x86_64-linux-gnu.so
Bio/cpairwise2.cpython-36m-x86_64-linux-gnu.so
Bio/motifs/_pwm.cpython-36m-x86_64-linux-gnu.so
Bio/trie.cpython-36m-x86_64-linux-gnu.so
```

peterjc:
Right now this is also (accidentally) removing a test from ``Tests/test_motifs.py``, starting:

```python
    def test_meme_parser_rna(self):
         """Test if Bio.motifs can parse MEME output files using RNA."""
         handle = open("motifs/meme.rna.oops.txt")
         ....
```

Bio/motifs/__init__.py
```diff
@@ -27,13 +27,17 @@ def parse(handle, format):
...
-     - pfm:           JASPAR-style position-frequency matrix
-     - jaspar:        JASPAR-style multiple PFM format
-     - sites:         JASPAR-style sites file
+     - AlignAce:         AlignAce output file format
+     - ClusterBuster:    Cluster Buster position frequency matrix format
+     - XMS:              XMS matrix format
+     - MEME:             MEME output file motif
+     - MAST:             MAST output file motif
+     - TRANSFAC:         TRANSFAC database file format
+     - pfm_four_columns: Generic position-frequency matrix format with four columns. (cisbp, homer, hocomoco, neph, tiffin)
```
peterjc: Style wise, in ``Bio.SeqIO`` etc we've used the convention ``format-variant`` with a minus sign not an underscore, and generally one word variants. How about ``pfm-4column`` or ``pfm-4col`` here or even ``pfm-cols``? Likewise for the row-based pfm format (next line of the documentation).

In any case, looking at the code, both call the same pfm parser, so I would think just exposing a single ``pfm`` format ought to be fine and easier for the user.

Bio/motifs/clusterbuster.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2015 by Gert Hulselmans.  All rights reserved.
+# This code is part of the Biopython distribution and governed by its
```
peterjc: Please use the dual license header as in ``Bio/SeqRecord.py``:

```python
# package.
```

Bio/motifs/xms.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2015 by Gert Hulselmans.  All rights reserved.
+# This code is part of the Biopython distribution and governed by its
```
peterjc: Please use the dual license header as in ``Bio/SeqRecord.py``:

```python
# package.
```

Bio/motifs/__init__.py
ghuls: @peterjc **pfm-cols** and **pfm-rows** are fine for me. I never liked it that JASPAR uses **pfm** instead of **pfm-jaspar**. It would be nicer if at least the **pfm-jaspar** alias is added.
The **cols** and **rows** option is important for the parsing of the motifs.
Sometimes it could be autodetectable which one to use, but for a 4x4 motif matrix without ACGT or position annoation, this is impossible.

peterjc:
@j-andrews7 are you OK with how to grab my updated version of the branch, and work from that (rather than your now out of date local copy)? i.e. how to do this with ``git``?

Probably something like this:

```
git checkout master
git branch -D clusterbuster  # this will delete your local branch & changes!
git fetch andrew # assuming that's what your remote is called
git checkout -t andrew/clusterbuster # this will checkout the updated branch
```

The safer but more round-about option (a good idea if you are still not confident with git) is to take a fresh clone of the entire repository (leaving your old copy as a backup until this is done).

Bio/motifs/__init__.py
peterjc: Does that mean something is missing in the code to pass the row/col information to the parser?

peterjc: I missed there already was already ``pfm`` defined for JASPAR style PFM. Is this different from the 4 row / 4 col PFM?

If @mdehoon does not object, adding ``pfm-jaspar`` as an alias makes sense.

j-andrews7:
@peterjc Yeah, checking out is no issue - I actually use the GitHub Desktop client whenever possible because I'm terrible at remembering the commands, and it's pretty dead-simple to use. Sorry for missing so many things. I'll update whenever I get a chance today.

I agree with the naming convention, I just tried to touch @ghuls original code as little as possible.

Bio/motifs/__init__.py
j-andrews7: The JASPAR style 'pfm' won't parse all of the other generic 4 row/column formats, just those from JASPAR. However, `pfm-rows` does handle the JASPAR style PFM, making the `pfm` command redundant and unnecessary really.

I'd suggest removing/deprecating the `pfm` command from the JASPAR module, as `pfm-rows` and `pfm-cols` will be much more flexible. If @mdehoon agrees, I will add a test or two for `pfm-rows` and `pfm-cols` as well.

ghuls: @j-andrews7 If I am not mistaken `pfm-rows` could also be used instead of most of the code in `_read_jaspar`. Only some code should be written to get the correct `identifier` (should be pfm `name`) and `name` (should be pfm `gene` or `name` if `gene` was not found) attributes: 
```
def _read_jaspar(handle):
    ...

        if head_match:
            identifier = head_match.group(1)
            if head_match.group(3):
                name = head_match.group(3)
            else:
                name = identifier

```

j-andrews7: @ghuls Yes, I also noticed that and you are correct that all that needs to be changed would be the identifier/name split. That function could also be deprecated with little work.

If @peterjc and/or @mdehoon thinks that's a good idea, I can also incorporate those changes.

peterjc:
This sounds like a good argument for making ``pfm`` an alias for ``pfm-rows`` and use the new parser. We could then deprecate the old Jasper PFM parser (rather than removing it, in case anyone is using it directly). Does that sound sensible @mdehoon / @barwil ?

peterjc:
New file ``Bio/motifs/pfm.py`` should also be dual licensed.

I'd like to have existing files ``Tests/test_motifs.py`` and ``Tests/test_motifs_online.py`` dual licensed, but that needs a contributor audit using #898 to confirm we have permission.

```
$ git log Tests/test_motifs_online.py | grep "^Author: " | sort | uniq
Author: Nader Morshed <XXX>
Author: peterjc <XXX>
```

That and the copyright statements means three contributors, me (fine), Nader (@naderm, agreed) and Antony Lee (@anntzer via #565, agreed).  So we can dual license ``Tests/test_motifs_online.py`` no problem.

```
$ git log Tests/test_motifs.py | grep "^Author: " | sort | uniq
Author: Antony Lee <XXX>
Author: Ariel Aptekmann <XXX>
Author: Bertrand Néron <XXX>
Author: Christian Brueffer <XXX>
Author: Gert Hulselmans <XXX>
Author: Jared Andrews <XXX>
Author: Peter Cock <XXX>
Author: Zaid Ur-Rehman <XXX>
Author: mdehoon <XXX>
Author: peterjc <XXX>
```

The copyright header also list Bartek Wilczynski (original file author, @barwil), and I think everyone has agreed to the dual licensing, except @zaidurrehman who I have just asked via https://github.com/biopython/biopython/issues/898#issuecomment-344545532

peterjc:
Update: That looks fine for all the contributors, so ``Tests/test_motifs.py`` can also be dual-licensed 👍

ghuls:
@j-andrews7 **clusterbuster.py** read method can also be replaced with `pfm-rows` parser (A small disadvantage is that in that case the clusterbuster method will parse motif files which contain invalid clusterbuster motifs (e.g. using spaces instead of TABS), but I think it shouldn't be a big problem as in that case the user can write the read clusterbuster motifs back to a file in the correct format). One thing that is currently not supported in the clusterbuster parser (as I didn't need this support) is reading `# WEIGHT X.X` attribute for a motif in ClusterBuster format.

j-andrews7:
Alright, I will change any licenses I haven't already. I'll also switch the aliases around as discussed and make the appropriate changes to `pfm-row` so that it properly grabs JASPAR identifiers and gene names as the old `pfm` does. 

@ghuls, I think I will just add support for `# WEIGHT x.x.` and `# GAP x.x` and leave the clusterbuster parser alone. We obviously can't support every motif format in the world, but if there have been concrete previous use cases and the code is more or less already complete, I'd rather leave it be and have complete support for the format when possible.

I should have some time to make those changes later today or tonight.

j-andrews7:
Okay, I have switch the `read` functions for `pfm` and `jaspar` to be picked up by the `pfm` module, rather than the `jaspar` one, though it is still used for writing for explicitness (though using the `pfm` format will match the `jaspar` format if it picked up on the `matrix_id` and `name` properties anyway).

Overall, I feel it definitely makes things more flexible. I added some tests, but it could probably use more. As far as I can tell, everything is more or less squared away, but it could definitely use another set of eyes. Thanks for your patience.

ghuls:
Shouldn't `if pfm_format == "pfm_cols":` and `elif pfm_format in ("pfm_rows", "pfm", "jaspar"):` be written with a "-" instead of "_" in `Bio/motifs/pfm.py`?

I think it would be better to not set the motif weight by default for Cluster-Buster.
So change the code to something like this (untested, but I think it is correct).
```python
    for line in handle:
        line = line.strip()
        if line:
            if line.startswith('>'):
                if motif_nbr != 0:
                    motif = motifs.Motif(alphabet=IUPAC.unambiguous_dna, counts=nucleotide_counts)
                    motif.name = motif_name
                    motif.weight = motif_weight
                    motif.gap = motif_gap
                    record.append(motif)

                motif_name = line[1:].strip()
                motif_weight = None
                motif_gap = None
                nucleotide_counts = {'A': [], 'C': [], 'G': [], 'T': []}
                motif_nbr += 1


...


def write(motifs):
    """Return the representation of motifs in Cluster Buster position frequency matrix format."""
    lines = []
    for m in motifs:
        line = '>{0}\n'.format(m.name)
        lines.append(line)

        if m.weight is not None:
            line = '# WEIGHT {0}\n'.format(m.weight)
            lines.append(line)

        if m.gap is not None:
            line = '# GAP {0}\n'.format(m.gap)
            lines.append(line)
```

j-andrews7:
@ghuls good catch, will change that and add another test or two.

The motif weight for ClusterBuster is 1.0 by default per their specifications. Is there any particular reason you don't think it should be written to output if someone is explicitly using `clusterbuster`?

ghuls:
@j-andrews7 
I know that the default motif weight is 1.0, but in the example file on:
http://cagt.bu.edu/page/ClusterBuster_download
https://github.com/weng-lab/cluster-buster/

`# WEIGTH 1.0` is not set by default. Also a lot of internal code I use, assumes there is no `# WEIGHT` attribute.

j-andrews7:
@ghuls yeah, I guess the weight is internally assumed by ClusterBuster itself. I'll make that change as well.

j-andrews7:
Docs still need to be written, but I probably won't have much time in the next week to do so. Will try to get to it as soon as I can, or someone else can hash it out and explain that most generic row or column-based multiple motif formats should have at least minimal support.

j-andrews7:
@peterjc is there a guideline somewhere for contributing to the docs? Or instructions for building them? I've looked around but I feel like I'm missing it. Sorry if it's somewhere obvious.

peterjc:
If you mean the main Tutorial, ``cd Docs`` and then ``make`` should work on a Mac or Linux machine. You will need LaTeX installed (``pdflatex``) and Hevea (ideally - I can explain how to bypass that if it is too painful to install on your system).

If you mean the API docs from the docstrings, currently we build them for each release using epydoc but I'm working on moving to sphinx-apidoc for that:
http://mailman.open-bio.org/pipermail/biopython-dev/2017-November/021884.html

Bio/motifs/xms.py
```diff
@@ -0,0 +1,95 @@
...
+                retlist.append(node.wholeText)
+            elif node.hasChildNodes:
+                retlist.append(self.get_text(node.childNodes))
+
+        return re.sub('\s+', ' ', ''.join(retlist))
+
+
+class Record(list):
+    """A Bio.motifs.xms.Record stores the information in a XMS matrix table.
+    The record inherits from a list containing the individual motifs.
```
peterjc: That needs a blank line to separate the one-line summary from the rest of the docstring, see https://github.com/biopython/biopython/issues/1170

(We now enforce this via TravisCI/Tox)

j-andrews7: Fixed. I don't know if I'm going to have time to add a section to the docs about this. A line or two saying to look at the API docs for `pfm.py` might be a good idea, as it should provide at least basic functionality for most common formats.

peterjc:
There are still a couple of trivial docstring tweaks needed - as noted earlier we didn't enforce this when you started the pull request, so apologies for the surprise:

```
$ flake8 Bio/
Bio/motifs/clusterbuster.py:13:1: D205 1 blank line required between summary line and description
Bio/motifs/pfm.py:15:1: D205 1 blank line required between summary line and description
```

(I don't mind fixing this myself if you've ticked the pull request option to allow edits from the Biopython maintainers)

ghuls:
Any update on this pull request?

j-andrews7:
I think it just needs a few spaces added and I haven't had time to add it.
If anybody else wants to, be my guest.

peterjc:
Looking over this again with fresh eyes, I can't find an explicit agreement from Bartek Wilczynski (@barwil) to dual license his contributions (which covers a lot of the motifs code and its tests). I will email him again.

peterjc:
I've tried emailing Bartek again, cross reference #898.

ghuls:
@peterjc Any update on this?

peterjc:
It needs updating for a merge conflict plus slightly stricter style checks, hang on...

peterjc:
Rebased with flake8 cleanup & merge conflicts fixed.

peterjc:
Yes, I'd made a simple merge conflict mistake - was missing a ``return record`` line for the minimal parser case.

peterjc:
@mdehoon can we go ahead with this as a rebase-and-merge? I've also done the dual licensing checks.

mdehoon:
@peterjc  OK

peterjc:
Thank you everyone - I'll just correct the placement of your names into the new release (should have spotted that earlier).

PR #1470, opened 2017-12-07
Add discoMegablast keywords to qblast + test

mxposed:

### DIFF
--- a/Tests/test_NCBI_qblast.py
+++ b/Tests/test_NCBI_qblast.py
@@ -48,15 +48,16 @@ class TestQblast(unittest.TestCase):
         # GI:160837788 aka NP_075631.2
         # the actin related protein 2/3 complex, subunit 1B [Mus musculus]
         self.run_qblast("blastp", "nr", "NP_075631.2", 0.001,
-                        "rat [ORGN]", ['9506405', '13592137', '37589612', '149064087', '56912225'])
+                        "rat [ORGN]", dict(megablast='FALSE'), ['9506405', '13592137', '37589612', '149064087', '56912225'])
 
     def test_pcr_primers(self):
         # This next example finds PCR primer matches in Chimpanzees, e.g. BRCA1:
         self.run_qblast("blastn", "nr", "GTACCTTGATTTCGTATTC" + ("N" * 30) + "GACTCTACTACCTTTACCC",
-                        10, "pan [ORGN]", ["XM_009432096.2", "XM_009432102.2", "XM_009432101.2",
-                                           "XM_016930487.1", "XM_009432104.2", "XM_009432099.2",
-                                           "XR_001710553.1", "XM_016930485.1", "XM_009432089.2",
-                                           "XM_016930484.1"])
+                        10, "pan [ORGN]", dict(megablast='FALSE'),
+                        ["XM_009432096.2", "XM_009432102.2", "XM_009432101.2",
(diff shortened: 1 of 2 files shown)


Tests/test_NCBI_qblast.py
```diff
@@ -71,23 +72,74 @@ def test_orchid_est(self):
...
+                        CTGAACAGCAACAGCAGCGGCAGCGGCAACAACAGCAACGACAACAGCGGCAGCAGCAGCCCCAGCAGCA
+                        GCAAGACCAACACCCTGAACCAGCAGAGCATCTGCATCAAGAGCGAGATCCAACGATACGTTGAAATTCG
+                        CTTGTGTGCCACTGGTAAATCCACCCCCCCTAAGCCTCTAATAGGGAGACCTTAG""",
+                        0.0000001, None, dict(
+                            template_type=0,
+                            template_length=18,
+                            megablast='on',
+                        ), ['XM_635681.1', 'XM_008496783.1'])
+
+    def run_qblast(self, program, database, query, e_value, entrez_filter, additional_args, expected_hits):
```
peterjc: You've inserted ``additional_args`` into the argument list, rather than adding it to the end. If this was a public API, that would be bad. As this is internal to the test script only, it should be fine.

Tests/test_NCBI_qblast.py
```diff
@@ -48,15 +48,16 @@ def test_blastp_nr_actin(self):
         # GI:160837788 aka NP_075631.2
         # the actin related protein 2/3 complex, subunit 1B [Mus musculus]
         self.run_qblast("blastp", "nr", "NP_075631.2", 0.001,
-                        "rat [ORGN]", ['9506405', '13592137', '37589612', '149064087', '56912225'])
+                        "rat [ORGN]", dict(megablast='FALSE'), ['9506405', '13592137', '37589612', '149064087', '56912225'])
```
peterjc: Style wise, avoid the very long line by using more line breaks.

Also why not write ``{'megablast': 'FALSE'}`` here (and likewise later on)?

peterjc:
I've noted some minor style things. @bow is this something you could review please?

Tests/test_NCBI_qblast.py
mxposed: I wanted to keep `expected_hits` the last parameter for semantic reasons. I can change this so `additional_args` would be an optional keyword argument.

Tests/test_NCBI_qblast.py
mxposed: No special reason. I'll change it to `{'megablast': 'FALSE'}` then

Tests/test_NCBI_qblast.py
peterjc: OK, yes - that is a good enough reason.

bow:
@peterjc Aside from the current comments, the rest seems fine ~.

mxposed:
Changed the `dict` syntax

peterjc:
@mxposed Do you want to be listed and thanked by name in the NEWS.rst and CONTRIB.rst files?

mxposed:
@peterjc I'll skip for now, thank you)

peterjc:
@mxposed OK, that's fine.

peterjc:
Merged, thank you for your contribution.

Issue #1473, opened 2017-12-11
Use imperative mood for one-line docstring summaries


peterjc:
This a low priority issue for better following the wider Python community style for our docstring documentation. This is a spin out from issue #1170.

For background, read PEP257 first https://www.python.org/dev/peps/pep-0257/ (and note we use reStructuredText for the docstrings, not just plain text).

To test this locally, first install ``pydocstyle``,

```
$ pip install pydocstyle
```

Then:

```
$ pydocstyle Bio/ --select D401 --count
Bio/_utils.py:38 in public function `read_forward`:
        D401: First line should be in imperative mood ('Read', not 'Reads')
...
Bio/Wise/__init__.py:26 in private function `_build_align_cmdline`:
        D401: First line should be in imperative mood; try rephrasing (found 'Helper')
416
```

If we do manage to make ``pydocstyle`` happy, then ``D401`` can be enforced on TravisCI/Tox by updating file ``Bio/.flake8`` which currently ignores it.

Given there are hundreds of docstrings to review, it makes sense to do this in batches, for example by top level folder. If you are going to work on this please comment here first to avoid overlap of effort (e.g. "I'm going to work on ``Bio/BLAST/`` next).

MarkusPiotrowski:
I'm going to work on ```Bio/SeqUtils```

MarkusPiotrowski:
I go on with ```Alphabet```, ```Compass``` and ```Data```

Gasta88:
I'll take `Bio/_utils.py`, `Bio/SeqRecord.py` and `Bio/phenotype`.

Gasta88:
I'll start with `Bio/Nexus`, `Bio/Wise` and `Bio/PopGene`

Gasta88:
I'll work on `Bio/Entrez`, `Bio/PDB`, `Bio/FSSP` and `Bio/Phylo`.

peterjc:
Wow - this is going quicker than I expected - thanks for all the work so far!

Gasta88:
Working on `Bio/Uniprot`, `Bio/motifs`, `Bio/KEGG`, `Bio/NMR` and `Bio/SCOP`.

Gasta88:
I'll take `Bio/SeqIO`, `Bio/Crystal` and `Bio/SearchIO`.

peterjc:
I'll do ``Bio/Sequencing/*.py`` now...

Gasta88:
[...]

PR #1475, opened 2017-12-13
SeqUtils module: PEP8 and PEP257

MarkusPiotrowski:
This pull request addresses issue #1473 D401: Use imperative mood... for the SeqUtils module


Bio/SeqUtils/MeltingTemp.py
```diff
@@ -716,8 +716,8 @@ def Tm_GC(seq, check=True, strict=True, valueset=7, userset=None, Na=50, K=0,
         sum(map(seq.count, ('B', 'V'))) * 66.67 / len(seq) + \
         sum(map(seq.count, ('D', 'H'))) * 33.33 / len(seq)
     if strict and tmp:
-        raise ValueError('ambiguous bases B, D, H, K, M, N, R, V, Y not ' +
-                         'allowed when strict=True')
+        raise ValueError('ambiguous bases B, D, H, K, M, N, R, V, Y not '
+                         'allowed when \'strict=True\'')
```
peterjc: Maybe use the double-quote for the string limits, to avoid escaping the single quotes you've added?

Bio/SeqUtils/ProtParam.py
```diff
@@ -4,8 +4,11 @@
 """Simple protein analysis.
 
 Example::
-
-    X = ProteinAnalysis("MAEGEITTFTALTEKFNLPPGNYKKPKLLYCSNGGHFLRILPDGTVDGTRDRSDQHIQLQLSAESVGEVYIKSTETGQYLAMDTSGLLYGSQTPSEECLFLERLEENHYNTYTSKKHAEKNWFVGLKKNGSCKRGPRTHYGQKAILFLPLPV")
+    protein = '''
+    MAEGEITTFTALTEKFNLPPGNYKKPKLLYCSNGGHFLRILPDGTVDGTRDRSDQHIQLQLSAESVGEVYIKSTE
+    TGQYLAMDTSGLLYGSQTPSEECLFLERLEENHYNTYTSKKHAEKNWFVGLKKNGSCKRGPRTHYGQKAILFLPL
+    PV'''
```
peterjc: That gives a different string - it would have new line characters embedded in it. The implicit line concatenation would be better if you really want to avoid the long line.

Since you are working on this, can you make it into a doctest too?

Bio/SeqUtils/ProtParam.py
```diff
@@ -231,21 +240,25 @@ def protein_scale(self, param_dict, window, edge=1.0):
...
+                                        subsequence[window - j - 1]))
 
             # Now add the middle value, which always has a weight of 1.
             middle = subsequence[window // 2]
             if middle in param_dict:
                 score += param_dict[middle]
             else:
-                sys.stderr.write('warning: %s  is not a standard amino acid.\n' % (middle))
+                sys.stderr.write('warning: %s  is not a standard amino acid.\n'
+                                 % (middle))
```
peterjc: Could just write ``middle`` rather than ``(middle)`` here.

MarkusPiotrowski: Yes

Bio/SeqUtils/ProtParam.py
MarkusPiotrowski: Yes.

Bio/SeqUtils/MeltingTemp.py
MarkusPiotrowski: I used the escaped single quotes elsewhere in the same code, so I stayed with them. No problem to replace them all.

peterjc: If you were following the existing module's style, that's fine then. Leave it.

MarkusPiotrowski:
@peterjc I wrote some doctests for ```ProtParam```, but not for all methods that where listed before, since some of them return large lists of floats, which are hard to test as docstyle.

The bad codecov/patch statistitics seems to be due to the fact that I did some PEP8 changes in error messages, and since error conditions are often not tested the relative amount of untested code in my changes is high.

peterjc:
The doctest looks perfect - good judgement in which methods to test and which to skip due to verbose output.

I have added ``Bio.SeqUtils.ProtParam`` to the list ``Tests/run_tests.py`` so that TravisCI etc will including the doctest.

If the automated tests still pass this is ready for merging, thank you!

MarkusPiotrowski:
Upps. Didn't update ```Tests/run_tests.py```...

peterjc:
As some point we might want to enable ``# doctest: +ELLIPSIS`` on all our doctests, doing this here as a one-off directive seems practical for now.

peterjc:
Maybe just remove the troublesome new test using the ellipsis?

MarkusPiotrowski:
OK ```ELLIPSIS``` wasn't the problem (it is turned on for doctests in ```Tests/run_tests.py```, but the output was in a different order. So the next commit will solve the problem (???)

MarkusPiotrowski:
Doctest for ```ProteinParam.py``` is running on TravisCI and everything is OK

peterjc:
Lovely, thank you.

PR #1476, opened 2017-12-14
Support PEP 519 in IO functions (anything using `File.as_handle`)

mruffalo:
This pull request is similar to others I made in PyTables (https://github.com/PyTables/PyTables/pull/640) and SciPy (https://github.com/scipy/scipy/issues/6854, https://github.com/scipy/scipy/pull/6855). This allows users to pass `pathlib.Path` objects to things like `SeqIO.parse`, and the logic changes required for this were quite small. As described in https://github.com/scipy/scipy/issues/6854, we can simply pass any unknown objects to `open` instead of checking for membership in a small set of predefined types. As before, `File.as_handle` ignores (returns unchanged) any objects it doesn't understand.

All tests passed locally under Python 3.6, and I'm going to depend on the Travis CI service to test other versions after I submit this pull request.

----------

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style checks pass with these changes.

I am happy be thanked by name in the ``NEWS.rst`` and ``CONTRIB.rst`` files, but this isn't very important to me, and though I already described this change in ``NEWS.rst``, I didn't feel compelled to add myself to the contributors for this release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ Doc/*/*/hevea.sty
 #Ignore IntelliJ IDEA directory and project files
 .idea
 *.iml
+
+#Ignore unittest cache dirctory
+.cache/
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -46,16 +46,20 @@ except ImportError:
 def as_handle(handleish, mode='r', **kwargs):
     r"""Context manager to ensure we are using a handle.
 
-    Context manager for arguments that can be passed to
-    SeqIO and AlignIO read, write, and parse methods: either file objects or strings.
+    Context manager for arguments that can be passed to SeqIO and AlignIO read, write,
+    and parse methods: either file objects or path-like objects (strings, pathlib.Path
+    instances, or more generally, anything that can be handled by the builtin 'open'
+    function).
 
-    When given a string, returns a file handle open to handleish with provided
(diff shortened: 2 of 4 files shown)


peterjc:
Looks good to me, but not having looked at this bit of Python 3 before, I'd like a second opinion. Maybe @bow or @etal since their modules would also be affected?

bow:
Thanks for the PR @mruffalo , this looks good to me as well. I've started to use `pathlib.Path` considerably in other projects and I could see that some people will prefer to use this than plain strings. Making sure they work with `as_handle` per PEP 519 should be a useful change for many.

peterjc:
Thanks @bow - and thank you @mruffalo, applied to master ready for our next release.

mruffalo:
My pleasure. It's always nice to be able to remove the extra `str` calls from my code when using convenience IO functions in third-party packages.

Also, I just realized that my change to `NEWS.rst` isn't particularly clear about when and how this change works, and that people might expect to be able to pass `pathlib.Path` objects to I/O functions under Python 3.4 and 3.5. A small tweak to my wording seems to be appropriate, and I'll make a new pull request for that momentarily.

peterjc:
Thank you for your attention to detail 👍

PR #1477, opened 2017-12-15
abi URL updates

dcroote:
This pull request addresses a couple of broken URLs.

I do not need to be thanked by name in the ``NEWS.rst`` and ``CONTRIB.rst`` files given how minor this is.


bow:
Good catch @dcroote & thanks for the PR!

In an ideal world we should parse all the URLs in our docs so we can periodically update dead links. But that's for another day perhaps :).

PR #1483, opened 2017-12-24
reserved word loop_ should be case insensitive

Ashafix:
Probably a minor issue but the reserved word loop_ should case insensitive. The code was changed accordingly and a few unit tests were added.

https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax
> In addition the following case-insensitive reserved words may not occur as unquoted data values.
> loop_ | identifies looped list of data

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -31,7 +31,7 @@ class MMCIF2Dict(dict):
             i = 0
             n = 0
             for token in tokens:
-                if token == "loop_":
+                if token.lower() == "loop_":
                     loop_flag = True
                     keys = []
                     i = 0
--- a/Tests/test_PDB_MMCIF2Dict.py
+++ b/Tests/test_PDB_MMCIF2Dict.py
@@ -107,6 +107,31 @@ class MMCIF2dictTests(unittest.TestCase):
         self.assertEqual(mmcif_dict["_test_key_value"], "foo")
         self.assertEqual(mmcif_dict["_test_loop"], list("abcdefg"))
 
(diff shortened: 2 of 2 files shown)


Bio/PDB/MMCIF2Dict.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, filename):
             i = 0
             n = 0
             for token in tokens:
-                if token == "loop_":
+                if token.lower() == "loop_":
```
peterjc: Should we raise an error (or at least a warning) if a non-lowercase "loop_" is found?

jgreener64: While a little unusual, the formal spec allows any use of upper/lower case in the word so I don't think it should give a warning or error.

https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax

peterjc: Agreed, https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax is very clear that any case is allowed - even ``lOoP_`` would be valid.

peterjc:
@jgreener64 since you've looked at this & know the format, do you think it is OK to be merged? Thanks.

jgreener64:
@peterjc looks okay to me.

peterjc:
@Ashafix would you like to add yourself to the ``NEWS.rst`` and ``CONTRIB.rst`` files as part of this pull request? Note the names are sorted alphabetically.

Ashafix:
@peterjc: Thanks but I am already on the contributors list, just with my real name.

peterjc:
Since no further acknowledgement is requested, merging - thank you!

PR #1485, opened 2017-12-27
improve docstyle for Bio/Entrez, Bio/PDB, Bio/FSSP and Bio/Phylo

Gasta88:
This pull request addresses issue #1473

### DIFF
--- a/Bio/Entrez/Parser.py
+++ b/Bio/Entrez/Parser.py
@@ -451,7 +451,7 @@ class DataHandler(object):
                 self.lists.append(name)
 
     def elementDecl(self, name, model):
-        """Callback function is called for each element declaration in a DTD.
+        """Return element according to its declaration in a DTD.
 
         This is used for each element declaration in a DTD like::
 
--- a/Bio/Entrez/Parser.py
+++ b/Bio/Entrez/Parser.py
@@ -451,7 +451,7 @@ class DataHandler(object):
                 self.lists.append(name)
 
     def elementDecl(self, name, model):
-        """Return element according to its declaration in a DTD.
+        """Call a call-back function for each element declaration in a DTD.
 
         This is used for each element declaration in a DTD like::
 
--- a/Bio/Phylo/_io.py
+++ b/Bio/Phylo/_io.py
@@ -34,7 +34,7 @@ except ImportError:
 
 
 def parse(file, format, **kwargs):
-    """Parse a file and return each of the trees it contains.
+    """Parse iteratively a file and return each of the trees it contains.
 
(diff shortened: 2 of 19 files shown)


Bio/Entrez/Parser.py
```diff
@@ -451,7 +451,7 @@ def parse_xsd(self, root):
                 self.lists.append(name)
 
     def elementDecl(self, name, model):
-        """Callback function is called for each element declaration in a DTD.
+        """Return element according to its declaration in a DTD.
```
peterjc: I don't think "Return ..." is correct for a call-back function. Maybe: "This is a call-back function for element declaration in a DTD"?

Bio/Phylo/_io.py
```diff
@@ -34,7 +34,7 @@
 
 
 def parse(file, format, **kwargs):
-    """Iteratively parse a file and return each of the trees it contains.
+    """Parse a file and return each of the trees it contains.
```
peterjc: It would be good to have something in the one line indicating this is an iterator, maybe "Parse a file iteratively, and yield each of the trees it contains."?

peterjc:
Most of this looked fine - I had two minor queries, but would prefer the relevant module owners  review these changes.

Bio/Entrez/Parser.py
Gasta88: Your suggestion is not imperative, therefore `pydocstyle` does not like it. 
I would opt for "Call a call-back function for each element declaration in a DTD", but I'll wait for the module owners opinion.

peterjc: I'm not 100% sure if this method itself is the callback function or not, and that is important for the phrasing (and meaning).

Bio/Phylo/_io.py
```diff
@@ -34,7 +34,7 @@
 
 
 def parse(file, format, **kwargs):
-    """Parse a file and return each of the trees it contains.
+    """Parse iteratively a file and return each of the trees it contains.
```
peterjc: How about:

```python
"""Parse a file iteratively, and yield each of the trees it contains.
```

In addition the the rephrasing, I suggest using yield rather than return given the Python specific meaning of these verbs.

Gasta88:
Made the amend on `Bio/Phylo/_io.py` as you suggested but I'm still waiting on a module owner review regarding the previous request.

peterjc:
I'm inclined to merge this now, and come back to ``elementDecl`` later if required.

peterjc:
TravisCI is failing due to #1496

peterjc:
Thank you!

PR #1388, opened 2017-09-12
Build API documentation with Sphinx via TravisCI/Tox [WIP]

peterjc:
This pull request would in part address issue #906, the goal of replacing epydoc for our API documentation.

Note in addition to Sphinx specific changes, there are some code changes in ``Bio.Restriction`` and ``Bio.Graphics.GenomeDiagram`` in order to avoid an exception and include the classes in the public documentation.

Still needs work, e.g. dealing with special methods like ``__init__`` which we probably want to be included by default.

PR #1486, opened 2017-12-28
Hack in Bio.Restriction for Sphinx api-doc support

peterjc:
This pull request addresses an issue found in #1388 under Sphinx and api-doc, but which otherwise I have not seen triggered other than in an artificial example like this, which now "works":

```
$ python3 -c "from Bio.Restriction.Restriction import RestrictionType; print(len(RestrictionType('RestrictionType', tuple(), {})))"
0
```

This hack was needed when using Sphinx and api-doc to build our API documentation.

@MarkusPiotrowski - can you see any downsides to this?

### DIFF
--- a/Bio/Restriction/Restriction.py
+++ b/Bio/Restriction/Restriction.py
@@ -346,9 +346,10 @@ class RestrictionType(type):
         try:
             return cls.size
         except AttributeError:
-            # Happens if the instance was not initialised as expected,
+            # Happens if the instance was not initialised as expected.
             # e.g. if instance created by a documentation framework
-            # like Sphinx trying to inspect the class automatically.
+            # like Sphinx trying to inspect the class automatically,
+            # Also seen within IPython.
             return 0
 
     def __hash__(cls):
--- a/Bio/Restriction/Restriction.py
+++ b/Bio/Restriction/Restriction.py
@@ -260,6 +260,8 @@ class RestrictionType(type):
             cls.compsite = re.compile(cls.compsite)
         except AttributeError:
             # Can happen if initialised wrongly.
+            # (This was seen when Sphinx api-doc imports the classes, and
+            # tried to automatically general documentation for them)
             pass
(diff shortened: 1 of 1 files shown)


Bio/Restriction/Restriction.py
```diff
@@ -258,6 +258,9 @@ def __init__(cls, name='', bases=(), dct=None):
         # super(RestrictionType, cls).__init__(cls, name, bases, dct)
         try:
             cls.compsite = re.compile(cls.compsite)
+        except AttributeError:
+            # Can happen if initialised wrongly.
+            pass
```
MarkusPiotrowski: Should we add in the comment the same explanation as below ('... Sphinx trying to inspect...')?

MarkusPiotrowski:
No, I can't see any downsides. I would just suggest to label this catches with something like "Hack for Sphinx" so that we remember why this was added.

MarkusPiotrowski:
@peterjc Does Sphinx explicitely asks for ```len()``` or would the implementation of ```__bool__``` (as in #1492) in combination with the change in ```compsite``` also work?

peterjc:
We should update the comments given it fixes #1491 (representation under iPython shell), and merge this.

MarkusPiotrowski:
Actually I wrote the last comment after it was clear that it also fixes #1491; if addition of ```__bool___``` (in combination with the change in ```compsite```) would also solve the Sphinx problem, I think it would be a 'cleaner' solution than 'hacking' ```len()```.
If not I agree that this should be merged soon.

peterjc:
OK, I'll have to test #1492 for the Sphinx issue then...

peterjc:
Based on https://travis-ci.org/peterjc/biopython/jobs/327744275 adding the ``__bool__`` definition from #1492 does not solve the Sphinx problem.

MarkusPiotrowski:
Then let us merge this PR

PR #1488, opened 2018-01-03
NEWS.rst: be clearer about PEP 519 support added in #1476

mruffalo:
It isn't quite accurate to say that "BioPython I/O methods support `pathlib.Path` objects" -- it's more like "BioPython no longer second-guesses whether an object can be passed to `__builtins__.open`.

Clarify that using `pathlib.Path` objects in these I/O functions requires Python 3.6 or newer, to avoid any bug reports from users who might expect this to work under 3.4 or 3.5.

<hr>

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style checks pass with these changes.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -54,10 +54,10 @@ In this release more of our code is now explicitly available under either our
 original "Biopython License Agreement", or the very similar but more commonly
 used "3-Clause BSD License".  See the ``LICENSE.rst`` file for more details.
 
-IO functions such as SeqIO.read now support objects conforming to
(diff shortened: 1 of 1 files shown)


mruffalo:
Odd -- looks like I didn't introduce that style check failure, since `master` failed in the same way: https://travis-ci.org/biopython/biopython/builds/323807774

peterjc:
Thank you!

(TravisCI failed on the copyright year check in the license file, should be fixed as of d5718bb7e3ee505b859b39c03f8ffad8a9a0be2f)

Issue #1224, opened 2017-05-10
NCBI Genetic Code Tables 27 to 31 + ambiguous codons


peterjc:
The NCBI has added more genetic code tables, now up to 31:
https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi#SG27

We have NCBI tables up to 26 already in https://github.com/biopython/biopython/blob/master/Bio/Data/CodonTable.py

There is a minor issue, first the NCBI have not yet updated the computer readable file ftp://ftp.ncbi.nih.gov/entrez/misc/data/gc.prt (which I've emailed them about), but we could add the tables by hand (as we had to for Table 26).

However there is also a larger issue: What to do with Tables 27, 28 and 31 where some codons can be an amino acid or a stop codon? I think we can store that in the existing codon table object, but it has troubling implications for down stream analysis like ORF finding or our translate functionality.

This gives a new meaning to the term *ambiguous codon*, because here the bases themselves are unambiguous.

Papers: 

* Swart *et al.* (2016) Genetic Codes with No Dedicated Stop Codon: Context-Dependent Translation Termination. https://doi.org/10.1016/j.cell.2016.06.020
* Záhonová *et al.* (2016) An Unprecedented Non-canonical Nuclear Genetic Code with All Three Termination Codons Reassigned as Sense Codons. https://doi.org/10.1016/j.cub.2016.06.064
* Heaphy *et al.* (2016) Novel Ciliate Genetic Code Variants Including the Reassignment of All Three Stop Codons to Sense Codons in Condylostoma magnum. https://doi.org/10.1093/molbev/msw166

peterjc:
Based on archive.org, the new tables 27 to 31 were added between 28 Oct and 26 Nov 2016.

peterjc:
I emailed the NCBI about this again, it looks like the source file has been updated based on https://www.ncbi.nlm.nih.gov/IEB/ToolBox/C_DOC/lxr/source/data/gc.prt - but the public copy on the FTP site has not.

peterjc:
Good news: The NCBI's GenBank taxonomy team got in touch by email, and the FTP site has been updated with version 4.2 of the genetic codes:

ftp://ftp.ncbi.nih.gov/entrez/misc/data/gc.prt

That still leaves the tricky question of how we're going to represent these new codons which can be stop of an amino acid in Biopython...

MarkusPiotrowski:
[...]

PR #1501, opened 2018-01-09
Update of CodonTable

MarkusPiotrowski:
- Small changes in ```update_ncbi_codon_table.py``` to give a PEP8 compliant output
 - Added codon tables 27 - 31 in ```Bio.Data.CodonTable.py``` (and some coding style issues)
 - Removed a bunch of ```assert``` statements in ```Bio.Data.Codon.Table.py```. They were moved to ```test_CodonTable.py```
 - Complete re-write of ```test_CodonTable.py``` to use the ```unittest``` framework. Added tests for each codon table.
 - Small change in ```Seq.py``` to raise a warning or exception when codon tables with 'dual coding' stop codons are used (see #1224). The use of such tables during translation will generally raise a ```BiopythonWarning```. If ```to_stop=True``` a ```ValueError``` is raised.
 - Small update in ```test_Seq.py``` to include these cases.
 - Small update in ```NEWS.rst``` for the addition of the new codon tables.

and have added myself to those files as part of this pull request.

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2387,19 +2387,20 @@ def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
     n = len(sequence)
 
     # Check for tables with 'ambiguous' (dual-coding) stop codons:
-    for codon in (c for c in stop_codons if c in forward_table):
+    dual_coding = [c for c in stop_codons if c in forward_table]
+    if dual_coding:
+        c = dual_coding[0]
         if to_stop:
-            raise ValueError("This table contains at least one stop codon "
-                             "('{}') which codes for both 'STOP' and an "
-                             "amino acid ('{}'). You can not use 'to_stop="
-                             "True' with this table."
-                             .format(codon, forward_table[codon]))
-        warnings.warn("This table contains at least one stop codon ('{}') "
-                      "which codes for both 'STOP' and an amino acid ('{}'). "
+            raise ValueError("You cannot use 'to_stop=True' with this table "
+                             "as it contains {} codon(s) which can be both "
+                             " STOP and an  amino acid (e.g. '{}' -> '{}' or "
+                             "STOP)."
+                             .format(len(dual_coding), c, forward_table[c]))
+        warnings.warn("This table contains {} codon(s) which code(s) for both "
+                      "STOP and an amino acid (e.g. '{}' -> '{}' or STOP). "
                       "Such codons will be translated as amino acid."
-                      .format(codon, forward_table[codon]),
+                      .format(len(dual_coding), c, forward_table[c]),
                       BiopythonWarning)
-        break
 
     if cds:
         if str(sequence[:3]).upper() not in table.start_codons:
@@ -2543,7 +2544,7 @@ def translate(sequence, table="Standard", stop_symbol="*", to_stop=False,
     >>> translate(coding_dna3, table=27, to_stop=True)
     Traceback (most recent call last):
        ...
-    ValueError: ... You can not use 'to_stop=True' with this table.
+    ValueError: You cannot use 'to_stop=True' with this table ...
     """
     if isinstance(sequence, Seq):
         return sequence.translate(table, stop_symbol, to_stop, cds)
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2387,26 +2387,19 @@ def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
     n = len(sequence)
 
     # Check for tables with 'ambiguous' (dual-coding) stop codons:
-    for codon in stop_codons:
-            try:
-                forward_table[codon]
-                if to_stop:
-                    raise ValueError("This table contains at least one stop "
-                                     "codon ('{}') which codes for both "
-                                     "'STOP' and an amino acid ('{}'). You "
-                                     "can not use 'to_stop=True' with this "
-                                     "table."
-                                     .format(codon, forward_table[codon]))
-                else:
-                    warnings.warn("This table contains at least one stop "
-                                  "codon ('{}') which codes for both 'STOP' "
-                                  "and an amino acid ('{}'). Such codons will "
-                                  "be translated as amino acid."
-                                  .format(codon, forward_table[codon]),
-                                  BiopythonWarning)
-                break
-            except KeyError:
-                pass  # KeyError is the expected behaviour for most tables
+    for codon in (c for c in stop_codons if c in forward_table):
+        if to_stop:
+            raise ValueError("This table contains at least one stop codon "
+                             "('{}') which codes for both 'STOP' and an "
+                             "amino acid ('{}'). You can not use 'to_stop="
+                             "True' with this table."
+                             .format(codon, forward_table[codon]))
+        warnings.warn("This table contains at least one stop codon ('{}') "
+                      "which codes for both 'STOP' and an amino acid ('{}'). "
+                      "Such codons will be translated as amino acid."
+                      .format(codon, forward_table[codon]),
+                      BiopythonWarning)
+        break
 
     if cds:
         if str(sequence[:3]).upper() not in table.start_codons:
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2550,9 +2550,7 @@ def translate(sequence, table="Standard", stop_symbol="*", to_stop=False,
     >>> translate(coding_dna3, table=27, to_stop=True)
     Traceback (most recent call last):
        ...
-    ValueError: This table contains at least one stop codon ('TGA') which codes
-    for both 'STOP' and an amino acid ('W'). You can not use 'to_stop=True'
-    with this table.
+    ValueError: ... You can not use 'to_stop=True' with this table.
     """
     if isinstance(sequence, Seq):
         return sequence.translate(table, stop_symbol, to_stop, cds)
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -2385,6 +2385,29 @@ def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
         valid_letters = set(IUPAC.ambiguous_dna.letters.upper() +
                             IUPAC.ambiguous_rna.letters.upper())
     n = len(sequence)
+
+    # Check for tables with 'ambiguous' (dual-coding) stop codons:
+    for codon in stop_codons:
+            try:
+                forward_table[codon]
+                if to_stop:
+                    raise ValueError("This table contains at least one stop "
+                                     "codon ('{}') which codes for both "
+                                     "'STOP' and an amino acid ('{}'). You "
+                                     "can not use 'to_stop=True' with this "
+                                     "table."
+                                     .format(codon, forward_table[codon]))
+                else:
+                    warnings.warn("This table contains at least one stop "
+                                  "codon ('{}') which codes for both 'STOP' "
+                                  "and an amino acid ('{}'). Such codons will "
+                                  "be translated as amino acid."
+                                  .format(codon, forward_table[codon]),
+                                  BiopythonWarning)
+                break
+            except KeyError:
+                pass  # KeyError is the expected behaviour for most tables
+
     if cds:
         if str(sequence[:3]).upper() not in table.start_codons:
             raise CodonTable.TranslationError(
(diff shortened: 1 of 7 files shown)


MarkusPiotrowski:
@peterjc Can we turn off ```W503```? PEP8 allows line breaks after **and** before binary operators and in fact it recommends line breaks *before*: 
 

> Following the tradition from mathematics usually results in more readable code:

```
# Yes: easy to match operators with operands
ncome = (gross_wages
         + taxable_interest
         + (dividends - qualified_dividends)
         - ira_deduction
         - student_loan_interest)
```

>In Python code, it is permissible to break before or after a binary operator, as long as the convention is consistent locally. For new code Knuth's style is suggested.

Knuth's style here means: line break *before* the binary operator
See also here: https://bugs.python.org/issue26763

Scripts/update_ncbi_codon_table.py
```diff
@@ -16,14 +19,20 @@
 import re
 
 INDENT = len("register_ncbi_table(")
+INDENT2 = INDENT + len("table={")
 
 
-def line_wrap(text, indent=0, max_len=78):
+def line_wrap(text, indent=0, max_len=78, string=False):
+    """Return a wrapped line if lenght is larger max_len."""
```
peterjc: Typo (lenght)

Scripts/update_ncbi_codon_table.py
```diff
@@ -32,13 +41,18 @@ def line_wrap(text, indent=0, max_len=78):
...
-print("# Start of auto-generated output from Scripts/update_ncbi_codon_table.py #")
-print("##########################################################################")
-print("")
-print("")
+print("""
+##########################################################################
+# Start of auto-generated output from Scripts/update_ncbi_codon_table.py #
+##########################################################################
+
+"""[1:])
```
peterjc: Personally I found the five separate print statements clearer, but don't have a strong view on this change.

Bio/Seq.py
```diff
@@ -2504,6 +2527,32 @@ def translate(sequence, table="Standard", stop_symbol="*", to_stop=False,
...
+    'cds=True', where the last codon will be translated as STOP.
+
+    >>> coding_dna3 = "ATGGCACGGAAGTGA"
+    >>> translate(coding_dna3)
+    'MARK*'
+
+    >>> translate(coding_dna3, table=27)  # Table 27: TGA -> STOP or W
+    'MARKW'
+
+    It will howver raise a BiopythonWarning (not shown)
```
peterjc: Typo (howver)

NEWS.rst
```diff
@@ -50,6 +50,8 @@ wrapper and include a new wrapper for fuzzpro.
 The restriction enzyme list in Bio.Restriction has been updated to the
 November 2017 release of REBASE.
 
+New codon tables 27-31 from NCBI were added to Bio.Data.CodonTable.
```
peterjc: Also mention these include some codons with unambiguous bases which can dependent on context encode an amino acid or a stop codon?

Scripts/update_ncbi_codon_table.py
peterjc: Also might be worth expanding the docstring to explain the use of the new string argument - I can see you use this to split string arguments over multiple lines, but it would only work if the strings are single quote delimited.

(The implementation is fine given it is specific to this helper script, but making this more explicit seems safer for future users of this code.)

peterjc:
You actually add all the missing tables 27-31, which is good. I've noted some minor things like typos, otherwise this looks fine.

Good work on reworking the test suite. Personally I would have split that into a second commit (making it easier to test the functional changes vs the test changes), but I take it you have confirmed the old tests still pass?

MarkusPiotrowski:
@peterjc Thank you, I will do the required changes. Regarding the tests, I think I included all the test from the old file in the new version. To be sure I will run the old testfile on this branch (I did in the beginning, but not at the final stage).

I also mailed to the e-mail list, asking for opinions how to deal with the problematic tables.

Do you have an opinion regarding ```W503```? As mentioned above I think we should turn it off, it violates the true meaning of ```PEP8``` regarding line breaks for dual operators.

peterjc:
I am sympathetic to line breaks before binary operators, and would be happy to adopt W504 in place of W503 to enforce this - but that is not yet merged into pydocstyle/flake8:

https://github.com/PyCQA/pycodestyle/pull/502

Given W503 is now not on the default pydocstyle ignore list, and not compatible with the updated PEP8, I am happy for us to ignore it. Please update ``Bio/.flake8`` as part of this pull request (as a separate commit).

MarkusPiotrowski:
@peterjc  +1 for agreeing to ignore W503. However, we need to do this separately, otherwise this PR will always fail the style checks on TravisCI ...

peterjc:
Sure, we can ignore W503 as its own PR (and merge that quickly), that's probably a better plan.

peterjc:
Cross reference pull request #1503

MarkusPiotrowski:
That's strange, obviously Python 2 has problems with line continuation in doctests? 
```
Failed example:
    translate(coding_dna3, table=27, to_stop=True)
Expected:
       ...
    ValueError: This table contains at least one stop codon ('TGA') which codes
    for both 'STOP' and an amino acid ('W'). You can not use 'to_stop=True'
    with this table.
Got:
      File "/opt/python/pypy-5.7.1/lib-python/2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest Bio.Seq.translate[14]>", line 1, in <module>
        translate(coding_dna3, table=27, to_stop=True)
      File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/Seq.py", line 2574, in translate
        gap=gap)
      File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/Seq.py", line 2399, in _translate_str
        .format(codon, forward_table[codon]))
    ValueError: This table contains at least one stop codon ('TGA') which codes for both 'STOP' and an amino acid ('W'). You can not use 'to_stop=True' with this table.
```
I ran ```Seq.py``` both from within IDLE and from the commandline under Python 2.7.14 without problems.

I'm going to use ELLIPSIS

peterjc:
If ellipsis doesn't work, just put the very long line directly in the doctest - crude but effective?

Scripts/update_ncbi_codon_table.py
```diff
@@ -16,29 +19,46 @@
...
-print("# Start of auto-generated output from Scripts/update_ncbi_codon_table.py #")
-print("##########################################################################")
-print("")
-print("")
+"""[1:])
 
+version = ''
 for line in open("gc.prt").readlines():
+    if not version and line.startswith("--  Version"):
+        version = line.split('Version')[1].strip()
```
bow: How likely is it that `Version` occurs more than once? I've found it is often better to be explicit with the number of expected `split()`s. In this case, if we only expect it once, I think simply doing `.split('Version', 1)` makes the code more robust.

Bio/Seq.py
```diff
@@ -2385,6 +2385,29 @@ def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
...
+                else:
+                    warnings.warn("This table contains at least one stop "
+                                  "codon ('{}') which codes for both 'STOP' "
+                                  "and an amino acid ('{}'). Such codons will "
+                                  "be translated as amino acid."
+                                  .format(codon, forward_table[codon]),
+                                  BiopythonWarning)
+                break
+            except KeyError:
+                pass  # KeyError is the expected behaviour for most tables
```
bow: This may be more of a style thing, but I often feel that `except <Exception>: pass` can be better rewritten to convey intent.

How about something like this?

```
for codon in stop_codons:
    if codon not in forward_table:
        continue
    if to_stop:
        <raise exception>
    <warn>
```

We can even remove the `continue` clause:

```
for codon in {c for c in stop_codons if c in forward_table}:
    if to_stop:
        <raise exception>
    <warn>
```

This would raise the exception on the first ambiguous codons if `to_stop` is `True`, but would raise warnings for all the ambiguous codons otherwise.

MarkusPiotrowski: Unfortunately, this doesn't work. ```codon in forward_table``` gives a ```TypeError``` for ambiguous tables, while ```forward_table[codon]``` works nicely (either returning an amino acid or raising a ```KeyError```, as expected). 
I don't know if this is a bug or intended, the ```__getitem__``` method in ```AmbiguousForwardTable``` is quite complex...
So if you don't insist, I would let it as it is.
Example:
```
>>> from Bio.Data import CodonTable
>>> table = CodonTable.ambiguous_dna_by_id[1]
>>> table.forward_table['ATG']
'M'
>>> 'ATG' in table.forward_table

  File "<pyshell#61>", line 1, in <module>
    'ATG' in table.forward_table
  File "C:\Program Files\Anaconda32\lib\site-packages\Bio\Data\CodonTable.py", line 375, in __getitem__
    self.ambiguous_nucleotide)
  File "C:\Program Files\Anaconda32\lib\site-packages\Bio\Data\CodonTable.py", line 209, in list_possible_proteins
    c1, c2, c3 = codon
TypeError: 'int' object is not iterable
```

Scripts/update_ncbi_codon_table.py
MarkusPiotrowski: OK

Bio/Seq.py
bow: Aha, I had thought earlier `forward_table` should behave like a plain `dict`. That is apparently not the case.

The error we're seeing is because `forward_table` (or rather, its class) doesn't define a `__contains__` method, [which makes the `in` operation resort to using `__getitem__` but using integers instead of `ATG`](https://docs.python.org/3/reference/expressions.html#membership-test-details).

Without an expected `__contains__` behavior, I can understand that the `exception <E>: pass` style is the one that does what we want ~ so let's keep it that way for now.

peterjc: Given it is dictionary like, do you think is the expected behaviour of the missing ``__contains__``method clearly defined? If so we could add it?

MarkusPiotrowski: I'm thinking about adding it. I'll see how it works

MarkusPiotrowski:
That was a bit to quick, lookups which return ```TranslationError``` should probably also give a positive membership testing. These are cases where ambiguous codons can either code for amino acid and stop or for several amino acids but there is no letter for this combination of amino acids. With my current implementation ```'xxx' in forward_table``` will return ```False``` in these cases (which allows you to do ```if 'xxx' in forward_table: return forward_table['xxx']``` without raising any error). However, since these codons do code (but can't be translated for some reason) ```['xxx'] in forward_table``` should probably return ```True``` (even if ```forward_table['xxx']``` would then raise a ```TranslationError```). Of course this should then be mentioned in the docstring of the ```___contains___```. 
@peterjc  @bow Your view?

peterjc:
Some of the past design choices were likely practical for implementation (e.g. the `X` fall back for traditionally ambiguous codons where it could be several amino acids)

bow:
I feel the latest change is ok as is. If `'xxx' in forward_table` returns `True`, then it's more intuitive to have `forward_table['xxx']` return an actual value rather than raise an exception, despite the more nuanced interpretation that `'xxx'` codes for an amino acid but can not be translated somehow.

If we ever want to tackle these corner cases, I don't think making the codon table behave as if it were a `dict` is the way to go. Instead we just treat it as a simple container object without `__getitem__` or `get`, that maybe has its own `get_amino_acid` is better. We could also still make it like a `dict`, but instead of ever raising a `TranslationError`, instead we either return values that mean the AA translation is known or unknown (so the `in` operator evaluating to `True` would still mean forward_table[_] always return a value).

But these feel like overengineering ~ I'm happy with the latest changes as is.

MarkusPiotrowski:
@peterjc Sorry, I don't understand your last comment.

peterjc:
@MarkusPiotrowski I meant it took me a while to understand the codon tables, and when they did and didn't raise a KeyError. For reference, this was originally used in ``Bio.Translate``, which we deprecated and removed in Biopython 1.57, last release here:

https://github.com/biopython/biopython/blob/biopython-156/Bio/Translate.py

Bio/Seq.py
```diff
@@ -2387,26 +2387,19 @@ def _translate_str(sequence, table, stop_symbol="*", to_stop=False,
...
+                             "('{}') which codes for both 'STOP' and an "
+                             "amino acid ('{}'). You can not use 'to_stop="
+                             "True' with this table."
+                             .format(codon, forward_table[codon]))
+        warnings.warn("This table contains at least one stop codon ('{}') "
+                      "which codes for both 'STOP' and an amino acid ('{}'). "
+                      "Such codons will be translated as amino acid."
+                      .format(codon, forward_table[codon]),
+                      BiopythonWarning)
+        break
```
peterjc: I had to study this for loop with the break some time to understand what it does. How about something like:

```python
dual_coding = [c for c in stop_codons if c in forward_table]
if dual_coding:
    if to_stop:
        raise ValueError("You cannot use 'tostop=True' with this table as it "
                         "contains %i codons which can be both stop and an "
                         "amino acid." % len(dual_coding))
    warnings.warn("This table contains %i codons which code for both "
                  "stop and an amino acid. Such codons will be translated ",
                  "as the amino acid." % len(dual_coding), BiopythonWarning)
```

(You could equally use ``c = dual_coding[0]`` and keep the existing proposed messages, which are helpfully explicit with an example codon)

peterjc:
Sorry, I forgot about this. Thanks for that final polish.

PR #1504, opened 2018-01-11
docstyle improved for Bio.Pathway, Bio.NeuralNetwork and Bio.Blast

Gasta88:
This pull request addresses issue #1473

### DIFF
--- a/Bio/Blast/Applications.py
+++ b/Bio/Blast/Applications.py
@@ -99,7 +99,10 @@ class _NcbibaseblastCommandline(AbstractCommandline):
         AbstractCommandline.__init__(self, cmd, **kwargs)
 
     def _validate_incompatibilities(self, incompatibles):
-        """Use _validate method by the BLAST+ (PRIVATE)."""
+        """Validate parameters for incompatibilities (PRIVATE).
+
+        Used by the _validate method.
+        """
         for a in incompatibles:
             if self._get_parameter(a):
                 for b in incompatibles[a]:
--- a/Bio/Blast/Applications.py
+++ b/Bio/Blast/Applications.py
@@ -99,7 +99,7 @@ class _NcbibaseblastCommandline(AbstractCommandline):
         AbstractCommandline.__init__(self, cmd, **kwargs)
 
     def _validate_incompatibilities(self, incompatibles):
-        """Used by the BLAST+ _validate method (PRIVATE)."""
+        """Use _validate method by the BLAST+ (PRIVATE)."""
         for a in incompatibles:
             if self._get_parameter(a):
                 for b in incompatibles[a]:
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -259,7 +259,7 @@ class BlastParser(_XMLparser):
         self._header.reference = self._value
 
     def _end_BlastOutput_db(self):
-        """Yield the database(s) searched (PRIVATE).
+        """Record the database(s) searched (PRIVATE).
 
         Save this to put on each blast record object
         """
@@ -355,7 +355,7 @@ class BlastParser(_XMLparser):
                                          int(self._value))
 
     def _end_Parameters_filter(self):
-        """Record a filtering options (-F) (PRIVATE)."""
+        """Record filtering options (-F) (PRIVATE)."""
         self._parameters.filter = self._value
 
     # def _end_Parameters_pattern(self):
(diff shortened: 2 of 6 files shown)


Bio/Blast/Applications.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, cmd=None, **kwargs):
         AbstractCommandline.__init__(self, cmd, **kwargs)
 
     def _validate_incompatibilities(self, incompatibles):
-        """Used by the BLAST+ _validate method (PRIVATE)."""
+        """Use _validate method by the BLAST+ (PRIVATE)."""
```
peterjc: The meaning has changed, how about:

```python
"""Validate parameters for incompatibilities (PRIVATE).

Used by the _validate method.
"""
```

Bio/Blast/NCBIXML.py
```diff
@@ -252,53 +252,53 @@ def _end_BlastOutput_version(self):
                 self._header.date = parts[2]
 
     def _end_BlastOutput_reference(self):
-        """A reference to the article describing the algorithm (PRIVATE).
+        """Yield the article describing the algorithm (PRIVATE).
```
peterjc: Yield implies a generator/iterator. Return seems wrong here - the function as no return value (Python default is just ``None``). How about:

```python
"""Record any article reference describing the algorithm (PRIVATE)."""
```

Bio/Blast/NCBIXML.py
```diff
@@ -252,53 +252,53 @@ def _end_BlastOutput_version(self):
...
-        """The database(s) searched (PRIVATE).
+        """Yield the database(s) searched (PRIVATE).
 
         Save this to put on each blast record object
         """
         self._header.database = self._value
 
     def _end_BlastOutput_query_ID(self):
-        """The identifier of the query (PRIVATE).
+        """Yield the identifier of the query (PRIVATE).
```
peterjc: Similar comment to above, how about  ``Record the identifier of the query (PRIVATE).`` instead?

Bio/Blast/NCBIXML.py
```diff
@@ -252,53 +252,53 @@ def _end_BlastOutput_version(self):
...
 
         Important in old pre 2.2.14 BLAST, for recent versions
         <Iteration_query-ID> is enough
         """
         self._header.query_id = self._value
 
     def _end_BlastOutput_query_def(self):
-        """The definition line of the query (PRIVATE).
+        """Yield the definition line of the query (PRIVATE).
```
peterjc: Same problem as above

Bio/Blast/NCBIXML.py
```diff
@@ -252,53 +252,53 @@ def _end_BlastOutput_version(self):
...
 
         Important in old pre 2.2.14 BLAST, for recent versions
         <Iteration_query-def> is enough
         """
         self._header.query = self._value
 
     def _end_BlastOutput_query_len(self):
-        """The length of the query (PRIVATE).
+        """Yield the length of the query (PRIVATE).
```
peterjc: Same as above, and applies to all the other private ``_end...`` methods which are used by the parse to capture/record information from the XML into our data structure.

Bio/Blast/NCBIXML.py
```diff
@@ -252,53 +252,53 @@ def _end_BlastOutput_version(self):
...
-        """A reference to the article describing the algorithm (PRIVATE).
+        """Record any article reference describing the algorithm (PRIVATE).
 
         Save this to put on each blast record object
         """
         self._header.reference = self._value
 
     def _end_BlastOutput_db(self):
-        """The database(s) searched (PRIVATE).
+        """Yield the database(s) searched (PRIVATE).
```
peterjc: Yield --> record

Bio/Blast/NCBIXML.py
```diff
@@ -355,7 +355,7 @@ def _end_Parameters_gap_extend(self):
                                          int(self._value))
 
     def _end_Parameters_filter(self):
-        """Filtering options (-F) (PRIVATE)."""
+        """Record a filtering options (-F) (PRIVATE)."""
```
peterjc: Options is plural, so "Record a filtering options" --> "Record filtering options"?

Gasta88: Well, this was pretty dumb xD

peterjc: Not really. You've been editing so many of these that little mistake can easily happen.

peterjc:
Thank you!

PR #1150, opened 2017-04-12
Updated the NCBIWWW module

yangwu91:
- "megablast" was added into "program" parameter, and "megablast"
parameter has been discarded.
- "num_threads" parameter was added, but
it is ONLY supported on the custom cloud.
- "matrix_name" parameter was
changed to "matrix".
- Official help URLs were updated.
- Some related
changes were made in the "test_NCBI_qblast.py" script.

PR #1025, opened 2016-12-13
Add "megablast" program into "NCBIWWW" module

yangwu91:
Revised NCBIWWW module:
- Add "megablast":
Megablast is intended to compare closely related sequences. It is much
faster than traditional blastn and recommended to use.
- Revise "filter" parameter:
"Filer" parameter now has different options.

PR #1513, opened 2018-01-22
Updated NCBIWWW.qblast and test_NCBI_qblast.py

yangwu91:
* added "organism" option to specify an organism for nr/nt databases, or
an SRA accession for SRA, EST and other databses.
* "num_threads" parameter was added, but it is ONLY supported on the
custom cloud.
* "matrix_name" parameter was changed to "matrix".
* Official help URLs were updated.
* Some related changes were made in the "test_NCBI_qblast.py" script.


yangwu91:
See also #1025 #1150 
I am sorry for the delay. Somehow I couldn't update my branch with `git fetch origin` and `git rebase origin/master`, so I have to re-folk.

Bio/Blast/NCBIWWW.py
```diff
@@ -67,6 +68,11 @@ def qblast(program, database, sequence, url_base=NCBI_BLAST_URL,
      - hitlist_size   Number of hits to return. Default 50
      - megablast      TRUE/FALSE whether to use MEga BLAST algorithm (blastn only)
      - service        plain, psi, phi, rpsblast, megablast (lower case)
+     - organism       Specify an organism while using nr/nt database, or an SRA accession
+                      for SRA, EST and other databases. The organism also could be specified
+                      by "entrez_query". For example: "entrez_query='rat [ORGN]'" is identical
```
peterjc: Shouldn't that be:  ``"entrez_query='rat[ORGN]'"`` (no space before the opening square bracket)?

Bio/Blast/NCBIWWW.py
```diff
@@ -76,12 +82,16 @@ def qblast(program, database, sequence, url_base=NCBI_BLAST_URL,
     import time
 
     assert program in ['blastn', 'blastp', 'blastx', 'tblastn', 'tblastx']
+    if url_base == "https://blast.ncbi.nlm.nih.gov/Blast.cgi":
+        assert num_threads is None
```
peterjc: I think you could give an explicit error here, e.g.

```python
if url_base == "https://blast.ncbi.nlm.nih.gov/Blast.cgi" and num_threads is not None:
    raise ValueError("The num_threads option cannot be used on the NCBI servers")
```

Tests/test_NCBI_qblast.py
```diff
@@ -48,7 +48,7 @@ def test_blastp_nr_actin(self):
         # GI:160837788 aka NP_075631.2
         # the actin related protein 2/3 complex, subunit 1B [Mus musculus]
         self.run_qblast("blastp", "nr", "NP_075631.2", 0.001,
-                        "rat [ORGN]", {'megablast': 'FALSE'},
```
peterjc: Does the old test still work? If so can we keep the old test using entrez_query, and add a similar new test using the organism field?

Bio/Blast/NCBIWWW.py
yangwu91: It looks llike both `rat[ORGN]` (without space) and `rat [ORGN]` (with space) could get expected results after I tested. `rat [ORGN]` (with space) was used by the original version.

Bio/Blast/NCBIWWW.py
yangwu91: Sure, it is way better.

Tests/test_NCBI_qblast.py
yangwu91: Yes, it works. I revised this part to prove two methods could get the same results, so they are equivalent. Of course I could add a similar test.

peterjc:
Why are there ``time.sleep(2)`` calls in the tests?

yangwu91:
I tried not to query too many times in a short period of time during testing.

peterjc:
Those delays should be handled within the main module, not in the test code.

Looking at this we're not being as strict as we should be when multiple BLAST calls are made, so I filled a new issue on that:

https://github.com/biopython/biopython/issues/1715

yangwu91:
I think the Travis CI build could not complete because of the latest version (v3.1.1) of `tox`. I tried `tox` v3.0.0, and it worked.

peterjc:
Yes, I noticed the Tox problem yesterday and just traced it to the their new release - see #1730

peterjc:
I think it would be easier for me to review and merge these as separate pull requests. i.e. Could you do the #1715 delay work on a new pull request instead (I have some suggestions for improving it), and remove it from this pull request which seemed almost ready? Thank you.

yangwu91:
Do you mean I should revert this branch to Commit f075468, which comes with the `organism` function but without the `delay` function?

peterjc:
f0754689ce487b81d6d30801f9a9503e41c41b34 is just some white space.

I meant remove the commits about #1715 and just focus on updating organism, threads, matrix, and the URLs.

PR #1514, opened 2018-01-24
Misc flake8 style fixes and stricter enforcement

peterjc:

and am already listed.


peterjc:
Cross reference #1515 which blocks enabling the flake8 E127 check at this point.

peterjc:
CodeCov thinks the coverage could be a little better, otherwise the automatic checks pass.

Bio/PDB/parse_pdb_header.py
```diff
@@ -103,8 +103,7 @@ def _nice_case(line):
         if c >= 'a' and c <= 'z' and nextCap:
             c = c.upper()
             nextCap = 0
-        elif c == ' ' or c == '.' or c == ',' or c == ';' or c == ':' or c == '\t' or\
-             c == '-' or c == '_':
+        elif c in ' .,;:\t-_':
```
peterjc: @etal @JoaoRodrigues @lennax could one of you OK this refactor in the PDB header parsing code?

(I wanted to fix the indentation warning by avoiding the slash-based line continuation, but going further this version seems more elegant)

Bio/Seq.py
```diff
@@ -2184,9 +2184,9 @@ def complement(self):
             d = ambiguous_rna_complement
         else:
             d = ambiguous_dna_complement
-        c = dict([(x.lower(), y.lower()) for x, y in d.items()])
-        d.update(c)
-        self.data = [d[c] for c in self.data]
+        mixed = d.copy()  # We're going to edit this to be mixed case!
+        mixed.update((x.lower(), y.lower()) for x, y in d.items())
+        self.data = [mixed[_] for _ in self.data]
```
peterjc: @MarkusPiotrowski you might be a good person to double check this change?

Tests/test_PAML_tools.py
```diff
@@ -130,4 +130,3 @@ def testYn00Binary(self):
 if __name__ == "__main__":
     runner = unittest.TextTestRunner(verbosity=2)
     unittest.main(testRunner=runner)
-    clean_up()
```
peterjc: @brandoninvergo is this change OK with you?

peterjc:
There are three non-trivial changes which I have commented on addressing specific people, otherwise everything else should be harmless.

Bio/PDB/parse_pdb_header.py
etal: This looks good to me. Much cleaner.

peterjc: Thanks - I was pretty sure I had the same list of symbols, but doubt we had 100% coverage here, so a second set of eyes on this is appreciated.

Bio/Seq.py
peterjc: @MarkusPiotrowski this was my commit comment for context:

Fix side effect on ambiguous_dna_complement etc

Spotted while reviewing flake8 F812, reuse of variables in a list comprension. Consider this example prior to the fix:

```python
>>> from Bio.Seq import MutableSeq
>>> from Bio.Data.IUPACData import ambiguous_dna_complement
>>> ambiguous_dna_complement
{'A': 'T', 'C': 'G', 'B': 'V', 'D': 'H', 'G': 'C', 'H': 'D', 'K': 'M', 'M': 'K', 'N': 'N', 'S': 'S', 'R': 'Y', 'T': 'A', 'W': 'W', 'V': 'B', 'Y': 'R', 'X': 'X'}
>>> MutableSeq('acgtACGT').complement()
>>> ambiguous_dna_complement
{'A': 'T', 'C': 'G', 'B': 'V', 'D': 'H', 'G': 'C', 'H': 'D', 'K': 'M', 'M': 'K', 'N': 'N', 'S': 'S', 'R': 'Y', 'T': 'A', 'W': 'W', 'V': 'B', 'Y': 'R', 'X': 'X', 'a': 't', 'c': 'g', 'b': 'v', 'd': 'h', 'g': 'c', 'h': 'd', 'k': 'm', 'm': 'k', 'n': 'n', 's': 's', 'r': 'y', 't': 'a', 'w': 'w', 'v': 'b', 'y': 'r', 'x': 'x'}
>>> quit()
```

Using the MutableSeq object's (in-place) complement method had the unexpected side effect of making the complement mapping dictionary mixed case.

MarkusPiotrowski: Good catch, looks good for me. I was shortly confused by the ```_``` variable, it reminds me of Perl's special variables.

Bio/Align/Applications/_Clustalw.py
```diff
@@ -129,47 +129,47 @@ def __init__(self, cmd="clustalw", **kwargs):
             # ***Fast Pairwise Alignments:***
```
MarkusPiotrowski: This whole bunch may start with (line 45):
```python
    self.parameters = [
        _Option(...
````

peterjc: That would be cleaner, yes. And it _probably_ wouldn't mean re-indenting everything (like I had to do on one of the other application wrappers in this pull request).

Bio/Seq.py
peterjc: Me too - Biopython historically has not used ``_`` as a dummy loop variable, but it seems to be a standard practice and is clear and concise once you are used to it.

peterjc:
Thanks all, merged.

Issue #1512, opened 2018-01-19
UnknownSeq.ungap() ignores the passed gap character


kblin:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609]
CPython
Linux-4.4.0-93-generic-x86_64-with-Ubuntu-16.04-xenial
1.71.dev0
0008a0d
```

I took the liberty of adding my git hash. :)

### Expected behaviour

`UnknownSeq.ungap()` passes on the gap character used to `Seq.ungap()` in https://github.com/biopython/biopython/blob/master/Bio/Seq.py#L1693

### Actual behaviour

Gap character is ignored.

### Steps to reproduce

Visual inspection of https://github.com/biopython/biopython/blob/master/Bio/Seq.py#L1693

kblin:
I'll do a patch for this next week, just can't be bothered today.

peterjc:
Opps. Good catch. The fix look easy (one word), but adding a unit test is slightly more work.

kblin:
One week later than intended, but there we go. :)

peterjc:
Thanks! I was wondering if I should do it if I got a lull in my day.

PR #1519, opened 2018-01-30
Seq: Pass on the gap character from UnknownSeq.ungap to Seq.ungap

kblin:
This avoids the "Gap character not given and not defined in alphabet" Value Error
that otherwise happens even if a gap character is given. Fixes issue #1512.

This pull request addresses issue #1512

### DIFF
--- a/Tests/test_Seq_objs.py
+++ b/Tests/test_Seq_objs.py
@@ -11,6 +11,7 @@ import unittest
 import sys
 
 from Bio import BiopythonWarning
+from Bio import SeqIO
 from Bio.Alphabet import generic_protein, generic_nucleotide
 from Bio.Alphabet import generic_dna, generic_rna
 from Bio.Alphabet.IUPAC import protein, extended_protein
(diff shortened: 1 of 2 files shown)


Tests/test_Seq_objs.py
```diff
@@ -10,7 +10,7 @@
 import unittest
 import sys
 
-from Bio import BiopythonWarning
+from Bio import BiopythonWarning, SeqIO
```
peterjc: I'd prefer this import on a separate line.

peterjc:
Rather than adding a new test file, would ``NT_019265.gb`` work here?

Tests/test_Seq_objs.py
kblin: I was following the multi-import per line style the rest of the file uses, but sure, changed. :)

kblin:
Yes, that file works as well, dropped the new file.

Tests/test_Seq_objs.py
peterjc: I think they were all multiple imports (classes, functions, etc) from the same python file - but Biopython's historical mixed case module names makes this harder to see.

peterjc:
Looks good - once the tests pass, this is ready to merge. You are already in the NEWS and CONTRIB files, so we're all set. Thank you!

Issue #1517, opened 2018-01-26
Enhancement? Support for Entrez on AWS Lambda and perhaps other serverless platforms


cparmet:
### Setup

Hi there. Huge huge fan of the Entrez module. 

This is more of an FYI, in case it's useful to know my experience. I defer to the team about whether this enhancement rises to the level of a revision to the package. 

I had a Flask web app that uses Entrez working on my local Windows 10 machine. When I deployed to Amazon Web Services as a Lambda function, it ran into trouble. 

```python
        Entrez.email = "<my email address>"
        handle = Entrez.esearch(db="pubmed", retmax=10, term=lookupID)
        record = Entrez.read(handle)
```

### Expected behaviour

A record to a PubMed entry, as it does on my local machine.

### Actual behaviour

On Lambda, esearch threw an exception and no record was read. 

The error occurs in parser.py, makedirs:

- 'OSError: [Errno 30] Read-only file system: '/home/sbx_user1060'

Full trace here:
[trace (showing error with Entrez module).docx](https://github.com/biopython/biopython/files/1668702/trace.showing.error.with.Entrez.module.docx).

### Steps to reproduce

Deploy a python project to AWS Lambda function that calls an Entrez method, such as esearch or esummary.

### Source of error
The trace shows the error happens when the Data Handler creates new directories for local_dtd and local_xsd. I learned that AWS Lambda only gives functions write access within a specific directory, 'tmp'. So Entrez is forbidden from creating the directories where it wants to.

```python
    if platform.system() == 'Windows':
        directory = os.path.join(os.getenv("APPDATA"), "biopython")
    else:  # Unix/Linux/Mac
        home = os.path.expanduser('~')
        directory = os.path.join(home, '.config', 'biopython')
        del home
    local_dtd_dir = os.path.join(directory, 'Bio', 'Entrez', 'DTDs')
    local_xsd_dir = os.path.join(directory, 'Bio', 'Entrez', 'XSDs')
    del directory
    del platform
    try:
        os.makedirs(local_dtd_dir)  # use exist_ok=True on Python >= 3.2
    except OSError as exception:
        # Check if local_dtd_dir already exists, and that it is a directory.
        # Trying os.makedirs first and then checking for os.path.isdir avoids
        # a race condition.
        if not os.path.isdir(local_dtd_dir):
            raise exception
    try:
        os.makedirs(local_xsd_dir)  # use exist_ok=True on Python >= 3.2
```

### Solution that worked for me
[...]

PR #1522, opened 2018-01-31
Expose Entrez cache to allow user to set custom cache directory (AWS Lambda deployment, etc)

cparmet:
This pull request addresses issue #1517 


Bio/Entrez/__init__.py
```diff
@@ -563,6 +565,9 @@ def _construct_params(params):
 In case of excessive usage of the E-utilities, NCBI will attempt to contact
 a user at the email address provided before blocking access to the
 E-utilities.""", UserWarning)
+    # Tell Entrez if we want to set a custom cache location.
+    if "cache" not in params:
+        params["cache"] = cache
```
peterjc: I don't think these lines are needed. IIRC, this is building a record of all the parameters to be encoding into the URL or post sent to the NCBI. They don't need to know about where we might be caching DTD files on the user's machine (and revealing this could leak information like their computer account username).

Bio/Entrez/__init__.py
```diff
@@ -36,6 +36,7 @@
 
     - email        Set the Entrez email parameter (default is not set).
     - tool         Set the Entrez tool parameter (default is ``biopython``).
+    - cache        Set the Entrez cache parameter (default is not set).
```
peterjc: How about ``Set the directory to cache NCBI DTD files for the Entrez XML parser (if not set the default is under your home directory).``

We could be even more explicit about the default path?

Bio/Entrez/__init__.py
cparmet: @peterjc Ah I see. Will delete. I mistakenly thought they were for internal parameters, and it makes sense that they're for NCBI given tool & email.

Bio/Entrez/__init__.py
cparmet: @peterjc Much better when it's explicit. I've replaced with your new one comment.

Bio/Entrez/Parser.py
```diff
@@ -169,12 +169,19 @@ def __str__(self):
...
-        directory = os.path.join(os.getenv("APPDATA"), "biopython")
-    else:  # Unix/Linux/Mac
-        home = os.path.expanduser('~')
-        directory = os.path.join(home, '.config', 'biopython')
-        del home
+    from Bio import Entrez
+
+    # Check if user has set a custom cache location
+    if Entrez.cache is not None:
+        directory = Entrez.cache
```
peterjc: Should we add a check here? Maybe:

```python
if not os.path.isdir(directory):
    raise ValueError("Please set Entrez.cache to a directory name (or None)")
```

peterjc:
I've only just noticed the test. While we do have some print-and-compare tests, they require the expected output be saved as a text file (see the chapter in the Tutorial on this). Using the Python ``unittest`` framework would be nicer.

I think to be complete, the test would need to set ``Entrez.cache`` to an empty folder created with ``tempfile``, and then do an Entrez XML parsing example which is expected to download a DTD file, and confirm not only that the parsing worked, but also that the DTD file was downloaded. That does seem quite complicated though... so something based on your current test ought to be enough.

Bio/Entrez/Parser.py
cparmet: Just confirming I understand, this would mean the cache must be an existing directory that user creates before running Entrez, right? In my test code, I used the cache parameter to have Entrez create the tmp folder ad hoc. If this is bad form, it makes sense to me to enforce that the cache must pre-exist.

cparmet:
@peterjc Thank you for this guidance on best practices for the Test script. I'll work on this!

Bio/Entrez/Parser.py
peterjc: True, this could be done either way since the code will create the directory if it does not exist... its a design choice. Now that you've pointed this out I have no strong view either way.

cparmet:
@peterjc - I'm working on setting up the 'complete test' as a unittest. Can
you confirm my strategy, in particular the best way to force Parser to
download a new DTD to the cache? This issue may be why you suggested such a
test is complicated, but I thought I'd give it a shot. Happy to proceed
with testing as you deem best.

*Possible unit test strategy*

   - In global DTD directory, rename this file: Bio/Entrez/DTDs/
esummary_v*.dtd.
   Change name to something Parser won't find, like holdout_for_cache_test.dtd.
   - Set cache to /temp
   - Call Entrez.esummary for a PMID.
   - Confirm parsing worked.
   - Confirm that Entrez downloaded a new esummary_v*.dtd file and put it
   in temp/Bio/Entrez/DTDs.
   - Delete /temp directory
   - Restore the original Global DTD file - Rename holdout_for_cache_test.dtd
   back to esummary_v*.dtd

I hesitate because this unit test would modify (renames) a working DTD
file, simply for the purposes of testing. Otherwise, Parser will find the
DTD in global DTDs, and not download a local one to the cache.

(On AWS Lambda, the new cache feature works successfully. There is of
course no global DTD directory then, and all DTDs are downloaded as needed
to a tmp cache. The above strategy mimics that environment, I suppose.)

Is there another Entrez call that always requires a fresh download to local
(no DTD in global DTDs, on any test environment a user might have)?


*If it's better to do a print-and-compare test *
I could remove the steps that involve renaming/restoring the Global DTD
file.

I'd leave a note at top of the test script to direct the user to perform
those steps manually before/after running the script, to test the cache
feature.

Would that be sufficient and consistent with best practices for Biopython
testing?

Thank you, Peter!

peterjc:
Changing working DTD files in the global DTD directory during the tests seems to me a risky idea.

I've not looked into this in enough detail to be confident how we could write a non-invasive test with no side effects.

For the test I was suggesting in https://github.com/biopython/biopython/pull/1522#issuecomment-362252203 , you are right - we'd need to find a valid NCBI DTD file which is not currently included in ``Bio/Entrez/DTDs``. Or, we deliberately stop providing one of the minor rarely used files (since in principle Biopython would download and cache it on demand in real world usage).

Perhaps you could make an artificial example with an invented DTD file, and an XML file which refers to the invented DTD file. With the cache set to a test folder with the invented DTD file in it, confirm that ``Entrez.parse`` can parse the XML file. 

I'm willing to be practical here - if we can't easily come up with an automated test, we could still include the new functionality (after manual testing)? However, @mdehoon is the module maintainer so his thoughts are important here.

peterjc:
@mdehoon do you have any input here please (on exposing the DTD cache folder as a user-configurable setting)? Thanks!

mdehoon:
@peterjc  @cparmet  I understand the need to set the directory, but I would reorganize the code a bit. Specifically, I would suggest to make "directory" a property of the DataHandler class (instead of a local variable when DataHandler is first initialized), so that it can be changed at any time. Then, instead of initializing local_dtd_dir etc. at the top of DataHandler class, they can be initialized whenever the directory property changes, or whenever local_dtd_dir etc. is actually used. Using directory as a property, it must be possible to set up the code such that (a) local_dtd_dir etc. always makes use of the current value of directory, and (b) local_dtd_dir etc. does not get initialized when the module is first imported (i.e. before the user had a chance to set local_dtd_dir).
If directory is a property of DataHandler, then there is no need for a separate Bio.Entrez.cache variable. I agree that Bio.Entrez.cache would be nicer than a directory variable hidden in Bio.Entrez.Parser.DataHandler, but I don't see an easy way right now to track if the value of Bio.Entrez.cache has changed.

cparmet:
@peterjc @mdehoon Sounds great. i will give it a shot and let you know any
questions.

Could you also suggest a preferred strategy for unit /other testing? Please
see discussion above.

cparmet:
@peterjc I'm sorry I don't understand the issues related to user-set
timeout values or online resources, with regards to the custom DTD/XSD
directories. I'm new to the guts of Biopython so forgive me!

To confirm, you'd be content with the strategy of the simple "print" type test
file
<https://github.com/biopython/biopython/blob/538223fdf0af149204f6c450185efa9467012b40/Tests/Entrez/test_Entrez_cache.py>I
included with the February commits? I'd, of course, update it after making
the directory changes recommended by @mdehoon. The test file calls esummary
with a custom directory (currently Bio.Entrez.cache) and checks if the call
worked and the DTD file was downloaded there.

I believe back in February we were discussing whether there was a
reasonable unit test I could write instead of print type tests. Sorry about
not being more specific, this is what I was hoping to double check in my
last comment.

I'm happy to do whatever you two advise would be most Biopython-ic!

peterjc:
Apologies about the timeouts thing - I commented on the wrong issue

peterjc:
OK, Entrez directory cache settings, and tests.

In terms of testing, no, please do not use a print-and-compare test.

Instead use the Python ``unittest`` framework. There is a chapter about our testing setup in the Biopython Tutorial, see http://biopython.org/DIST/docs/tutorial/Tutorial.html or http://biopython.org/DIST/docs/tutorial/Tutorial.pdf

This would probably fit within ``Tests/test_Entrez.py``

Note that as I wrote earlier:

> I'm willing to be practical here - if we can't easily come up with an automated test, we could still include the new functionality (after manual testing)? However, @mdehoon is the module maintainer so his thoughts are important here.

cparmet:
OK makes sense. @mdehoon let us know what you think about testing? And note
@peterjc suggested some unit test options in our Feb comments here.

Thanks, both!

mdehoon:
@cparmet  I agree with @peterjc  's comments to use Python's `unittest` framework, and to use Python's `tempfile` module to create the temporary file. Creating a file under `/tmp` directly may work, but it can be fragile on some platforms, especially Windows.

cparmet:
@mdehoon Hi - I've implemented directory as a property of DataHandler, and added an offline unittest. The revised module works on a test app I deployed to AWS Lambda, which was my initial use case for this PR. (Lambda only gives the app write access to a specific tmp directory, so Entrez calls fail without this modification.) 

See what you think.

cc: @peterjc

Bio/Entrez/Parser.py
```diff
@@ -640,3 +623,49 @@ def externalEntityRefHandler(self, context, base, systemId, publicId):
         handle.close()
         self.dtd_urls.pop()
         return 1
+
+    def _initialize_directory(self, directory_type):
```
peterjc: It isn't clear from the docstring what argument ``directory_type`` should be. You use a mix of strings like ``directory_type='XSD'`` and a list like ``directory_type=['DTD', 'XSD']``

In anyway case, why not always initialise both the DTD and XSD paths? That would make the internal API simpler too.

peterjc:
I had a quibble about the internal API, but the user facing API seems nice 👍

peterjc:
What do you think @mdehoon? OK to merge (I'd suggest squash-and-merge given how much this as changed)?

This lets the user set a private attribute, but the cache directory could in future be exposed as an explicit argument to ``__init__`` instead.

@cparmet Do you want to be thanked by name in the ``NEWS.rst`` and ``CONTRIB.rst`` files?

cparmet:
@peterjc Hi - Per your comment, I removed that directory_type parameter and committed a new version. 

I would be happy to be thanked by name in those files. Much appreciated!

PS: My original thought on the directory_name was whether there could be an efficiency to specifying DTD or XSD, if a setup would only need/allow the creation of 1 such directory. Such as a serverless deployments that delete the tmp folder immediately. But I see now from your comment on this, any efficiency would be tiny and not worth the added user complexity.

peterjc:
I've not worked with a server which aggressively removes removes temp files, so I had not thought of that.

cparmet:
On AWS Lambda at least (I’m not sure about other serverless platforms from Google etc), code/apps are forbidden from creating files that persist on the app server. The temp directory is purged immediately after the Python code finishes.

cparmet:
@peterjc @mdehoon  Hi. Just checking, is the codecov patch test an expected fail since this unit test is, by design, an offline test only? So we're not planning to test all the online pathways through the DTD/XSD resourcing steps? 

Do I understand correctly, or are there more unit tests I should add?

Thanks!

peterjc:
Currently our TravisCI and AppVeyor setups deliberately do not run our online tests, which means that our CodeCov information does not track that.

In any case, we treat the coverage as advisory rather than a hard requirement for pull requests.

cparmet:
Very cool that you have it set up to distinguish offline vs online tests. Thanks for explaining, @peterjc !

mdehoon:
Looks good to me. I would suggest to remove the empty `return`s at the end of `_initialize_directory` and `directory.setter`, as they are unnecessary. Otherwise I'm OK to squash and merge.

cparmet:
@mdehoon @peterjc Empty returns removed and pushed new commit.

Thank you both for all the guidance you gave me along the way of this PR. I learned a lot!

Bio/Entrez/Parser.py
```diff
@@ -535,14 +514,15 @@ def count(model):
             self.structures.update({name: multiple})
 
     def open_dtd_file(self, filename):
-        path = os.path.join(DataHandler.local_dtd_dir, filename)
+        self._initialize_directory()
```
mdehoon, line 517: Actually why was this line needed here? We are already calling `self._initialize_directory` from `directory.setter`.

Same for the calls to `self._initialize_directory` below.

cparmet: @mdehoon Great question. I'm refreshing my memory. 

Am I correct that the `directory.setter` function will be called:
1. when the DataHandler object is created (__init__)
2. again if the user sets a custom directory (local cache)
?

If that's true, then I see your point. There shouldn't be a need to call _initialize_directory yet again in the open* and save* functions, if self_directory should not have changed in the interim? It will already have gotten initialized one way or another? It's possible those 4 instances were added for another reason I can't recall. I apologize, I haven't been in this codebase since the PR.

mdehoon: It's easy to set it up like that. But should `directory` be a class attribute or an object attribute? Right now it is an object attribute, meaning that it has to be set separately each time a `DataHandler` is created.

cparmet: Sorry for the delay in responding, @mdehoon . 

Can you catch me up? It looks like the `_initialize_directory` method from this PR has since evolved into the `directory` method decorated by `@directory.setter`. But it sounds like we are still making this unnecessary call to that code (formerly in `self.initialize_directory`) in the 4 open and save methods? Can you trace the story for me? 

Apologies, I'm rusty from being away from this repo. I'm very happy to help clean up my PR! 

When I look at the latest Master, `open_dtd_file` and cousin methods don't seem to initialize anymore or set `directory` directly. From your comment I am guessing each method somehow invokes a `directory=`, etc, and I'm just missing it now. Thank you!

PR #1526, opened 2018-02-08
Improve fasta io

Gasta88:

Probably I went the wrong way here, because the original test script is opening Fasta-m10 files with `AlignIO.parse()`. 

Here I used `FastaIO.FastaM10Iterator()` like in the self test attached to the `AlignIO.FastaIO` module.

Any problem, I'll edit as soon as I can.


Tests/test_AlignIO_FastaIO.py
```diff
@@ -2,73 +2,255 @@
...
-    assert t_format == "fasta-m10" and t_per == 2
+    def test_output001(self):
+        """Check output001.m10 file."""
+        fasta_file = 'Fasta/output001.m10'
+        with open(fasta_file, "r") as handle:
+            alignments = list(FastaIO.FastaM10Iterator(handle))
+            self.assertEqual(len(alignments), 4)
+            self.assertEqual(len(alignments[0]), 2)
+            self.assertEqual(alignments[0].get_alignment_length(), 108)
+            self.assertEqual(str(alignments[0][0].seq) + " " + alignments[0][0].id + " " + str(alignments[0][0].annotations["original_length"]), "SGSNT-RRRAISRPVRLTAEED---QEIRKRAAECGKTVSGFLRAAALGKKVNSLTDDRVLKEVM-----RLGALQKKLFIDGKRVGDREYAEVLIAITEYHRALLSR gi|10955263|ref|NP_052604.1| 107")
```
peterjc: That line and the others like it is far too long. Also it would be much clearer as three lines, and you shouldn't need ``str(...)`` on the sequence object.

```python
self.assertEqual(alignments[0][0].seq, "SGSNT-RRRAISRPVRLTAEED---QEIRKRAAE"
                                       "CGKTVSGFLRAAALGKKVNSLTDDRVLKEVM---"
                                       "--RLGALQKKLFIDGKRVGDREYAEVLIAITEYHRA"
                                       "LLSR")
self.assertEqual(alignments[0][0].id, "gi|10955263|ref|NP_052604.1|")
self.assertEqual(str(alignments[0][0].annotations["original_length"]), "107")
```

Do you need ``str(...)`` for the original length? Does this work?

```python
self.assertEqual(alignments[0][0].annotations["original_length"], 107)
```

Also since this becomes very repetitive and boring, I personally wouldn't check the exact sequence of every pairwise alignment. Focus on the interesting ones (e.g. gaps in odd places).

peterjc:
It should be fine to focus on the lower level ``FastaIO.FastaM10Iterator()`` here since we test the ``AlignIO.parse()`` usage elsewhere.

Tests/test_AlignIO_FastaIO.py
Gasta88: Repetition is not a problem since I generate the test methods via a script.
If I don't use the `str(...)` I get back a `Seq` object type, which is not comparable with a string.

I can remove the `str(...)` from the original length.

peterjc: This should work on a Biopython 1.65 or later

```python
from Bio.Seq import Seq
assert Seq("ACGT") == "ACGT"
assert "ACGT" == Seq("ACGT")
```

peterjc:
You should also delete the old print-and-compare expected output file, ``Tests/output/test_AlignIO_FastaIO`` - and after that I think this is ready to merge.

Thanks for working on this (and good to know you automated creating the test asserts!)

peterjc:
Thank you!

PR #1529, opened 2018-02-13
remove self-test from Bio.PDB.PDBIO

Gasta88:

The module is already tested in `test_PDB.py` and it has a coverage of 83.73% on CodeCov.


peterjc:
Thank you

Issue #1536, opened 2018-02-20
Questions: interproscan suport


azneto:
I have to parse a xml from Inteproscan and couldn't find a python module for that. 
Should I just deal with the XML directly or is it worth to implement a biopython module + the XML parser?
If the best option is to create the module, should it be independent, like Bio/Blast, or should it integrate Bio/SearchIO?

peterjc:
I would think integration into ``Bio.SearchIO`` would be best, what do you think @bow ?

bow:
It does sound like a fit for SearchIO. It's also nice that the InterproScan team [has put up a proper documentation on what we can expect from the output](https://github.com/ebi-pf-team/interproscan/wiki/OutputFormats).

> Should I just deal with the XML directly or is it worth to implement a biopython module + the XML parser?

I would say it depends on your use case. Plain XML parsing seems to be the simplest and quickest way to get whatever you need from the file. But if you're interested in contributing an `interproscan-xml` parser, I would be happy to give some pointers and help along the way. This would require more work, though it would surely help others who need to parse such a file as well.

azneto:
We're already using blast-xml, therefore interpro-xml would be a really nice solution for handling interproscan files. I'll study your code and give it a try. @bow , do you mind if I write you directly to have your feedback on my fork and discuss the progress?

bow:
@azneto Not at all. Feel free to ping me or write to me anytime.

azneto:
The first draft is in my fork https://github.com/azneto/biopython
There's a test file in the Tests directory.

There are some issues we need address:

- go-ref and pathway-ref: I was planning to load them as hit._id_alt and hit._description_alt but it didn't work since they must be unique even between hits and it's common to find different hits with the same GO term. Should I add a xref data structure in SearchIO._model.hit.py?
- hit database: Whereas blast has the database info for the query_result, intepro has such information available for every hit (signature-library-release). I guess I should implement a 'target' data structure for hit.py, the same way query.py has.
- models: I'm planning to ignore them for now.

bow:
Closing this issue since the proposed changes has been merged.

PR #1543, opened 2018-02-24
Bio.SearchIO parser for InterProScan XML

azneto:
This pull request addresses issue #1536 


Tests/test_SearchIO_interpro_xml.py
```diff
@@ -0,0 +1,76 @@
...
+TEST_DIR = 'InterProScan'
+FMT = 'interpro-xml'
+
+
+def get_file(filename):
+    """Returns the path of a test file."""
+    return os.path.join(TEST_DIR, filename)
+
+
+class BlastnXmlCases(unittest.TestCase):
```
bow: This should be `InteproscanXmlCases`, I believe :).

Bio/SearchIO/InterproIO/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""InterProScan."""
+from .interpro_xml import InterproXmlParser
+
+
```
bow: Would it also be useful to add more docstrings here about the format in general? For example, you can say which InterProScan output type is supported or how to access certain values. The docstrings in `BlastIO/__init__.py` and/or `HmmerIO/__init__.py` may be useful as a template.

Bio/SearchIO/__init__.py
```diff
@@ -232,6 +232,7 @@
         'hmmscan3-domtab': ('HmmerIO', 'Hmmer3DomtabHmmhitParser'),
         'hmmsearch3-domtab': ('HmmerIO', 'Hmmer3DomtabHmmqueryParser'),
         'phmmer3-domtab': ('HmmerIO', 'Hmmer3DomtabHmmqueryParser'),
+        'interproscan-xml': ('InterproscanIO', 'InterproscanXmlParser'),
```
peterjc: The dictionary keys seem to be alphabetical here.

peterjc:
My query to @bow is about the new attributes being added, and if this the best solution long term? Should we favour a dictionary approach over adding new format specific attributes?:

```
 +        self.type = None
 +        self.target = None
 +        self.target_version = None
 +        self.dbxrefs = []
```

We have the precedent of ``.dbxrefs`` in the ``Seq*`` objects so the naming of that is fine.

Tests/test_SearchIO_interpro_xml.py
azneto: done

Bio/SearchIO/InterproIO/__init__.py
azneto: done

Bio/SearchIO/__init__.py
azneto: done

azneto:
@bow, it's good to know it's in the right track.
I'll address each topic:

1. matchTypes and locationTypes have always the same prefix. So I've decided to store the string excluding the '-match' or '-location'. The variable is hit.type, but I'd like to hear from you about Peter's questioning. I'm not quite sure about how to implement it in another way (dict).
2. InterProScan does not store 'no hits' and I didn't find flags to modify the xml output. 
3. Nice observation. I've just changed every mention to the software to 'interproscan'.

bow:
@peterjc Hmm, a dictionary sound more preferable. This way the `__str__` display can be made generic for other format attributes as well.

@azneto For the dict, we can have one new attribute (e.g. `format_attributes`, which is of type `dict`). Format-specific parsers could then add their own attributes there.

For your responses:

1. In that case, related to the current naming of the types also: `type` is perhaps better replaced with a synonym (`kind`? `database_type`? `search_type`? `hit_type`?). This is because `type` is a Python keyword and it's generally a good idea to avoid shadowing the builtin keywords.

2 & 3. Ok.

azneto:
I think I've got it.

azneto:
Hi @bow, did you have the chance to review it?

bow:
@azneto ~ Ah, I missed the notification for the last changeset, sorry. I seems ready for merging now. @peterjc, anything else that should be added?

peterjc:
From a usability perspective, is ``.format_attributes`` as the dict name the best choice? Is ``.attributes`` enough? Or maybe ``.qualifiers`` or ``.annotations`` better to match other usage in Biopython?

azneto:
Such name refers to format specific attributes. But I agree with you, .attributes will definitely improve usability.

peterjc:
I think this would deserve to be mentioned in the ``NEWS.rst`` file, along with adding your name for the release and to the ``CONTRIB.rst`` file.

Also once this is merged, we'd want to add an entry to the website listing the new format in Biopython 1.71:
http://biopython.org/wiki/SearchIO
https://github.com/biopython/biopython.github.io/edit/master/wiki/SearchIO.md

azneto:
Sounds good. My name is already in the CONTRIB.rst and NEWS.rst files. We just need to add the information about the tool.

bow:
Great! If there are no further objections, I will merge this today, no sooner than 12 noon CET. Thank you @azneto and for @peterjc for reviewing.

peterjc:
Is it easy to add indexing as well as a future addition? Or is there little need for it?

bow:
There is some resemblance to the `blast-xml` structure here, so I'm guessing the approaches we use in `blast-xml` indexing can be used if we do need indexing.

But for now, I guess this can be merged as-is (unless @azneto wants to work on indexing before this is merged to `master`?). This already covers the majority of the use cases it seems :).

peterjc:
Sure, merge as is - indexing can be added later.

azneto:
I think it might be useful add indexing, but I never worked with this and I'll need time to study it. We can add that later on.

bow:
Great, ok, I'm merging this now then :+1:. Thanks!

Issue #1544, opened 2018-02-26
Incorrect error propagation in Bio.File.as_handle


bow:
### Setup

Hi everyone, I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
3.5.3 (default, Feb 20 2017, 14:45:39) 
[GCC 6.3.1 20170109]
CPython
Linux-4.14.20-1-lts-x86_64-with-arch
1.71.dev0
```
(this is done on the tip of the current master: 27e345755).

### Expected behaviour

When using `Bio.File.as_handle` in a `with` block, TypeErrors from inside the `with` block should be propagated correctly.

### Actual behaviour

In addition to the actual exception traceback, another internal error is shown:

```
<actual exception traceback>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "as_handle.py", line 11, in <module>
    write_to(target, "foo")
  File "/home/bow/.pyenv/versions/3.5.3/lib/python3.5/contextlib.py", line 78, in __exit__
    raise RuntimeError("generator didn't stop after throw()")
RuntimeError: generator didn't stop after throw()
```

### Steps to reproduce

You can run the following script:

```
#!/usr/bin/env python

from Bio.File import as_handle


with as_handle("test.txt", "w") as target:
    raise TypeError("foo")
```

### Additional Notes

I stumbled on this when trying to add keyword arguments for `SeqIO.*` in [my branch](https://github.com/bow/biopython/tree/enh/seqio-kwargs). It seems to be caused by the `except TypeError` block at the end of the `as_handle` function. [Exceptions inside the `with` block is reraised at the point where `yield` occured inside the decorated function](https://docs.python.org/3.6/library/contextlib.html#contextlib.contextmanager), so in our case this means the TypeError is then re-caught by the final `except` block, which then yields another object. This incorrect, however, and I suppos is the reason the `contextlib` authors added a check on it (which raises the second additional exception).

I have a fix already, [in that branch](https://github.com/bow/biopython/blob/enh/seqio-kwargs/Bio/File.py#L85) ~ but I'd just like another set of eyes to confirm the bug and/or the fix in the branch. It simply ensures that `TypeError`s encountered in the `yield` points are tracked correctly and are re-raised in the last `except` block.

If everything looks ok, I'll open a PR shortly.

peterjc:
[...]

PR #1545, opened 2018-02-28
Fix for as_handle exception propagation

bow:
This pull request addresses issue #1544 

### DIFF
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -82,17 +82,20 @@ def as_handle(handleish, mode='r', **kwargs):
     been deprecated (this happens automatically in text mode).
 
     """
-    try:
-        if sys.version_info[0] >= 3 and "U" in mode:
-            mode = mode.replace("U", "")
-        if 'encoding' in kwargs:
-            with codecs.open(handleish, mode, **kwargs) as fp:
-                yield fp
-        else:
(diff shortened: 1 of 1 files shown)


Bio/File.py
```diff
@@ -82,16 +82,23 @@ def as_handle(handleish, mode='r', **kwargs):
     been deprecated (this happens automatically in text mode).
 
     """
-    try:
-        if sys.version_info[0] >= 3 and "U" in mode:
-            mode = mode.replace("U", "")
+    # Workaround for https://github.com/biopython/biopython/issues/1544.
+    # TODO: Remove when Python 3.5 dependency is dropped.
+    if sys.version_info[:2] < (3, 6):
```
peterjc: I think this needs to be also checking at least Python 3? Otherwise:

```python
ImportError: No module named pathlib
```

peterjc: Correction: At least Python 3.4
https://docs.python.org/3.4/library/pathlib.html

Specifically we don't want to do this on Python 2

bow: Ah, indeed ~ I missed the version 2 check. It's updated now, also with a more granular check for the minor version.

bow:
Closed in favor of #1555

Issue #1546, opened 2018-03-04
Improper reliance on local encoding of remote resource


EdwardBlakeMiller:
This is in response to issue [#1402](https://github.com/biopython/biopython/issues/1402) which was addressed with commit [1a32c50](https://github.com/biopython/biopython/commit/1a32c50fb91b9461eb4c0d0d427823dca81b41ee).

The issue there was regarding platforms where UTF-8 is not the default encoding, for example, Windows, where cp1252 may be the default.

This only occurs with python3 as only there is `_binary_to_string_handle` needed.

The solution to that issue was to add the following to `_binary_to_string_handle`:

```
+        try:
 +            # If this is a network handle from urllib,
 +            # the HTTP headers may tell us the encoding.
 +            encoding = handle.headers.get_content_charset()
 +        except AttributeError:
 +            encoding = locale.getpreferredencoding(False)
 +        wrapped = io.TextIOWrapper(io.BufferedReader(handle), encoding=encoding)
```

This has a number of problems:

1. `encoding = handle.headers.get_content_charset()` may return `None` rather than raise an `AttributeError`. In this instance, the call to `io.TextIOWrapper` with `encoding=None` will default to the system's default encoding, which on Windows may be cp1252, which is a revert back to the initial issue.
2. A reliance on ` locale.getpreferredencoding(False)` is reverting to the default encoding of the user's system, which has no relation to the encoding of a remote resource. According to the HTTP protocol parameters from [W3C](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1):

> When no explicit charset parameter is provided by the sender, media subtypes of the "text" type are defined to have a default charset value of "ISO-8859-1" when received via HTTP. 

`ISO-8859-1` is also known as `latin-1`. Therefore, when no encoding information is available, the default should revert to `latin-1` and not the user's preferred encoding or the user's default encoding.

fabianegli:
Assuming text transfers to be `ISO-8859-1` a.k.a. `latin-1` encoded seems to be obsolete for version HTTP 1.1 (contrary to HTTP 1.0). 

From https://tools.ietf.org/html/rfc7231 `Appendix B.  Changes from RFC 2616`:

> The default charset of ISO-8859-1 for text media types has been
   removed; the default is now whatever the media type definition says.
   Likewise, special treatment of ISO-8859-1 has been removed from the
   Accept-Charset header field.  (Section 3.1.1.3 and Section 5.3.3)

[...]

PR #1547, opened 2018-03-04
Default to 'latin-1' encoding for HTTP text requests.

EdwardBlakeMiller:
This is following the W3C protocol documentation here:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1

This pull request addresses issue #1546 ...

### DIFF
--- a/Bio/_py3k/__init__.py
+++ b/Bio/_py3k/__init__.py
@@ -106,6 +106,13 @@ if sys.version_info[0] >= 3:
         except AttributeError:
             encoding = None
         if encoding is None:
+            # The W3C recommendation is:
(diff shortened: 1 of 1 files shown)


peterjc:
This seems sensible for the network handles - although if we want to restrict this to network handles only this may not be the best place to do the encoding selection? Note someone might well be parsing an NCBI file previously saved to disk, so this is not clear cut.

@mdehoon what do you think?

@EdwardBlakeMiller I'd like a comment on this if statement linking to the W3C recommendation explaining why ``latin-1`` aka ``ISO-8859-1`` is used.

peterjc:
Thanks!

PR #1553, opened 2018-03-09
expose timeout parameter in Bio.KEGG.REST

ryhaberecht:
This exposes the timeout parameter from urllib.request.urlopen (imported through Bio._py3k) to each function in Bio.KEGG.REST.
If timeout == None, it defaults to the system's default socket timeout, just as before.
Timout is in seconds.

**Why:**
When downloading many files over a long period of time, things will go mad. Especially on a wireless connection.
Until now, the query just stalled for an undefined time. Theoretically, the socket's default timeout would have fired eventually, but I have witnessed several cases where a thread executing the query simply stalled forever, with no chance of recovery.
While this might have been caused by a totally unrelated bug, I feel confident that exposing the socket's timeout parameter can help solve networking issues for many users and use cases.

**Beware:**
On my system (Linux 4.9, cPython 3.4.6), for some reason, sockets double the timeout passed to them. If you tell REST.kegg_get() to timeout after 3 seconds, it will always timeout after 6, etc. To solve this issue I added this to my global settings store:
downloadTimeoutSocket = math.floor(downloadTimeout / 2)
If this issue exists on more systems, we might want to add my global fix into _q(). However, I was reluctant to do so because of probably many unnecessary math operations.

### DIFF
--- a/Bio/KEGG/REST.py
+++ b/Bio/KEGG/REST.py
@@ -31,7 +31,7 @@ from Bio._py3k import urlopen as _urlopen
 from Bio._py3k import _binary_to_string_handle
 
 
-def _q(op, arg1, arg2=None, arg3=None):
+def _q(op, arg1, arg2=None, arg3=None, timeout=None):
     URL = "http://rest.kegg.jp/%s"
     if arg2 and arg3:
         args = "%s/%s/%s/%s" % (op, arg1, arg2, arg3)
@@ -39,7 +39,11 @@ def _q(op, arg1, arg2=None, arg3=None):
         args = "%s/%s/%s" % (op, arg1, arg2)
     else:
         args = "%s/%s" % (op, arg1)
-    resp = _urlopen(URL % (args))
+
+    if timeout is not None:
+        resp = _urlopen(URL % (args), timeout=timeout)
+    else:
+        resp = _urlopen(URL % (args))
 
     if "image" == arg2:
         return resp
@@ -48,7 +52,7 @@ def _q(op, arg1, arg2=None, arg3=None):
 
 
 # http://www.kegg.jp/kegg/rest/keggapi.html
-def kegg_info(database):
+def kegg_info(database, timeout=None):
     """KEGG info - Displays the current statistics of a given database.
 
     db - database or organism (string)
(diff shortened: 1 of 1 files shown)


peterjc:
Does this work? It seems like a simpler general solution which we could document (e.g. FAQ entry, and directly mention in the docstrings for the various online modules):

```python
import socket
socket.setdefaulttimeout(5)
from Bio.KEGG import REST
# or from Bio import Entrez
# or from Bio import TogoWS
# or ...
...
```

peterjc:
I had assumed this would only change the timeout for the currently running Python instance, which is possibly a problem for some complicated programs.

ryhaberecht:
As far as I know, only Windows initialises a new python instance for each new process, on Unixoids fork() is used on the parent process, copying the whole process address space. Therefore it seems as if on non-Windows systems, socket.setdefaulttimeout() even affects child processes, if it happened before the fork. I forgot to mention that in the list of concerns above.

Of course, I might well misunderstand how the internal "_socket" class (this is the underlying module written in C, which actually implements setdefaulttimeout) behaves, as I did not look at its code. But it would really surprise me if the python developers implemented any kind of transparent barrier between modules/packages/threads/processes, seeing that they generally refrain from implementing any kind of complex guard rail, as this would be deemed "not pythonic".

peterjc:
Isn't all this abstraction wonderful - normally it just works and we don't need to know or care. In this case I don't know how localised this "global" setting is.

@mdehoon what do you think as maintainer of another of the online modules in Biopython - should we expose the timeout settings like this on each and every function?

ryhaberecht:
Really is great! Had a ton of fun implementing abortable Futures in Python. Hence my sarcastic hint at "pythonism" ;-)

Another solution short of exposing a timeout parameter on each function is keeping track of an optional timeout parameter per object of class REST. Obviously, there is no such class yet, we simply use functions of a python module named REST.py. Would it be possible to keep the current API for the REST.py module, but implement it via a singleton object in the background (of class REST), which could then store a timeout parameter? So normally, users don't care about timeouts and use REST.py as they always did, but if they do care, they can explicitly create a REST object, give it their desired timeout, and use it for their most beloved, fault-tolerant HTTP requests.
The singleton would be unique to the whole process then, while an explicitly created object is subject to standard scoping and garbage collection, right?

peterjc:
A singleton class would work but my gut reaction is it seems over-engineered. I can see another way to minimally expand our API to cover this.

What we have in ``Bio.Entrez`` is module level constants for settings (like the user's email address which the NCBI require). We could have a module level timeout, so the user only has to set it once - e.g.

```python
from Bio.KEGG import REST
REST.timeout = 5  # Default was too short on my network
...
```

The same pattern could be applied to ``Bio.Entrez`` etc as needed.

ryhaberecht:
Agreed, this would be the best solution. A willing user in need of it could quickly expand the module constant with toggling, or thread-safety measures, etc.

peterjc:
I'm going to close this as stale, thank you though.

PR #1555, opened 2018-03-09
File.as_handle: fix incorrect error propagation (#1544)

peterjc:
This pull request addresses issue #1544, commit from @mruffalo as an alternative to PR #1545.

### DIFF
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -82,7 +82,16 @@ def as_handle(handleish, mode='r', **kwargs):
     been deprecated (this happens automatically in text mode).
 
     """
-    try:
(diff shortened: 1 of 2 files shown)


mruffalo:
I'd be quite happy with this PR being merged, but I would also recommend including https://github.com/mruffalo/biopython/commit/1aa46de81f0222bacaf2256d5714f7a1d0d98357. It seems appropriate to ensure that we're as permissive as possible with custom path-like objects, and that no one accidentally breaks this in the future.

bow:
Ok, this looks good now ~ thank you everyone!

peterjc:
Thanks everyone.

Issue #1532, opened 2018-02-16
The 'pairwise2.align.globalms' function is sometimes not returning any alignments


dhauge:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 12:01:12) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
CPython
Darwin-17.4.0-x86_64-i386-64bit
1.70
```

Calls to `Bio.pairwise2.align.globalms` with some sequences and weights return no alignments. Unless I am missing some subtlety of the algorithm it should always return something because there are always alignments, so there must be at least one best one.

One problematic set of arguments is demonstrated by
```python
import Bio.pairwise2
print(Bio.pairwise2.align.globalms('GGCGCTGGGACGCCTGCTCACTGGGCGCT', 'GGAGCAGGGAGCCAGTCACCCTGGGAGCCTGGACCA', 1.0, -0.5, -1.75, -0.25, penalize_end_gaps=(True, False)))
```

This is new in `1.68`; version `1.67` did return results.

### Expected behaviour

In v1.67, this prints
```python
[...]

PR #1563, opened 2018-03-13
Solve pairwise2 is sometimes not returning an alignment

MarkusPiotrowski:
This pull request addresses issue  #1532

Since Biopython 1.68 alignments are possible where a gap in one sequence is followed by a gap in the other sequence. At the same time, to reduce redundancy, only one of two redundant cases are allowed:  a gap in sequence A may follow a gap in sequence B, but not the other way round.
Issue #1532 describes cases where this behaviour may result in returning no alignments. The reason and a workaround is described in the issue.

This pull request implements the described workaround in ```Bio.pairwise2```: If an empty alignment is returned, the traceback will be repeated with inversed sequence input and inversed matrices.


peterjc:
You've added a sensible test, and all our automated checks pass - so merged, thank you.

Issue #1564, opened 2018-03-14
SearchIO hmmer2-text parser incorrectly handles "end of sequence" marker if that is in a line break


kblin:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609]
CPython
Linux-4.4.0-93-generic-x86_64-with-Ubuntu-16.04-xenial
1.71.dev0
git revision: eeaa270
```

### Expected behaviour

Parsing the hmmpfam input using `Bio.SearchIO.parse(file, 'hmmer2-text')` works.

### Actual behaviour

Parsing the hmmpfam input causes a `ValueError`:

```
Traceback (most recent call last):
  File "/home/kai/dev/libs/biopython/Tests/test_SearchIO_hmmer2_text.py", line 221, in test_hmmpfam_23_break_in_end_of_seq
    res = next(results)
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/__init__.py", line 314, in parse
    for qresult in generator:
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/HmmerIO/hmmer2_text.py", line 45, in __iter__
    for qresult in self.parse_qresult():
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/HmmerIO/hmmer2_text.py", line 126, in parse_qresult
    self.parse_hsp_alignments()
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/HmmerIO/hmmer2_text.py", line 298, in parse_hsp_alignments
    frag.query = otherseq
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/_model/hsp.py", line 849, in _query_set
    self._query = self._set_seq(value, 'query')
  File "/home/kai/dev/libs/biopython/build/lib.linux-x86_64-3.5/Bio/SearchIO/_model/hsp.py", line 817, in _set_seq
    len(seq), seq_type))
ValueError: Sequence lengths do not match. Expected: 297 (hit); found: 296 (query).
```

### Steps to reproduce

[line_parsing_break.txt](https://github.com/biopython/biopython/files/1810253/line_parsing_break.txt)

Using the above input file, run:
```python
[...]

PR #1565, opened 2018-03-14
SearchIO/hmmer2-text: Deal with line breaks in the end of sequence marker

kblin:

This pull request addresses issue #1564

### DIFF
--- a/Bio/SearchIO/HmmerIO/hmmer2_text.py
+++ b/Bio/SearchIO/HmmerIO/hmmer2_text.py
@@ -259,16 +259,12 @@ class Hmmer2TextParser(object):
                         break
 
                 # skip the *-> start marker if it exists
-                if self.line[19] == '*':
+                if self.line[19:22] == '*->':
                     seq = self.line[22:]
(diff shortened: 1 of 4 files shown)


kblin:
As mentioned in the bug report, I don't like the dance in lines 286..288, but couldn't come up with a better way of making sure we're not parsing the `-` end coordinate marker as part of `otherseq`.

peterjc:
Over to you @bow - can you see a cleaner fix, or should we go with this?

bow:
This looks ok ~ sometimes ugly corners just pop up of course, at least I think here it is commented properly. (By the way, I mistakenly added my comments on the [issue page](https://github.com/biopython/biopython/issues/1564) rather than here).

Thanks again @kblin!

kblin:
I just ran across an end marker with a `<\n-*` style break and can confirm that my "fake" version matches. So we should be good for both cases. :)

peterjc:
Great. Are you happy with the fake test, or you do want to replace it with the new real one?

Issue #1551, opened 2018-03-07
Cannot change IDs for Residue


jadolfbr:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)]
CPython
Darwin-17.3.0-x86_64-i386-64bit
1.70

### Expected behaviour

`Residue.id(value) or res.id = value`
changes the id of the residue. This is used for renumbering a PDB.
This used to work fine for many years and now results in an error, with no obvious override mechanism. 

### Actual behaviour

```
ValueError: Cannot change id from `(' ', 8, ' ')` to `(' ', 9, ' ')`. The id `(' ', 9, ' ')` is already used for a sibling of this entity.
```

In addition, using `detach_child()` on the residue for the 'already in use sibling' does not work and results in this error:

```
  File "/Library/Python/2.7/site-packages/Bio/PDB/Entity.py", line 120, in detach_child
    child = self.child_dict[id]
KeyError: (' ', 9, ' ')
```

### Steps to reproduce

Change the id of a residue to another residue number, that is a 'child'

peterjc:
Can you make a self contained example, and/or confirm an older version of Biopython where this worked?

CC: @etal @JoaoRodrigues @lennax

jadolfbr:
Version 1.68 works fine.  I'll get a self-contained snippet for you today @peterjc

peterjc:
Using one of the ``Tests/PDB/`` files would probably be easiest for us - it might even end up as a new unit test?

JoaoRodrigues:
Hi @jadolfbr , any luck with the snippet?

I would wager this is because of the "propagation" of parent ids, but hard to look at without a specific case.

jadolfbr:
Hey @JoaoRodrigues - Thanks for the push. I'm still cloning, but here is a quick tested snippet you should be able to run on 1A80, which is in Test/PDB. 

```
import Bio
parser = Bio.PDB.PDBParser()
pdb = parser.get_structure("1A80", "1A80.pdb")
res = pdb[0]['L'][( ' ', 152, ' ')]
res.id = (' ', 153, ' ')
```

peterjc:
Now that #1567 has been merged, does this problem still occur in the latest master branch?

[...]

PR #1567, opened 2018-03-14
Update operators for Residue

schaarj:
The == operator for residues currently seems to only compare identity
of the `id` field which changes behaviour to v1.70. 

This was observed by different results obtained with following code:
```python
ns = NeighborSearch(atom_list)
all_list = ns.search_all(radius=5.5, level='R')
```
which yields different results between 1.70 an 1.71 if two chains with overlapping residue IDs within the returned set are present.

The same applies for the other operators, as residues with identical "id" fields are considered
equal. This e.g. causes NeighborSearch to report the same residue pair twice if the id field is identical.
Last the  `__le__` and `__lt__` operator were identical which I assume is not intended.


peterjc:
Do you have a test case you could share? Ideally using files under ``Tests/PDB/`` but any public structure would be helpful.

Does this link to #1551 perhaps?

Also are you comparing the current release Biopython 1.70 with the current code in GitHub (since we have not yet released Biopython 1.71)?

CC: @etal @JoaoRodrigues @lennax

schaarj:
Yes I'll post the example I already shared with @JoaoRodrigues ,
I'm not sure if it's related to the mentioned issue but could well be.
I indeed compared the current master with v1.70

```python
from Bio.PDB import NeighborSearch, FastMMCIFParser, PDBList, PDBParser
#
# import sys,Bio
# print(sys.version)
# print(Bio.__version__)

def analyse_contacts(contact_list):
    """
    Enumerates and classifies contacts based on the chemical characteristics
    of the participating amino acids.
    """

    bins = {
        'AA': 0, 'PP': 0,
        'CC': 0, 'AP': 0,
        'CP': 0, 'AC': 0,
    }

    _data = {'ALA': 'A', 'CYS': 'A', 'GLU': 'C', 'ASP': 'C', 'GLY': 'A', 'PHE': 'A', 'ILE': 'A', 'HIS': 'C', 'LYS': 'C',
             'MET': 'A', 'LEU': 'A', 'ASN': 'P', 'GLN': 'P', 'PRO': 'A', 'SER': 'P', 'ARG': 'C', 'THR': 'P', 'TRP': 'A',
             'VAL': 'A', 'TYR': 'A', }

    for (res_i, res_j) in contact_list:
        contact_type = (_data.get(res_i.resname), _data.get(res_j.resname))
        contact_type = ''.join(sorted(contact_type))
        bins[contact_type] += 1

    return bins

def run(s): 
    chains = list(s.get_chains())
    for c in chains:
        if c.id not in {'A','B'}:
            c.parent.detach_child(c.id)
    
    res_list = list(s.get_residues())
    
    def _ignore(r):
        return r.id[0][0] == 'W' or r.id[0][0] == 'H'
    
    for res in res_list:
        if _ignore(res):
            chain = res.parent
            chain.detach_child(res.id)
    
    atom_list = list(s.get_atoms())
    def _ignore(x):
        return x.element == 'H'
    
    
    for atom in atom_list:
        if _ignore(atom):
            print('removing atom: {}'.format(atom.name))
            residue = atom.parent
            residue.detach_child(atom.name)
    
    ns = NeighborSearch(atom_list)
    all_list = ns.search_all(radius=5.5, level='R')
    print("# all contacts: {}".format(len(all_list)))
    ic_list = [c for c in all_list if c[0].parent.id != c[1].parent.id]
    print("# iface contacts: {}".format(len(ic_list)))
    print(analyse_contacts(ic_list))
    
p = FastMMCIFParser(QUIET=True)
s = p.get_structure('1xwd', PDBList().retrieve_pdb_file('1xwd', file_format='mmCif'))
run(s)
# p = PDBParser(QUIET=True)
# s = p.get_structure('1xwd', PDBList().retrieve_pdb_file('1xwd', file_format='pdb'))
# run(s)
```

peterjc:
If this broke since Biopython 1.70, it most likely was from *Implements Py3-compatible rich comparison for all Bio.PDB objects* #1360, merged August 2017

https://github.com/biopython/biopython/commit/107fc004d007e7c2464b32e8422c9bc7ec8d951c#diff-9c26d1d62823268c48ba4300a5efe28d

schaarj:
Yes I suspected that it might be linked, #1356 which was addressed by this PR was the reason I actually switched to the latest master from v1.70

schaarj:
I quickly looked at the unit test for comparison of SMCRA objects, and the example structure used there indeed doesn't catch the issue observed (same residue ID different chain). @JoaoRodrigues what's your suggestion, adding a separate unit test for residue or extending that one. Either way a second structure matching the criteria (e.g. `2BEG` from the test folder) or a modification of `a_structure` will be required to test for this scenario.

JoaoRodrigues:
@schaarj I would be more in favor of modifying `a_structure.pdb` to avoid adding yet another file to the test suite. You don't need to add an entire residue, just one atom is enough to trigger that particular code. Thanks!

JoaoRodrigues:
I have a question about using `full_id` (no need to use the `get_full_id()` method) for sorting. When sorting residues, will the same rules still apply, i.e. first by HETATM flag, then by residue name, then by insertion code? If it does, we're good and this is a very simple general solution. Otherwise, I'd stick with the `full_id` for equality checking and stick with the sorting the way it's done.

schaarj:
As mentioned I did a few tests and it appears to sort according to the order in the tuple in python2 and 3, which would be the anticipated sorting order. I also added the parent check for Atom (not yet committed) and all tests pass. Should this be added here as well or should Atom routines remain the way they are (I guess the additional parent check in atom would have the biggest impact on performance)

peterjc:
Our automated PEP8 style checks are failing, mostly just including the trailing full stop on the new docstrings:

```
$ flake8 Bio/
Bio/PDB/Atom.py:78:1: D400 First line should end with a period
Bio/PDB/Atom.py:85:1: D400 First line should end with a period
Bio/PDB/Atom.py:88:19: E271 multiple spaces after keyword
Bio/PDB/Atom.py:92:1: D400 First line should end with a period
Bio/PDB/Atom.py:108:1: D400 First line should end with a period
Bio/PDB/Atom.py:124:1: D400 First line should end with a period
Bio/PDB/Atom.py:140:1: D400 First line should end with a period
Bio/PDB/Atom.py:285:5: E303 too many blank lines (2)
Bio/PDB/Entity.py:58:1: D400 First line should end with a period
ERROR: InvocationError: '/home/travis/build/biopython/biopython/.tox/style/bin/flake8 Bio/'
```

Bio/PDB/Atom.py
```diff
@@ -76,19 +76,24 @@ def __init__(self, name, coord, bfactor, occupancy, altloc, fullname, serial_num
...
-            return (self.id, self.altloc) == (other.id, other.altloc)
+            return self.full_id == other.full_id
         else:
             return NotImplemented
 
     def __ne__(self, other):
+        """Test inequality."""
         if isinstance(other, Atom):
-            return (self.id, self.altloc) != (other.id, other.altloc)
+            return  self.full_id != other.full_id
```
peterjc: Two spaces after ``return``, should be one - that's giving this error on the style checks:

```
Bio/PDB/Atom.py:88:19: E271 multiple spaces after keyword
```

schaarj:
@peterjc Thank you for the notifications and apologies for not catching these style violations in the first place. I indeed missed the other two issues between the full stop warnings. Hopefully everything is in order now.

peterjc:
We currently have to run the style checks on ``Bio/`` and ``BioSQL`` separately from those on ``Tests/`` (due to different stringencies), anyway - a few more little things to fix:

```
$ flake8 Tests/
Tests/test_PDB.py:557:43: E203 whitespace before ','
Tests/test_PDB.py:557:44: E231 missing whitespace after ','
Tests/test_PDB.py:557:48: E203 whitespace before ','
Tests/test_PDB.py:579:32: E226 missing whitespace around arithmetic operator
Tests/test_PDB.py:607:9: E303 too many blank lines (2)
```

schaarj:
Already took care of it when I saw the tests failing. Now the Basic tests passed. Since the unit tests (without dependencies) passed in python 2 and 3 on my machine I'm optimistic that the other tests will pass as well.

peterjc:
Some real failures in ``test_mmtf.py``, Python 2.7:

```
FAIL: test_4CUP (test_mmtf.ParseMMTF)
Compare parsing 4CUP.mmtf and 4CUP.cif
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 78, in test_4CUP
    self.check_mmtf_vs_cif("PDB/4CUP.mmtf", "PDB/4CUP.cif")
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 62, in check_mmtf_vs_cif
    self.check_atoms()
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 33, in check_atoms
    mmcif_atom.full_id)  # (structure id, model id, chain id, residue id, atom id)
AssertionError: Tuples differ: ('4CUP', 0, u'A', (' ', 1856, ... != ('example', 0, 'A', (' ', 1856...
First differing element 0:
'4CUP'
'example'
- ('4CUP', 0, u'A', (' ', 1856, ' '), ('N', ' '))
?   ^^^^      -
+ ('example', 0, 'A', (' ', 1856, ' '), ('N', ' '))
?   ^^^^^^^
```

Python 3.6,

```
FAIL: test_4CUP (test_mmtf.ParseMMTF)
Compare parsing 4CUP.mmtf and 4CUP.cif
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 78, in test_4CUP
    self.check_mmtf_vs_cif("PDB/4CUP.mmtf", "PDB/4CUP.cif")
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 62, in check_mmtf_vs_cif
    self.check_atoms()
  File "/home/travis/build/biopython/biopython/Tests/test_mmtf.py", line 33, in check_atoms
    mmcif_atom.full_id)  # (structure id, model id, chain id, residue id, atom id)
AssertionError: Tuples differ: ('4CUP', 0, 'A', (' ', 1856, ' '), ('N', ' ')) != ('example', 0, 'A', (' ', 1856, ' '), ('N', ' '))
First differing element 0:
'4CUP'
'example'
- ('4CUP', 0, 'A', (' ', 1856, ' '), ('N', ' '))
?   ^^^^
+ ('example', 0, 'A', (' ', 1856, ' '), ('N', ' '))
?   ^^^^^^^
```

The other versions of Python are likely also affected, but we try to have a mix of the optional dependencies - these are the only once with ``mmtf-python`` installed under TravisCI/Tox:

https://github.com/biopython/biopython/blob/biopython-170/.travis-tox.ini#L62

schaarj:
Yes, I noticed. Unfortunately test_mmtf (skipped on my system) fails due to the entity ID being set differently when comparing Atoms. I discussed with @JoaoRodrigues  what should be considered in the comparison. For me it makes sense to also consistently check the parent but for Atom and Residues  ignoring the structure ID and maybe also the model ID might be valid as well. Otherwise I can also adapt the test to have matching Structure ID's when comparing the atoms

peterjc:
You should be able to do ``pip install mmtf-python`` to try this locally.

schaarj:
Thanks, just tested locally and noticed that this test was most likely not doing what it thought it was doing  as it compared uninitialised values, I'll submit a fix, once it is passing locally (hopefully without having to fix formatting this time).

peterjc:
I think the required MMTF test change reflects a subtle API change - previously the user-supplied model ID was ignored for atom level comparison, but now it gets checked. I never really understood the intended usage of the user-supplied identifier, so am unsure what to think here.

schaarj:
Yes that's the major issue in this PR, previously Atom comparison only looked at ID and altloc while comparison of other entities was comparing something else. The current implementation of the operators in Entity added in PR #1360  compares also IDs only but changed the behaviour compared to biopython 1.70. So the question is, whether two Entities/Atoms with identical ID but different parents should be considered equal. I modified Atom to be consistent with the other entities but this causes a change in behaviour compared to previous versions, which might be problematic (which is however also true for different behaviour e.g. between residue and atom). The mmtf test was nevertheless flawed as it explicitly stated to compare full IDs (including the structure ID) while it actually tested `None == None`

schaarj:
Regarding the comment "previously the user-supplied model ID was ignored for atom level comparison, but now it gets checked":
Actually commit c8157c6 changed the comparison in Atom and Entity to ignore the structure ID but there is indeed an API change in Atom comparison.

JoaoRodrigues:
If this passes all tests, it should be good to go. Thanks for adding more tests @schaarj , it will really help us writing better stuff in the future.

peterjc:
Good work.

Would the current ``NEWS.rst`` entry need any further updates for this work compared to Biopython 1.70?

Can you add yourself to the ``NEWS.rst`` and ``CONTRIB.rst`` files please (sorted alphabetically).

schaarj:
Great, I updated `NEWS.rst` and `CONTRIB.rst` including a paragraph describing the changes and  mentioning the API change, Hopefully this is what you had in mind. Looking forward  to Version 1.71, Thank you!

JoaoRodrigues:
@peterjc Is there a way to force the tests to run again? Seems like it just timed out ..

peterjc:
Re-running the AppVeyor  pull request test - it seems its all or nothing, unlike TravisCI which lets you re-run a specific job.

https://ci.appveyor.com/project/biopython/biopython/build/0.0.816

IIRC don't AppVeyor didn't GitHub teams for authentication/permissions, so sharing the account is probably the only way to share this ability.

JoaoRodrigues:
@peterjc Thanks, seems all good to go, shall I merge or you do it?

peterjc:
Excellent - I let you merge this @JoaoRodrigues - thank you!

JoaoRodrigues:
Thanks everyone, specially @schaarj  !

PR #1590, opened 2018-04-02
Rephrase acknowledgement by name in PR template

peterjc:
Hopefully this wording is more inclusive/positive, and being practical will result in more people adding their own names as part of a pull request, and thus make dealing with merges easier in the long run.


peterjc:
What do you think @OscarMaestre (and did you want to be retrospectively added to the acknowledgements?)

OscarMaestre:
No, but thank you for taking it into consideration :)

PR #1615, opened 2018-04-16
Fix bug in mmCIF parsing for mid-line comments

StefansM:
Comments in mmCIF files are only started if the "#" character is at the
beginning of a line or preceded by whitespace:
https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax#lex.
This means that an unquoted valid token can contain a "#".

This commit adds a check to the `Bio.PDB.MMCIF2Dict.MMCIF2Dict._splitline`
method to account for this. Previously, a "#" character would always
start a mid-line comment, which broke the parsing of (e.g.) 1EMB at the
line `_entity_src_gen.pdbx_host_org_vector TU#58`. I have also added
some tests to hopefully prevent regressions here.

I introduced this bug in 781441b :)

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -84,7 +84,7 @@ class MMCIF2Dict(dict):
                     quote_open_char = None
                     in_token = False
                     yield line[start_i:i]
-            elif c == "#" and not quote_open_char:
+            elif not quote_open_char and c == "#" and ((i == 0) or (i > 0 and line[i - 1] in self.whitespace_chars)):
                 return
             elif not in_token:
                 in_token = True
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -84,7 +84,10 @@ class MMCIF2Dict(dict):
                     quote_open_char = None
                     in_token = False
                     yield line[start_i:i]
-            elif not quote_open_char and c == "#" and ((i == 0) or (i > 0 and line[i - 1] in self.whitespace_chars)):
+            elif c == "#" and not in_token:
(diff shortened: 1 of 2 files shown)


peterjc:
It looks like you've done a good job with the test coverage 👍

jgreener64:
Nice one @StefansM, this change looks good to me.

Bio/PDB/MMCIF2Dict.py
```diff
@@ -84,7 +84,7 @@ def _splitline(self, line):
                     quote_open_char = None
                     in_token = False
                     yield line[start_i:i]
-            elif c == "#" and not quote_open_char:
```
lennax: The change is well-tested but it'd be nice to have a comment here or even break up the conditional into some descriptive chunks like "at_beginning_of_line".

peterjc:
Additional comments in the code as suggested by @Lennax still welcome, but I'm OK with merging this now anyway.

@StefansM may we list you in the ``NEWS.rst`` file for the upcoming release?

StefansM:
@peterjc Sorry, I forgot about this. I agree that it could benefit from a comment as it is. But with the benefit of fresh eyes I think the nasty conditional can just be replaced with `elif c == "#" and not in_token:`. Unless I've forgotten what I was doing at the time, I think I was just duplicating the work that the lexer is already doing.

peterjc:
That does look much cleaner, and and extra test too. Nice.

You are not in the ``NEWS.rst`` contributors list for the upcoming Biopython 1.72 release, so if you want to add that here, that'd be great (as I can merge this from GitHub without needing the command line).

Leaving this open for a day or so in case there are any further comments.

peterjc:
Thank you!

PR #1619, opened 2018-04-18
Preserve on-disk order in SeqIO & SearchIO indexes

peterjc:
Since the standard dict object provided in Python 3.6 onwards preserves order, the in-memory dictionary-like indexing in SeqIO and SearchIO will now preserve record order there.

This pull request generalises this to apply to on older versions of Python, and for the SQLite based on-disk dictionary-like indexes.

This means looping over the index will load the records in the on-disk order, which will be much faster (previously it would be effectively at random, based on the key hash sorting).

Essentially this works by explicitly using ``OrderedDict`` on Python 3.5 or older for the in-memory indexes, and adding explicit sorting to the SQL command for the on-disk SQLite indexes.

Potentially the SQL schema could add an index to the offset column to speed this up further?

--


peterjc:
@bow do you have a favourite multi-record SearchIO file I should use for a unit test on the order?

bow:
@peterjc, the `xml_2226_tblastn_001.xml` is one you can use, but IIRC anything annotated with 'multiple queries' in `Tests/Blast/README.txt` or `Tests/Hmmer/README.txt` should be usable here.

peterjc:
Thanks.

I've not run any detailed timings yet with an extra index, but that seems prudent.

As to the Python 3.6 case, given we only care about CPython and PyPy, these _should_ all be ordered. However, yes, we can't promise this on Jython or IronPython. Do you have a better wording suggestion and/or check in the if-statement for this?

Finally, with this work in place, it ought to be possible to expand the dictionary-like interface to also accept integer indexes (0, 1, 2, ..., n-1 for a file with n records). I can see this being useful, although have not thought about exactly what this mean for implementation on the SQLite side.

bow:
For the wording, how about:

```
Since Python 3.7, the default dict class maintains key order, meaning
this dictionary will reflect the order of records given to it. For CPython,
this was already implemented in 3.6.

As of Biopython 1.72, we explicitly use OrderedDict for CPython older than
3.6 (and for other Python older than 3.7) so that you can always assume the
record order is preserved.
```

And the `if` block check can include the [`platform.get_implementation()`](https://docs.python.org/3.6/library/platform.html#platform.python_implementation) check:
```
# Ordered dict is a language feature from Python 3.7 onwards.
# CPython 3.6 also already has ordered dicts.
if sys.version_info > (3, 7) or (sys.version_info == (3, 6) and platform.python_implementation() == "CPython"):
    _dict = dict
else:
    from collections import OrderedDict as _dict
```

peterjc:
I forgot about this - I'll rebase it now to deal with the minor merge conflicts, and say Biopython 1.73 not 1.72 as we missed the last release.

Issue #1624, opened 2018-04-19
SeqIO.write(..., 'fasta') gives error with SeqRecord description that contains only whitespace


aowi:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

```
>>> import sys; print(sys.version)
3.4.3 (default, Nov 28 2017, 16:41:13) 
[GCC 4.8.4]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-3.19.0-77-generic-x86_64-with-Ubuntu-14.04-trusty
>>> import Bio; print(Bio.__version__)
1.70
```

### Expected behaviour

Biopython creates my fasta file

### Actual behaviour

Biopython gives a stacktrace

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/z/datalake/virtual_envs/20180419T162906/lib/python3.4/site-packages/Bio/SeqIO/__init__.py", line 491, in write
    count = writer_class(fp).write_file(sequences)
  File "/z/datalake/virtual_envs/20180419T162906/lib/python3.4/site-packages/Bio/SeqIO/Interfaces.py", line 214, in write_file
    count = self.write_records(records)
  File "/z/datalake/virtual_envs/20180419T162906/lib/python3.4/site-packages/Bio/SeqIO/Interfaces.py", line 199, in write_records
    self.write_record(record)
  File "/z/datalake/virtual_envs/20180419T162906/lib/python3.4/site-packages/Bio/SeqIO/FastaIO.py", line 190, in write_record
    if description and description.split(None, 1)[0] == id:
IndexError: list index out of range
```

### Steps to reproduce

```
>>> from Bio import SeqIO, SeqRecord, Seq
>>> SeqIO.write([SeqRecord.SeqRecord(seq=Seq.Seq('MMMM'), id='FOO', description=' ')], '/tmp/wat.fasta', 'fasta')
```cd

peterjc:
Wow. That code has been in use for years without anyone finding that bug - the FASTA writer does not like the description to be white space only. The short fix on your end would be to add ``.strip()`` or similar to your code where you create the record.

Or our end, I think it would be practical to do something similar - or do you actually want a trailing space in the output?

peterjc:
[...]

PR #1625, opened 2018-04-20
fix handling of whitespace-only SeqRecord descriptions

aowi:
This pull request addresses issue #1624 

### DIFF
--- a/Bio/AlignIO/StockholmIO.py
+++ b/Bio/AlignIO/StockholmIO.py
@@ -454,7 +454,16 @@ class StockholmIterator(AlignmentIterator):
                 elif line[:5] == '#=GS ':
                     # Generic per-Sequence annotation, free text
                     # Format: "#=GS <seqname> <feature> <free text>"
-                    seq_id, feature, text = line[5:].strip().split(None, 2)
+                    parts = line[5:].strip().split(None, 2)
+                    if len(parts) not in (2, 3):
+                        raise ValueError(
+                            "Could not split line in seqname, "
+                            "feature, and text:\n" + line)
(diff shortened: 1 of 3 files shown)


peterjc:
There seems to be an unrelated stockholm fix on this pull request for SeqRecord descriptions?

aowi:
It isn't unrelated. Adding the test-case with a whitespace description broke the Stockholm test as mentioned since it also doesn't handle this case. This was my attempt at doing something sensible, but I'm not sure at all if it's correct.

Alternatively, I could just make the test accept the exception generated by the stockholm writer. The underlying bug would just remain for somebody else to stumble upon.

peterjc:
Ah. OK. thanks for the clarification - that does make sense now, SeqIO and AlignIO are somewhat inter-linked.

Bio/AlignIO/StockholmIO.py
```diff
@@ -454,7 +454,16 @@ def __next__(self):
                 elif line[:5] == '#=GS ':
                     # Generic per-Sequence annotation, free text
                     # Format: "#=GS <seqname> <feature> <free text>"
-                    seq_id, feature, text = line[5:].strip().split(None, 2)
```
peterjc: Taking the specification literally, the free text could be spaces? If so, does this cause problems elsewhere?

```
seq_id, feature, text = line[5:].split(None, 2)
```

Also, should we not make a similar change to the ``"#=GF `` case above?

peterjc:
Apologies that this fell off my to-do list, and there are now multiple conflicts so a rebase wouldn't be trivial. It might be worthwhile revisiting, but I'm closing the stale pull request.

Issue #846, opened 2016-06-09
Enforce --offline switch in run_tests.py


peterjc:
Can we do something like monkey-patching urllib for `run_tests.py --offline` so that attempting use an online resource triggers our own exception making the tests fail? This would have caught #819 much earlier for example, and might help if for example Linux packagers try to run the test suite in a network-less chroot jail or similar.

Gasta88:
Hi,
could you elaborate a bit better the issue ? What do you mean with "monkey-patching"?

peterjc:
Changing a function at run time is often called "monkey patching" in Python, https://stackoverflow.com/questions/5626193/what-is-monkey-patching

We could (within ``run_tests.py``) do something like this to replace the ``urlopen`` function (etc) with a stub which just raise an exception:

```python
from Bio._py3k import urlopen

def dummy_urlopen(...):
    raise RuntimeError("Bad test! You are not allowed to go online!")

urlopen = dummy_urlopen  # replace with error raising stub!
```

I've not tested this - but something like that ought to work.

Gasta88:
Thanks, I wasn't sure that "monkey patching" was a real thing :)

I'll take a look at this.

peterjc:
My idea was if ``run_tests.py`` is called with ``--offline``, then it should do the monkey patching to disable ``urlopen`` (etc).

Any test known to need the network should be using this code snippet, which means it get skipped if ``--offline`` is used:

```python
import requires_internet
requires_internet.check()
```

Any test using the network anyway would then fail (because of the altered ``urlopen``), so we'd know about this problem, and add that snippet of code marking the test as needing to go online.

peterjc:
Closed via #1643, thank you.

PR #1643, opened 2018-05-09
Monkey patch to enforce --offline switch in run_tests.py

Gasta88:
This pull request addresses issue #846

This is far from being the final version because I'm not sure about the end result.

Monkey patching `urlopen()` throws a `MissingExternalDependencyError` each time the method is called inside the unittest script.

The end result via a `run_tests --offline` call, is a `FAIL` flag associated to the online test.

Is this informative enough or would it be better to have again a message that explain why the test has failed? (ex: like before `skipping. No internet available`)

### DIFF
--- a/Tests/run_tests.py
+++ b/Tests/run_tests.py
@@ -38,7 +38,6 @@ import doctest
 import distutils.util
 import gc
 from io import BytesIO
-from Bio import MissingExternalDependencyError
 
 
 # Note, we want to be able to call run_tests.py BEFORE
@@ -256,11 +255,14 @@ def main(argv):
             return 0
         if opt == "--offline":
             print("Skipping any tests requiring internet access")
+            # This is a bit of a hack...
+            import requires_internet
+            requires_internet.check.available = False
             # Monkey patch for urlopen()
             import Bio._py3k
 
             def dummy_urlopen(url):
-                raise MissingExternalDependencyError("internet not available")
+                raise RuntimeError("internet not available")
 
             Bio._py3k.urlopen = dummy_urlopen
         if opt == "-g" or opt == "--generate":
--- a/Tests/run_tests.py
+++ b/Tests/run_tests.py
@@ -262,7 +262,7 @@ def main(argv):
             import Bio._py3k
 
             def dummy_urlopen(url):
-                raise RuntimeError("internet not available")
+                raise RuntimeError("Internal test suite error, attempting to use internet despite --offline setting")
 
             Bio._py3k.urlopen = dummy_urlopen
         if opt == "-g" or opt == "--generate":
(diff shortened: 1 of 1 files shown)


Tests/run_tests.py
```diff
@@ -254,10 +256,13 @@ def main(argv):
             return 0
         if opt == "--offline":
             print("Skipping any tests requiring internet access")
-            # This is a bit of a hack...
-            import requires_internet
-            requires_internet.check.available = False
```
peterjc: I think leave this in place so that tests can check and abort immediately if we are running offline.

Tests/run_tests.py
```diff
@@ -254,10 +256,13 @@ def main(argv):
...
-            # The check() function should now report internet not available
+            # Monkey patch for urlopen()
+            import Bio._py3k
+
+            def dummy_urlopen(url):
+                raise MissingExternalDependencyError("internet not available")
```
peterjc: I think this should be a runtime error - if this happens it means the test did not respect the ``requires_internet`` which (assuming that bit of code remains above).

Gasta88:
I'm happy to preserve the original check for unit test. I might have misunderstood but, is the monkey patch an added level of security here?

I have a questions: most of the online tests alias `urlopen` as `_urlopen`. Would that bypass the patch since the masked function is `Bio.py3k.urlopen`?

Running the edited `run_test.py --offline` I get back the original messages for online unit tests. How can I make sure that the patch works?

peterjc:
Yes, my thinking here was to detect any test files using the network, but not doing the ``requires_internet`` check first.

I think the "import as" thing should be fine.

The simplest test of that would be to (temporarily) deliberately remove the ``requires_internet`` check from one of the test files, and confirm the new error is triggered.

Gasta88:
Thanks, it is clearer now.
Unfortunately my finger slipped and I pushed the wrong edit into the PR. (I'm currently learning how to use VCS inside PyCharm).

I need to re-edit the code to place the `requires_internet` part after the monkey patch.

Gasta88:
OK, I did a couple of more tests and it looks ok on my end.

Tests/run_tests.py
```diff
@@ -257,7 +258,13 @@ def main(argv):
             # This is a bit of a hack...
             import requires_internet
             requires_internet.check.available = False
-            # The check() function should now report internet not available
+            # Monkey patch for urlopen()
+            import Bio._py3k
+
+            def dummy_urlopen(url):
+                raise RuntimeError("internet not available")
```
peterjc: Can you make that error message something like: ``Internal test suite error, attempting to use internet despite --offline setting.``

Issue #1651, opened 2018-05-14
Avoid single quotes in Python examples in Tutorial


peterjc:
Currently our Tutorial is written in LaTeX to produce HTML and PDF output.

The code examples using the double-quote are fine in both HTML and PDF.

The code examples using the single-quote work fine in HTML, but pdflatex replaces them with a pretty quote symbol in the PDF output. This breaks when copy-and-pasted into Python, e.g. #1647 and pull request #1648 for SearchIO.

Long term we'd like to move away from LaTeX (see #907), but in the short term any Python snippet the user would type should avoid the single quotes in favour of double quotes.

We could enforce this in the TravisCI/Tox style checks with grep?

peterjc:
This grep command seems to catch a lot of examples, looking for anything with the leading ``>>> `` indicating a docstyle style Python snippet:

```
$ grep ">>> .*\'" Doc/Tutorial/*.tex
Doc/Tutorial/chapter_advanced.tex:>>> ftab = FreqTable.read_count(open('myCountFile'))
Doc/Tutorial/chapter_advanced.tex:>>> ftab = FreqTable.read_frequency(open('myFrequencyFile'))
Doc/Tutorial/chapter_appendix.tex:>>> my_info = 'A string\n with multiple lines.'
Doc/Tutorial/chapter_cluster.tex:>>> genetree = record.treecluster(method='s')
Doc/Tutorial/chapter_cluster.tex:>>> exptree = record.treecluster(dist='u', transpose=1)
Doc/Tutorial/chapter_entrez.tex:>>> handle = Entrez.esearch(db="nlmcatalog", term="computational[Journal]", retmax='20')
Doc/Tutorial/chapter_entrez.tex:>>> print("The first 20 are\n{}".format(record['IdList']))
Doc/Tutorial/chapter_entrez.tex:>>> info = record[0]['TitleMainList'][0]
Doc/Tutorial/chapter_entrez.tex:>>> for record in records['PubmedArticle']:
Doc/Tutorial/chapter_kegg.tex:>>> open("ec_5.4.2.2.txt", 'w').write(request.read())
Doc/Tutorial/chapter_motifs.tex:>>> m.counts['A']
Doc/Tutorial/chapter_motifs.tex:>>> m.counts['T', 0]
Doc/Tutorial/chapter_motifs.tex:>>> m.counts['T', 2]
Doc/Tutorial/chapter_motifs.tex:>>> m.counts['T', 3]
[...]

PR #1668, opened 2018-05-31
Change single to double quotes on Doc/Tutorial

Gasta88:
This pull request addresses issue #1651

Some single quotes were used for Regex function (ex: `re.findall` in `Doc/Tutorial/uniprot.tex`, line 406 and 408).
I left them like so in case the function was compromised.


peterjc:
That looks good - and you are right to be cautious about the regular expressions.

I expect the automated tests to all pass...

peterjc:
Thank you

PR #1678, opened 2018-06-05
PyPI metadata - HTTPS for homepage; drop download_url

peterjc:
I noticed this since on the recently refreshed PyPI pages the left hand column of https://pypi.org/project/biopython/ has prominent entries under "Project Links" for:

* Homepage - http://www.biopython.org/
* Download - http://biopython.org/DIST/

The homepage now supports HTTPS (https://github.com/biopython/biopython.github.io/issues/133), so we might as well update that link.

The download URL "works" but is not helpful.

Since moving the webserver to GitHub pages we have not had a traditional directory contents listing
on http://biopython.org/DIST/ and in any case all the documentation on this parameter suggests it
should point at the specific version's source code, typically as a tar-ball.

Nowadays with installation from PyPI via pip, the wheel or tar-ball will be hosted on PyPI anyway, although we have continued to host the tar-ball and zip files on our website under http://biopython.org/DIST/ as well.

See also:
https://www.python.org/dev/peps/pep-0314/#download-url
https://docs.python.org/3.1/whatsnew/2.5.html#pep-314-metadata-for-python-software-packages-v1-1

--


peterjc:
I've requested reviews from a few of you who have other projects that you maintain on PyPI for a second opinion.

Issue #1682, opened 2018-06-07
about NCBIWWW.qblast connection


voidxb:
### A issue

I am reporting a problem about NCBIWWW in Bio.Blast

```python
previous = time.time()
    while True:
        current = time.time()
        wait = previous + delay - current
        if wait > 0:
            time.sleep(wait)
            previous = current + wait
        else:
            previous = current
        if delay + .5 * delay <= 120:
            delay += .5 * delay
        else:
            delay = 120

        request = _Request(url_base,
                           message,
                           {"User-Agent": "BiopythonClient"})
        handle = _urlopen(request,context=context)
        results = _as_string(handle.read())

        # Can see an "\n\n" page while results are in progress,
        # if so just wait a bit longer...
        if results == "\n\n":
            continue
        # XML results don't have the Status tag when finished
        if "Status=" not in results:
            break
        i = results.index("Status=")
        j = results.index("\n", i)
        status = results[i + len("Status="):j].strip()
        if status.upper() == "READY":
            break
```

I think 120s is too long....I change it to 20s and it can work well.

peterjc:
Are you saying if we wait 120s (2 minutes) BLAST fails (perhaps the NCBI drops the connection)?

The current value has literally worked for years, see 950696fb7662b924e0e8c9c1795ebaffc74bf489, but perhaps there has been a recent NCBI change.

Do you have more information? e.g. time of day (and timezone)? This might only happen during USA working hours (i.e. peak usage times)

voidxb:
Emm，I 'm in China.When the value is 120s,I always wait for  connecting success for a long time (>>120s),which I can not bear.Then I change it to 20s........Ok，work well.It could prove to connect NCBI in about one minute.

peterjc:
So the current code does work for you, it is just taking a very long time?

I decided to try to see if there was any specific for the NCBI documentation on this, and found as often seems to happen that the old pages I was looking for seem to have been replaced.

However, https://blast.ncbi.nlm.nih.gov/Blast.cgi?CMD=Web&PAGE_TYPE=BlastDocs&DOC_TYPE=DeveloperInfo seems to cover it:

> Usage Guidelines
[...]

PR #1684, opened 2018-06-12
Update remote BLAST timings (less frequent)

peterjc:
This pull request addresses issue #1682, although rather than relaxing our historic limits this does the opposite and is now stricter.

### DIFF
--- a/Bio/Blast/NCBIWWW.py
+++ b/Bio/Blast/NCBIWWW.py
@@ -168,9 +168,10 @@ def qblast(program, database, sequence, url_base=NCBI_BLAST_URL,
     # 4. Run scripts weekends or between 9 pm and 5 am Eastern time
     #    on weekdays if more than 50 searches will be submitted.
     # --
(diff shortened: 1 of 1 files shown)


peterjc:
Not sure who is best to review this, maybe @chapmanb or @mdehoon?

chapmanb:
Peter -- looks good to me. If you wanted to make it even simpler you could do a 20 second query (for quick results, based on #1682) then fall back to polling every minute. That complies with NCBI's request, makes the code simpler, and might help some with responsiveness. I'm sure the initial 2 minutes was arbitrary. Either way is cool and thanks for tackling this.

peterjc:
Interesting idea Brad, yes, starting with a 20s wait might overall be faster than an optimistic 10s gap which might rarely succeed first time, thus always taking at least 70s.

peterjc:
The automated tests broke, this was a mistake on the master, already fixed.

Issue #1741, opened 2018-07-23
Loading .ab1 files fails if the sample_id is not defined. Maybe define sample_id by default to some value?


ananswam:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

3.5.5 |Anaconda custom (64-bit)| (default, Apr 26 2018, 08:11:22) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
CPython
Darwin-15.4.0-x86_64-i386-64bit
1.70

### Expected behaviour

When reading in Sanger files (.ab1), I get an error that says local variable sequence_id referenced before assignment.

I was able to get around this by manually setting sample_id as follows in AbiIO.py:

```
if 'sample_id' not in locals():
    sample_id = 'NULL'
```

### Actual behaviour

This is what I was gettign before making a change.

```In [3]: a = SeqIO.read(open('EX25-1-A11-Sanger-For_A11.ab1','rb'),'abi')
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-3-c58003d62089> in <module>()
----> 1 a = SeqIO.read(open('EX25-1-A11-Sanger-For_A11.ab1','rb'),'abi')

~/anaconda3/lib/python3.5/site-packages/Bio/SeqIO/__init__.py in read(handle, format, alphabet)
    680     iterator = parse(handle, format, alphabet)
    681     try:
--> 682         first = next(iterator)
    683     except StopIteration:
    684         first = None

~/anaconda3/lib/python3.5/site-packages/Bio/SeqIO/__init__.py in parse(handle, format, alphabet)
    615             raise ValueError("Unknown format '%s'" % format)
    616         # This imposes some overhead... wait until we drop Python 2.4 to fix it
--> 617         for r in i:
    618             yield r
    619 

~/anaconda3/lib/python3.5/site-packages/Bio/SeqIO/AbiIO.py in AbiIterator(handle, alphabet, trim)
    424             file_name = ""
    425         record = SeqRecord(Seq(seq, alphabet),
--> 426                            id=sample_id, name=file_name,
    427                            description='',
    428                            annotations=annot,

UnboundLocalError: local variable 'sample_id' referenced before assignment
```

### Steps to reproduce

[...]

PR #1743, opened 2018-07-25
Define a default sample_id in AbiIO

bow:
This pull request addresses issue #1741. A little surprising (but not completely unexpected), the issue is caused by an ABI file without any sample IDs present. This PR simply defines a default `sample_id` value so that the variable will always be defined and ready for use in creating the `SeqRecord` object.

There are still other values (e.g. PBAS2 for the sequence and PCON2 for the qualities) that we assume always exist (and thus will raise a similar `UnboundLocalError` if not present). But if files without these tags do exist, I would like to see what they look like first before trying to generalize here. I think it is much less likely for an ABI file to exist without these tags, as the sequence and its quality is the essential information that should be present in the file.

Of course the file may be malformed and the tags may be missing then. But then we should raise the appropriate error, instead of still returning an object (like we do here).

Anyway, comments welcome as usual. If there are no objections, I will merge this no sooner than 26/7 8 AM CEST.


peterjc:
Spaces in the ``SeqRecord`` object's `.id`` are bad idea for interconversion to other formats, so I don't like this default value, but ``"<unknown id>"`` is the class' default so using it here is reasonable.

Do you have a real world test case we can include for this fix?

bow:
Hmm..indeed it is not preferable. I don't mind using `""`, though I was trying to follow what other SeqIO parsers are doing with  `<unknown id>`.

In any case, I have to ask for @ananswam for permission to use the test case he sent me. Could you let us know whether we can add this to our test suite, to prevent regressions in the future?

bow:
@ananswam Thank you!
@peterjc The PR should be updated with test case soon :).

peterjc:
Looks good - feel free to merge this, and to add update the names in NEWS.rst 👍

Issue #1731, opened 2018-07-10
Addition of a `Seq.join()` method


widdowquinn:
### Setup

I would like to be able to use the string-like concatenation/addition behaviour of `Seq` objects in the same way as using `'some_string'.join()`, but this behaviour is not currently available.

### Expected behaviour

```python
spacer = Seq('NNNNN')
seqlist = list(SeqIO.parse('myfile.fasta', 'fasta'))
concatenated = spacer.join(seqlist)
```

### Actual behaviour

Not implemented.

I appreciate that this is a minefield for resolving alphabets…

widdowquinn:
Yeah, sorry about the mix-up of classes with `SeqIO.parse()` there… ;)

kaskales:
@peterjc I see this issue is still open. Is it ok if I give it a shot as well?

peterjc:
@kaskales did you have any thoughts on examples like ``Seq("").join(some_dna_seqs)`` vs ``Seq("", some_generic_alphabet).join(some_dna_seqs)`` where in some senses the default/generic alphabet of the empty sequence space is not important for the combined sequence (which in this example might be expected to be DNA from the alphabet attributes set in ``some_dna_seqs``)

peterjc:
Yes, I'm talking about the special case where the spacer is a zero length sequence.

peterjc:
Applied as e856cd3d1896814ed3f83b73391098d539e4ff8c from #1746. Thank you all.

PR #1746, opened 2018-08-01
Added the Seq.Join() method, issue #1731

kaskales:
Added the Seq.Join() method, issue #1731 .

Allows user to concatenate Seq objects with a spacer Seq object.


Bio/Seq.py
```diff
@@ -1259,6 +1260,47 @@ def ungap(self, gap=None):
             raise ValueError("Unexpected gap character, {0!r}".format(gap))
         return Seq(str(self).replace(gap, ""), alpha)
 
+    def join(self, other):
+        """Return a concatination of all sequences in the iterable 'other', spaced with the sequence from self.
```
peterjc: Can you shorten the one line summary to try to fit within 80 characters?

Bio/Seq.py
```diff
@@ -1259,6 +1260,47 @@ def ungap(self, gap=None):
...
+
+        Accepts Seq objects and Strings as objects to be concatinated with the spacer
+
+        >>> spacer = Seq('NNNNN')
+        >>> seqlist = list([Seq("AAA"),Seq("TTT"),Seq("PPP")])
+        >>> concatenated = spacer.join(seqlist)
+        >>> concatenated
+        Seq('AAANNNNNTTTNNNNNPPP')
+
+        Throws error if other is not an iterable and if objects inside of the iterable are not Seq or String objects
```
peterjc: Please line wrap the paragraph, ideally at 78 characters.

Bio/Seq.py
```diff
@@ -1259,6 +1260,47 @@ def ungap(self, gap=None):
...
+        """
+        if not isinstance(other, collections.Iterable):
+            raise ValueError("Input must be an iterable")
+        from Bio.SeqRecord import SeqRecord  # Lazy to avoid circular imports
+        temp_data = ""
+        if self._data == "":  # if the spacer is empty it will initially default to the type of the first sequence
+            a = None
```
peterjc: What if the arguments are all alphabet-less strings? Add ``if a is None: a = self.alphabet`` logic at the end?

Doc/examples/tree1.nwk
```diff
@@ -0,0 +1 @@
+((A:0.10200,B:0.23000):0.06000,C:0.40000):0.00000;
```
peterjc: This file should not be checked it - if you can see which test is creating the file and leaving it behind, fixing that would be a welcome separate pull request.

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,27 @@ def test_init_typeerror(self):
         self.assertRaises(TypeError, Seq, (1066))
         self.assertRaises(TypeError, Seq, (Seq("ACGT", generic_dna)))
 
+    def test_join(self):
+        """Checks if Seq join correctly concatinates sequence with the spacer."""
+        # Only expect it to take Seq objects only!
```
peterjc: No, we also expect strings (although if you mean in this test, make that clearer)

peterjc:
Do you not want to be in the NEWS and CONTRIB listings? If you do, can you think how to make the pull request template clearer that we'd like you to add yourself?

In general the changes look good - although I've not had a chance to test them yet. I would like to see more tests, including the error conditions.

kaskales:
Thank you for the comments, I made changes based on your suggestions and added a few more test conditions.

At this moment I do not wish to have my name listed, I only checked the box as an acknowledgement that I read the option. Thanks for double checking with me.

peterjc:
Would you be happy for add ``.join`` for the ``UnknownSeq``? Even if it just calls the ``Seq`` method that would be useful.

kaskales:
If it calls the `Seq` `.join` method, does that mean the `UnkownSeq` `.join` should accept both `UnkownSeq`  and `Seq `as input? Or You should still be only allowed to use join with objects from the same class?

kaskales:
I've added join() for MutableSeq objects as well and tested the functionality. Implementation was done based on how the __add__ method functions for Unknown and MutableSeq.

peterjc:
Aside from the alphabet special cases with an empty spacer, the behaviour of ``spacer.join([A, B, C, ..., Z])`` should match ``A + spacer + B + spacer + C + spacer + ... + Z`` (and it makes sense to check this explicitly in the tests).

Bio/Seq.py
```diff
@@ -1263,7 +1263,7 @@ def ungap(self, gap=None):
     def join(self, other):
         """Return a merge of the sequences in other, spaced by the sequence from self.
 
-        Accepts Seq objects and Strings as objects to be concatinated with the spacer
+        Accepts all Seq objects and Strings as objects to be concatinated with the spacer
```
peterjc: Typo: concatenated

Bio/Seq.py
```diff
@@ -2527,6 +2527,55 @@ def toseq(self):
         """
         return Seq("".join(self.data), self.alphabet)
 
+    def join(self, other):
+        """Return a merge of the sequences in other, spaced by the sequence from self.
+
+        Accepts all Seq objects and Strings as objects to be concatinated with the spacer
+
+        >>> spacer = MutableSeq('NNNNN')
+        >>> seqlist = list([Seq("AAA"),Seq("TTT"),Seq("PPP")])
```
peterjc: Spaces after the commas to follow PEP8 style please. Also there is no reason to have ``list(...)`` here.

Tests/test_Seq_objs.py
```diff
@@ -789,6 +789,60 @@ def test_join(self):
         self.assertRaises(ValueError, spacer.join, ["ATG", "ATG", 5, "ATG"])
         self.assertEqual(NotImplemented, spacer.join(SeqIO.parse(file, 'fasta')))
 
+    def test_join_UnknownSeq(self):
+        """Checks if Seq join correctly concatinates sequence with the spacer."""
+        # Only expect it to take Seq objects and/or strings in an iterable!
+
+        file = 'Fasta/f003'
```
peterjc: Avoid ``file`` as a variable name - it is the name of a built in function. How about ``filename`` here?

Bio/Seq.py
```diff
@@ -1849,6 +1849,67 @@ def ungap(self, gap=None):
...
+        type_is_unknown = True
+        for c in other:
+            if isinstance(c, SeqRecord):
+                return NotImplemented
+            elif hasattr(c, "alphabet"):
+                if a is None:  # if spacer is empty alphabet defaults to type of the first sequence
+                    a = c.alphabet
+                else:
+                    if not Alphabet._check_type_compatible([a,
+                                                                c.alphabet]):
```
peterjc: This can be one line, can't it?

peterjc:
I wonder if the ``Seq`` object ``.join`` should be using the string's ``.join`` method internally, rather than a for-loop and concatenation. I expect that for large test cases the performance difference would be noticeable (but have not yet tried to measure this).

kaskales:
I tested the difference between using join and concatenation using one of the test arrays multiplied by 3000 to have a large sample pool and ran it X20. I tried two implementations using join. One created an array which appends each string and everything gets joined at the end and the other using join in place of concatenation inside the for-loop. Both are only slightly slower on average, .15-.31seconds, per test run than using normal concatenation. Maybe my implementation was just poor, but it seems to me to be best the way it currently is.

peterjc:
I'm glad you checked the performance. If the difference was small, it is likely that the overhead of the alphabet work is far more significant.

peterjc:
@MarkusPiotrowski would you be interested in testing this code out for a second opinion? Thanks!

peterjc:
There seems to be a lot of code duplication here.

I noticed that other than ``len(self.data)`` vs ``self._data == ""`` (which could be written as ``len(self)`` and handle either), then the ``Seq`` and ``MutableSeq`` new ``.join(...)`` methods would be identical.

Can the ``MutableSeq`` method off load to the ``Seq`` object (even though it does not subclass it), something like this?

```python
    def join(self, others):
        """Return a merge of the sequences in other, spaced by the sequence from self.
        ....
        """
        return Seq.join(self, others)  # Have I got the syntax right here?
```

If that isn't possible then a private module level join function called from both the ``Seq`` and ``MutableSeq`` new ``.join(...)`` methods would solve this instead.

Or, if having the ``MutableSeq`` new ``.join(...)`` methods returning a ``Seq`` is fine, which to me seems acceptable, even simpler:

```python
    def join(self, others):
        """Return a merge of the sequences in other, spaced by the sequence from self.
        ....
        """
        return self.toseq().join(others)
```

kaskales:
I'll have time this weekend to look into it.

Bio/Seq.py
```diff
@@ -2540,7 +2540,7 @@ def join(self, other):
         are not Seq or String objects
         """
         seq_joined = self.toseq().join(other)  # call to Seqs join method and then back to MutableSeq
-        return seq_joined.tomutable()
+        return seq_joined
```
peterjc: The comment about is now out of date

peterjc: Also the doctest would need to be updated as well,

```python
>>> concatenated
Seq('AAANNNNNTTTNNNNNPPP')
```

peterjc:
I can see why the Python 2.7 and PyPy tests fail, the doctest needs updating:

```
FAIL: join (Bio.Seq.MutableSeq)
Doctest: Bio.Seq.MutableSeq.join
  File "/home/travis/pypy-6.0.0-linux_x86_64-portable/lib-python/2.7/doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for Bio.Seq.MutableSeq.join
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/Seq.py", line 2596, in join
File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/Seq.py", line 2604, in Bio.Seq.MutableSeq.join
Failed example:
    concatenated
Expected:
    MutableSeq('AAANNNNNTTTNNNNNPPP')
Got:
    Seq('AAANNNNNTTTNNNNNPPP')
Ran 272 tests in 339.001 seconds
```

Currently the ``Bio.Seq`` docstring tests are being skipped on Python 3,
https://github.com/biopython/biopython/blob/biopython-172/Tests/run_tests.py#L176 - pull request #1813 will fix that.

kaskales:
Thank you for the help, I'm sorry about the errors. It's been a while and I had forgotten the things I needed to check before pushing.

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,165 @@ def test_init_typeerror(self):
         self.assertRaises(TypeError, Seq, (1066))
         self.assertRaises(TypeError, Seq, (Seq("ACGT", generic_dna)))
 
+    def test_join_Seq(self):
+        """Checks if Seq join correctly concatinates sequence with the spacer."""
```
peterjc: typo

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,165 @@ def test_init_typeerror(self):
...
+        # Only expect it to take Seq objects and/or strings in an iterable!
+
+        filename = 'Fasta/f003'
+        seqlist = SeqIO.parse(filename, 'fasta')
```
peterjc: Can you move the file based example into a separate set method, e.g. ``def test_join_with_file(self):``

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,165 @@ def test_init_typeerror(self):
...
+        # seq objects with empty spacer
+        concatenated1 = spacer1.join(record.seq for record in SeqIO.parse(filename, 'fasta'))
+        # strings with empty spacer
+        concatenated2 = spacer1.join(example_strings)
+        # strings and seqs
+        concatenated3 = spacer.join(example_strings_seqs)
+        concatenated4 = spacer1.join(example_strings_seqs)
+        concatenated5 = spacer2.join(example_strings_seqs)
+
+        temp_data = temp_data1 = temp_data2 = temp_data3 = ""
```
peterjc: You have named things ``concatenatedN`` and ``temp_dataN`` but they are out of sync.

Using ``str_concatenatedN`` and ``seq_concatenatedN`` would be much clearer, where we expect ``str_concatenatedN == str(seq_concatenatedN)`` for each ``N``.

Tests/test_Seq_objs.py
peterjc: Also, ``seqlist`` is not a list of ``Seq`` objects. It is a generator yielding ``SeqRecord`` objects. How about:

```python
seq_list = [record.seq for record in SeqIO.parse(filename, 'fasta')]
```

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,165 @@ def test_init_typeerror(self):
...
+
+        for seq in seqlist:
+            temp_data += seq.seq._data + spacer._data
```
peterjc: Using ``seq`` for a ``SeqRecord`` is very confusing, we'd usually use ``record`` as the variable name. Also as noted above, the variable name ``seqlist`` is confusing as it gives ``SeqRecord`` objects not ``Seq`` objects.

Why are you using ``._data`` here? If you want the ``Seq`` object as a string, use ``str(...)``

peterjc:
Hi @kaskales, I'd not looked at this for a while either. I think the tests need some reworking.

Something like this is more what I had in mind:

```python
class StringMethodTests(unittest.TestCase):
    ...
    def test_join(self):
        examples_as_strings = [str(_) for _ in self._examples]
        for seq in self._examples:
            if isinstance(seq, MutableSeq):
                # This does not have a join method 
                continue
            # Can  seq.join(...) cope with strings?
            self.assertEqual(str(seq.join(examples_as_strings)),
                             str(seq).join(examples_as_strings))
             # Can seq.join(...) cope with Seq/UnknownSeq/MutableSeq objects?
            self.assertEqual(str(seq.join(self._examples)),
                             str(seq).join(examples_as_strings))
```

Then ideally some specific examples checking the alphabet behaviour.

Bio/Seq.py
```diff
@@ -1280,6 +1281,54 @@ def ungap(self, gap=None):
...
+        >>> concatenated = spacer.join(seqlist)
+        >>> concatenated
+        Seq('AAANNNNNTTTNNNNNPPP')
+
+        Throws error if other is not an iterable and if objects inside of the iterable
+        are not Seq or String objects
+        """
+        if not isinstance(other, collections.Iterable):  # doesn't detect single strings
+            raise ValueError("Input must be an iterable")
+        if isinstance(other, basestring):
```
peterjc: Should you also check for a single ``Seq`` object or ``MutableSeq`` object here? i.e.

```python
if isinstance(other, (basestring, Seq, MutableSeq)):
```

kaskales: This should already be accounted for in the `if not isinstance(other, collections.Iterable)`, as the test already has a line checking for a ValueError when a Seq object is an input on line 786 of test_Seq_objs.py: 
`self.assertRaises(ValueError, spacer.join, test_seq)`.
I can add another assert for MutableSeq.

peterjc: Ah - I was wrongly assuming the ``Seq`` etc objects would pass the Iterable check. Thanks.

kaskales:
@peterjc I've taken your suggestion and reworked the tests so that they loop through the spacers. At the same time I think I've managed to correct the other issues you have pointed out. I also took the opportunity to make separate ValueError and TypeError tests at the same time as separating out the file tests.

The Travis build seems to be failing on a GenomeDiagram file:
`Bio/Graphics/GenomeDiagram/_Track.py:226:5: F811 redefinition of unused 'range' from line 27`
I am not sure if I inadvertently changed something, but to my knowledge I pulled the most recent BioPython version before I pushed.

peterjc:
Thanks for the reads up about flake8, that's not your fault - the master is broken right now, should be fixed by 78a73b99856b92327ecfef632781f0f5fde0f5d3

peterjc:
The tests look much clearer now, although there are still uses of ``._data`` rather than ``str(...)`` which I'd like you to change.

You also still need to remove the accidentally checked in file ``Doc/examples/tree1.nwk``.

Any change should get re-tested against the updated master with the GenomeDiagram problem fixed.

Bio/Seq.py
```diff
@@ -1280,6 +1281,54 @@ def ungap(self, gap=None):
...
+        from Bio.SeqRecord import SeqRecord  # Lazy to avoid circular imports
+        temp_data = ""
+        # if the spacer is empty it will initially default none
+        if self._data == "":
+            a = None
+        else:
+            a = self.alphabet
+        for c in other:
+            if isinstance(c, SeqRecord):
+                return NotImplemented
```
peterjc: I think you should only use ``return NotImplemented`` when implementing binary special methods.

Should this be ``raise NotImplementedError`` or better ``raise TypeError`` instead?

Tests/test_Seq_objs.py
```diff
@@ -737,6 +737,184 @@ def test_init_typeerror(self):
...
+        # seq objects with empty spacer
+        seq_concatenated1 = spacer1.join(seqlist)
+
+        ref_data = ref_data1 = ""
+        ref_data = str(spacer).join(seqlist_as_strings)
+        ref_data1 = str(spacer1).join(seqlist_as_strings)
+
+        self.assertEqual(str(seq_concatenated), ref_data)
+        self.assertEqual(str(seq_concatenated1), ref_data1)
+        self.assertEqual(NotImplemented, spacer.join(SeqIO.parse(filename, 'fasta')))
```
peterjc: See comment on the implementation, this should probably be ``self.assertRaises(...)`` instead.

Bio/Seq.py
kaskales: In the original issue posted, widdowquinn used 
`seqlist = list(SeqIO.parse('myfile.fasta', 'fasta'))
concatenated = spacer.join(seqlist)` 
in his example usage of the method, so I thought it might be nice to keep a NotImplemented error( will be changed to raise NotImplementedError) instead of TypeError so there is an open door to use SeqRecords in the future.

peterjc: I thought we agreed in the discussion on #1731 that the ``Seq`` & ``SeqRecord`` mixing was unintentional. Using ``NotImplementedError`` suggests we do plan to implement this at some point, and right now I don't think we should. So a ``TypeError`` seems best here.

kaskales: Ok, my error. I changed it to a TypeError.

peterjc:
@MarkusPiotrowski or @mdehoon would you be able to look at this for a second opinion?

There are some minor changes I might still make like shortening the doctest examples, but I think this would benefit from fresh eyes.

kaskales:
In addition to shortening the doctest what other changes did you have in mind?

peterjc:
Doctest example,

```python
        >>> spacer = Seq('NNNNN')
        >>> seqlist = [Seq("AAA"), Seq("TTT"), Seq("PPP")]
        >>> concatenated = spacer.join(seqlist)
        >>> concatenated
        Seq('AAANNNNNTTTNNNNNPPP')
```

Could be:

```python
        >>> Seq('NNNNN').join([Seq("AAA"), Seq("TTT"), Seq("PPP")])
        Seq('AAANNNNNTTTNNNNNPPP')
```

peterjc:
Did we discuss the pros and cons of doing the join internally with string concatenation (as now), versus calling the string join method? I suspect the later will scale better, but have not benchmarked this - and what you have now is a useful first implementation which we can always revise later.

kaskales:
I think the discussion of string concatenation vs an internal string join was already spoken about, though I can't find the relevant comments anymore. I did a short test for a join with 300 sequences and I for the moment I see no differences as the most intensive part in both cases seems to be converting everything into a string from a Seq object for the returned Seq() and iterating the list to make sure everything is an allowed object.

peterjc:
OK, that makes sense - I thought we'd talked about it too, but also couldn't find the discussion.

Bio/Seq.py
```diff
@@ -1280,6 +1281,53 @@ def ungap(self, gap=None):
...
+            elif isinstance(c, basestring):
+                temp_data += str(self) + str(c)
+            else:
+                raise ValueError("Input must be an iterable of Seqs or Strings")
+
+        if a is None:  # arguments are all alphabet-less strings
+            a = self.alphabet
+        if self._data == "":
+            return self.__class__(temp_data, a)
+        return self.__class__(temp_data[len(self._data):], a)  # remove the last addition of the spacer
```
MarkusPiotrowski: In the comment, you mean "remove the **first** addition of the spacer"?

Bio/Seq.py
```diff
@@ -1280,6 +1281,53 @@ def ungap(self, gap=None):
...
+        else:
+            a = self.alphabet
+        for c in other:
+            if isinstance(c, SeqRecord):
+                raise TypeError("Iterable cannot contain SeqRecords")
+            elif hasattr(c, "alphabet"):
+                if a is None:  # if spacer is empty alphabet defaults to type of the first sequence
+                    a = c.alphabet
+                else:
+                    if not Alphabet._check_type_compatible([a, c.alphabet]):
```
MarkusPiotrowski: I found that all these alphabet checks are intrinsicely slow. Also, I would think that often the alphabets are identical. So you can speed this up considerably if you only do the following checks if the alphabets are not identical:
```python
if a != c.alphabet:
    if not Alphabet._ ....
        ...
```
In a test run with 1000 'AAA' seq objects, this check made the method nearly 9 times faster.

Bio/Seq.py
```diff
@@ -1280,6 +1281,53 @@ def ungap(self, gap=None):
...
+                        raise TypeError(
+                            "Incompatible alphabets {0!r} and {1!r}".format(
+                                a, c.alphabet))
+                    a = Alphabet._consensus_alphabet([a, c.alphabet])
+                temp_data += str(self) + str(c)
+            elif isinstance(c, basestring):
+                temp_data += str(self) + str(c)
+            else:
+                raise ValueError("Input must be an iterable of Seqs or Strings")
+
```
MarkusPiotrowski: If you remove line 1319, replace 1321 with ```pass``` and add below 1323:

```python
temp_data = str(self).join([str(z) for z in other])
```
you can gain another 1.6x in speed.

peterjc: Good - I did think that would be quicker, but it was masked by the seq/alphabet overhead until your suggestion for the common case of everything having the same alphabet object.

Bio/Seq.py
```diff
@@ -2452,6 +2558,21 @@ def toseq(self):
...
+        >>> concatenated = MutableSeq('NNNNN').join([Seq("AAA"), Seq("TTT"), Seq("PPP")])
+        >>> concatenated
+        Seq('AAANNNNNTTTNNNNNPPP')
+
+        Throws error if other is not an iterable and if objects inside of the iterable
+        are not Seq or String objects
+        """
+        seq_joined = self.toseq().join(other)  # returns Seq object instead of MutableSeq
+        return seq_joined
```
MarkusPiotrowski: It's not clear to me, why you return a ```Seq``` object here, instead of a ```MutableSeq```. Somewhere in the discussion you said that you tried to be consistent with the ```add``` method (you mean ```__add__```?), but for ```MutableSeq``` ```__add__``` returns a new ```MutableSeq``` object.

peterjc: An earlier iteration did return a ``MutableSeq``, but required a lot of code duplication - returning a ``Seq`` avoided that neatly.

MarkusPiotrowski: But can't you finally make a new ```MutableSeq``` with the joined sequence and alphabet and return this? E.g. using ```tomutable()```. Like in the ```__add__``` method?

On the other hand, I wonder if I would expect the result of a join operation, where the MutableSeq acts "only" as spacer, to be a MutableSeq. So this is probably fine, but should be mentioned in the docstring.

peterjc: You could argue either way on the the return type, personally I think a ``Seq`` is more likely to be useful as when I have needed a ``MutableSeq`` it has only been briefly. The doctest does show a ``Seq``, but you are right, saying so in the text of the docstring would be clearer.

Bio/Seq.py
kaskales: You're correct, I previously misunderstood and tested this idea poorly. I've updated it to perform this join.

peterjc:
The merge conflict is due to another test being added for the MutableSeq object.

If you've never dealt with a conflict before, just leave it as is - I can do the merge manually.

kaskales:
I don't think I understand, where is the merge conflict? GitHub is showing me that I've passed all 4 checks and that there are no conflicts with the base branch. Maybe it's better if I leave this up to you.

peterjc:
I see "This branch cannot be rebased due to conflicts" and the merge buttons are greyed out. You may not see this message because you won't see the merge buttons either.

I pulled your changes into a temporary branch on my fork, and if that passes TravisCI, will merge it:

https://github.com/peterjc/biopython/commit/e856cd3d1896814ed3f83b73391098d539e4ff8c

peterjc:
The conflict itself is from fa9a625f11940ff681bfea12f91c21899a50046d - you both added tests to the class at around line 740 - happily very easy to resolve the conflict as we want both that test and your new tests.

kaskales:
Ok, I see. Thank you!

peterjc:
Do you want to be added to the NEWS.rst or CONTRIB.rst files? Just as "kaskales" is fine if you want to remain pseudonymous.

peterjc:
Applied as e856cd3d1896814ed3f83b73391098d539e4ff8c - thank you!

kaskales:
Thank you! 

I would like to make more contributions so I'll remember to add my name for the next time to the NEWS.rst file.

Issue #1013, opened 2016-11-29
Test suite leaving behind extra files


peterjc:
This came up on pull request #983 where some stray files from the test suite were accidentally checked in. Ideally out test suite should not leave any stray files behind.

Where this is impractical to avoid, they should be added to ``.gitignore`` as in 2b0f33fd4bc775e812b17ad18a3ce5d291e72516 for:

```
Doc/examples/other_trees.nwk
```

We also use the ``.gitignore`` approach for graphical output from test files which ought to be manually checked occasionally, since we have not automated that.

For the Tutorial doctest system via ``Tests/test_Tutorial.py`` perhaps we need a mechanism for removing output files? The doctests are a particular problem because natural examples will produce files in the current directory (here using a temp file is an unwanted distraction). This leads to nasty workarounds like https://github.com/biopython/biopython/blob/biopython-168/Bio/SeqIO/SffIO.py#L229

However, in general the ``Tests/test_*.py`` scripts should clean up after themselves and/or use ``tempfile`` rather than writing output in the current directory. e.g. ``test_TreeConstruction.py`` and ``test_Consensus.py`` needed fixing - see da4ed3831fd3d9fc7cef6c8fdd2f07bc0e689842

Note before closing this issue, someone needs to audit this with a full test-suite run with all the optional dependencies - the command line tool wrappers often generate small test output files.

peterjc:
OK, I've fixed the problems I knew about - help wanted to see what else needs cleaning up.

peterjc:
With #1747 merged, another group of unwanted files from the test suite have been fixed.

Leaving this issue open pending a complete audit.

jvfe:
Don't really know for sure because I can't run all tests locally, but from what I gather PR #2203 closes this issue, @peterjc

peterjc:
Hopefully - I'm working from home and also don't have absolutely everything installed right now. Closing.

PR #1747, opened 2018-08-05
removes files created during run of test_Tutorial.py

kaskales:
This pull request addresses issue discussed in the comments of a previous pull request, 1746.

Removes two from created from chapter_phylo.tex and six from chapter_cluster.tex during the run of test_Tutorial.py

### DIFF
--- a/Tests/test_Tutorial.py
+++ b/Tests/test_Tutorial.py
@@ -259,12 +259,12 @@ class TutorialTestCase(unittest.TestCase):
     def tearDown(self):
         global original_path
         os.chdir(original_path)
-        #  remove files created from chapter_phylo.tex
-        # files don't get created during test with python3.5
+        # files currently don't get created during test with python3.5 and pypy
+        # remove files created from chapter_phylo.tex
         if os.path.exists(os.path.join(tutorial_base, "examples/tree1.nwk")):
             os.remove(os.path.join(tutorial_base, "examples/tree1.nwk"))
             os.remove(os.path.join(tutorial_base, "examples/other_trees.nwk"))
-        #  remove files created from chapter_cluster.tex
+        # remove files created from chapter_cluster.tex
         tutorial_cluster_base = os.path.abspath("../Tests/")
         if os.path.exists(os.path.join(tutorial_cluster_base, "Cluster/cyano_result.atr")):
             os.remove(os.path.join(tutorial_cluster_base, "Cluster/cyano_result.atr"))
--- a/Tests/test_Tutorial.py
+++ b/Tests/test_Tutorial.py
@@ -55,7 +55,6 @@ import unittest
 import doctest
 import os
 import sys
-import platform
 import warnings
 from Bio import BiopythonExperimentalWarning, MissingExternalDependencyError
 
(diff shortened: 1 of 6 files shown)


kaskales:
For some reason the files don't get created while running test_Tutorial.py in python3.5 or pypy, but get created in all the other python versions. I haven't managed to work out this discrepancy.

peterjc:
Most likely you have an optional dependency missing on those versions of Python?

--

Does it look simple to edit the examples in Tutorial itself to clean up after themselves, rather than hard coding special case cleanup in ``test_Tutorial.py``?

We already try to do this in some of the other doctests, e.g. 

https://github.com/biopython/biopython/blob/biopython-172/Bio/SeqIO/QualityIO.py#L1414

However, this can be an unwanted distraction from the flow of the example - so this is a judgement call.

kaskales:
The way I've implemented it currently is that it checks of the files exist first, before deleting them. So no special cases, like which python version is being used, are hard-coded in. 

The only lines of code I could find creating the unnecessary files are in `.tex` documents,` chapter_phylo.tex `and `chapter_cluster.tex` . So I don't think adding commands to delete the files there would look nice.

peterjc:
How about adding a comment to the examples in those ``*.tex`` files that the extra files are deleted in ``test_Tutorial.py``, and the other way round too? This would help with the long term maintenance.

Tests/test_Tutorial.py
```diff
@@ -55,6 +55,7 @@
 import doctest
 import os
 import sys
+import platform
```
peterjc: You don't need this import line.

Tests/test_Tutorial.py
```diff
@@ -258,6 +259,20 @@ def test_doctests(self):
     def tearDown(self):
         global original_path
         os.chdir(original_path)
+        # files currently don't get created during test with python3.5 and pypy
```
peterjc: Right now you assume the unwanted files come in batches, and only test the existence of one file from each group. Would it be cleaner to build a list of unwanted potential files, and then loop over the list deleting any present?

kaskales:
@peterjc I've updates so the files are in lists. Its looks a little cleaner to loop over each file. In addition I keep on getting build errors, on both my pull requests, where it doesn't build because TravisCI couldn't properly install the correct python version : 

```
3.4 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.4.tar.bz2
$ curl -sSf -o python-3.4.tar.bz2 ${archive_url}
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
Unable to download 3.4 archive. The archive may not exist. Please consider a different version.
```
This has happened with Python3.4, pypy, and pypy3 previously for me. I'm not sure if this is something which needs to be noted.

peterjc:
Looks like https://github.com/travis-ci/travis-ci/issues/9744

peterjc:
This would partly address #1013

peterjc:
You also need to remove ``Doc/examples/tree1.nwk`` if present,
https://github.com/biopython/biopython/blob/biopython-172/Doc/Tutorial/chapter_phylo.tex#L284

I noticed this via it accidentally being checked in on #1746

kaskales:
When I run test_Tutorial.py on my computer `Doc/examples/tree1.nwk` does get deleted if it's created. But the file wasn't added to .gitignore yet if that's what you're asking.

peterjc:
Good idea to update the ``.gitignore`` list 👍 

Sorry yes, with your changes ``tree1.nwk`` will get deleted.

peterjc:
Merged, thank you.

kaskales:
Thank you

Issue #1727, opened 2018-07-05
Incremental parsing in Bio.Emboss.PrimerSearch


peterjc:
Migrated from https://redmine.open-bio.org/issues/3057 which I filed on 2010-04-13,

> The Bio.Emboss.PrimerSearch module has a single function "read" which loads and parses an entire output file from the EMBOSS tool primersearch into memory at once, returning what is essentially a dictionary keyed by primer name, with as values lists of amplimer information objects.
>
> Even though this still seems to work with "large" output files for thousands of primer pairs, I think it would be useful to provide an iterator function "parse" returning the amplimers for each primer. The current "read" function could be retained for backward compatibility.
>
> The parsing code itself could be extended to extract information like the forward and reverse primer sequences, where the hit (location and strand) and with how many mismatches. This information is currently all held in a long string.

Issue #1737, opened 2018-07-17
TravisCI: ReportLab 3.5.0 breaks tests under Python 2.7 and 3.5


peterjc:
Noticed on some recent pull requests, e.g. https://github.com/biopython/biopython/pull/1735#issuecomment-405063386

Python 2.7.14,

```
======================================================================
FAIL: test_partial_diagram (test_GenomeDiagram.DiagramTest)
Construct and draw SVG and PDF for just part of a SeqRecord.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/test_GenomeDiagram.py", line 770, in test_partial_diagram
    == gdd.write_to_string('PDF').replace(b"\r\n", b"\n")
AssertionError
----------------------------------------------------------------------
```

and Python 3.5,

```
======================================================================
FAIL: test_partial_diagram (test_GenomeDiagram.DiagramTest)
Construct and draw SVG and PDF for just part of a SeqRecord.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/test_GenomeDiagram.py", line 770, in test_partial_diagram
    == gdd.write_to_string('PDF').replace(b"\r\n", b"\n")
AssertionError
----------------------------------------------------------------------
```

Given the timing it looks like the release of ReportLab 3.5.0 earlier this week is to blame, see https://pypi.org/project/reportlab/#history

e.g. Our weekly TravisCI cron job failed on the previously passing master branch:

https://travis-ci.org/biopython/biopython/builds/404526057

peterjc:
Testing a workaround by pinning ReportLab 3.4
https://travis-ci.org/peterjc/biopython/builds/404681877

Note we currently only install ReportLab under Python 2.7 and 3.5, so that is why only those failed.

rtf-const:
I finally realised the issue.
The assertion that fails is the one that simple checks that PDF file saved in a file binary equals to PDF saved as string. But reportlab add unique ID to each file, value of ID is depending on current timestamp, so the file generated now and after millisecod will be different because of different PDF ID.
But there is a way to switch out generating ID based on timestamp by setting "invariant" variable in reportlab config to true.
I just created the pull request #1753 that passes the tests without holding down reportlab version

peterjc:
Good work @rtf-const :)

peterjc:
This seems to affect AppVeyor as well,

```
======================================================================
[...]

PR #1752, opened 2018-08-10
Issue 1727

rtf-const:
This pull request addresses issue #1727

### DIFF
--- a/Bio/Emboss/PrimerSearch.py
+++ b/Bio/Emboss/PrimerSearch.py
@@ -45,35 +45,48 @@ class OutputRecord(object):
 class Amplifier(object):
     """Represent a single amplification from a primer."""
 
-    def __init__(self):
+    def __init__(self, primer_name):
         """Initialize the class."""
+        self.primer_name = primer_name
         self.hit_info = ""
         self.length = 0
+
+
+def parse(handle):
+    current_aplifier = None
+    current_primer_name = None
+    for line in handle:
+        print('AAAAAAA')
+        if not line.strip():
+            continue
+
+        elif line.startswith("Primer name"):
+            current_primer_name = line.split()[-1]
+        elif line.startswith("Amplimer"):
+            if current_aplifier is not None:
+                current_aplifier.hit_info = current_aplifier.hit_info.rstrip()
+                yield current_aplifier
+            current_aplifier = Amplifier(current_primer_name)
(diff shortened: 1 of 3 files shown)


peterjc:
It looks like your changes also modify how the parsed information is stored - I've not looked at the file format in a while so the details are not fresh in my mind. Perhaps @widdowquinn might have some useful comments?

peterjc:
(And AppVeyor is failing on a ReportLab test, which is strange but should not be related to this work)

rtf-const:
I didn't change file format, I just added one more aplifier in sample file for increase test coverage
AppVeyor fails because of #1737, it is unrelated with this task

peterjc:
I meant you changed the class ``Amplifier`` by adding attributes like ``.forward_seq``, which does not seem to be directly related to iteratively parsing the files.

rtf-const:
It was also in the issue description

> The parsing code itself could be extended to extract information like the forward and reverse primer sequences, where the hit (location and strand) and with how many mismatches. This information is currently all held in a long string.

peterjc:
Sorry, I must have missed that when copy-and-pasting the old issue from RedMine into GitHub.

I'm hoping @widdowquinn can take a look (e.g. see his code in https://github.com/widdowquinn/find_differential_primers which uses this). I've not used PrimerSearch for a long time, so it would take me a while to get back up to speed on this.

rtf-const:
I checked the project mentioned. Actually it currently doesn't use biopython Bio/Emboss/PrimerSearch functions and classes, instead it uses its own classes for represent primer and amplimers info. https://github.com/widdowquinn/find_differential_primers/blob/diagnostic_primers/diagnostic_primers/primersearch.py
I think it is exactly because current imlementation than represents amplimer search info as a plain string is not very useful for any practical usage. 
But there is a main difference in my implementation and @widdowquinn 's version. I return a iterator by amplimers found where primer name is an attrubute. In @widdowquinn implementation every parsed item is an PrimerSearchRecord with a list of amplimers. I have no expertise to say that approach is more reasonable.

peterjc:
@rtf-const An iterator based parser is more flexible - you can call ``list(...)`` to get an in memory list.

You are probably right that @widdowquinn wrote his own code if the Biopython code was not flexible enough. It puts him in a strong position to review this (and he said he hoped to if time allows).

PR #1759, opened 2018-08-20
Add a function for searching UniProt

JureFabjan:
The added function performes a search on UniProt and returns the hits in aspecified format.


peterjc:
This looks reasonable, although it would need some tests and perhaps inclusion in the Tutorial somewhere to help people find it and start using it.

e.g. to start with, add something to ``test_SeqIO_online.py``, and then perhaps look at the style of ``test_TogoWS.py`` for a more in depth set of tests.

Does the UnitProt API documentation talk about the encoding they use? Right now the code will assume the local system's default encoding when turning the bytes received into (unicode) strings.

JureFabjan:
The adding of tests is no problem. I will do it when I resolve the points I bring up in this comment. And of course I will gladly also contribute to the tutorial.

The UniPort API documentation does not talk specifically about the encoding they use. As far as I tested it, the latin-1 encoding works great where the encoding is not pre-specified (where it is utf-8). The further change I would make to the function is in the type of object it returns. The user can specify the format as one of the following: html | tab | xls | fasta | gff | txt | xml | rdf | list | rss. In case of fasta, for example, it would make more sense, if the function returned the results as the Bio.SeqIO.FastaIO.SimpleFastaParser object.

peterjc:
If the data is returned as FASTA, then we still want to get it as a handle. This is the most flexible API, and is consistent with other online resources in Biopython like TogoWS and the NCBI Efetch.

The user can then easily give that handle, to ``SeqIO.parse(...)`` to get ``SeqRecord`` objects, ``Bio.SeqIO.FastaIO.SimpleFastaParser(...)`` to get tuples of strings, write it to a file, or anything else.

JureFabjan:
I see. Then I will not change it anymore. I have added the test in the correct file and created a new pull request for that.

peterjc:
Please include the new test here (and not on a separate pull request like #1763 because the test is not useful on its own - it is part of this enhancement).

Tests/test_SeqIO_online.py
```diff
@@ -95,6 +96,19 @@ def funct(d, f, e, l, c):
     del funct
 del database, formats, entry, length, checksum
 
+class UniProtTests(unittest.TestCase):
+    """Test for UniPort module."""
+
+    def test_search(self):
```
peterjc: This generic UniProt search test would be better in a new file I think, say ``test_UniProt_online.py``.

This file (``test_SeqIO_online.py``) was intended for parsing various sequence files from online resources with Bio.SeqIO, as a test of the parser (e.g. has the file format changed in a new release) and of the online wrapper (has the API changed, or is it offline). Could you use the UniProt FASTA return mode here, and any other sequence formats ``SeqIO`` can parse?

peterjc: That looks good, thanks.

Tests/test_SeqIO_online.py
```diff
@@ -95,6 +96,26 @@ def funct(d, f, e, l, c):
     del funct
 del database, formats, entry, length, checksum
 
+class UniProtTests(unittest.TestCase):
+    """Test for UniPort module."""
```
peterjc: Typo, and maybe say UniProt search here?

JureFabjan:
Indeed. I will update it and also create a separate test file for testing if the function itself works.

Bio/UniProt/__init__.py
```diff
@@ -11,3 +11,33 @@
...
+from Bio._py3k import urlopen as _urlopen
+from Bio._py3k import urlencode as _urlencode
+from Bio._py3k import _binary_to_string_handle
+
+
+def search(text, output_format="tab", sort="score", oragnism="", columns=(),
+           isoform=False, compress=False, offset=0, limit=0):
+    """Perform a query over the UniProt API.
+
+    More at: https://www.uniprot.org/help/api_queries
```
peterjc: We should probably provide a little more help here, specifically that columns should be a list/tuple of strings, and perhaps that isoform and compress are treated as booleans, with offset and limit expected to be integers.

peterjc:
The TravisCI style checks are failing,

```
$ flake8 Tests/
Tests/test_SeqIO_online.py:99:1: E302 expected 2 blank lines, found 1
Tests/test_SeqIO_online.py:112:18: F821 undefined name 'search'
Tests/test_SeqIO_online.py:119:1: E305 expected 2 blank lines after class or function definition, found 1
```

This is easiest to fix locally if you have ``flake8`` installed, see https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

JureFabjan:
Sorry for a bit of a long pause in adapting the pull request.
Now I added a piece of documentation explaining the possibilities for columns and output format (as this also has limitations on the formats, which can be chosen.

rwst:
I would try to fix the codecov fails myself but AFAIK this works only by submitting a PR. @JureFabjan do you agree with me opening a new PR with your commits?

peterjc:
We're not enforcing the CodeCov metrics, but they are a useful guide to when changes need more tests.

peterjc:
There is no merge conflict here (currently), the only issue I can see is the typo UniPort

JureFabjan:
@peterjc do I need to change anything else, or can this branch be merged?

Issue #1701, opened 2018-07-02
BGZF RecursionError


peterjc:
There appears to be an issue in the ``.read()`` and/or ``.readline()`` methods as per these issues using ``Bio/bgzf.py`` within ``pyfaidx``:

https://github.com/mdshw5/pyfaidx/issues/125
https://github.com/mdshw5/pyfaidx/issues/131

Steps to reproduce from @mdshw5 

```
$ wget http://ftp.ensembl.org/pub/current_fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
--2018-01-06 20:21:07--  http://ftp.ensembl.org/pub/current_fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
Resolving ftp.ensembl.org... 193.62.193.8
Connecting to ftp.ensembl.org|193.62.193.8|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 881214396 (840M) [application/x-gzip]
Saving to: 'Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'

Homo_sapiens.GRCh38.dna.primary_assemb 100%[============================================================================>] 840.39M   393KB/s    in 24m 34s 

2018-01-06 20:45:41 (584 KB/s) - 'Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz' saved [881214396/881214396]

$ gunzip Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz 
$ bgzip Homo_sapiens.GRCh38.dna.primary_assembly.fa 
$ python
Python 2.7.10 (default, Feb  7 2017, 00:08:15) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyfaidx import Fasta
>>> pyfaidx_handle = Fasta('Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz')
>>> pyfaidx_handle['12'][111803961:111803962]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Development/pyfaidx/pyfaidx/__init__.py", line 698, in __getitem__
    return self._fa.get_seq(self.name, start + 1, stop)[::step]
  File "/Development/pyfaidx/pyfaidx/__init__.py", line 887, in get_seq
[...]

PR #1767, opened 2018-08-22
BGZF RecursionError

rtf-const:
This pull request addresses issue  #1701 

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -633,59 +633,55 @@ class BgzfReader(object):
         """Read method for the BGZF module."""
         if size < 0:
             raise NotImplementedError("Don't be greedy, that could be massive!")
-        elif size == 0:
-            if self._text:
-                return ""
+
+        result = "" if self._text else b""
+        while size and self._buffer:
+            if self._within_block_offset + size <= len(self._buffer):
+                # This may leave us right at the end of a block
+                # (lazy loading, don't load the next block unless we have too)
+                data = self._buffer[self._within_block_offset:self._within_block_offset + size]
+                self._within_block_offset += size
+                assert data  # Must be at least 1 byte
+                result += data
+                break
             else:
-                return b""
-        elif self._within_block_offset + size <= len(self._buffer):
-            # This may leave us right at the end of a block
-            # (lazy loading, don't load the next block unless we have too)
-            data = self._buffer[self._within_block_offset:self._within_block_offset + size]
-            self._within_block_offset += size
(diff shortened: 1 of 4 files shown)


Tests/test_bgzf.py
```diff
@@ -416,6 +416,29 @@ def test_write_tell(self):
 
         h.close()
 
+    def test_bug_1701(self):
```
peterjc: Can you rename that to something like ``test_many_blocks_in_one_read_call``?

peterjc:
This looks great - the fix is along the lines I was expecting/hoping would work, but you've made a nice reproducible test case which it the bit I was blocked on.

I noticed you haven't added your self to the NEWS and CONTRIB file, and don't put your name on your GitHub profile. If you want to be acknowledged as just @rtf-const that's fine - or leave it as is.

@mdshw5 is it easy for you to test this fix locally with pyfaidx, or would you prefer I merge it first?

mdshw5:
I'll test this locally, but it looks like this will provide a fix for BGZF in pyfaidx, so feel free to merge.

peterjc:
Thanks @mdshw5 - I'll look forward to your test results 👍

Bio/bgzf.py
```diff
@@ -633,59 +633,55 @@ def read(self, size=-1):
...
-                return data + self.read(size)
+
+        result = "" if self._text else b""
+        while size and self._buffer:
+            if self._within_block_offset + size <= len(self._buffer):
+                # This may leave us right at the end of a block
+                # (lazy loading, don't load the next block unless we have too)
+                data = self._buffer[self._within_block_offset:self._within_block_offset + size]
+                self._within_block_offset += size
+                assert data  # Must be at least 1 byte
```
mdshw5: In the future it might be nice to have an `Exception` here in case the assertion is false. Something like `ValueError("Expected data at byte %d." % self.tell())`. For now I think this looks fine.

Tests/test_bgzf.py
mdshw5: Thanks for writing this test. I've had this in the back of my mind for almost a year and always procrastinated. This is really helpful!

rtf-const:
> If you want to be acknowledged as just @rtf-const that's fine - or leave it as is.

No, I have no special issues with that. I added my name to CONTRIB and NEWS and renamed the test method

Bio/bgzf.py
peterjc: Fair point, yes.

peterjc:
Thank you 👍

Issue #1745, opened 2018-07-30
Automatically fix bad origin wrapping locations in GenBank/EMBL?


peterjc:
Given a circular plasmid of say 7000bp, some tools producing GenBank format files are known to represent origin spanning locations as ``6800..100`` rather than the expected more explicit notation ``join(6800..7000,1..100)`` as used by the NCBI GenBank files. Likewise for reverse complement strand origin wrapping.

* Biopython up to 1.70 would parse locations like ``6800..100`` without error or warning, but create inconsistent location objects (e.g. ``len(...)`` would fail).
* Biopython 1.71 would error when parsing these locations due to #1497 which enforces the location object model expectation that start <= end.
* Biopython 1.72 would give a warning when parsing these locations, and leave the problem feature(s) with a location of ``None`` (see pull request #1657). This followed discussion on issues #1608 and other duplicate reports stemming from the same changes.

The complete failure to parse these invalid files in Biopython 1.71 was an unfortunate regression. I believe that the behaviour in Biopython 1.72 is reasonable and consistent with other problematic parsing issues where we do something reasonable and give a warning.

This issue is about if we should go further and recognise these invalid origin wrapping features as a special case, giving a warning, but then automatically upgrade the invalid locations to the presumed intended meaning using as if an explicit join had been used.

Although this has only been reported with GenBank format, EMBL flat files share the same INSDC table format, so the same issue likely applies there too.

Referring to the Zen of Python, https://www.python.org/dev/peps/pep-0020/

> *Special cases aren't special enough to break the rules.*
> *Although practicality beats purity.*
> *Errors should never pass silently.*

There is a case for practicality here - provided the problematic location usage is really widespread. @Zulko is unlikely to be alone in finding this a problem.

nimne:
I think that it is reasonable to throw a warning, then try to fix it. I've done my best to [implement this](https://github.com/biopython/biopython/compare/master...nimne:Origin_wrapping), and it seems like it works well. Are there any unusual situations where a notation like `6800..100` would be ambiguous, or cases where an origin wrap would have anything except the two positions? It seems like the intention is clear enough that an automatic fix should be okay.

Implicitly fixing the poorly formatted origin wrapping should probably only occur when the genome is explicitly defined as circular, although this doesn't address the recent comment in #1608 where the topology is undefined and there is a feature spanning the presumed origin. 

[...]

PR #1770, opened 2018-08-24
Automatically fix bad origin wrapping

nimne:
This pull request addresses issue #1745.

This will attempt to automatically fix (and warn about) incorrectly formatted location for features that span the origin of a circular genome eg `6800..100`. This fix works for both GenBank and EMBL files because the share the same location format.

I've limited this to only fix the formatting in circumstances where the genome is defined as being circular.

### DIFF
--- a/Bio/GenBank/__init__.py
+++ b/Bio/GenBank/__init__.py
@@ -1055,21 +1055,26 @@ class _FeatureConsumer(_BaseGenBankConsumer):
                 # Could be non-integers, more likely bad origin wrapping
                 import warnings
                 from Bio import BiopythonParserWarning
-                print(self._seq_type)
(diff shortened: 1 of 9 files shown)


Bio/GenBank/__init__.py
```diff
@@ -1051,13 +1051,31 @@ def location(self, content):
                 cur_feature.location = SeqFeature.FeatureLocation(int(s) - 1,
                                                                   int(e),
                                                                   strand)
-            except ValueError as e:
+            except ValueError:
```
peterjc: Good change - I can see why single letter variables names are frowned on, here ``e`` is for ``error`` was it was also used just above for ``end``.

PR #1786, opened 2018-09-06
MMCIFIO.set_structure KeyError for pdb_object.level=="R" with chain!='A'

Bernhard10:
chain.id is now a property that updates the parent's child_dict.

Before this PR, using a residue with chain!="A" in `set_structure` caused a KeyError.

- [ x] I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and

- [ x] I have added my name to the alphabetical contributors listings in the files

### DIFF
--- a/Bio/PDB/mmcifio.py
+++ b/Bio/PDB/mmcifio.py
@@ -85,8 +85,8 @@ class MMCIFIO(object):
                             parent_id = pdb_object.parent.id
                             sb.structure[0]['A'].id = parent_id
                         except ValueError:
-                            pass
-                        sb.structure[0]['A'].add(pdb_object)
+                            parent_id="A"
+                        sb.structure[0][parent_id].add(pdb_object)
                     else:
                         # Atom
                         sb.init_residue('DUM', ' ', 1, ' ')
--- a/Bio/PDB/mmcifio.py
+++ b/Bio/PDB/mmcifio.py
@@ -85,7 +85,7 @@ class MMCIFIO(object):
                             parent_id = pdb_object.parent.id
                             sb.structure[0]['A'].id = parent_id
                         except ValueError:
-                            parent_id="A"
+                            parent_id = "A"
                         sb.structure[0][parent_id].add(pdb_object)
                     else:
                         # Atom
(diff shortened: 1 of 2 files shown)


peterjc:
Could you review this please @jgreener64, cross reference #968

jgreener64:
I think this is okay if tests pass.

My only question is should the class from PDBIO that this was copied from change too? https://github.com/biopython/biopython/blob/master/Bio/PDB/PDBIO.py#L139-L140

peterjc:
(The pypy3 test timed out, I restarted it)

jgreener64:
Would you mind adding that change to this PR?

peterjc:
@Bernhard10 are you able to update the class in ``PDBIO.py`` to match?

peterjc:
Good - you've avoided code duplication with subclassing.

Bernhard10:
Now the bug should be fixed for PDBIO and MMCIFIO, for residues and atoms (I had not fixed it before for atoms)

I changed the code to first get the chain_id and then initiate the chain, because I find it easier to read this way.

I also took the set_structure method out and put it into a base-class that provides this function for PDBIO and MMCIFIO, to remove code-duplication.

jgreener64:
That subclassing is definitely a good idea. I'm happy for this to merge if tests pass.

Bio/PDB/PDBIO.py
```diff
@@ -42,7 +42,47 @@ def accept_atom(self, atom):
         return 1
 
 
-class PDBIO(object):
+class _PDBIOBase(object):
+    """
+    Provides the set_structure method for PDBIO and MMCIFIO
+    """
```
peterjc: This is breaking the style tests,

```
$ flake8 Bio/
Bio/PDB/PDBIO.py:45:1: D200 One-line docstring should fit on one line with quotes
Bio/PDB/PDBIO.py:45:1: D204 1 blank line required after class docstring
Bio/PDB/PDBIO.py:45:1: D400 First line should end with a period
```

Use something like this:

```python
class _PDBIOBase(object):
    """Base class to provide the set_structure method for PDBIO and MMCIFIO.""""

```

Issue #1788, opened 2018-09-07
Adding makeblastdb to the BLAST wrappers


nimne:
I'd like to work on adding `makeblastdb` to the BLAST command wrappers. This is to address a comment in #1463 as well as my own needs. As far as I'm aware, I haven't seen anyone else working on this!

Opening this issue as a discussion location separate from #1463

peterjc:
Sounds sensible, thank you for working on this.

peterjc:
Closed via #1800

PR #1800, opened 2018-09-21
Add makeblastdb wrapper

nimne:
This pull request addresses issue #1788, with the purpose of adding a makeblastdb wrapper.

In this pull request I have added ncbi-blast+ to the TravisCI checks. It doesn't seem to add more than a few seconds to the build time, but I'm happy to turn these off if the build time starts to get too high.

When I added the `NcbimakeblastdbCommandline` class, I didn't subclass from `_NcbibaseblastCommandline` because so few of the command line arguments were the same between them. The other option would have been to split `_NcbibaseblastCommandline` into another superseding class with only `-h` `-help` `-version` and `-out`. I thought this was a little more convoluted than not subclassing at all but it doesn't fit with the existing style as nicely, I'm happy to change this if needed.

This should show up in the generated docs, but is there anywhere else that this should be documented? 

### DIFF
--- a/Bio/Blast/Applications.py
+++ b/Bio/Blast/Applications.py
@@ -1341,11 +1341,7 @@ class NcbimakeblastdbCommandline(AbstractCommandline):
         AbstractCommandline.__init__(self, cmd, **kwargs)
 
     def _input_type_checker(command, x):
-        expression = (x == 'asn1_bin' or
-                      x == 'asn1_txt' or
-                      x == 'blastdb' or
-                      x == 'fasta')
-        return expression
+        return x in ('asn1_bin', 'asn1_txt', 'blastdb', 'fasta')
 
     def _validate(self):
         incompatibles = {"mask_id": ["gi_mask"],
--- a/Bio/Blast/Applications.py
+++ b/Bio/Blast/Applications.py
@@ -1235,16 +1235,17 @@ class NcbideltablastCommandline(_Ncbiblast2SeqCommandline):
         _Ncbiblast2SeqCommandline.__init__(self, cmd, **kwargs)
 
 
-class NcbimakeblastdbCommandline(_NcbibaseblastCommandline):
+class NcbimakeblastdbCommandline(AbstractCommandline):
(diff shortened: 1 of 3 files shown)


nimne:
Looks like the codecov report showed some deficiencies. I don't understand the `Bio/GA/Evolver.py` or `Bio/Graphics/BasicChromosome.py` changes, but I can make the `Bio/Blast/Applications.py` coverage better.

peterjc:
Sadly we do have some stochastic test coverage in the GA code and some of the graphics (the later should be easy to fix - rather than a random list for the scatter plot, just hard code it), and as result the CodeCov numbers are often a tiny fraction negative. This is one reason why we don't require the CodeCov checks to pass.

peterjc:
Looking at your coverage output, I'd be happy with it as it is.

Bio/Blast/Applications.py
```diff
@@ -1331,13 +1338,21 @@ def __init__(self, cmd="makeblastdb", **kwargs):
...
         ]
-        _NcbibaseblastCommandline.__init__(self, cmd, **kwargs)
+        AbstractCommandline.__init__(self, cmd, **kwargs)
+
+    def _input_type_checker(command, x):
+        expression = (x == 'asn1_bin' or
+                      x == 'asn1_txt' or
+                      x == 'blastdb' or
+                      x == 'fasta')
+        return expression
```
peterjc: I see you've only moved this old code, but I think we should simplify it to just:

```python
return x in ('asn1_bin', 'asn1_txt', 'blastdb', 'fasta')
```

peterjc:
I am a bit surprised BLAST+ wasn't already being installed on TravisCI - I thought I would have tried that long ago, but maybe not? It does not seem to affect the build run time much, so let's leave it in.

Minor comment on a helper function, otherwise this looks ready to merge. Thank you.

Bio/Blast/Applications.py
nimne: This is much simpler! I like the change.

peterjc:
Merged, thank you!

PR #1808, opened 2018-09-26
Speed up simple FASTA parsers 

chris-rands:
This pull request addresses issue #1805

Basically, the PR replaces `.readline()` calls by iterating over file handles for FASTA parsing with the `SimpleFastaParser` and `FastaTwoLineParser `. This speeds up the parsers, at least on my system, maybe someone else can test this independently also?

Here are timings *before* the PR, suggesting a speed improvement for the new implementation:

```
Timings: 1 large sequence
old simple parser 76.89216220937669
new simple parser 57.97007081564516
regular parser 76.77094562910497

Timings: 1,000,000 small sequences
old simple parser 43.77311799954623
new simple parser 34.3902100995183
regular parser 206.45853164047003

Timings 1,000,000 small sequences, 2 line parsers
old simple two-line parser 43.24465982336551
new simple two-line parser 23.533120669424534
```

Here are timings after the PR, old and new timings should be the same here since the new methods are now implemented (I am not sure why they are not for the 2-line parser case, but there is still a speed improvement):
```
Timings: 1 large sequence
old simple parser 58.63168554287404
new simple parser 58.73430686816573
regular parser 58.55501845944673

Timings: 1,000,000 small sequences
old simple parser 34.61581619922072
new simple parser 34.34436380956322
regular parser 191.3278057463467

Timings 1,000,000 small sequences, 2 line parsers
old simple two-line parser 34.29244772717357
new simple two-line parser 23.71209624875337
```

The code is on pastebin for reproducibility:
https://pastebin.com/MmhBVfFu 

- [X] I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ **AND** the _BSD 3-Clause License_.

- [X] I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

- [X] I have added my name to the alphabetical contributors listings in the files
``NEWS.rst`` and ``CONTRIB.rst`` as part of this pull request, am listed
already, or do not wish to be listed. (*This acknowledgement is optional.*)

PR #1810, opened 2018-09-28
Adding SimpleFastaParser and FastaTwoLineParser unit tests

chris-rands:
Apologies Peter for giving you so many reviewing tasks, but I just wanted to tie up the lose end from PR #1808 while it is fresh in my mind, the extra unit tests.  `SimpleFastaParser` and `FastaTwoLineParser` were previously not tested directly anywhere in the Biopython test suite, this addresses that and includes some non-intuitive edge-cases. Hopefully, it is fairly straightforward.

### DIFF
--- a/Tests/test_SeqIO_FastaIO.py
+++ b/Tests/test_SeqIO_FastaIO.py
@@ -12,6 +12,7 @@ from Bio._py3k import StringIO
 from Bio import SeqIO
 from Bio.SeqIO.FastaIO import FastaIterator
 from Bio.Alphabet import generic_nucleotide, generic_dna
+from Bio.SeqIO.FastaIO import SimpleFastaParser, FastaTwoLineParser
 
 
 def title_to_ids(title):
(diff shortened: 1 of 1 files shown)


chris-rands:
Changed and simplified `FastaTwoLineParser ` to correctly handle the case of where a new-line character is the only thing on the sequence line (it still throws an exception in the absence of a sequence line, unlike `SimpleFastaParser`). Also improved the unit tests. Note that although the code coverage isn't really changing, some of these edge cases were not tested before, not even indirectly. Hopefully this is okay now?

peterjc:
What happened to the exceptions for a two-line-per-record FASTA file ending after the ``>`` line without a sequence line? i.e. Why does the last commit remove these exceptions?

peterjc:
I should have started with: The additional explicit tests are very welcome, thank you.

chris-rands:
Thanks Peter, if I understand you correctly, then I think this case still raises an exception. When the file ends on a title/header line starting with `>`, then I think this should be caught by the final index check:

```
if idx % 2 == 0:  # on a header line
    raise ValueError("Should be at end of file, but line= '{}'".format(line))
```

For example:

```
>>> from Bio._py3k import StringIO
>>> from Bio.SeqIO.FastaIO import FastaTwoLineParser
>>> handle = StringIO(">1")
>>> list(FastaTwoLineParser(handle))
  File "/path/.local/lib/python3.6/site-packages/Bio/SeqIO/FastaIO.py", line 117, in FastaTwoLineParser
    raise ValueError("Should be at end of file, but line= '{}'".format(line))
ValueError: Should be at end of file, but line= '>1'
```

This is tested in the new tests I added for `test_exceptions_FastaTwoLineParser`.

Would it be clearer and more explicit to replace the final index check with this?

```
if idx == -1:
    pass  # empty file
elif idx % 2 == 0:  # on a title line
    assert line[0] == '>', "line[0] != '>' ; this should be impossible!"
    raise ValueError("Missing sequence line at end of file "
                     "if this is strict two-line-per-record FASTA format. "
                     "Have title line '{}'".format(line))
else:
    assert line[0] != '>', "line[0] == '>' ; this should be impossible!"
```

Or did I make a mistake in my logic?

peterjc:
That looks clearer I think (although remove the first assert, just do the raise).

peterjc:
I must have missed this in the original PR, but I focused on the multi-line FASTA parser.

Looking at this test case:

```python
    ins_two_line_edges = [">\nACGT", ">1\n\n",
                          ">1>1\n\n>1\n\n", ""]
    outs_two_line_edges = [[("", "ACGT")], [("1", "")],
                           [("1>1", ""), ("1", "")], []]
...

     def test_regular_FastaTwoLineParser(self):
         """Test regular FastaTwoLineParser cases."""
         for inp, out in zip(self.ins_two_line, self.outs_two_line):
             handle1 = StringIO(inp)
             handle2 = StringIO(inp + '\n')
             self.assertEqual(list(FastaTwoLineParser(handle1)), out)
             self.assertEqual(list(FastaTwoLineParser(handle2)), out)
```

The result of parsing ``inp`` and ``inp + '\n'`` should not in general be the same, but it seems the original parser was also tolerant of a missing new line at the end of the file, or trailing blank line at the end of the file.

Looking at this now:

* ``">\nACGT"`` <-- missing tailing new line, I expect a warning or error
* ``">\nACGT\n"`` <-- good
* ``">1\n\n"`` <- good (zero length sequence)
* ``">1\n\n\n"`` <- bad, extra blank line, expect an error
* ``">1>1\n\n>1\n\n"`` <-- good (two zero length seqs)
* ``">1>1\n\n>1\n\n\n"`` <-- bad, extra blank line, expect an error
* ``""`` <-- good (empty file, no sequences)
* ``"\n"`` <-- bad, single blank line, expect an error

Given I've accidentally set a tolerant precedent (which you have sensibly followed), perhaps we should leave it as is, or at most add warnings for these?

chris-rands:
Thanks, sorry if my tests added some confusion; you quoted the `ins_two_line_edges` list but the tests you quoted are for the regular `ins_two_line`. Note the `test_edgecases_FastaTwoLineParser` does not include `inp + '\n'`. So apart from the  first bullet point `">\nACGT"`, all those tests do perform as you expect (see below). 

For this case where the last line is a valid sequence line but lacks a new-line `\n` character (which also is tested in `test_regular_FastaTwoLineParser` as you said), maybe we should add a warning then, but it's your call of course, what kind of warning would it be?

```
>>> from Bio._py3k import StringIO
>>> from Bio.SeqIO.FastaIO import FastaTwoLineParser
>>> 
>>> #">\nACGT" <-- missing tailing new line, I expect a warning or error
... list(FastaTwoLineParser(StringIO(">\nACGT")))
[('', 'ACGT')]
>>> 
>>> #">\nACGT\n" <-- good
... list(FastaTwoLineParser(StringIO(">\nACGT\n")))
[('', 'ACGT')]
>>> 
>>> #">1\n\n" <- good (zero length sequence)
... list(FastaTwoLineParser(StringIO(">1\n\n")))
[('1', '')]
>>> 
>>> #">1\n\n\n" <- bad, extra blank line, expect an error
... list(FastaTwoLineParser(StringIO(">1\n\n\n")))
ValueError: Expected FASTA record starting with '>' character. Perhaps this file is using FASTA line wrapping? Got: '
'
>>> 
>>> #">1>1\n\n>1\n\n" <-- good (two zero length seqs)
... list(FastaTwoLineParser(StringIO(">1>1\n\n>1\n\n")))
[('1>1', ''), ('1', '')]
>>> 
>>> #">1>1\n\n>1\n\n\n" <-- bad, extra blank line, expect an error
... list(FastaTwoLineParser(StringIO(">1>1\n\n>1\n\n\n")))
ValueError: Expected FASTA record starting with '>' character. Perhaps this file is using FASTA line wrapping? Got: '
'
>>> 
>>> #"" <-- good (empty file, no sequences)
... list(FastaTwoLineParser(StringIO("")))
[]
>>> 
>>> #"\n" <-- bad, single blank line, expect an error
... list(FastaTwoLineParser(StringIO("\n")))
ValueError: Expected FASTA record starting with '>' character. Perhaps this file is using FASTA line wrapping? Got: '
'
```

chris-rands:
Also,  just to be sure, this is only for the `FastaTwoLineParser` yes? The `SimpleFastaParser` (multi-line) currently never raises an error or warning for any inputs and will always treat `inp` and `inp + '\n'` the same I think.

chris-rands:
One last thing, should we add a note to these unit test lines like this?

```
handle2 = StringIO(inp + '\n')  #  This could raise a warning in future? see PR #1810
```

peterjc:
Maybe just open a new issue for a new warning when parsing plain text files without a final new line character? e.g. FASTA, FASTQ, EMBL, GenBank, Swiss, ....

The rational here is files in all those file formats can in principle be concatenated to make a larger multi-record file, but that breaks if a child file's final ``\n`` is missing.

peterjc:
Merged, thank you!

Issue #1719, opened 2018-07-05
Nicer GC and AT content and skew functions


peterjc:
Migrated from https://redmine.open-bio.org/issues/2705 which I filed on 2008-12-09

This bug started out as a discussion on Bug 2671, based on some nucleotide scoring functions in GenomeDiagram which were used for plotting sequence properties along a sequence using a sliding window. The basic underlying functions could make a nice addition under Bio.SeqUtils (rather than hiding them under Bio.Graphics.GenomeDiagram).

In particular, GenomeDiagram's Utilities.py included the following (non-windowed) nucleotide composition functions:

calc_gc_content - returns a float in the range 0 to 1.
calc_at_content - returns a float in the range 0 to 1.
calc_gc_skew - returns a float [*]
calc_at_skew - returns a float [*]

[*] As discussed on Bug 2671, these currently give zero if there is no AT content, which was a reasonable shortcut given these functions were originally used for plotting only. They should instead raise an exception or return None or NaN instead.

Also, as implemented in GenomeDiagram, these functions do not cope with mixed case sequences (easily rectified). Also, for GC and AT content these do not deal with ambiguous nucleotides (where we could follow the existing Bio.SeqUtils convention).

Bio.SeqUtils already has several related functions including:

GC - returns a float (a percentage in the range 0 to 100)
GC123 - returns a tuple of four floats (percentages between 0 and 100)

GC_skew - returns a list of floats using a default window size of 100bp. Gives
a floating point exception if there is no GC content in any window.

Personally I don't like the fact that the existing GC function returns a number
between 0 and 100 (rather than 0 and 1). Leighton agreed.

I don't think the current GC_skew function is intuitive and doesn't cover the
non-windowed use-case where you want the GC_skew of the whole sequence passed
in. This is important if you want to do your own windowing (e.g. comparing GC
skew of individual genes to the whole genome).

Because they differ from the existing Bio.SeqUtils code, I think there is a
case for adding the four non-windowed functions from GenomeDiagram's
Utilities.py under Bio.SeqUtils. Each would take a single argument, a sequence (coping with a string, Seq object or MutableSeq object). I have no particularly strong views on the naming of these functions. Perhaps they could be located under a sub module like Bio.SeqUtils.Nucleotides or Bio.SeqUtils.NucUtils? The existing GC functions in Bio.SeqUtils could be deprecated or at least declared obsolete.

[...]

PR #1829, opened 2018-10-12
New module below SeqUtils with GC-content related functions

dzarmola:
This pull request addresses issue #1719 

### DIFF
--- a/Bio/SeqUtils/NucUtils.py
+++ b/Bio/SeqUtils/NucUtils.py
@@ -18,29 +18,29 @@ __GC_CONTENT_VALUE = {
     "N": 0.5
 }
 __GC_CONTENT_VALUE_AMBG = {
-    "M" : 0.5,
-    "K" : 0.5,
-    "R" : 0.5,
-    "Y" : 0.5,
+    "M": 0.5,
+    "K": 0.5,
+    "R": 0.5,
+    "Y": 0.5,
 
-    "B" : 0.66,
-    "D" : 0.33,
-    "H" : 0.33,
-    "V" : 0.66,
+    "B": 0.66,
+    "D": 0.33,
+    "H": 0.33,
+    "V": 0.66,
 }
 
 
 def calc_gc_content(seq, interpret_all=False):
-    """Calculates G+C content, returns a fraction (float between 0 and 1).
+    """Calculate G+C content, returns a fraction (float between 0 and 1).
 
     Copes mixed case sequences, and with the ambiguous nucleotides S (G or C)
-    and N ( A or C or G or T ) when counting the G and C content.  
+    and N ( A or C or G or T ) when counting the G and C content.
 
     When interpret_all is True, includes probability of other ambiguous
     to be C or G:
-        - for M,K,R,Y it is 0.5
-        - for B,V it is 0.66
-        - for D,H it is 0.33.
+    - for M,K,R,Y it is 0.5
+    - for B,V it is 0.66
+    - for D,H it is 0.33.
 
     The percentage is calculated against the full length, e.g.:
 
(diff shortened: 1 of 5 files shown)


Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+    "D": 0.33,
+    "H": 0.33,
+    "V": 0.66,
+}
+
+
+def calc_gc_content(seq, interpret_all=False):
+    """Calculate G+C content, returns a fraction (float between 0 and 1).
+
+    Copes mixed case sequences, and with the ambiguous nucleotides S (G or C)
```
peterjc: Typo: "Copes with mixed ..."

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+    and N ( A or C or G or T ) when counting the G and C content.
```
peterjc: There meaning for S and N is isn't clear to me as written.

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+    >>> from Bio.SeqUtils.NucUtils import calc_gc_content
+    >>> calc_gc_content("ACTGN")
+    0.5
+
+    >>> from Bio.SeqUtils.NucUtils import calc_gc_content
+    >>> calc_gc_content("ACTGB")
+    0.4
+
+    >>> from Bio.SeqUtils.NucUtils import calc_gc_content
+    >>> calc_gc_content("ACTGB",True)
```
peterjc: Style wise please include a space after the comma (PEP8 guidelines)

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+    0.532
+
+    Note that this will return zero for an empty sequence.
+    """
+    len_seq = len(seq)
+    if not len_seq:
+        return 0.
+
+    gc = sum(__GC_CONTENT_VALUE.get(x.upper(),
```
peterjc: I worry that calling ``x.upper()`` for every letter in a large sequence like an entire chromosome would be far slower than the existing approach in ``Bio/SeqUtils/__init__.py`` which has both cases in the dictionary. Have you benchmarked this?

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
+# Copyright 2018 Aleksandra Jarmolinska.  All rights reserved.
```
peterjc: This should probably acknowledge being based on ``Bio/SeqUtils/___init__.py`` and if used Leighton Pritchard's code in GenomeDiagram's ``Utilities.py``.

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+# This file is part of the Biopython distribution and governed by your
+# choice of the "Biopython License Agreement" or the "BSD 3-Clause License".
+# Please see the LICENSE file that should have been included as part of this
+# package.
+"""Functions for dealing with GC/AT content."""
+
+
+from __future__ import print_function
+
+__GC_CONTENT_VALUE = {
```
peterjc: Why the double leading underscore, isn't the convention one leading underscore for a private variable?

Bio/SeqUtils/NucUtils.py
```diff
@@ -0,0 +1,191 @@
...
+    "M": 0.5,
+    "K": 0.5,
+    "R": 0.5,
+    "Y": 0.5,
+
+    "B": 0.66,
+    "D": 0.33,
+    "H": 0.33,
+    "V": 0.66,
+}
```
peterjc: Here I would be much happier if the magic values here computed on the fly when first used from the ambiguity tables in ``Bio/Data/IUPACData.py`` - or at very least use those values in the unit tests.

This would also fix the rounding issues which would occur by defining ``1/3`` and ``2/3`` to only two decimal plates, which is not enough.

Bio/SeqUtils/NucUtils.py
dzarmola: You are right - I benchmarked it just now, and it increases the times approx. by half. I'm now adding the probabilities calculation based on IUPAC tables, and will correct the dictionaries accordingly.

Bio/SeqUtils/NucUtils.py
dzarmola: I've expanded this a bit

Bio/SeqUtils/NucUtils.py
dzarmola: I wasn't sure about the proper form so added authors of __init__.py as "Based on code by ...". Didn't use GenomeDiagram

Bio/SeqUtils/NucUtils.py
dzarmola: I've modified it so that the values are calculated based on IUPAC tables on module import

Bio/SeqUtils/NucUtils.py
dzarmola: fixed

Bio/SeqUtils/NucUtils.py
dzarmola: I believe PEP8 allows double leading underscore for "very private" variables, and there is some name mangling involved for use in classes which are supposed to be subclassed. Although you are - here the "normal" one underscore is absolutely sufficient

dzarmola:
I didn't take into account, that when some bases are not interpreted, the at and gc contents would not match. I have now fixed this with a separate AT content dictionary computed at import

Issue #1693, opened 2018-06-21
Compatibility with Python 3.7.0


sagitter:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
Python 3.7.0
GCC 8.1.1
Linux 4.18.0-0.rc1 fedora
Biopython 1.71
```

Probably, Biopython is not ready yet for Python 3.7.0 but unfortunately on Fedora 29 (devel branch) we are already using it. 

### Expected behaviour

All tests passed.

### Actual behaviour

At least, `test_AlignIO` ([link](https://kojipkgs.fedoraproject.org//work/tasks/2859/27752859/build.log)) and `test_bgzf` ([link](https://kojipkgs.fedoraproject.org//work/tasks/4520/27764520/build.log) on ARM64) are failing.

### Steps to reproduce

Compiling with Python 3.7.0.

peterjc:
Part of this is ``RuntimeError: generator raised StopIteration`` from ``test_AlignIO``,
https://www.python.org/dev/peps/pep-0479/

The BGZF problem might be ARM specific - I've not run Biopython on an Raspberry Pi recently, but that might be the easiest way for me to reproduce this.

peterjc:
There is an (internal) use of ``StopIteration`` in ``Bio/bgzf.py`` which could be in a round-about-way causing the failing on ARM...

peterjc:
Our tests seem to pass on TravisCI under Python 3.7.0a4+
https://travis-ci.org/peterjc/biopython/jobs/395027004

Which version of Python exactly are you using? Python 3.7.0 is not out yet, just a release candidate, Python 3.7.0rc1 
https://www.python.org/dev/peps/pep-0537/
https://www.python.org/downloads/release/python-370rc1/

peterjc:
The reported AlignIO failure,

```
test_AlignIO ... ERROR
======================================================================
ERROR: test_AlignIO
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-biopython-1.71/python3/build/lib.linux-x86_64-3.7/Bio/AlignIO/NexusIO.py", line 42, in NexusIterator
    n = Nexus.Nexus(handle)
  File "/builddir/build/BUILD/python-biopython-1.71/python3/build/lib.linux-x86_64-3.7/Bio/Nexus/Nexus.py", line 623, in __init__
    self.read(input)
[...]

PR #1635, opened 2018-04-29
Add cif-seqres and cif-atom SeqIO drivers

StefansM:
This commit adds the `cif-seqres` and `cif-atom`
drivers to the Bio.SeqIO module. These are implemented by the
`CifSeqresIterator` and `CifAtomIterator` functions in the
`Bio.SeqIO.CifIO` module.

Most of the logic for the `cif-atom` driver is identical to the
`pdb-atom` driver, as they both read from a Bio.PDB Structure object.
The shared code has been refactored into the `AtomIterator` function in
the `Bio.SeqIO.structure_io` module.

The cif-seqres iterator reads three categories: `_pdbx_poly_seq_scheme`,
which contains the actual sequence; `_struct_ref`, which contains
database cross-references; and `_struct_ref_seq`, which is linked to
_struct_ref via the ref_id field and contains the PDB and chain ID of
the cross-reference.

Since so much of the test code is shared between the pdb-* and cif-*
SeqIO readers, the `test_SeqIO_PdbIO.py` file has been renamed to
`test_SeqIO_StructureIO.py`, and now contains the tests for both the
pdb-* and cif-* readers. For the sake of consistency, I have also added
2XHE.cif and a_structure.cif to the Tests/PDB directory.

Shared test code is implemented in a superclass returned from a
function in order to convince unittest to run the child test suites but
not the parent suite.

This pull request addresses issue #1573.

Comments are very welcome. I haven't marked this as a WIP because it works as-is, but I am expecting some suggestions about the naming conventions.

---

I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ **AND** the _BSD 3-Clause License_.

I have read the ``CONTRIBUTING.rst`` file and understand that AppVeyor and
TravisCI will be used to confirm the Biopython unit tests and ``flake8`` style
checks pass with these changes.

I have added my name to the alphabetical contributors listings in the files
``NEWS.rst`` and ``CONTRIB.rst`` as part of this pull request, am listed
already, or do not wish to be listed.

PR #1835, opened 2018-10-26
Possible fix for Python 3.7 RuntimeError in MMCIF2Dict

chris-rands:
This pull request might address issue #1693 (see also #1635), but it is untested, please don't consider merging until the tests are complete.

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -26,7 +26,10 @@ class MMCIF2Dict(dict):
             loop_flag = False
             key = None
             tokens = self._tokenize(handle)
-            token = next(tokens)
+            try:
+                token = next(tokens)
+            except StopIteration:
+                return  # for Python 3.7 and PEP 479
             self[token[0:5]] = token[5:]
             i = 0
             n = 0


chris-rands:
@peterjc a new failure, now across all Python versions:

```
ERROR: test_SeqIO
  File "run_tests.py", line 459, in runTest
  File "/opt/python/3.6.3/lib/python3.6/unittest/loader.py", line 153, in loadTestsFromName
  File "/home/travis/build/biopython/biopython/Tests/test_SeqIO.py", line 499, in <module>
    records = list(SeqIO.parse(handle, t_format))
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.6/Bio/SeqIO/__init__.py", line 647, in parse
    for r in i:
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.6/Bio/SeqIO/PdbIO.py", line 468, in CifAtomIterator
    struct = MMCIFParser().get_structure(pdb_id, buffer)
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.6/Bio/PDB/MMCIFParser.py", line 64, in get_structure
    self._build_structure(structure_id)
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.6/Bio/PDB/MMCIFParser.py", line 78, in _build_structure
    atom_id_list = mmcif_dict["_atom_site.label_atom_id"]
KeyError: '_atom_site.label_atom_id'
```

It looks like the test failure relates to testing for an [empty handle](https://github.com/biopython/biopython/blob/master/Tests/test_SeqIO.py#L492), but I have no familiarity with the PDB module I'm afraid, and I'm confused as to how the test didn't fail before but does not.

peterjc:
Good question, I don't immediately see it either. Wild guess, might this work instead (``continue`` vs ``return``)?

```python
            try:
                token = next(tokens)
            except StopIteration:
                continue  # for Python 3.7 and PEP 479
```

chris-rands:
`continue` (or `break`) fail since there isn't yet a loop a this point.

[Some formats](https://github.com/biopython/biopython/blob/master/Tests/test_SeqIO.py#L495) are excused from these empty handle tests, I don't know if `cif-atom` (where the test appears to crash) should be in that category. Either way, it doesn't explain why this test didn't fail previously.

jgreener64:
Is the issue on Python 3.7 the call to `next`?

If so, the code could be refactored to remove this call:

```python
tokens = self._tokenize(handle)
i = 0
n = 0
for i, token in enumerate(tokens):
    if i == 0:
        self[token[0:5]] = token[5:]
        continue
    if token.lower() == "loop_":
        ...
```

NB I haven't checked this works.

chris-rands:
I think I mostly understand now. The current implementation actually has a subtle bug, which coincidentally allows it correctly handle the empty file test case!  Consider the following toy example in Py 3.6:

```
class A(list):
    it = iter([0])
    def __init__(self):
        while True:
            self.append(next(self.it))

class B(list):
    it = iter([0])
    def __init__(self):
        while True:
            try:
                self.append(next(self.it))
            except StopIteration:
                return

def wrapper(MyClass):
    lst = MyClass()
    print('lst', lst)
    for item in lst:
        print('iterating')
        yield item

print('A...')
print(list(wrapper(A)))
print('\nB...')
print(list(wrapper(B)))
```

Outputs:

```
A...
[]

B...
lst [0]
iterating
[0]
```

For `A`, the output is wrong, it *always* produces an empty list. However, for the test case we always want an empty iterable here so it works by accident. For `B` the output is correct in the sense that the `wrapper` function doesn't prematurely terminate, but in our test case this fails because no-one thought to handle the empty file case I assume.

So if I'm right then refactoring the code into a `for` loop should lead to the same `KeyError`?  @jgreener64

@peterjc Is that clear? I think the solution is either exclude the relevant formats from the test cases [here](https://github.com/biopython/biopython/blob/master/Tests/test_SeqIO.py#L495) if empty files are not allowed, or by addressing the `KeyError` if empty files are allowed?

peterjc:
That sounds sensible @chris-rands.

@StefansM Should the CIF formats added to SeqIO in #1635 accept empty files as meaning no sequence (like FASTA files), or an error (like an XML format)?

StefansM:
I don't think that empty mmCIF files make any sense. It's been a while since I took a proper look at the spec, but I think valid mmCIF files must contain at least a `data_` block.

I think `MMCIF2Dict` should raise an exception (probably a `ValueError`, or some subclass `EmptyFileEncounteredError` or similar) when it encounters a file with no tokens. While we're at it, I think that `PDBParser` should do the same. In the meantime, adding the `cif-*` parsers to the whitelist in `test_SeqIO` makes sense to me.

jgreener64:
Your analysis seems sound @chris-rands but if `tokens` is empty then the `for` loop never executes, so I don't think it would throw a `KeyError`.

chris-rands:
It does raise a `KeyError` with the `for` loop suggested as the resulting dict is empty, much like [the earlier test failure](https://github.com/biopython/biopython/pull/1835#issuecomment-433393158).

@peterjc @jgreener64 Hopefully the build passes now, but travis isn't checking Python 3.7 for some reason? Is the added commit sufficient? If we want to add a custom traceback for empty files across various formats and appropriate unit tests, they perhaps they could form a separate PR?

jgreener64:
> It does raise a KeyError with the for loop suggested as the resulting dict is empty, much like the earlier test failure.

Ah I see, fair enough.

PR #1843, opened 2018-11-08
Updated docstrings for Bio.SearchIO. Issue #1203

svalqui:
This pull request addresses issue #1203 

### DIFF
--- a/Bio/SearchIO/BlastIO/blast_tab.py
+++ b/Bio/SearchIO/BlastIO/blast_tab.py
@@ -212,6 +212,7 @@ class BlastTabParser(object):
         self.line = self.handle.readline().strip()
 
     def __iter__(self):
+        """Iterate over BlastTabParser, yield query result."""
         # stop iteration if file has no lines
         if not self.line:
             return
--- a/Bio/SearchIO/BlastIO/blast_tab.py
+++ b/Bio/SearchIO/BlastIO/blast_tab.py
@@ -212,7 +212,7 @@ class BlastTabParser(object):
         self.line = self.handle.readline().strip()
 
     def __iter__(self):
-        """Iterate over BlastTabParser, yield query result."""
+        """Iterate over BlastTabParser, yields query results."""
         # stop iteration if file has no lines
         if not self.line:
             return
--- a/Bio/SearchIO/_model/hit.py
+++ b/Bio/SearchIO/_model/hit.py
@@ -147,7 +147,7 @@ class Hit(_BaseSearchObject):
             self.append(hsp)
 
     def __repr__(self):
-        """Return Hit id, query id, and number of hsps."""
+        """Return string representation of Hit object."""
         return "Hit(id=%r, query_id=%r, %r hsps)" % (self.id, self.query_id,
                                                      len(self))
 
@@ -172,7 +172,7 @@ class Hit(_BaseSearchObject):
         return hsp in self._items
 
     def __str__(self):
-        """Return the string of hit and its values recursively."""
+        """Return a human readable summary of the Hit object."""
         lines = []
 
         # set query id line
@@ -236,7 +236,7 @@ class Hit(_BaseSearchObject):
         return '\n'.join(lines)
 
     def __getitem__(self, idx):
-        """Return item of index idx."""
+        """Return the HSP object at the given index."""
         # if key is slice, return a new Hit instance
         if isinstance(idx, slice):
             obj = self.__class__(self.hsps[idx])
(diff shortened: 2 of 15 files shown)


svalqui:
Added docstrings for Bio.SearchIO.

Bio/SearchIO/BlastIO/blast_tab.py
```diff
@@ -212,6 +212,7 @@ def __init__(self, handle, comments=False, fields=_DEFAULT_FIELDS):
         self.line = self.handle.readline().strip()
 
     def __iter__(self):
+        """Iterate over BlastTabParser, yield query result."""
```
peterjc: You've not been consistent with yield/yields, and I would expect results plural given the context. How about "..., yields query results." here (and on the other entries)?

svalqui: @peterjc you are right, amended.

peterjc:
Would you like to merge this @bow, or are there further improvements you'd like to see first? Thanks!

Bio/SearchIO/_model/hit.py
```diff
@@ -147,26 +147,32 @@ def __init__(self, hsps=(), id=None, query_id=None):
             self.append(hsp)
 
     def __repr__(self):
+        """Return Hit id, query id, and number of hsps."""
```
bow: Technically `__repr__` is not returning those values, but rather it returns a string representation of the object.

Bio/SearchIO/_model/hit.py
```diff
@@ -147,26 +147,32 @@ def __init__(self, hsps=(), id=None, query_id=None):
...
 
     # Python 2:
     __nonzero__ = __bool__
 
     def __contains__(self, hsp):
+        """Return True if hsp in items."""
         return hsp in self._items
 
     def __str__(self):
+        """Return the string of hit and its values recursively."""
```
bow: I would rather state the intention of `__str__`, rather than what it does per-se.

For example, this could be something like `Returns a human readable summary of the hit object.`

Bio/SearchIO/_model/hit.py
```diff
@@ -230,6 +236,7 @@ def __str__(self):
         return '\n'.join(lines)
 
     def __getitem__(self, idx):
+        """Return item of index idx."""
```
bow: Same thing here. I would phrase this as `Return the HSP object at the given index`.

Bio/SearchIO/_model/hsp.py
```diff
@@ -1098,12 +1122,14 @@ def _query_span_get(self):
...
 
     def _hit_range_get(self):
+        """Return the start and end of a hit (PRIVATE)."""
         return (self.hit_start, self.hit_end)
 
     hit_range = property(fget=_hit_range_get,
                          doc="Tuple of hit start and end coordinates.")
 
     def _query_range_get(self):
+        """Return the star and end of a query (PRIVATE)."""
```
bow: This should be `start` instead of `star`.

svalqui:
Hi @bow , I have amended it as per your feedback, I think is better, what do you think?

svalqui:
Hi @peterjc , @bow ; reviewing this I noticed that flake8 is not finding some missing docstrings, like on file hsp.py, line 377, def _validate_fragment(self, fragment):; tested with flake8 3.5 and 3.6 is the same passes some missing docstrings; I need few days to review these files and push a complete update.

svalqui:
Yes, this seems related to pydocstyle.

svalqui:
Hi @peterjc ,
I am not sure how to proceed here, I hope you can advise.
PEP257 states all modules and public methods should have docs strings, implying docstrings on private modules, such Bio/SearchIO/_model/ and below the structure, are not needed.

pydocstyle seems to continue parsing for modules and public methods under private modules, but ignoring private methods and fuctions, a current open issue highlights this. which has risen some missing docstrings addressed on this PR, that it should had not.

https://github.com/PyCQA/pydocstyle/issues/323

So the current issue on pydocstyle, makes this PR over docstrings documented, but is just few extra docstrings, I think that should be fine. Provided Bow is okay with the docstrings this should be merged.

The other option will be to add docstrings to all private modules, out of the scope of PEP 257, however there is no tool for that, it requires someone to proof read it all.

What do you think?

peterjc:
In this case, while the objects are defined in private files ``Bio/SearchIO/_model/*.py`` they are exposed as a public API via the imports in ``Bio/SearchIO/__init__.py``, so we do want them all documented:

```python
from Bio.SearchIO._model import QueryResult, Hit, HSP, HSPFragment
```

So in this case, I think pydocstyle is doing the right thing (which is not to say there isn't a bug in the logic it applies).

In any case, adding documentation to our private methods and functions (with a single leading underscore) is still welcome, even if pydocstyle is not checking for their absence.

It looks like you have addressed @bow's comments, so this should be OK to merge now. Let's give Bow a few more just days in case.

bow:
It's merged now, thanks again @svalqui.

Issue #1867, opened 2018-12-03
[Discussion] handle too many request of Entrez


hsiaoyi0504:
### Setup

I am reporting a problem with Biopython 1.72, 3.6, and CentOS 7 as follows:

``` shell
Traceback (most recent call last):
  File "create_fixture.py", line 127, in <module>
    summary_text, chr_id = search(g)
  File "/var/www/html/Melvin/utils/__init__.py", line 11, in search
    handle = Entrez.esummary(db='gene', id=record['IdList'][0])
  File "/var/www/html/Melvin/env/lib64/python3.6/site-packages/Bio/Entrez/__init__.py", line 320, in esummary
    return _open(cgi, variables)
  File "/var/www/html/Melvin/env/lib64/python3.6/site-packages/Bio/Entrez/__init__.py", line 533, in _open
    raise exception
  File "/var/www/html/Melvin/env/lib64/python3.6/site-packages/Bio/Entrez/__init__.py", line 531, in _open
    handle = _urlopen(cgi)
  File "/usr/lib64/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.6/urllib/request.py", line 532, in open
    response = meth(req, response)
  File "/usr/lib64/python3.6/urllib/request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python3.6/urllib/request.py", line 570, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.6/urllib/request.py", line 650, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 429: Too Many Requests
```

### Expected behaviour

Provide a option to avoid this

### Actual behaviour

raise HTTPError

### Steps to reproduce

invoke request frequently

It's a feature request and I think I can contribute this part.
The reason why I open this is just to have a discussion how should I do that (spec ...).

peterjc:
That should not happen - we already try to enforce most of the NCBI usage restrictions - although as documented, not the *For any series of more than 100 requests, do this at weekends or outside USA peak times.* rule. See "Entrez Guidelines" in our tutorial:

http://biopython.org/DIST/docs/tutorial/Tutorial.html
[...]

PR #1869, opened 2018-12-05
increase delay time without API key to 0.37

hsiaoyi0504:
To fix issue #1867.

### DIFF
--- a/Bio/Entrez/__init__.py
+++ b/Bio/Entrez/__init__.py
@@ -520,7 +520,9 @@ def _open(cgi, params=None, post=None, ecitmatch=False):
     # Equivalently, at least a third of second between queries
     params = _construct_params(params)
     options = _encode_options(ecitmatch, params)
-    delay = 0.1 if api_key else 0.333333334
+    # Using just 0.333333334 seconds sometimes hit the NCBI rate limit,
+    # the slightly longer pause of 0.37 seconds has been more reliable.
+    delay = 0.1 if api_key else 0.37
     current = time.time()
     wait = _open.previous + delay - current
     if wait > 0:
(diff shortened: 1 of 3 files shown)


Bio/Entrez/__init__.py
```diff
@@ -520,7 +520,7 @@ def _open(cgi, params=None, post=None, ecitmatch=False):
     # Equivalently, at least a third of second between queries
     params = _construct_params(params)
     options = _encode_options(ecitmatch, params)
-    delay = 0.1 if api_key else 0.333333334
+    delay = 0.1 if api_key else 0.37
```
peterjc: Can you add a comment like this here:

```python
# Using just 0.333333334 seconds sometimes hit the NCBI rate limit,
# the slightly longer pause of 0.37 seconds has been more reliable.
```

Otherwise the value of 0.37 is confusing and given the code comments earlier looks like a typo.

peterjc:
You didn't agree to the licensing terms - we will not merge you pull request without this. Also do you want to be listed as a contributor? You can read the template again at https://github.com/biopython/biopython/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Note that in this case this is an online test, so currently we are not running it on TravisCI or AppVeyor - so only the Python style checks would be run automatically.

hsiaoyi0504:
Got it. I will update the PR then.

Bio/Entrez/__init__.py
hsiaoyi0504: Fixed in 9d39b276c4c900ce4c3e5322ebea83777240b5e3

hsiaoyi0504:
I have updated the PR and accepted the license terms. Thanks !

peterjc:
Lovely, thank you!

Issue #1308, opened 2017-07-04
MafIO: Concatenation of negative strand exons


TBradley27:
Hello,

My problem relates to the case in which the user invokes the 'MafIndex.get_spliced' method in the case of multiple negative strand exons.

I use the following example:

**Single Exon**

`new_multiple_alignment = idx.get_spliced([22073644],[22073727],-1)`
`AlignIO.write(new_multiple_alignment, "path/bar.fa", "fasta")`

Which returns the following sequence (non-human species omitted for the sake of brevity):

`>hg38.chrY`
`tctggaatccctgtcctcaggtgatctggccgcttcgacctccaaaagtgctgagattac
aagcttgaactgccatgcccAAT`

**Multiple Exons**

`new_multiple_alignment = idx.get_spliced([22073644, 22071756],[22073727, 22072581],-1)`
`AlignIO.write(new_multiple_alignment, "path/foo.fa", "fasta")`

Which returns the following sequence (non-human species omitted for the sake of brevity, and gaps removed for the sake of clarity and formatting):

`>hg38.chrY
[...]

PR #1885, opened 2018-12-28
MafIO: Multiple negative strand exons fix

HopedWall:
This pull request addresses issue #1308

### DIFF
--- a/Bio/AlignIO/MafIO.py
+++ b/Bio/AlignIO/MafIO.py
@@ -728,15 +728,48 @@ class MafIndex(object):
         # finally, build a MultipleSeqAlignment object for our final sequences
         result_multiseq = []
 
-        for seqid, seq in subseq.items():
-            seq = Seq(seq)
+        # previously, calling get_spliced(starts,ends,-1) with len(starts) > 1 and len(ends) > 1
+        # would return the exons in the wrong order, see issue #1308
+        # basically, the whole sequence would be reverse complemented; instead
+        # each exon has to be reverse complemented, and then concatenated to the resulting sequence
+        if strand != ref_first_strand:
+            #find how long each exon was
+            offsets = [end-start for start,end in zip(starts,ends)];
 
-            seq = seq if strand == ref_first_strand else seq.reverse_complement()
+            #find the starting position of each exon
+            positions = [0];
+            for offset in offsets:
+                positions.append(positions[-1] + offset);
 
-            result_multiseq.append(SeqRecord(seq,
-                                             id=seqid,
-                                             name=seqid,
-                                             description=""))
+            for seqid, seq in subseq.items():
+                temp_seq = "";
+
+                #split the sequence into exons
+                exons = [seq[i:j] for i,j in zip(positions,positions[1:])]
+
+                for exon in exons:
+                    #convert exon (string) to seq in order to reverse and complement it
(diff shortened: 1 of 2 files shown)


peterjc:
Are you implementing what @blaiseli described here?:
https://github.com/biopython/biopython/issues/1308#issuecomment-312909627

HopedWall:
Correct! In a few tests I had done, calling `get_spliced(start,ends) #start,ends either sorted or not sorted` would return the exons concatenated in the correct order (I can include these tests in `test_MafIO_index.py` if you want). 
However, this was not the case when calling `get_spliced(starts, ends, -1)`, because the whole sequence would be reverse complemented (that's why the exons would appear in the opposite order!)

Bio/AlignIO/MafIO.py
```diff
@@ -728,15 +730,60 @@ def get_spliced(self, starts, ends, strand=1):
...
+
+            for seqid, seq in subseq.items():
+                temp_seq = ""
+
+                # split the sequence into exons
+                exons = [seq[i:j] for i, j in zip(positions, positions[1:])]
+
+                for exon in exons:
+                    # convert exon (string) to seq in order to reverse and complement it
+                    seq = Seq(exon).reverse_complement()
```
peterjc: If you are working with strings, just use the ``reverse_complement()`` function instead or converting to ``Seq`` object and back again.

i.e. ``from Bio.Seq import reverse_complement as _rc`` or similar.

peterjc: Are you sure ``subseq`` isn't already ``Seq`` objects? If it is, you are discarding the alphabet here.

peterjc:
The docstring for ``get_spliced`` would need expanding to describe how it is (now) intended to be used with reverse strand features.

Bio/AlignIO/MafIO.py
HopedWall: 1) Done!
2) Yes, I tried `print(type(seq))` and got `<type 'str'>`

HopedWall:
Changed the docstring for get_spliced, let me know if it's ok.

Bio/AlignIO/MafIO.py
```diff
@@ -565,13 +565,15 @@ def get_spliced(self, starts, ends, strand=1):
         """Return a multiple alignment of the exact sequence range provided.
 
         Accepts two lists of start and end positions on target_seqname, representing
-        exons to be spliced in silico.  Returns a *MultipleSeqAlignment* of the
-        desired sequences spliced together.
+        exons to be spliced in silico. A strand may also be provided.
+        Returns a *MultipleSeqAlignment* of the desired sequences spliced together.
+        When strand equals -1, the sequences are reverse complemented before the splice.
```
peterjc: This does not tell me what order to put the starts & ends in. For forward strand it is implicit and obvious enough, but not for the reverse strand.

Perhaps "..., and the start and end ordinates should be in numerically increasing rather than translation order."

Or the opposite, "...,  and the start and end ordinates should be in translation rather than numerically increasing order."

HopedWall: I changed the docstring as you suggested, it should be good now.

peterjc:
I agree, @blaiseli's point needs addressing.

Issue #1884, opened 2018-12-28
Bio.PDB.MMCIFParser.MMCIFParser exception


Minys233:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)

3.6.7 | packaged by conda-forge | (default, Nov 20 2018, 18:20:05) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)]
CPython
Darwin-18.2.0-x86_64-i386-64bit
1.73
```

(*Please copy and run the above in your Python, and copy-and-paste the output*)

### Expected behaviour

`Bio.PDB.MMCIFParser.MMCIFParser.get_structure()` parses structure `3HR6` normally.

### Actual behaviour

It will try to parse a string which representing a float number in a char to char manner, thus throw exception: `ValueError: could not convert string to float: '-'`.

### Steps to reproduce

First, download the mmCIF structure of `3HR6` from RCSB PDB: https://www.rcsb.org/structure/3HR6
Then write the following script and run
```python
from Bio.PDB.MMCIFParser import MMCIFParser
model = MMCIFParser().get_structure("", "path/to/3HR6/mmCIF")
```
Then you got this exception:
```python
Traceback (most recent call last):
  File "/Users/minys/Downloads/test/test.py", line 2, in <module>
    model = MMCIFParser().get_structure("", "test/3hr6.cif")
  File "/Users/minys/anaconda3/envs/pymol/lib/python3.6/site-packages/Bio/PDB/MMCIFParser.py", line 64, in get_structure
    self._build_structure(structure_id)
  File "/Users/minys/anaconda3/envs/pymol/lib/python3.6/site-packages/Bio/PDB/MMCIFParser.py", line 202, in _build_structure
    mapped_anisou = [float(_) for _ in u ]
  File "/Users/minys/anaconda3/envs/pymol/lib/python3.6/site-packages/Bio/PDB/MMCIFParser.py", line 202, in <listcomp>
    mapped_anisou = [float(_) for _ in u ]
ValueError: could not convert string to float: '-'
```
This exception results from the following code from `Bio/PDB/MMCIFParser.py`:

```python
if aniso_flag == 1 and i < len(aniso_u11):
    u = (aniso_u11[i], aniso_u12[i], aniso_u13[i],
[...]

Issue #1533, opened 2018-02-17
inconsistent types of MMCIF2Dict data values


speleo3:
### Setup

`Bio.__version__`: 1.71.dev0

### Expected behaviour

Consistent types for `MMCIF2Dict` data values.

### Actual behaviour

It's unpredictable whether a key may hold `list` or `str` type.

If an item has more than one value, it only can be represented in a `loop_`. But if it has a single value, it could either be in a loop or not in a loop. To my understanding, both representations are equivalent.

### Example

Example of a 3 atom file and a 1 atom file. The former represents the `_chem_comp_atom` items in a loop, the latter doesn't:

```python
from urllib2 import urlopen
from Bio.PDB.MMCIF2Dict import MMCIF2Dict

a = MMCIF2Dict(urlopen('https://files.rcsb.org/ligands/download/HOH.cif'))
b = MMCIF2Dict(urlopen('https://files.rcsb.org/ligands/download/MG.cif'))

print(type(a['_chem_comp_atom.model_Cartn_x']))
# <type 'list'>
print(type(b['_chem_comp_atom.model_Cartn_x']))
# <type 'str'>
```

### Suggested Solutions

1. Store all values as lists
2. Or add a `get_list` or `get_iter` method to `MMCIF2Dict`

peterjc:
CC @JoaoRodrigues @jgreener64 @etal

speleo3:
[gemmi](https://github.com/project-gemmi/gemmi) library has a **find_values** method that always returns a list:
http://gemmi.readthedocs.io/en/latest/cif-parser.html#pair-or-loop

jgreener64:
I follow your logic and guess this is an inconsistency, I don't think I'll have time to look at it now though.

I guess as a workaround use:
```python
if isinstance(val, str):
    ...
elif isinstance(val, list):
    ...
```

Of your suggested solutions, the first seems better long term but is breaking. I suppose the second would be implemented by the following in `MMCIF2Dict`:

```python
def get_list(self, key):
    if isinstance(self[key], str):
        return [self[key]]
    elif isinstance (self[key], list):
        return self[key]
```

This could be a simple-ish PR?

JoaoRodrigues:
@jgreener64 why would make things into a list break, could you give an example? Strings are also indexable, for example. I'd think moving to lists would be the best overall.

jgreener64:
[...]

PR #1886, opened 2019-01-02
Ensure aniso value is a list

jgreener64:
This pull request addresses issue #1884.

The issue here was that the anisotropic values would be read in as strings if they only had a single value. Indexing into the string and converting to a float then errors when the first character of the string is "-".

This change ensures the value is a list before continuing. It means the failing PDB entries in the issue are all read in without throwing an error. I think the mmCIF files are incorrect in these instances, as there is only a single anisotropic value. This gets mapped to the first atom. But I guess that's a PDB problem, not our problem?

This is also related to discussion at https://github.com/biopython/biopython/issues/1533. It's an example of running into the list v string problem.

### DIFF
--- a/Bio/PDB/MMCIFParser.py
+++ b/Bio/PDB/MMCIFParser.py
@@ -67,6 +67,12 @@ class MMCIFParser(object):
 
     # Private methods
 
+    def _ensure_list(self, val):
+        if isinstance(val, list):
+            return val
+        else:
+            return [val]
+
     def _build_structure(self, structure_id):
 
         # two special chars as placeholders in the mmCIF format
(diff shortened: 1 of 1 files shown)


jgreener64:
On further investigation, the PDB mmCIF files are correct, as the atom to map the aniso value to is specified (`_atom_site_anisotrop.pdbx_label_comp_id` etc.). We ignore this and just assume that all atoms are be in the aniso list in the correct order. Seems like that is not the case. I expect this means we are reading in many anisotropic factors incorrectly.

peterjc:
So does this pull request need further work before merging?

jgreener64:
My preference would be to merge this PR to make mmCIF parsing work for the erroring structures. Then I could open a separate issue to mark the potential error in aniso value reading.

peterjc:
TravisCI failure should be fixed by https://github.com/biopython/biopython/commit/e3f69e08bd0bb4fdd49f23dcdaf42bf5b485c1b4

peterjc:
AppVeyor failed after the tests while setting up the coverage report, you can ignore that.

jgreener64:
If #2186 is merged this PR is redundant so I'll close it.

jgreener64:
Closing as redundant because #2186 is merged.

Issue #1870, opened 2018-12-05
Updating GenBank parser for extra long accession


peterjc:
Quoting this from the NCBI,

https://ncbiinsights.ncbi.nlm.nih.gov/2018/12/03/adapting-flatfile-parsers-genbank-new-accession-formats/

> As [previously announced](https://ncbiinsights.ncbi.nlm.nih.gov/2018/09/19/genbank-expanded-accession-formats/), GenBank and other INSDC members will expand the accession formats used for sequencing projects by the end of this year. We’re introducing these new formats to accommodate the growth of Whole Genome Shotgun (WGS), Transcriptome Shotgun Assembly (TSA), and Targeted Locus Study (TLS) sequencing sequences. More details about those changes are available on [NCBI Insights](https://ncbiinsights.ncbi.nlm.nih.gov/2018/09/19/genbank-expanded-accession-formats/).
> 
> You may have to adjust your code and databases to accommodate the new formats’ longer length. In particular, the first line of the flatfile format, referred to as the LOCUS line, includes the “Locus Name” (usually identical to the accession number), which may now grow to as long as 20 characters. See section 3.4.4 of the [GenBank release notes](https://ftp.ncbi.nlm.nih.gov/genbank/gbrel.txt) for examples of how the LOCUS line might change.
> 
> Since 2003, the GenBank release notes have recommended that flatfile parsers use a whitespace-separated tokens approach to accommodate changes like the one described in section 3.4.4. If your flatfile parsers rely solely on position, you may have to make modifications. From our internal testing, it appears BioPython and BioPerl properly handle most of the examples shown in section 3.4.4, and only have issues with the last theoretical examples where the sequence length no longer ends at position 40. We do recommend adjusting code to accommodate those theoretical examples for future-proofing.
> 
> Please write to the helpdesk with any questions about the new formats.
>

And from the release notes,

> 3.4.4.1 : Important notice about parsing the LOCUS line
> 
>   Users who process the data elements of the LOCUS line should use a
> token-based parsing approach rather than parsing its content based on
> fixed column positions.
> 
>   Historically, the LOCUS line has had a fixed length and its elements have
> been presented at specific column positions. A complete description of the
> data fields and their typical column positions is provided in Section 3.4.4.2 .
> But with the anticipated increases in the lengths of accession numbers, and
> the advent of sequences that are gigabases long, maintaining the column
> positions will not always be possible and the overall length of the LOCUS
> line could exceed 79 characters.
> 
[...]

PR #1888, opened 2019-01-03
Update the GenBank parser for extra long accession

nimne:
This pull request addresses issue #1870

The genbank parser was essentially ready to handle the extra long accession, I've changed some changes to turn errors in to warnings.

The genbank writer will also handle the longer accession length which resulted in quite a few changes in the test suite - where there was a 'name too long' error, there now is none.

There are warnings on both parsing, and writing genbank files with extra long accession numbers. I think these should stay for a while, as I expect a long accession field will cause problems in other software until it is updated.

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1407,8 +1407,8 @@ class GenBankScanner(InsdcScanner):
             if int(splitline[2]) > sys.maxsize:
                 raise ValueError("Tried to load a sequence with a length %s, "
                                  "your installation of python can only load "
-                                 "sesquences of length %s" % splitline[2],
-                                 sys.maxsize)
+                                 "sesquences of length %s" % (splitline[2],
+                                 sys.maxsize))
             else:
                 consumer.size(splitline[2])
 
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1177,6 +1177,15 @@ class GenBankScanner(InsdcScanner):
         the column based layout.
 
         We also try to cope with GenBank like files with partial LOCUS lines.
+
+        As of release 229.0, the columns are no longer strictly in a given
+        position. See:
+          Historically, the LOCUS line has had a fixed length and its elements
+          have been presented at specific column positions...
+          But with the anticipated increases in the lengths of accession
+          numbers, and the advent of sequences that are gigabases long,
+          maintaining the column positions will not always be possible and the
+          overall length of the LOCUS line could exceed 79 characters.
         """
         #####################################
         # LOCUS line                        #
@@ -1387,6 +1396,8 @@ class GenBankScanner(InsdcScanner):
                 and line.split()[5] in ('linear', 'circular'):
             # Cope with invalidly spaced GenBank LOCUS lines like
             # LOCUS       AB070938          6497 bp    DNA     linear   BCT 11-OCT-2001
+            # This will also cope with extra long accession numbers and
+            # sequence lengths
             splitline = line.split()
             consumer.locus(splitline[1])
             consumer.size(splitline[2])
@@ -1394,7 +1405,8 @@ class GenBankScanner(InsdcScanner):
             consumer.topology(splitline[5])
             consumer.data_file_division(splitline[6])
             consumer.date(splitline[7])
-            warnings.warn("Attempting to parse malformed locus line:\n%r\n"
+            warnings.warn("Attempting to parse locus line that is extra "
+                          "long or malformed  :\n%r\n"
                           "Found locus %r size %r residue_type %r\n"
                           "Some fields may be wrong."
                           % (line, splitline[1], splitline[2], splitline[4]),
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1179,14 +1179,15 @@ class GenBankScanner(InsdcScanner):
         We also try to cope with GenBank like files with partial LOCUS lines.
 
         As of release 229.0, the columns are no longer strictly in a given
-        position. See:
+        position. See GenBank format release notes:
 
-        Historically, the LOCUS line has had a fixed length and its elements
-        have been presented at specific column positions...
-        But with the anticipated increases in the lengths of accession
-        numbers, and the advent of sequences that are gigabases long,
-        maintaining the column positions will not always be possible and the
-        overall length of the LOCUS line could exceed 79 characters.
+            "Historically, the LOCUS line has had a fixed length and its
+            elements have been presented at specific column positions...
+            But with the anticipated increases in the lengths of accession
+            numbers, and the advent of sequences that are gigabases long,
+            maintaining the column positions will not always be possible and
+            the overall length of the LOCUS line could exceed 79 characters."
+
         """
         #####################################
         # LOCUS line                        #
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1401,7 +1401,17 @@ class GenBankScanner(InsdcScanner):
             # sequence lengths
             splitline = line.split()
             consumer.locus(splitline[1])
-            consumer.size(splitline[2])
+            # Provide descriptive error message if the sequence is too long
+            # for python to handle
+            import sys
+            if int(splitline[2]) > sys.maxsize:
+                raise ValueError("Tried to load a sequence with a length %s, "
+                                 "your installation of python can only load "
+                                 "sesquences of length %s" % splitline[2],
+                                 sys.maxsize)
+            else:
+                consumer.size(splitline[2])
+
             consumer.residue_type(splitline[4])
             consumer.topology(splitline[5])
             consumer.data_file_division(splitline[6])
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1180,12 +1180,13 @@ class GenBankScanner(InsdcScanner):
 
         As of release 229.0, the columns are no longer strictly in a given
         position. See:
-          Historically, the LOCUS line has had a fixed length and its elements
-          have been presented at specific column positions...
-          But with the anticipated increases in the lengths of accession
-          numbers, and the advent of sequences that are gigabases long,
-          maintaining the column positions will not always be possible and the
-          overall length of the LOCUS line could exceed 79 characters.
+        
+        Historically, the LOCUS line has had a fixed length and its elements
+        have been presented at specific column positions...
+        But with the anticipated increases in the lengths of accession
+        numbers, and the advent of sequences that are gigabases long,
+        maintaining the column positions will not always be possible and the
+        overall length of the LOCUS line could exceed 79 characters.
         """
         #####################################
         # LOCUS line                        #
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1408,7 +1408,7 @@ class GenBankScanner(InsdcScanner):
                 raise ValueError("Tried to load a sequence with a length %s, "
                                  "your installation of python can only load "
                                  "sesquences of length %s" % (splitline[2],
-                                 sys.maxsize))
+                                                              sys.maxsize))
             else:
                 consumer.size(splitline[2])
 
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1417,12 +1417,12 @@ class GenBankScanner(InsdcScanner):
             consumer.topology(splitline[5])
             consumer.data_file_division(splitline[6])
             consumer.date(splitline[7])
-            warnings.warn("Attempting to parse locus line that is extra "
-                          "long or malformed  :\n%r\n"
-                          "Found locus %r size %r residue_type %r\n"
-                          "Some fields may be wrong."
-                          % (line, splitline[1], splitline[2], splitline[4]),
-                          BiopythonParserWarning)
+            if len(line) < 80:
+                warnings.warn("Attempting to parse malformed locus line:\n%r\n"
+                              "Found locus %r size %r residue_type %r\n"
+                              "Some fields may be wrong."
+                              % (line, splitline[1], splitline[2], splitline[4]),
+                              BiopythonParserWarning)
         elif len(line.split()) == 7 and line.split()[3] in ["aa", "bp"]:
             # Cope with EnsEMBL genbank files which use space separation rather
             # than the expected column based layout. e.g.
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -1180,7 +1180,7 @@ class GenBankScanner(InsdcScanner):
 
         As of release 229.0, the columns are no longer strictly in a given
         position. See:
-        
+
         Historically, the LOCUS line has had a fixed length and its elements
         have been presented at specific column positions...
         But with the anticipated increases in the lengths of accession
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -602,9 +602,14 @@ class GenBankWriter(_InsdcWriter):
         if len(locus) > 16:
             if len(locus) + 1 + len(str(len(record))) > 28:
                 # Locus name and record length to long to squeeze in.
-                raise ValueError("Locus identifier %r is too long" % locus)
-            else:
-                warnings.warn("Stealing space from length field to allow long name in LOCUS line", BiopythonWarning)
+                # Per updated GenBank standard (Dec 15, 2018) 229.0
+                # the Locus identifier can be any length, and a space
+                # is added after the identifier to keep the identifier
+                # and length fields separated
+                warnings.warn("Increasing length of locus line to allow "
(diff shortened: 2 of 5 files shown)


nimne:
Looks like AppVeyor is reporting a problem on Windows. I'll investigate.

peterjc:
The comments on https://bugs.python.org/issue26423 suggest this might be specific to the Anaconda Windows  Python 2 builds, which is annoying as that is what we are using on AppVeyor:

```
ERROR: test_longer_locus_line (test_GenBank_unittest.GenBankTests)
Check that we can read and write files with longer locus lines
  File "C:\projects\biopython\Tests\test_GenBank_unittest.pySkipping any tests requiring internet access
Python version: 2.7.15 |Anaconda, Inc.| (default, Dec 10 2018, 21:40:17) [MSC v.1500 32 bit (Intel)]
Operating system: nt win32
", line 459, in test_longer_locus_line
    record = SeqIO.read(path.join("GenBank", "long_header_genbank_test.gb"), 'genbank')
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\SeqIO\__init__.py", line 720, in read
    first = next(iterator)
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\SeqIO\__init__.py", line 655, in parse
    for r in i:
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\GenBank\Scanner.py", line 489, in parse_records
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\GenBank\Scanner.py", line 473, in parse
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\GenBank\Scanner.py", line 455, in feed
    consumer.record_end("//")
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\GenBank\__init__.py", line 1327, in record_end
    self.data.seq = UnknownSeq(self._expected_size, seq_alphabet)
  File "C:\projects\biopython\build\lib.win32-2.7\Bio\SeqRecord.py", line 314, in _set_seq
    self._per_letter_annotations = _RestrictedDict(length=len(self.seq))
OverflowError: long int too large to convert to int
Ran 272 tests in 728.454 seconds
```

If the GenBank file does not have a sequence at the end, we make an ``UnknownSeq`` object using the length from the ``LOCUS`` line - it doesn't actually make the long string of Ns in memory unless explicitly requested. So, ``len(self.seq)`` is calling ``UnknownSeq.__len__(...)`` to get the length of the record's (implicit) sequence, to set the length for entries in the per-letter annotations restricted dictionary.

Do you have access to a Windows machine where you can try the Anaconda Python 2.7? e.g. do ``int('10000000000')`` and ``int(10000000000)`` work?

Bio/SeqIO/InsdcIO.py
```diff
@@ -615,7 +620,10 @@ def _write_the_first_line(self, record):
         if len(record) > 99999999999:
             # Currently GenBank only officially support up to 350000, but
             # the length field can take eleven digits
```
peterjc: This comment needs updating

Tests/test_GenBank_unittest.py
```diff
@@ -432,6 +433,43 @@ def test_genbank_date_invalid(self):
             gb = SeqIO.read(handle, "gb")
             self.assertEqual(gb.annotations["date"], "01-JAN-1980")
 
+    def test_longer_locus_line(self):
+        """Check that we can read and write files with longer locus lines"""
+        # Create file
+        with open(path.join("GenBank", "DS830848.gb"), 'r') as inhandle:
+            data = inhandle.readlines()
+            data[0] = "LOCUS       AZZZAA02123456789 10000000000 bp    DNA     linear   PRI 15-OCT-2018\n"
+        with open(path.join("GenBank", "long_header_genbank_test.gb"), 'w') as outfile:
```
peterjc: Please don't use a named temp file under the ``Tests/`` folder, they tend not to get cleaned up properly and occasionally get checked into version control.

I would use a ``StringIO`` handle (via ``from Bio._py3k import StringIO``), or use ``tempfile`` if you really want an actual file.

Tests/test_GenBank_unittest.py
```diff
@@ -432,6 +433,43 @@ def test_genbank_date_invalid(self):
...
+            warnings.simplefilter("ignore")
+            record = SeqIO.read(path.join("GenBank", "long_header_genbank_test.gb"), 'genbank')
+            SeqIO.write(record, path.join("GenBank", "long_header_genbank_test_out.gb"), 'genbank')
+
+            record_in = SeqIO.read(path.join("GenBank", "long_header_genbank_test_out.gb"), 'genbank')
+            self.assertEqual(record_in.id, "DS830848.1")
+            self.assertEqual(record_in.name, "AZZZAA02123456789")
+            self.assertEqual(len(record_in.seq), 10000000000)
+
+        # Clean up
```
peterjc: These never get called if the test fails, this kind of thing would be better in the ``tearDown`` method (perhaps with the ``setUp`` method used to setup a temp folder or filenames). Or even better, never create these files in the first place.

nimne:
Thanks for the code review and suggestions, I've cleaned up the testing code.

I will setup a windows testing environment in the next day or two and see if I can figure out a solution.

nimne:
It looks like `int(10000000000)` works correctly in that it returns a `long`.
```python
Python 2.7.15 |Anaconda, Inc.| (default, Dec 10 2018, 21:40:17) [MSC v.1500 32 bit (Intel)] on win32
>>> int(10000000000)
10000000000L
```

Unfortunately, it looks like `__len__` must return an `int` rather than a `long`, there is a good explaination for the reason [in an answer here](https://stackoverflow.com/questions/42521449/how-does-python-ensure-the-return-value-of-len-is-an-integer-when-len-is-cal).

I made a small test example in both python2 and python3 on windows. I will use the following test class for all of the examples:
```python
>>> class test_class():
...     def __init__(self, value):
...         self.value = value
...
...     def __len__(self):
...         return self.value
...
>>>
```
```python
Python 2.7.15 |Anaconda, Inc.| (default, Dec 10 2018, 21:40:17) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.\
>>> test_class_100 = test_class(100)
>>> test_class_10000000000 = test_class(10000000000)
>>>
>>> type(test_class_100.value)
<type 'int'>
>>> type(len(test_class_100))
<type 'int'>
>>>
>>> type(test_class_10000000000.value)
<type 'long'>
>>> type(len(test_class_10000000000))
TypeError: __len__() should return an int
>>> import sys
>>> sys.maxsize
2147483647
```
```python
Python 3.7.1 (default, Dec 10 2018, 22:09:34) [MSC v.1915 32 bit (Intel)] :: Anaconda, Inc. on win32
>>> test_class_100 = test_class(100)
>>> test_class_10000000000 = test_class(10000000000)
>>>
>>> type(test_class_100.value)
<class 'int'>
>>> type(len(test_class_100))
<class 'int'>
>>>
>>> type(test_class_10000000000.value)
<class 'int'>
>>> type(len(test_class_10000000000))
OverflowError: cannot fit 'int' into an index-sized integer
>>> import sys
>>> sys.maxsize
2147483647
```

It appears that 64-bit python3 on windows fixes this 'problem', but it still exists in python2 as `sys.maxsize` is a `long`.
```python
Python 2.7.15 |Anaconda, Inc.| (default, Dec 10 2018, 21:57:18) [MSC v.1500 64 bit (AMD64)] on win32
>>> class test_class():
...     def __init__(self, value):
...         self.value = value
...
...     def __len__(self):
...         return self.value
...
>>> test_class_100 = test_class(100)
>>> test_class_10000000000 = test_class(10000000000)
>>>
>>> type(test_class_100.value)
<type 'int'>
>>> type(len(test_class_100))
<type 'int'>
>>>
>>> type(test_class_10000000000.value)
<type 'long'>
>>> type(len(test_class_10000000000))
TypeError: __len__() should return an int
>>> import sys
>>> sys.maxsize
9223372036854775807L
```
```python
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> test_class_100 = test_class(100)
>>> test_class_10000000000 = test_class(10000000000)
>>>
>>> type(test_class_100.value)
<class 'int'>
>>> type(len(test_class_100))
<class 'int'>
>>>
>>> type(test_class_10000000000.value)
<class 'int'>
>>> type(len(test_class_10000000000))
<class 'int'>
>>> import sys
>>> sys.maxsize
9223372036854775807
```

This means that on a 64-bit windows install of python3, the test passes as expected. I don't have immediate access to a 32-bit python install on linux, but I would guess that the test would fail.
```
(base) C:\Users\User\Desktop\git\biopython\Tests>python run_tests.py test_GenBank_unittest
Python version: 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)]
Operating system: nt win32
test_GenBank_unittest ... ok
Ran 1 test in 2.016 seconds
```

The only solution that I can see is to raise a `RuntimeError` (or similar) if trying to set `size` in `_FeatureConsumer` to anything greater than `sys.maxsize`. This woud essentially preserve the current behavior, because the code already isn't capible of handling sequences with a length greater than `sys.maxsize`.

peterjc:
Cross reference https://www.biostars.org/p/12560/ we should be OK with 9 digit lengths for the largest known chromosomes, and in practise even the Python 2.7 limit of ``sys.maxsize = 2147483647`` (10 digits, 2.1 giga-base-pairs) for integers should be fine.

So, how about leaving the code as it is, making sure we have a 10 digit test case (under the 2147483647 limit), and then under Python 2.7 skip the 10000000000 (11 digit) test case.

(I'm trying to be pragmatic here, the limit is currently theoretical rather than practical, and we're dropping Python 2.7 support soon anyway - in the year 2020).

nimne:
I agree that the sequence length limitation of `2147483647 bp` is probably not something that will be biologically relevant in the near future.

The current code works with the more flexible locus line and raises an informative error if the input sequence length is too long. Let me know if there was a different approach that you had in mind.

peterjc:
Don't worry about the CodeCov "red cross", we're not enforcing that strictly although better test coverage is a long term goal.

Bio/GenBank/Scanner.py
```diff
@@ -1177,6 +1177,16 @@ def _feed_first_line(self, consumer, line):
         the column based layout.
 
         We also try to cope with GenBank like files with partial LOCUS lines.
+
+        As of release 229.0, the columns are no longer strictly in a given
+        position. See:
+
+        Historically, the LOCUS line has had a fixed length and its elements
```
peterjc: This doesn't flow nicely from the previous paragraph. Is this a quote from the NCBI? If so putting opening and closing quote marks round it would be enough.

You could explore making it a nice block quote when rendered as RST using http://rst.ninjs.org/ or similar but that isn't required.

Bio/GenBank/Scanner.py
```diff
@@ -1387,14 +1397,27 @@ def _feed_first_line(self, consumer, line):
...
+            # This will also cope with extra long accession numbers and
+            # sequence lengths
             splitline = line.split()
             consumer.locus(splitline[1])
-            consumer.size(splitline[2])
+            # Provide descriptive error message if the sequence is too long
+            # for python to handle
+            import sys
+            if int(splitline[2]) > sys.maxsize:
+                raise ValueError("Tried to load a sequence with a length %s, "
```
peterjc: Good - this is a much more helpful error message than the cryptic exception we were getting before 👍

Bio/SeqIO/InsdcIO.py
```diff
@@ -602,9 +602,14 @@ def _write_the_first_line(self, record):
         if len(locus) > 16:
             if len(locus) + 1 + len(str(len(record))) > 28:
                 # Locus name and record length to long to squeeze in.
-                raise ValueError("Locus identifier %r is too long" % locus)
-            else:
-                warnings.warn("Stealing space from length field to allow long name in LOCUS line", BiopythonWarning)
```
peterjc: I had to read this part of the diff several times.

You're removing the old warning about stealing spaces (where the later fields would not change position), since this is now clearly allowed by the NCBI and ought not to break other parsers trying to read our output.

You're downgrading the old ValueError into a new warning (because this NCBI change is new, a warning makes sense as not every parser out there will cope).

I'm happy 👍

Tests/test_GenBank_unittest.py
```diff
@@ -294,15 +295,16 @@ def test_long_names(self):
...
         original = SeqIO.read("GenBank/iro.gb", "gb")
         self.assertEqual(len(original), 1326)
+        # Acceptability of LOCUS line with length > 80 invalidates some of these tests
         for name, seq_len, ok in [
                 ("short", 1, True),
                 ("max_length_of_16", 1000, True),
                 ("overly_long_at_17", 1000, True),
                 ("excessively_long_at_22", 99999, True),
-                ("excessively_long_at_22", 100000, False),
+                # ("excessively_long_at_22", 100000, False),
```
peterjc: Can you switch this to:

```python
                ("excessively_long_at_22", 100000, True),
```

Tests/test_GenBank_unittest.py
```diff
@@ -294,15 +295,16 @@ def test_long_names(self):
...
                 ("pushing_the_limits_at_24", 999, True),
-                ("pushing_the_limits_at_24", 1000, False),
-                ("longest_possible_len_of_26", 10, False),  # 2 digits
+                # ("pushing_the_limits_at_24", 1000, False),
+                # ("longest_possible_len_of_26", 10, False),  # 2 digits
```
peterjc: Again, rather than commenting out, can you change ``False`` to ``True``?

Also, any ideas for renaming the 26 letter long ID of ``"longest_possible_len_of_26"`` (since this is no longer the maximum possible)? How about ``"old_max_name_length_was_26"`` which is also 26 characters?

Tests/test_GenBank_unittest.py
```diff
@@ -432,6 +434,102 @@ def test_genbank_date_invalid(self):
...
+            record_in = SeqIO.read(out_tmp, 'genbank')
+            self.assertEqual(record_in.id, "DS830848.1")
+            self.assertEqual(record_in.name, "AZZZAA021234567891234")
+            self.assertEqual(len(record_in.seq), 2147483647)
+
+    if sys.maxsize > 2147483647:
+        def test_extremely_long_sequence(self):
+            """Tests if extremely long sequences can be read.
+
+            This is only run if sys.maxsize > 2147483647.
```
peterjc: Can you add a note here about Python 2, it would help for removing all the Python 2 specific hacks in 2020 if we can find it with ``grep -i "Python 2"`` or ``grep -i "Python 3"``

peterjc:
This looks good - only some minor comments about the docstrings and some minor ones on the tests. We're at the final polishing stage now :)

Tests/test_GenBank_unittest.py
nimne: As far as I can tell, this will still be a limitation in 32-bit Python3. See the documentation [here ](https://docs.python.org/3/reference/datamodel.html#object.__len__) and [here](https://docs.python.org/3/library/sys.html#sys.maxsize). This is why I decided to have the conditional statement based on the `sys.maxsize` rather than the python version.

peterjc: OK, good point. I've not knowingly used a 32bit Python for some time, so that didn't cross my mind.

peterjc:
Great work, I'm happy with the code. I intend to squash-and-merge this to give a simpler history.

Would you like to add yourself to the ``NEWS.rst`` entry?

Should we add a  ``NEWS.rst`` entry something like this too:

```
The NCBI now allows longer accessions in the GenBank file LOCUS line, meaning
the fields may not always follow the historical column based positions. We
no longer give a warning when parsing these. We now allow writing such files
(although with a warning as support for reading them is not yet widespread).
```

nimne:
It seems sensible to add a news entry. One clarification, there is still a warning on parsing ([as was present previously](https://github.com/biopython/biopython/blob/cabc27851ccf1fa83b2ec876ab271dcb117a98ea/Bio/GenBank/Scanner.py#L1397)). I left this warning because it was already in the parser that functioned based on white space. Now that this kind of parsing is the recommended specification, do you think this warning should be removed?

The previous behavior was to trigger the warning on GenBank files with unusual locus line spacing (too few / too many spaces in the wrong place). Should this warning be ignored only if the length is >80 and displayed otherwise?

peterjc:
Ah. We have a ``NEWS.rst`` merge conflict. Do you want to try a ``git rebase origin/master``, or would you prefer I do the merge at the command line?

nimne:
I'll try the rebase!

nimne:
I think I've resolved the conflict correctly. I've updated the NEWS.rst and clarified the warning message.

peterjc:
It looks like you did a merge rather than a rebase... let me have a go now.

peterjc:
Squash-merged as a single commit 406cb2eb07335a256ca2540d03a9cd97dd7050fa

Thank you Nick!

nimne:
Thanks for fixing my git issue and explaining the process!

peterjc:
Helping newcomers now will usually pay off in the long run - be it your future contributions to Biopython or elsewhere.

Issue #1323, opened 2017-07-13
Bio.Phylo: "neighbor joining" tree construction fails for minimal alignment


hendrikweisser:
The "neighbor joining" algorithm for the construction of phylogenetic trees, implemented in `Bio.Phylo.TreeConstruction.DistanceTreeConstructor.nj` fails with the following error, if applied to a minimal sequence alignment containing only two sequences:
```
  File "...\Bio\Phylo\TreeConstruction.py", line 703, in nj
    if clades[0] == inner_clade:
UnboundLocalError: local variable 'inner_clade' referenced before assignment
```
This is with BioPython version 1.68, but looking at the code, the cause of the problem seems to be still present in the "master" branch: Following from [here (1.68)](https://github.com/biopython/biopython/blob/biopython-168/Bio/Phylo/TreeConstruction.py#L656) or [here ("master")](https://github.com/biopython/biopython/blob/master/Bio/Phylo/TreeConstruction.py#L694), the code that sets the `inner_clade` variable is only executed for distance matrices of size > 2.

A simple check could be added to avoid the error. However, since the UPGMA algorithm is able to successfully generate a (trivial) tree for such a minimal alignment, it would be great if the NJ algorithm could be adapted to do the same.

peterjc:
CC @lijax, @etal

peterjc:
This should be closed via #1891.

PR #1891, opened 2019-01-08
Added special case for minimum alignment matrices in the NJ algorithm

ZykeDev:
This pull request addresses issue #1323 


Bio/Phylo/TreeConstruction.py
```diff
@@ -692,63 +692,79 @@ def nj(self, distance_matrix):
...
-            clades[1].branch_length = dm[1, 0]
-            clades[0].clades.append(clades[1])
+            clades[0] = inner_clade
             root = clades[0]
         else:
-            clades[0].branch_length = dm[1, 0]
-            clades[1].branch_length = 0
-            clades[1].clades.append(clades[0])
-            root = clades[1]
+            while len(dm) > 2:
```
peterjc: I think rather than indenting all the old code one level using:

```
# special case for Minimum Alignment Matrices
if len(dm) == 2:
    ... # new code
else:
    while len(dm) > 2:
     ... # old code
```

you could leave it as:

```
if len(dm) == 2:
    ... # new code
while len(dm) > 2:
    ... # old code
```

This should make the diff much smaller, and be easier to review.

peterjc:
Much clearer now - what do you think @etal?

peterjc:
@etal or @lijax does this look ready to merge?

ZykeDev:
Appveyor seems to run into this error, but I doubt it's in any way related to TreeConstruction:

`File "C:\Py\lib\site-packages\PIL\Image.py", line 94, in <module>
    from . import _imaging as core
ImportError: DLL load failed: The specified module could not be found`

Should I fork from the latest commit (710fc1b), open another Pull Request and retry from there?
I was also thinking about adding tests for minimal matrices to the NJ part of the Test Suite.

peterjc:
A similar PIL DLL problem showed up on another pull request, it may have been sorted out now...

The latest TravisCI flake8 problems (F117 and F632) were caused by a recent flake8 update, just fixed on the master branch. Any minor update to your branch should trigger a rebuild merged into the latest master branch (and thus fix that).

I don't think you need to, but the way to 'move' all your work to be on top of the a recent commit from the official master branch (or simply the latest master branch) would be with ``git rebase``. Something like this:

```
git fetch origin  # get latest changes from official repository
git checkout patch-3  # make sure we're on your branch
git rebase origin/master  # rewrite branch history as if was starting from latest master
git push ZykeDev patch-3 --force  # update this pull request
```

You'd need to adjust ``origin`` and ``ZykeDev`` here if your remotes are named something else, perhaps ``upstream`` and ``origin`` respectively?

Note this is about the only time you'd ever use ``git push`` with ``--force`` as it will re-write the git history of the branch.

ZykeDev:
If flake8 and the other issues have been fixed then I guess I don't need to rebase. Thanks for the info tho.

peterjc:
Thank you.

If there are no comments/objections from @etal, @lijax (or anyone else familiar with this area of the code), I will aim to merge this at the end of the coming week.

peterjc:
Squash-and-merged, thank you.

I'm assuming you don't want to be acknowledged by name as a contributor - but if you do, please just ask.

lijax:
Sorry for the late reply. And thanks for your contribution @ZykeDev . As it's been 5 years since the first version of this library, and I'm currently not working in this field, it would be nice if you or someone else can maintain it with more clear knowledge in this field.

peterjc:
Great - thanks @lijax

Issue #1896, opened 2019-01-17
SeqFeature.translate ignores CDS type


rwst:
This is master with Python3. The doctest of `SeqFeature.translate` follows:
```
        >>> seq = Seq("GGTTACACTTACCGATAATGTCTCTGATGA", generic_dna)
        >>> f = SeqFeature(FeatureLocation(0, 30), type="CDS")
        >>> f.qualifiers['transl_table'] = [11]
        >>> f.translate(seq)
```
 I would expect that giving a type of `CDS` make the translation behave as if I would have said:
```
In [10]: Seq('GGTTACACTTACCGATAATGTCTCTGATGA'[0:30], generic_dna).translate(table=11, cds=True)
---------------------------------------------------------------------------
TranslationError                          Traceback (most recent call last)
<ipython-input-10-08236f76b040> in <module>()
----> 1 Seq('GGTTACACTTACCGATAATGTCTCTGATGA').translate(table=11, to_stop=False, cds=True, gap=None)

/usr/lib64/python3.7/site-packages/Bio/Seq.py in translate(self, table, stop_symbol, to_stop, cds, gap)
   1183 
   1184         protein = _translate_str(str(self), codon_table, stop_symbol, to_stop,
-> 1185                                  cds, gap=gap)
   1186 
   1187         if gap and gap in protein:

/usr/lib64/python3.7/site-packages/Bio/Seq.py in _translate_str(sequence, table, stop_symbol, to_stop, cds, pos_stop, gap)
   2592         if str(sequence[:3]).upper() not in table.start_codons:
   2593             raise CodonTable.TranslationError(
-> 2594                 "First codon '{0}' is not a start codon".format(sequence[:3]))
   2595         if n % 3 != 0:
   2596             raise CodonTable.TranslationError(

TranslationError: First codon 'GGT' is not a start codon
```
But the output is:
```
Seq('GYTYR*CL**', HasStopCodon(ExtendedIUPACProtein(), '*'))
```
Reason is that the `type` property of `SeqFeature` is ignored in `SeqFeature.translate()` and a default `cds` parameter of `False` is used.

I'd like to do a PR where the translation uses cds=`True` if the feature type is `CDS` and the parameter is not explicitly set to `False`. Doctests change accordingly.

peterjc:
I thought we had something related as an existing issue, on using more of the NCBI style annotation typically stored in the ``SeqFeature`` object's ``.qualifiers`` annotation dictionary, but I could only find:

[...]

PR #1897, opened 2019-01-17
SeqFeature.translate ignores CDS type

rwst:
This pull request addresses issue #1896

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -338,7 +338,7 @@ class SeqFeature(object):
         return self.location.extract(parent_sequence)
 
     def translate(self, parent_sequence, table="Standard", start_offset=None,
-                  stop_symbol="*", to_stop=False, cds=False, gap=None):
+                  stop_symbol="*", to_stop=False, cds=None, gap=None):
         """Get a translation of the feature's sequence.
 
         This method is intended for CDS or other features that code proteins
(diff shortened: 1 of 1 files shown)


peterjc:
This looks good, apart from the annoying Python 2 vs 3 doctest problem. From memory I think you can shorten the expected exception from ``Bio.Data.CodonTable.TranslationError`` to just ``TranslationError``

Separately, do you think this warrants a NEWS entry, or is it too niche?

rwst:
Really? Simply `TranslationError` fails here both with `python3 Bio/SeqFeature.py` and `python3 setup.py test` but let's see. As to NEWS, I feel its too special.

rwst:
Now Py3 is failing. The issue is mentioned in http://python3porting.com/problems.html#handling-expected-exceptions.

peterjc:
You were right - that fixed it for Python 2, but broke it on Python 3. So the good news is the code change works on both (although that was expected).

We've had this kind of pain with exceptions in doctests before, but I am not immediately sure how we fixed it - perhaps by the practical route of moving the test out of the docstring, which might be worth doing anyway from a documentation clarity point of view.

Maybe move the tests to ``Tests/test_SeqFeature.py`` instead?

rwst:
Maybe we could use `IGNORE_EXCEPTION_DETAIL` as Py2.6 is not an issue.

peterjc:
Found the old example I was thinking of - 4a66a74e08a4b21fe2808a26cd1f8c670501ec79 - at the time I tried ``IGNORE_EXCEPTION_DETAIL`` and it didn't solve things.

So, one option is to follow that ``Bio/Seq.py`` precedent, and leave the doctests in Python 3 style, but skip them under Python 2 via the hack in ``Tests/run_tests.py``.

I would still favour moving these tests into ``Tests/test_SeqFeature.py`` so as not to have the docstrings overly long and busy.

peterjc:
We're note yet enforcing the dots at the end of docstring lines via flake8, but that is complaining about some other minor style issues:

```
$ flake8 Tests/
Tests/test_SeqFeature.py:140:1: E302 expected 2 blank lines, found 1
Tests/test_SeqFeature.py:151:1: W391 blank line at end of file
```

Note we have stricter flake8 settings on ``Bio/`` than on ``Tests/``, updating the coding style there was seen as a lower priority.

peterjc:
Style tests passed 👍

PR #1901, opened 2019-01-19
Update run_tests.py

rwst:


rwst:
I would be grateful for a squash after all this.

peterjc:
I agree a squash commit would be sensible here.

Were we really skipping that many doctests? Wow. Perhaps rather a positive list like this, we can switch to testing them all by default and listing modules to exclude from doctests.

rwst:
https://stackoverflow.com/questions/42340772/how-can-i-integrate-doctests-with-unittests-test-discovery

Python is not my primary language so I can't recommend one of these solutions. Perhaps we can implement exclusion via `doctest.DocTestSuite` in this PR and you open a ticket for a different implementation using one of the newer tools mentioned in that link. Agree?

peterjc:
I meant at some point we'll have increased ``DOCTEST_MODULES`` to cover more than half the modules with working doctests in Biopython (very long include or white list), at which point it would be simpler to have a shorter list of excluded modules (aka a black list).

Handling that as a separate issue/pull request make sense.

peterjc:
In answer to my earlier question, no - not all of the modules being added to ``DOCTEST_MODULES`` here actually have any doctests. e.g. ``Bio.AlignIO.MafIO`` has none. Including it here is harmless.

However, others do, like ``Bio.AlignIO.FastaIO`` and ``Bio.AlignIO.MauveIO`` (although the later is not currently included here?)

rwst:
`MauveIO` is one of the failing ones (#1902). An obstacle to merging this PR would be that other already tested but unmerged PRs might introduce failing doctests.

peterjc:
Does that mean the list on #1902 is the only failures (plus special cases handled already like Python 3 specific)? That's much better than I thought.

rwst:
Yes the #1902 list and the list  `run_tests` should add to a list of modules (or at least of modules with doctests).

rwst:
Thanks.

peterjc:
Thank you - and hopefully my comments on #1902 make sense too.

Issue #1914, opened 2019-01-31
Vector division broken in python 3.x


sconstable:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

```
python
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys; print(sys.version)
3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-4.18.0-13-generic-x86_64-with-Ubuntu-18.04-bionic
>>> import Bio; print(Bio.__version__)
1.73
```

### Expected behaviour

I should be able to write `v = Vector(2, 2, 2) / 2` to divide a vector by a scalar.

### Actual behaviour

An error is raised: `TypeError: unsupported operand type(s) for /: 'Vector' and 'int'`
It appears to be due to the fact that BioPython supplied `__div__` for Vectors, but in python 3.x we need to supply `__truediv__` and `__floordiv__` instead.  My current work-around is to call `Vector(2, 2, 2).__div__(2)`

### Steps to reproduce

simply divide a vector by an int.

platinhom:
You can add `Bio.PDB.Vector.__truediv__ = Bio.PDB.Vector.__div__` before use.  

 Biopython should modify the `Bio/PDB/vectors.py`, add `from __future__ import division` and modify `__div__` to `__truediv__`.

peterjc:
This clearly highlights that vector division is missing in our unit tests, visible here (the ``__contains__`` method also looks untested):

https://codecov.io/gh/biopython/biopython/src/master/Bio/PDB/vectors.py

Ideally we'd add something like @sconstable's example to ``Tests/test_PDB_vectors.py`` as part of fixing this.

@platinhom Your fix sounds perfect - do you want to make a pull request?

peterjc:
It turns out a Python 2 and 3 fix was a little different, but you got there with #1920. Well done.

Note for next time, GitHub lets you update a pull request by updating the associated branch, so you don't have to make new pull requests.

peterjc:
Fixed by #1920, thanks for reporting this issue.

PR #1920, opened 2019-02-02
Vector division broken in python 3.x #1914

platinhom:
This pull request addresses issue #1914 

### DIFF
--- a/Bio/PDB/vectors.py
+++ b/Bio/PDB/vectors.py
@@ -281,6 +281,8 @@ class Vector(object):
         a = self._ar / numpy.array(x)
         return Vector(a)
 
+    __truediv__ = __div__
+
     def __pow__(self, other):
         """Return VectorxVector (cross product) or Vectorxscalar."""
         if isinstance(other, Vector):
--- a/Bio/PDB/vectors.py
+++ b/Bio/PDB/vectors.py
@@ -276,12 +276,13 @@ class Vector(object):
         """Return Vector.Vector (dot product)."""
         return sum(self._ar * other._ar)
 
-    def __div__(self, x):
+    def __truediv__(self, x):
         """Return Vector(coords/a)."""
         a = self._ar / numpy.array(x)
         return Vector(a)
 
-    __truediv__ = __div__
+    # For Python 2:
+    __div__ = __truediv__
 
     def __pow__(self, other):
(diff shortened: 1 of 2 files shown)


peterjc:
Looks good - I'll allow a few days in case @etal or any of the other PDB experts wants to comment, but hope to merge this during the coming week.

platinhom:
I'm not familiar with the github pull requests and the code check 😑

peterjc:
@platinhom it was new to me once, and Github improves it over time - so I keep learning by doing.

Bio/PDB/vectors.py
```diff
@@ -281,6 +281,8 @@ def __div__(self, x):
         a = self._ar / numpy.array(x)
         return Vector(a)
 
+    __truediv__ = __div__
```
lennax: What if we do the opposite? Rename the current method to `__truediv__` and assign `__div__ = __truediv__`. That way, when we drop 2.x support, we can just remove `__div__`.

Tests/test_PDB_vectors.py
```diff
@@ -36,6 +36,16 @@ def test_legacy_imports(self):
...
 
 
+class VectorTests(unittest.TestCase):
+    """Tests for the Vector class."""
+
+    def test_division(self):
+        """Confirm division works."""
+        from Bio.PDB.Vector import Vector
+        v = Vector(2, 2, 2) / 2
+        self.assertEqual(repr(v), "<Vector 1.00, 1.00, 1.00>")
```
lennax: I'm probably being too picky but this test wouldn't fail if the vector was somehow divided using floor division then cast to int. I'd use example numbers that would look even more strange if floor division were used (e.g. the classic `1 // 2 == 0`).

peterjc: I wondered that too - this would be clearer, good idea.

Bio/PDB/vectors.py
peterjc: That seems a worthwhile improvement.

peterjc:
Sorry, I would have merged this as is, but if you're willing to polish this further based on @lennax's useful comments, that'd be even better.

peterjc:
The tests passed, but GitHub seems a bit slow to show that. Odd.

peterjc:
Changes applied - thank you both!

Issue #1921, opened 2019-02-02
Get PDB idcode (a.k.a. pdbid) from PDB header


rob-miller:
### Setup

This is an enhancement request.  I am using Biopython version, Python version, and operating
system as follows:

3.7.2 (default, Jan 13 2019, 12:50:01) 
[Clang 10.0.0 (clang-1000.11.45.5)]
CPython
Darwin-18.2.0-x86_64-i386-64bit
1.74.dev0   (I fixed mine already)

### Desired behaviour

The 4-character idCode (PDBID) is a unique reference for a specific structure within the PDB.  See [PDB format docs](http://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#HEADER).  Sources such as the Biopython tutorial pass this as an argument to PDBParser as id, 'the id that will be used for the structure'.  For bulk processing of lists of PDB files it would be nice to have it inside the data structure if it is available.

### Actual behaviour

The idCode is not part of the header dictionary when loaded by PDBParser()

### Steps to reproduce

    PDB_parser = PDBParser(PERMISSIVE=False, QUIET=True)
    structure = PDB_parser.get_structure('prot', f)
    for k,v in structure.header.items():
        print(k,v)

### Fix

The following diff adds what I believe is the desired behaviour to PDB/parse_pdb_header.py:

    diff --git a/Bio/PDB/parse_pdb_header.py b/Bio/PDB/parse_pdb_header.py
    index 9d86c039d..1166ac4d7 100644
    --- a/Bio/PDB/parse_pdb_header.py
    +++ b/Bio/PDB/parse_pdb_header.py
    @@ -183,6 +183,7 @@ def _parse_pdb_header_list(header):
         dict = {
             'name': "",
             'head': '',
    +        'idcode': '',
             'deposition_date': "1909-01-08",
             'release_date': "1909-01-08",
             'structure_method': "unknown",
    @@ -219,6 +220,9 @@ def _parse_pdb_header_list(header):
                 rr = re.search(r"\d\d-\w\w\w-\d\d", tail)
                 if rr is not None:
                     dict['deposition_date'] = _format_date(_nice_case(rr.group()))
    +            rr = re.search(r"\s+\w\w\w\w\s*\Z",tail)
    +            if rr is not None:
    +                dict['idcode'] = rr.group(0)
                 head = _chop_end_misc(tail).lower()
                 dict['head'] = head
             elif key == "COMPND":

[...]

PR #1933, opened 2019-02-17
Extract PDB ID in Bio/PDB/parse_pdb_header.py

rob-miller:
This pull request addresses issue #1921 

Add the idCode (a.k.a. PDBid) to the header dictionary when loaded by PDBParser()

### DIFF
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -183,6 +183,7 @@ def _parse_pdb_header_list(header):
     dict = {
         'name': "",
         'head': '',
+        'idcode': '',
         'deposition_date': "1909-01-08",
         'release_date': "1909-01-08",
         'structure_method': "unknown",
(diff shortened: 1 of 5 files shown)


peterjc:
The code seems clear, although I've not looked at the PDB header line format to see if a regular expression is the best approach here or not. @lennax @etal @jgreener64  etc any comment? If not, I'll merge this at the end of the week.

Rob: Do you want to add yourself to the CONTRIB and NEWS files?

Note we'd want to do a squash-and-merge to clean up the git history.

Bio/PDB/parse_pdb_header.py
```diff
@@ -219,6 +220,9 @@ def _parse_pdb_header_list(header):
             rr = re.search(r"\d\d-\w\w\w-\d\d", tail)
             if rr is not None:
                 dict['deposition_date'] = _format_date(_nice_case(rr.group()))
+            rr = re.search(r"\s+\w\w\w\w\s*\Z", tail)
```
lennax: The first character of a PDB ID is always digit 1-9:

http://www.wwpdb.org/documentation/file-format-content/format33/sect1.html#Field

Also, `\w` matches _, which isn't a valid character in a PDB ID.

I'd suggest `r"\s+([1-9][0-9A-Z]{3})\s*\Z"` and using `group(1)` to exclude the whitespace.

rob-miller: Yes definitely, thank you for both suggestions.  Also updated NEWS and CONTRIB as suggested.

lennax: Thanks for updating the regex. It would also be nice to add one line to an existing unittest in `test_PDB_parse_pdb_header.py`.

rob-miller: Done.  Thank you for getting me to learn about the testing framework :-), I think implementation I've done is correct..

peterjc:
Thank you all.

Issue #1882, opened 2018-12-24
mysql keyword rank


lhovo:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

```3.6.7 (default, Dec 20 2018, 01:14:33) 
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)]
CPython
FreeBSD-11.2-RELEASE-p4-amd64-64bit-ELF
1.73

mysql  Ver 8.0.12 for FreeBSD11.2 on amd64 (Source distribution)
```

Running code

```from Bio import Entrez
from Bio import SeqIO
handle = Entrez.efetch(db="nuccore", id="6273291,6273290,6273289", rettype="gb", retmode="text")
for seq_record in SeqIO.parse(handle, "genbank") :
    print(seq_record.id, seq_record.description[:50] + "...")
    print("Sequence length %i," % len(seq_record.seq),)
    print("from: %s" % seq_record.annotations['source'])
handle.close()
```

### Expected behaviour

To insert records as per example
https://biopython.org/wiki/BioSQL

### Actual behaviour

Failed to insert record due to table name "rank", is now a keyword in mysql?

### Steps to reproduce

I believe in mysql 8.0 'rank' is now a keyword,
by escaping all instances of `rank` in https://github.com/biopython/biopython/blob/master/BioSQL/Loader.py
I was able to get this example to work.

Unsure if this will cause issues with other sql databases if the Insert table names are escaped.

peterjc:
This makes sense, according to https://dev.mysql.com/doc/refman/8.0/en/keywords.html

> RANK (R); added in 8.0.2 (reserved)

How exactly did you escape the table's column name? In words or better yet as a pull request would be great?

If this is simple we probably don't need to change the schema itself - which would be annoying as BioSQL is shared across Bio* projects.

lhovo:
Sorry my code was rather hackey as I only escaped the rank key words.

I would expect to do this correctly the code should escape all table names like so.

```
sql = "INSERT INTO bioentry_reference (`bioentry_id`, `reference_id`," \
              " `start_pos`, `end_pos`, `rank`) VALUES (%s, %s, %s, %s, %s)"
```

peterjc:
[...]

PR #1934, opened 2019-02-17
Escape rank in BioSQL for MySQL 8 support

nimne:
This pull request addresses issue #1882

These changes escape the new MySQL keyword `rank`. This works locally on MySQL 8.0.15 installed on  linux. Unfortunately I don't think that it is simple or straightforward to enable MySQL 8 on Travis CI.

Changes were needed to the BioSQL file itself. I've chosen to put `rank` in double quotes, this required adding `SET sql_mode='ANSI_QUOTES';` to the top of the file. Alternatively, `rank` could go in backticks without needing to set `ANSI_QUOTES` as this is a MySQL specific file. I'm not sure what is the best style.

### DIFF
--- a/BioSQL/BioSeq.py
+++ b/BioSQL/BioSeq.py
@@ -252,7 +252,7 @@ def _retrieve_dbxrefs(adaptor, primary_id):
         "SELECT dbname, accession, version"
         " FROM bioentry_dbxref join dbxref using (dbxref_id)"
         " WHERE bioentry_id = %s"
-        " ORDER BY rank", (primary_id,))
+        " ORDER BY \"rank\"", (primary_id,))
     for dbname, accession, version in dbxrefs:
         if version and version != "0":
             v = "%s.%s" % (accession, version)
@@ -263,10 +263,10 @@ def _retrieve_dbxrefs(adaptor, primary_id):
 
 
 def _retrieve_features(adaptor, primary_id):
-    sql = "SELECT seqfeature_id, type.name, rank" \
+    sql = "SELECT seqfeature_id, type.name, \"rank\"" \
           " FROM seqfeature join term type on (type_term_id = type.term_id)" \
           " WHERE bioentry_id = %s" \
-          " ORDER BY rank"
+          " ORDER BY \"rank\""
     results = adaptor.execute_and_fetchall(sql, (primary_id,))
     seq_feature_list = []
     for seqfeature_id, seqfeature_type, seqfeature_rank in results:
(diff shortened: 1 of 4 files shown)


peterjc:
First a minor stylistic point, rather than escaping the double quotes:

```python
    sql = "SELECT seqfeature_id, type.name, \"rank\"" \
          " FROM seqfeature join term type on (type_term_id = type.term_id)" \
          " WHERE bioentry_id = %s" \
          " ORDER BY rank"
```

how do you feel about just switching the outer quotes to single quotes:

```python
    sql = 'SELECT seqfeature_id, type.name, "rank"' \
          " FROM seqfeature join term type on (type_term_id = type.term_id)" \
          " WHERE bioentry_id = %s" \
          " ORDER BY rank"
```

peterjc:
My main query - do you have to modify ``Tests/BioSQL/biosqldb-mysql.sql`` for this to work?

That file is meant to be an exact copy of the standard schema from https://github.com/biosql/biosql so it is not something we (Biopython) should change unilaterally. We'd need to have any such change approved as part of https://github.com/biosql/biosql/issues/22

nimne:
In terms of the quotes, I'm happy to use the single quote 'wrappers' to make things more readable. Do you think that all of the quote blocks should be single quotes (for consistency) or only the ones with the escaped `rank`? I'm happy to do either.

As for the `Tests/BioSQL/biosqldb-mysql.sql`, it does need to be modified. `rank` can no longer be non-escaped in any MySQL query. I'm happy to submit a PR over at [BioSQL](https://github.com/biosql/biosql/) before this gets fixed here. Perhaps someone there would have more insight into weather the MySQL file should use backticks or set `ANSI_QUOTES` and use double quotes for `rank`.

nimne:
Made pull request [over at BioSQL](https://github.com/biosql/biosql/pull/23). After thinking about it, the backticks might make more sense in the MySQL specific file than double quotes. I'll wait to make changes here until that PR is resolved.

peterjc:
With your BioSQL pull request https://github.com/biosql/biosql/pull/23 applied, and the BioSQL issue about the reserved keyword ``rank`` https://github.com/biosql/biosql/issues/22 resolved, this needs a minor update.

i.e. Use the newly updated MySQL schema as accepted in BioSQL

nimne:
Just updated files to address the quotation style, and the MySQL schema!

peterjc:
Looks good - the automated testing should be a formality...

peterjc:
I could have merged that earlier, but done now. Thank you!

PR #1936, opened 2019-02-19
Remove redundant self test from Bio.GenBank.Scanner 

svalqui:


peterjc:
Thank you for working on this - the initial pull request is a good start, but I'd like to see some large changes (in terms of lines of code - conceptually the change is small).

Currently we do use ``EmblScanner`` and ``GenBankScanner`` in ``Tests/test_GenBank_unittest.py``, but only in a slightly artificial way to test parsing first line. We don't directly test the ``parse_records`` method nor ``parse_cds_features`` method which is done in the self test. That's the bit I think we need to move from the self test into the unittest framework, and your pull request does that :)

However, I don't think we necessarily have to use the same GenBank and EMBL examples to do this - probably many of the existing examples under ``Tests/GenBank/`` which could be used instead?

If there is something interesting about these test files which makes them worth retaining, then I don't see any reason to keep them as text embedded in the Python code, when as part of the main test framework they could more usefully be files under ``Tests/GenBank/``. That way they can be used in other tests more easily.

So, please either remove the embedded files completely (and test using existing GenBank or EMBL test files), or move the embedded files to files on disk.

svalqui:
Ok, no worries I'll review the files available and update.

Tests/test_GenBank_unittest.py
```diff
@@ -688,6 +689,145 @@ def test_000_write_invalid_but_parsed_locus_line(self):
...
+        return l_cds_f
+
+    def gb_to_l_r(self, filename, do_features=False):
+        """Gb file to Seq list parse records."""
+
+        handle = open(filename, 'r')
+        if do_features:
+            l_gb_r = list(self.gb_s.parse_records(handle, do_features=True))
+        else:
+            l_gb_r = list(self.gb_s.parse_records(handle, do_features=False))
```
peterjc: Wouldn't one line instead of four be enough:

```python
l_gb_r = list(self.gb_s.parse_records(handle, do_features= do_features))
```

Tests/test_GenBank_unittest.py
```diff
@@ -688,6 +689,145 @@ def test_000_write_invalid_but_parsed_locus_line(self):
...
+            l_cds_f = list(self.gb_s.parse_cds_features(handle, tags2id=tags2id))
+        else:
+            l_cds_f = list(self.gb_s.parse_cds_features(handle))
+        handle.close()
+        return l_cds_f
+
+    def gb_to_l_r(self, filename, do_features=False):
+        """Gb file to Seq list parse records."""
+
+        handle = open(filename, 'r')
```
peterjc: Could you use a context manager here please,

```python
    # code
```

not:

```python
handle = open(filename)
# code
handle.close()
```

Also there is no need to specify the mode since "r" is the default.

Tests/test_GenBank_unittest.py
```diff
@@ -688,6 +689,145 @@ def test_000_write_invalid_but_parsed_locus_line(self):
...
+        self.assertEqual(l_r[0].name, 'NC_000932')
+        self.assertEqual(l_r[0].description, 'Arabidopsis thaliana chloroplast, '
+                                             'complete genome')
+        self.assertEqual(len(l_r[0].features), 259)
+
+    def test_embl_cds_interaction(self):
+        """Test EMBL CDS interaction, parse CDS features on embl files."""
+
+        # Test parse CDS features on embl_file
+        handle_embl7046 = open("EMBL/AE017046.embl", "r")
```
peterjc: Again, context manager please:

```python
        with open("EMBL/AE017046.embl") as handle:
            embl_s = Scanner.EmblScanner()
            l_cds_f = list(embl_s.parse_cds_features(handle))
```

Tests/test_GenBank_unittest.py
```diff
@@ -688,6 +689,145 @@ def test_000_write_invalid_but_parsed_locus_line(self):
...
+        self.assertEqual(len(l_cds_f), 10)
+        # Seq ID
+        self.assertEqual(l_cds_f[0].id, 'AAS58758.1')
+        self.assertEqual(l_cds_f[0].description, 'putative transposase')
+
+    def test_embl_record_interaction(self):
+        """Test EMBL Record interaction on embl files."""
+
+        #  Test parse records on embl_file
+        handle_embl7046 = open("EMBL/AE017046.embl", "r")
```
peterjc: Again, please use a context manager.

peterjc:
Are you happy to do a little more polishing? Using context managers is recommend style over explicitly closing file handles.

Tests/test_GenBank_unittest.py
svalqui: yes, amended.

Tests/test_GenBank_unittest.py
svalqui: Using context manager for all now; all good, I think is ready now.

peterjc:
Merged, thank you!

Issue #1982, opened 2019-03-26
UnicodeDecodeError  when reading ABI file with Python 3.6 but not with Python 2.7


zqfang:
### Setup

I am reporting a problem with Biopython version 1.73, Python version 3.6, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```

3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 02:16:08) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
CPython
Darwin-17.7.0-x86_64-i386-64bit
1.73

### Expected behaviour

These code works in python 2:
```python
In [8]: from Bio import SeqIO

In [9]: record = SeqIO.read("./8s11-KO-F1_E08.ab1","abi")

In [10]: record.seq
Out[10]: Seq('NNNNNNNNNNCTCGAGCGCTCGGCGGAGACACTAAGGCGGCCCGGGGCGCGGCG...CAG', IUPACUnambiguousDNA())
```

### Actual behaviour
The same code caused errors in python 3:
```
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-27-79ec5da732f7> in <module>
----> 1 SeqIO.read("./8s11-KO-F1_E08.ab1",format='abi')

~/miniconda/envs/snakeflow/lib/python3.6/site-packages/Bio/SeqIO/__init__.py in read(handle, format, alphabet)
    718     iterator = parse(handle, format, alphabet)
    719     try:
--> 720         first = next(iterator)
    721     except StopIteration:
    722         first = None

~/miniconda/envs/snakeflow/lib/python3.6/site-packages/Bio/SeqIO/__init__.py in parse(handle, format, alphabet)
    653             raise ValueError("Unknown format '%s'" % format)
    654         # This imposes some overhead... wait until we drop Python 2.4 to fix it
--> 655         for r in i:
    656             yield r
    657 

~/miniconda/envs/snakeflow/lib/python3.6/site-packages/Bio/SeqIO/AbiIO.py in AbiIterator(handle, alphabet, trim)
    373 
    374     raw = dict()
--> 375     for tag_name, tag_number, tag_data in _abi_parse_header(header, handle):
    376         key = tag_name + str(tag_number)
    377 

[...]

PR #1984, opened 2019-03-27
Ignore ABIF tags with unicode encoding problems

peterjc:
Instead, just give a warning. Unless in a critical tag like sequence identifier, should be OK.

This pull request is a partial work around for issue #1982

### DIFF
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -18,9 +18,11 @@ http://www6.appliedbiosystems.com/support/software_community/ABIF_File_Format.pd
 
 import datetime
 import struct
+import warnings
 
 from os.path import basename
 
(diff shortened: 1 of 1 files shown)


peterjc:
@bow do you want to apply this as is, as a short term solution until you have time to work on the proposed long term solution?

bow:
@peterjc Sorry for the delay ~ I have been away the last few days. I am pushing a new change sometime today. Most of the changes, as you have mentioned, has to do with the tests. All of the string value are kept as bytes (these are data types with element codes 2, 18, and 19 as described in [the specs](https://projects.nfstc.org/workshops/resources/articles/ABIF_File_Format.pdf)),  except for where we need the values as strings to conform to the `SeqRecord` arguments.

peterjc:
I think this is no longer needed with #2022 applied.

PR #1987, opened 2019-03-28
Add parse_index attribute to HSPs

bow:
This pull request is related to #1965 and adds an HSP attribute for SearchIO objects to store the order by which the HSP appears in the output file given to SearchIO.

### DIFF
--- a/Bio/SearchIO/_model/hsp.py
+++ b/Bio/SearchIO/_model/hsp.py
@@ -211,7 +211,7 @@ class HSP(_BaseHSP):
     +------------+---------------------------------------------------+
 
     In addition to the objects listed above, HSP objects also provide the
-    following properties:
+    following properties and/or attributes:
 
     +--------------------+------------------------------------------------------+
     | Property           | Value                                                |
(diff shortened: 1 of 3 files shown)


bow:
@linucks, continuing from your point in #1965, here I am using `output_index`.

I wouldn't mind `rank` actually, but I worry that `rank` may be used as an actual attribute of an HSP result. So I prepended `parse_` because it seems less likely to be used as an actual attribute name. I do understand that the parsing order itself may be different, but I would still like to reduce the chance name collision.

How does `output_index` sound?

linucks:
Hi @bow. That's definitely better, but it still seems somewhat tied to how the results were generated. The best I could think of was global_rank or global_index as it's global within the entire search, but not within the hit. However, I'm not sure that's much better then output_index and maybe it's easier for people to understand if the name is actually related to how the results were produced, so I'd be fine with output_index.

bow:
Ok then, let's start with `output_index`. I've just updated `NEWS.rst` with a short description of the change and will merge this as soon as the build succeeds. We could then continue back in #1965.

PR #1990, opened 2019-03-29
Increase test coverage for pairwise2

MarkusPiotrowski:
Added some more tests to increase coverage of pairwise2. Including one additional test module, which allows to prevent the import of the C modul for testing the implemented fallback Python functions.

### DIFF
--- a/Tests/test_pairwise2_no_C.py
+++ b/Tests/test_pairwise2_no_C.py
@@ -14,7 +14,7 @@ from Bio.SubsMat.MatrixInfo import blosum62
 
 sys.modules['Bio.cpairwise2'] = None
 
-from Bio import pairwise2
+from Bio import pairwise2  # noqa: E402
 
 
 class TestPairwiseGlobal(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


MarkusPiotrowski:
I don't understand Codecov.... The main change should have an effect on lines ~600 - 700 in ```pairwise2```, but these lines are simply missing in the report: https://codecov.io/gh/biopython/biopython/pull/1990/src/Bio/pairwise2.py#L407

peterjc:
Not sure what Codecov is doing with the missing lines. On the master right now it should missing tests for function ``_make_score_matrix_fast`` in that part of the file just fine:

https://codecov.io/gh/biopython/biopython/src/master/Bio/pairwise2.py

Any strong views on filenames? ``test_pairwise2_wo_C.py`` where ``wo`` without? How about just ``test_pairwise2_no_C.py``?

MarkusPiotrowski:
@peterjc Just to be sure: I don't need to 'register' a new test file somewhere to get it included in Codecov?

PR #1993, opened 2019-04-01
Improve test coverage for module Bio.NMR

svalqui:


Bio/NMR/xpktools.py
```diff
@@ -179,18 +179,18 @@ def residue_dict(self, index):
...
-            outfile.write("\012")
-            outfile.write(self.dataset)
-            outfile.write("\012")
-            outfile.write(self.sw)
-            outfile.write("\012")
-            outfile.write(self.sf)
-            outfile.write("\012")
-            outfile.write(self.datalabels)
-            outfile.write("\012")
+            outfile.write(self.firstline.encode())
```
peterjc: This looks wrong - why add ``.encode()`` to all the lines, can't you write the file in text mode?

This file dates back well before Python 3 - most likely using binary mode was an oversight, or is to do with Windows new lines?

Doc/examples/nmr/simplepredict.py
```diff
@@ -123,14 +122,14 @@
...
-        # specific fields in the that make the information
+        # specific fields in the file that make the information
         # more readable and suitable for creating data tables
         # This output will be printed to the screen.
         # The data table contains the assignment, coordinates and
         # intensity of the resonance.
 
-        print(string.split(entry1.fields["15N2.L"], ".")[0], "-->",
-              string.split(entry1.fields["N15.L"], ".")[0], "\t",
+        print(str.split(entry1.fields["15N2.L"], ".")[0], "-->",
```
peterjc: You are right the old code using the ``string`` module ought to be updated, but why not use the string method?:

```
    print(entry1.fields["15N2.L"].split("."), "-->",
```

Bio/NMR/xpktools.py
svalqui: Because it rises an Type Error; I had to change for it to run, I'm using python 3.6.

Doc/examples/nmr/simplepredict.py
svalqui: It causes an AttributeError: module 'string' has no attribute 'split'; so had to change it.

svalqui: Sorry, I had a typo amending now.

peterjc: Heh. I had a typo too, missed out the ``[0]`` bit.

peterjc:
TravisCI is failing due to #1995, not your fault.

peterjc:
We've fixed the #1995 issue. In the meantime, there have been changes causing a merge conflict.

Are you familiar with ``git rebase`` to update your branch onto the latest version of the master branch?

svalqui:
Re-based, I think it has updated all. Cheers,

peterjc:
I'm not sure what exactly you did, but GitHub reports 304 files edited which is not what I expected... I just did a rebase and force updated the branch and thus this pull request.

Bio/NMR/xpktools.py
peterjc: Does changing ``with open(outfn, 'wb') as outfile:`` to ``with open(outfn, 'w') as outfile:`` work (instead of adding the ``.encode()`)?

svalqui: It doesn't work, it seems as there is no difference of using 'wb'or 'b', if I use open(outfn, 'w') raises TypeError as below:
TypeError: a bytes-like object is required, not 'str'

svalqui:
Sorry about that, and thank you for fixing it; I had some issues with rebase in the past so I tried rebase from Pycharm, did not showed issues but created all those entries, I'll keep using rebase from CLI in the future.

Bio/NMR/xpktools.py
peterjc: I'm confused, ``self.firstline`` should be string not bytes since the parsing is done in text mode.

As an aside, the use of ``"\012"`` is unusual, ``"\n"`` would be preferable  style wise.

peterjc: The only reason I can think of to explicitly use "\012" or similar is to avoid getting CR LF on Windows - does the command line tool work on Windows? Can you test that? Perhaps leaving it as "\012" would be safer after all.

svalqui: I'll test on Windows, is going to take few days.

svalqui: I manage to remove the "encode" and "b", I think that was my mistake, as I have been running from install while testing DocTest, I've changed the test to check the content instead, as the file size reports different, from the reference created in linux with the file created on Windows, I think  due to CR LF, as content is the same. Now is working on Linux Ubuntu 18.04, python 3.6 and Windows 7 Python 3.7.
Could you please check the merge; the updated commits appear now on my branch, I do not remember seen this on previous merges.

peterjc:
The history is a little confused again, the merge might be part of it. I'll try rebasing it again now...

peterjc:
The rebase proved painful, a squash merge into master seems to have worked - branch updated.

peterjc:
Sorry, should have merged this already. Thank you!

Issue #1839, opened 2018-10-31
Add support for NCBI BLAST XML v2


abkomissarov:
Hi, I'm now learning how to use BioPython for parsing BLAST XML output files. I have got a problem with parsing. Below is my code and the output error - "AttributeError: 'BlastParser' object has no attribute '_blast'". I have no idea how to fix it. Could anyone help me? Thanks a lot in advance.
Sorry if issue formatting is not OK. It's my first issue raised.
XML example file is attached.

python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609]
CPythonLinux-4.4.0-127-generic-x86_64-with-Ubuntu-16.04-xenial
BioPython 1.72
[XML-example.zip](https://github.com/biopython/biopython/files/2535265/XML-example.zip)

### My code
`from Bio.Blast import NCBIXML

try:
	handle=open(input('Please select XML file: '))
except IOError:
	print('Selected file does not exist')

blast_records=NCBIXML.parse(handle)
print(type(blast_records))

for record in blast_records:

	for alignment in blast_record.alignments:
		for hsp in alignment.hsps:
			if hsp.expect <0.01:
				print('****Alignment****')
				print('sequence:', alignment.title) 
				print('length:', alignment.length)
				print('score:', hsps.score)
				print('gaps:', hsps.gaps)
				print('e value:', hsp.expect)
				print(hsp.query[0:90] + '...')
				print(hsp.match[0:90] + '...')
				print(hsp.sbjct[0:90] + '...')`

### Output

<class 'generator'>
Traceback (most recent call last):
  File "./XML_open.py", line 15, in <module>
    for record in blast_records:
  File "/home/andrey/.local/lib/python3.5/site-packages/Bio/Blast/NCBIXML.py", line 631, in parse
    expat_parser.Parse(text, False)
  File "../Modules/pyexpat.c", line 414, in StartElement
  File "/home/andrey/.local/lib/python3.5/site-packages/Bio/Blast/NCBIXML.py", line 67, in startElement
    getattr(self, method)()
  File "/home/andrey/.local/lib/python3.5/site-packages/Bio/Blast/NCBIXML.py", line 373, in _start_Hit
    self._blast.alignments.append(Record.Alignment())
AttributeError: 'BlastParser' object has no attribute '_blast'

peterjc:
You do not have the original BLAST XML output, you have BLAST XML v2 which
we do not yet support:

```xml
[...]

Issue #1995, opened 2019-04-02
Tests failing with AttributeError: module 'PIL.Image' has no attribute 'VERSION'


peterjc:
Noticed on TravisCI, it looks like a new release of Pillow and/or ReportLab broke things:

e.g. Python 2.7 and 3.5 here:
https://travis-ci.org/biopython/biopython/builds/514538956

```
======================================================================
ERROR: test_render_KGML_basic (test_KGML_graphics.KGMLPathwayTest)
Basic rendering of KGML: write to PDF without modification.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/test_KGML_graphics.py", line 115, in test_render_KGML_basic
    kgml_map.draw(p.output_stem + '_original.pdf')
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-2.7/Bio/Graphics/KGML_vis.py", line 158, in draw
    self.drawing.drawImage(imfilename, 0, 0)
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/pdfgen/canvas.py", line 947, in drawImage
    imgObj = pdfdoc.PDFImageXObject(name, image, mask=mask)
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/pdfbase/pdfdoc.py", line 2087, in __init__
    self.loadImageFromA85(src)
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/pdfbase/pdfdoc.py", line 2095, in loadImageFromA85
    imagedata = pdfutils.makeA85Image(source,IMG=IMG,detectJpeg=True)
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/pdfbase/pdfutils.py", line 38, in makeA85Image
    raw = img.getRGBData()
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/lib/utils.py", line 902, in getRGBData
    annotateException('\nidentity=%s'%self.identity())
  File "/home/travis/build/biopython/biopython/.tox/py27-cover/lib/python2.7/site-packages/reportlab/lib/utils.py", line 1406, in annotateException
    rl_reraise(t,v,b)
[...]

PR #1997, opened 2019-04-03
Add support for NCBI BLAST XML v2

rtf-const:
This pull request addresses issue #1839


peterjc:
This looks impressive - I see you've split the existing BLAST XML parser into two using subclasses, did that seem easiest?

Does this mean adding support for BLAST XML v2 to ``Bio.SearchIO`` should be really simple to include?

rtf-const:
I do not use subclasses for parsing different XML versions, because parser doesn't know XML version beforehand. Instead actuall version is determined in the process of parsing based on root tag name. Blast XML v1 and v2 are very similar in structure and mainly differ just by tags names. So the process is very simple: start reading XML, look at the name of root node, determine version and then code setups the dictionary that matches XML tag name with the method that process it.

As I can see Bio.SearchIO already uses some constant dictionaries that match tag name with a resulting objects properties. So for adding  xml v2 support in SerachIO we can use different dictionaries depending on XML version.

peterjc:
TravisCI failed on a ReportLab issue, since fixed on the master - see #1995. If you add yourself to the ``NEWS.rst`` file or made any other change that would trigger a rebuild.

@bow are you available to review this work?

bow:
@peterjc I'm less familiar with the internals of `Bio.Blast`, but from what I understand, the changes look ok.

And yes, SearchIO does indeed match XML tag names with object properties, since XML parsing is handled by `xml.etree.ElementTree`. This can be extended to v2 as well.

peterjc:
Rebased to deal with a trivial merge conflict in the NEWS file with the contributor names.

I think this deserves a short ``NEWS.rst`` entry too - although if it is now trivial to add BLAST v2 XML to SearchIO, we could merge this as is, and do that next?

bow:
Sounds like a good idea. It's also nice that @rtf-const already added the test files for v2 ~ which we can use for SearchIO as well :).

PR #2000, opened 2019-04-04
Added docstrings D102 to public method on Bio.Crystal

svalqui:
This pull request addresses issue #1963 

### DIFF
--- a/Bio/Crystal/__init__.py
+++ b/Bio/Crystal/__init__.py
@@ -18,10 +18,13 @@ from Bio._py3k import basestring
 
 
 class CrystalError(Exception):
+    """Class to manage errors."""
+
     pass
 
 
 def wrap_line(line):
+    """Add end of line."""
     output = ''
     for i in range(0, len(line), 80):
         output += '%s\n' % line[i: i + 80]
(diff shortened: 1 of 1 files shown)


Bio/Crystal/__init__.py
```diff
@@ -18,17 +18,21 @@
 
 
 class CrystalError(Exception):
+    """Class to manage errors."""
+
     pass
 
 
 def wrap_line(line):
+    """Add end of line."""
```
peterjc: The point of this function is to insert line breaks for line wrapping at 80characters (presumably a requirement of the file format).

svalqui: I have re-word it, I hope is better now; just to note this module seems incomplete, I haven't found something like 'get_structure' to fill the structure, and the code is from 2002, this might be getting  obsolete.

peterjc:
You may be right - having worked on this do you think it makes sense to deprecate ``Bio.Crystal`` now?

svalqui:
Yes, from here http://rna.bgsu.edu/rna3dhub/pdb it seems atlas works having PDB and moving to mmCIF, so any works for crystals should go on Bio.PDB.

peterjc:
OK. Do you want to submit a pull request to deprecate ``Bio.Crystal``? See https://biopython.org/wiki/Deprecation_policy

PR #2003, opened 2019-04-07
fixed PDB header tail trim to match comment.

rob-miller:
This pull request changes _chop_end_misc in parse_pdb_header.py to match the docline better.  

The current implementation is very general (matching 4 spaces optionally followed by anything at the end of the line, `r"\s\s\s\s+.*\Z"`) and fails on pdb2yxy.ent.  Alternatively just change the '+' to a '*' in the current implementation, but this matches the specification in the routine's docline better (`"Chops lines ending with  '     14-JUL-97  1CSA' and the like (PRIVATE)."`)

*Peter: Fixed checkboxes by removing spaces*

### DIFF
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -93,7 +93,7 @@ def _chop_end_codes(line):
 
 def _chop_end_misc(line):
     """Chops lines ending with  '     14-JUL-97  1CSA' and the like (PRIVATE)."""
-    return re.sub(r"\s\s\s\s+.*\Z", "", line)
+    return re.sub(r"\s+\d\d-\w\w\w-\d\d\s+[1-9][0-9A-Z]{3}\s*\Z", "", line)
 
 
 def _nice_case(line):
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -60,6 +60,7 @@ possible, especially the following contributors:
 - Nick Negretti
 - Peter Cock
 - Ralf Stephan
+- Rob Miller
 - Sergio Valqui
 - Antony Lee
 - Wibowo 'Bow' Arindrarto
--- a/Tests/PDB/header.pdb
+++ b/Tests/PDB/header.pdb
@@ -1,0 +1,5 @@
+HEADER    STRUCTURAL GENOMICS, UNKNOWN FUNCTION   08-SEP-08   3EFG              
+TITLE     STRUCTURE OF SLYX PROTEIN FROM XANTHOMONAS CAMPESTRIS PV. CAMPESTRIS  
+TITLE    2 STR. ATCC 33913                                                      
+COMPND    MOL_ID: 1;                                                            
+COMPND   2 MOLECULE: PROTEIN SLYX HOMOLOG;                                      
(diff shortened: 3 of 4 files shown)


peterjc:
Could you clarify the motivating use case? By ``pdb2yxy.ent`` do you mean a recent full PDB download?

rob-miller:
- Sorry, yes the 'pdbNXXX.ent' reference is to a PDB file downloadable from the repository at http://www.rcsb.org/ amongst other sources.
- The use case is I want to read the header line and write it back out so that it can be parsed as a header line again.  To do this in the normal case I combine attributes 'head', 'deposition_date' and 'idcode' from pdbStructure.header with spacing as specified for the PDB format.  The subroutine this PR patches is supposed to remove idcode and depostion_date from the PDB HEADER record, but in the example above there are not enough spaces before deposition_date so nothing gets trimmed (and my code adds redundant deposition_date and idcode fields)
- Is this related to the recent change to capture the 'idcode' field?  No, this is a different subroutine (though without the 'idcode' captured I couldn't write the HEADER record out at all, so in that sense 'yes' :-))
- Are the failing cases valid according to the PDB format specification?  Yes, in fact there is no requirement for any space at all between the 'classification' and the 'deposition_date' fields, and a stricter regex would leave out the initial '\s+'.  I'd be happy to tweak this if there is desire for it, but (personally) I'm more comfortable checking for at least one intervening space.
- How big an issue is this?  In a sample list of 5,824 nonredundant PDB files with resolution 2.2 angstroms or better, I find 567 failure cases.  512 of these have the header text 'STRUCTURAL GENOMICS, UNKNOWN FUNCTION'.

peterjc:
A very detailed reply, thank you 👍 

I wonder if any of those 567 failures are small enough to add as a bundled test case, or if we can tweak an existing test PDB file to cover this? With test coverage I'd be happy to merge this, although I'd prefer input from one of the Biopython developers actively using Bio.PDB - @etal, @lennax, @jgreener64 perhaps?

jgreener64:
Looks like a sensible change.

Regarding testing, you wouldn't necessarily need to have the whole PDB file present. You could just have the header (e.g. of 2yxy) in a file and read it in with `parse_pdb_header`, adding the test to the `test_PDB_parse_pdb_header.py` file.

rob-miller:
I can do either the header alone or a small PDB file - I would propose pdb3efg.ent which is only 69kb (58 residues plus some waters) but it has the offending structural genomics header line and two disordered residues (Glu and Arg backbone CA and sidechain atoms with altlocs).  

I don't see any other real PDB files (that's excluding a_structure.pdb) in Tests/PDB which have disorder, and from my inspection I think a_structure.pdb is only checked for disordered residues and a count of disordered atoms, not actually testing that the atoms are loaded correctly - but perhaps that's a given if the disordered count is correct.

peterjc:
69kb seems an awfully large file for testing just this - the code snippet idea from @jgreener64 seems enough.

rob-miller:
OK, test submitted.  Idea with the 3EFG PDB file was to improve some tests around disordered atoms, but that would be better done in its own PR.

peterjc:
Ah, #1995 is now affecting AppVeyor (was just TravisCI earlier)

peterjc:
You are in the ``CONTRIB.rst`` listing already, do you want to be added to the ``NEWS.rst`` for the upcoming release?

rob-miller:
Sure, thanks.
Rob.

peterjc:
Thank you

PR #2005, opened 2019-04-10
Added docstrings D102 to public method on Bio.Wise

svalqui:
This pull request addresses issue #1963 

### DIFF
--- a/Bio/Wise/dnal.py
+++ b/Bio/Wise/dnal.py
@@ -105,6 +105,7 @@ class Statistics(object):
             self.coords = [(0, 0), (0, 0)]
 
     def identity_fraction(self):
+        """Calculate the fraction of matches."""
         return self.matches / (self.matches + self.mismatches)
 
     header = "identity_fraction\tmatches\tmismatches\tgaps\textensions"
@@ -116,6 +117,7 @@ class Statistics(object):
 
 
 def align(pair, match=_SCORE_MATCH, mismatch=_SCORE_MISMATCH, gap=_SCORE_GAP_START, extension=_SCORE_GAP_EXTENSION, **keywds):
+    """Align a pair of DNA files using dnal and calculate the statistics of the alignment."""
     cmdline = _build_dnal_cmdline(match, mismatch, gap, extension)
     temp_file = Wise.align(cmdline, pair, **keywds)
     try:
@@ -129,6 +131,7 @@ def align(pair, match=_SCORE_MATCH, mismatch=_SCORE_MISMATCH, gap=_SCORE_GAP_STA
 
 
 def main():
+    """Command line implementation."""
     import sys
     stats = align(sys.argv[1:3])
     print("\n".join("%s: %s" % (attr, getattr(stats, attr))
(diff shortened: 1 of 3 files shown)


Bio/Wise/dnal.py
```diff
@@ -116,6 +117,7 @@ def __str__(self):
 
 
 def align(pair, match=_SCORE_MATCH, mismatch=_SCORE_MISMATCH, gap=_SCORE_GAP_START, extension=_SCORE_GAP_EXTENSION, **keywds):
```
MarkusPiotrowski: Can you break the line after ```extension=_SCORE_GAP_EXTENSION```? (To make it <80 char; we do not enforce this yet)

Bio/Wise/psw.py
```diff
@@ -133,7 +143,7 @@ def align(pair,
           gap_start=None,
           gap_extension=None,
           *args, **keywds):
```
MarkusPiotrowski: This may be a matter of taste, but I would put all the arguments in one or two lines (not exceeding 79 characters).

peterjc: Largely personal taste, and we have no compelling reason to change this as part of this pull request.

(Separately I am becoming a fan of the code formatting tool ``black`` which would decide this kind of thing automatically)

Bio/Wise/dnal.py
peterjc: That would be nice, but isn't essential for this pull request focused on the docstrings.

peterjc:
Looks good - Markus is right there are some unrelated style issues we could work on, but nothing which blocks merging your docstring work. Thank you!

MarkusPiotrowski:
A different discussion and yes, offen a matter of personal taste, but I hate what ```black``` would do to with expressions in matching brackets, e.g.:

```
# in:

ImportantClass.important_method(exc, limit, lookup_lines, capture_locals, extra_argument)

# out:

ImportantClass.important_method(
    exc, limit, lookup_lines, capture_locals, extra_argument
)
```
And I find it even more terrible, what it would do to function arguments:
```
# out:

def very_important_function(
    template: str,
    *variables,
    file: os.PathLike,
    engine: str,
    header: bool = True,
    debug: bool = False,
):
    """Applies `variables` to the `template` and writes to `file`."""
    with open(file, "w") as f:
        ...
```

peterjc:
As you say a matter of taste - I had some niggles with black's style, I still don't agree with all the choices, but it has grown on me - and delegating this is liberating (both on solo projects, and likely more so on group projects).  I've filed #2008 as a more visible place to have this discussion.

svalqui:
@MarkusPiotrowski , My apologies, I couldn't include your feedback on time I will consider for future PR.
This PR had take me more than expected as I had difficulties installing wise 2.4.1, to give me a better sense of what the module does, It has not work on Ubuntu 18.04;  wise 2.4.1 seems stale, so I am not sure about the future of Bio.Wise.
Cheers,

MarkusPiotrowski:
@svalqui No problem, everything's fine! I know it's often hard to add meaninful docstrings, because you need to understand what the code is really doing. I also made this experience several times!

Issue #2014, opened 2019-04-18
Emphasize more strongly that contributors should use pre-commit hooks


MarkusPiotrowski:
While Biopython is enforcing many/most coding-style conventions described in PEP-8 and PEP-257, many pull requests get delayed, because contributors (often unintentionally) violate these style conventions.

A discussion about a Biopython specific pre-commit git hook to enforce PEP-8 style has already started some time ago (e.g. #493, #1188).

Using a pre-commit hook in the git environment is an easy way to make contributors aware of style-violations and give them the chance to address these before submitting a pull request.  

However, it seems that many contributors are not using pre-commit hooks, and one reason for this may be that they are not aware of them and/because we are not explicit enough about that they *should* be used. 

There are some places, where we could emphasize this more strongly:

- ```README.rst``` should contain, under 'Contributing', a link to ```CONTRIBUTING.rst```.
- ```CONTRIBUTING.rst``` should strongly recommend the use of a pre-commit hook and show how to install and set it up.
- There should be a clear recommendation, which tools should be used. 
- PR #1188 addresses already some of these topics and should be merged. It contains a script that will setup the git environment with our folder-specific ```.flake8``` settings.
- The suggested changes also apply to our Biopython home page: ```GitUsage.md``` (https://biopython.org/wiki/GitUsage). 
- We can also think about putting another (optional) paragraph in our PR template https://github.com/biopython/biopython/blob/master/.github/PULL_REQUEST_TEMPLATE.md or rephrase the second paragraph to mention pre-commit hooks.

peterjc:
This is a very good idea in principle, but the complexities of our different ``.flake8`` settings in different folders is an annoyance requiring the bespoke git commit script.

I would very much like to address that first, which bring the enormous advantage of letting us use an off-the shelf widely used hook, e.g. that provided by ``flake8`` itself:

```
$ flake8 --install-hook git
$ git config --bool flake8.strict true
```

Currently (leaving aside optional extensions we don't currently require):

https://github.com/biopython/biopython/blob/biopython-173/BioSQL/.flake8 - quite strict
- ignores W504,D101,D102,D103,D105,

https://github.com/biopython/biopython/blob/biopython-173/Doc/examples/.flake8 - quite strict
- ignores D100,D101,D102,D103

https://github.com/biopython/biopython/blob/biopython-173/Bio/.flake8 - fairly strict
 - ignores E501,E731,F401,F841,D100,D101,D102,D103,D105,D203,

[...]

PR #2016, opened 2019-04-18
 flake8 D300 """triple double quotes""" in Tests

peterjc:
This pull request is somewhat related to issue #883 / #2014.

### DIFF
--- a/Tests/test_NCBITextParser.py
+++ b/Tests/test_NCBITextParser.py
@@ -2,6 +2,7 @@
 # This code is part of the Biopython distribution and governed by its
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
+"""Test for the plain text NCBI BLAST parser."""
 
 import os
 import unittest
@@ -17,15 +18,16 @@ warnings.filterwarnings('ignore', r'Parsing BLAST plain text output '
 
 
 class TestBlastRecord(unittest.TestCase):
+    """Test the BLAST Record object."""
 
     # - incredibly incomplete. Just testing what I'm adding -- Brad.
 
     def setUp(self):
+        """Initialise the parser."""
         self.parser = NCBIStandalone.BlastParser()
 
     def test_conversion(self):
-        "Test converting a Blast record multiple alignment"
-
+        """Test converting a Blast record multiple alignment."""
         from Bio.Alphabet import IUPAC
 
         path = os.path.join("Blast", 'text_2010L_blastp_006.txt')
@@ -39,16 +41,17 @@ class TestBlastRecord(unittest.TestCase):
 
 
 class TestNCBITextParser(unittest.TestCase):
+    """Tests for the text BLAST parser."""
 
     reference = 'Altschul, Stephen F., Thomas L. Madden, Alejandro A. Schaffer, \nJinghui Zhang, Zheng Zhang, Webb Miller, and David J. Lipman (1997), \n"Gapped BLAST and PSI-BLAST: a new generation of protein database search\nprograms",  Nucleic Acids Res. 25:3389-3402.'
 
     def setUp(self):
+        """Initialise the parser."""
         self.parser = NCBIStandalone.BlastParser()
         self.pb_parser = NCBIStandalone.PSIBlastParser()
 
     def test_text_2010L_blastp_001(self):
-        "Test parsing BLASTP 2.0.10 output (text_2010L_blastp_001)"
-
+        """Test parsing BLASTP 2.0.10 output (text_2010L_blastp_001)."""
         path = os.path.join('Blast', 'text_2010L_blastp_001.txt')
         handle = open(path)
         record = self.parser.parse(handle)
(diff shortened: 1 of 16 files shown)


peterjc:
Forgot to check in ``test_SearchIO_hmmer3_text.py``, hang on...

Tests/test_SearchIO_hmmer3_text.py
```diff
@@ -43,7 +43,7 @@ def test_31b1_hmmscan_001(self):
         self.assertEqual(0, len(qresult))
 
         # test fifth result
-        for qresult in qresults:
+        for _ in qresults:
```
peterjc: Ah, shouldn't have fixed that 'unused' loop variable.

Tests/test_NCBITextParser.py
```diff
@@ -167,8 +170,7 @@ def test_text_2010L_blastp_001(self):
         self.assertAlmostEqual(record.blast_cutoff[1], 28.2)
 
     def test_text_2010L_blastp_002(self):
-        "Test parsing BLASTP 2.0.10 output without hits (text_2010L_blastp_002)"
-
+        """Test parsing BLASTP 2.0.10 output without hits (text_2010L_blastp_002)."""
```
MarkusPiotrowski: In a case like this, where the line length is only slightly above 80 characters, putting the one-line comment on an extra line would keep the line length under 80 characters.

```python
        """
        Test parsing BLASTP 2.0.10 outpu without hits (text_2010L_blastp_002).
        """
```

Do you think this is worth considering? (What would ```black``` do?)

Tests/test_ProtParam.py
```diff
@@ -14,59 +15,62 @@
...
+    """Tests for ProtParam."""
+
     def setUp(self):
+        """Initialise objects."""
         self.seq_text = "MAEGEITTFTALTEKFNLPPGNYKKPKLLYCSNGGHFLRILPDGTVDGTRDRSDQHIQLQLSAESVGEVYIKSTETGQYLAMDTSGLLYGSQTPSEECLFLERLEENHYNTYTSKKHAEKNWFVGLKKNGSCKRGPRTHYGQKAILFLPLPV"
         self.analysis = ProtParam.ProteinAnalysis(self.seq_text)
 
     def test_count_amino_acids(self):
-        "Test getting amino acid counts"
+        """Calculating amino acid counts."""
```
MarkusPiotrowski: So we don't enforce "imperative mood" in ```Tests```?

Tests/test_ProtParam.py
```diff
@@ -14,59 +15,62 @@
...
     def test_get_monoisotopic_molecular_weight(self):
-        "Test calculating the monoisotopic molecular weight"
+        """Calculating monoisotopic molecular weight."""
         self.analysis = ProtParam.ProteinAnalysis(self.seq_text, monoisotopic=True)
         self.assertAlmostEqual(round(self.analysis.molecular_weight(), 2),
                                17092.61)
 
     def test_get_molecular_weight_identical(self):
-        "Test calculating the protein molecular weight agrees with calculation from Bio.SeqUtils"
+        """Calculating the protein molecular weight agrees with calculation from Bio.SeqUtils."""
```
MarkusPiotrowski: Not related to this PR, but I wonder if this and the following test make sense, since ```molecular_weight``` of ```ProtParam.ProteinAnalysis``` is just a call to the module's function:

```python
 def molecular_weight(self):
        """Calculate MW from Protein sequence."""
return molecular_weight(self.sequence, monoisotopic=self.monoisotopic)
```

Tests/test_phenotype.py
```diff
@@ -3,9 +3,11 @@
...
 # license.  Please see the LICENSE file that should have been included
 # as part of this package.
+"""Tests for the Bio.phenotype module."""
 
 try:
     import numpy
+    del numpy
 except ImportError:
     from Bio import MissingExternalDependencyError
     raise MissingExternalDependencyError(
```
MarkusPiotrowski: Again not direclty related to this PR, but I would sort the import stuff like this (PEP-8):

```python
import json
import unittest
import warnings

try:
    import numpy
    del numpy
except ImportError:
    from Bio import MissingExternalDependencyError
        "Install NumPy if you want to use Bio.phenotype.")

from Bio import BiopythonExperimentalWarning

with warnings.catch_warnings():
    warnings.simplefilter('ignore', BiopythonExperimentalWarning)
    from Bio import phenotype
```

Tests/test_NCBITextParser.py
peterjc: I'm happy to leave it - we don't current enforce a strict 80 limit (indeed the PEP8 wording is flexible on this), and also black uses 88 characters.

That said, I did rename some of the tests docstrings from "Test parsing..." to "Parsing..." and that would help here.

Tests/test_ProtParam.py
peterjc: Not yet, but we should. I was initially running all the desired flake8 checks on the files I changed, but ran out of energy. Well spotted, that was an unfortunate rewording on my part...

Tests/test_ProtParam.py
peterjc: Not sure, haven't looked at that for a long time.

Tests/test_phenotype.py
peterjc: You mean first standard library imports, then own library imports (here Bio, BioSQL), finally any local imports (e.g. the BioSQL tests do that). I agree, we should work towards that.

peterjc:
@MarkusPiotrowski Lots of good questions, I have addressed the two specific issues in the changes on the pull request, and replied to the others. Thank you for looking at this so carefully!

Tests/test_NCBITextParser.py
peterjc: Shorted by removing "Test " from the start (done for the whole file).

PR #2018, opened 2019-04-19
Move Bio/.flake8 to top level

peterjc:
Work intended to help with #493 and #2014 - fairly sure the automated tests will work without changing the TravisCI/Tox setup, about to find out...

Still have other folder specific .flake8 files, but plan is to improve the compliance in Tests/ etc and have a single set of flake8 settings here.

This will allow use to simplify how we run flake8, and use its provided git pre-commit hook.


MarkusPiotrowski:
See my last comment on #2014 regarding the ```per-file-ignores``` in newer ```flake8``` versions. This should allow to keep some folder specific ignores (for now) with only one ```.flake8``` file.

peterjc:
@MarkusPiotrowski has noticed a better solution with flake 3.7 onwards offering more configuration options: https://github.com/biopython/biopython/issues/2014#issuecomment-484929844

PR #2021, opened 2019-04-19
pairwise2: Test pure Python fallback methods

MarkusPiotrowski:
This is another try to include the pure Python backfall functions ```_make_score_matrix_fast``` and ```rint``` into the unit tests. In my last PR #1990 I tried to block the import of the respective C functions by playing around with ```sys.modules['Bio.pairwise2'] = None```. While this seemed to work in my local respository, it does not seem to work here on Github, since Codecov does not report the adressed code lines to be covered. I also expected to see a ```BiopythonWarning``` about a failed import during the Travis CI tests, which does not come up.

Now I have added two handles to the pure Python methods in ```pairwise2``` which allows to access them directly. 


MarkusPiotrowski:
A new ReportLab problem in Appveyour under Python 3.6

peterjc:
Not really ReportLab, just our very simple graph plotting - ``test_simple_scatter_plot_1`` sometimes picks bad random points and gets a divide by zero error. Test restarted. I would be tempted to deprecate those functions, but in the short term we could just hard code a particular random set of numbers, or set the seed?

peterjc:
I was wondering while working on flake8 issues on Tests/ if the two pairwise test files were duplicating a lot of code needless. Perhaps importing common tests like we do with BioSQL would be better?

MarkusPiotrowski:
> I was wondering while working on flake8 issues on Tests/ if the two pairwise test files were duplicating a lot of code needless. Perhaps importing common tests like we do with BioSQL would be better?

I will have a look at the ```BioSQL```tests. The reason why I made a second file was that I tried to invoke an ```ImportError``` and you cannot (? at least easily) make an import fail and not fail within the same module. I'd appreciate any hints.
I only kept those tests in the second file which address the implemented ```_make_score_matrix_fast``` function, but you are most likely right, that there are needless code duplications. I wanted to be sure that each code line (so each if-branch) is covered. Reducing the tests would be the next step, but first I want to know if I get the code covered in Codecov. 
Now, that I'm using a different way to get access to the pure Python functions, I can think about running the same tests twice with different settings. 
So, I'll have a look at the ```BioSQL``` test, but in the meanwhile I'd like to see from Codecov if my approach is working now.
Is it possible to check for test coverage offline?

peterjc:
Yes, you need to do ``pip install coverage`` and then ``coverage run run_tests.py --offline`` (the command ``coverage run`` is effectively an alias for running ``python`` with coverage support).

See https://coverage.readthedocs.io/en/v4.5.x/#quick-start for how to use it at the command line.

For the online reporting, we get the XML output and upload it with the ``codecov`` tool.

https://github.com/biopython/biopython/blob/biopython-173/.travis-tox.ini#L82
https://github.com/biopython/biopython/blob/biopython-173/.appveyor.yml#L51

MarkusPiotrowski:
Re-structured `test_pairwise2.py` to include checks for C and Python-fallback functions, removed second test file (`test_pairwise2_no_C.py`).

Bio/pairwise2.py
```diff
@@ -1160,7 +1160,12 @@ def format_alignment(align1, align2, score, begin, end):
 # Try and load C implementations of functions. If I can't,
 # then throw a warning and use the pure Python implementations.
 # The redefinition is deliberate, thus the no quality assurance
-# flag for when using flake8:
+# flag for when using flake8.
+# Before, we secure access from outside to the Python functions:
+
+python_make_score_matrix_fast = _make_score_matrix_fast
+python_rint = rint
```
peterjc: Give these leading underscores as implied private, or do you deliberately want them in the public API?

MarkusPiotrowski: Good point, will change

peterjc:
The way you've got the python vs C testing working seems rather repetitive, essentially:

```python
class ClassNameHere(unittest.TestCase):
     def test_method_name_here(self):
        for _mode in ('python', 'C'):
            switch_python_C(_mode)
            ...
```

It seems too complicated for long term maintenance?

Does this seem workable? More the test classes into ``common_pairwise2.py`` and then ``test_pairwise2_no_C.py`` becomes:

```python
# Really do want "import *" to get all the test classes:
from common_pairwise2 import *  # noqa: F403

# Explicitly use pure Python implementation
from Bio import pairwise2
pairwise2._make_score_matrix_fast = pairwise2._python_make_score_matrix_fast
pairwise2.rint = pairwise2._python_rint

if __name__ == '__main__':
```

while ``test_pairwise2.py`` becomes something like this (you can perhaps improve the check to ensure it is using C):

```python
# Really do want "import *" to get all the test classes:
from common_pairwise2 import *  # noqa: F403

# Implicitly using using C implementation
from Bio import pairwise2
assert pairwise2._make_score_matrix_fast != pairwise2._python_make_score_matrix_fast

if __name__ == '__main__':
```

i.e. Using the same idea as the BioSQL test framework which has to try multiple backends.

MarkusPiotrowski:
I didn't understand that 'just' importing makes the TestRunner aware of the TestCases. Your simplified example did help very much, thank you.
I restructured the test as you have suggested.

peterjc:
Yep, the TestRunner must do something clever like look at the class of all the objects imported into the local namespace. I've not exampled the implementation details. I honestly forget now how I came up with this for the BioSQL tests - I assume I based it on something.

PR #2022, opened 2019-04-21
Update AbiIO to preserve bytes value

bow:
This pull request addresses issue #1982 and is a continuation from PR #1984.

@peterjc This PR incorporates all the changes we have discussed so far (including the fallback to system encoding when we can't extract string values for `SeqRecord.id` and `SeqRecord.description). There is also a test case included, with the abif file containing the non-ascii tag that @zqfang reported earlier.


Bio/SeqIO/AbiIO.py
```diff
@@ -331,6 +334,23 @@
...
+def _get_string_tag(opt_bytes_value, default=None):
+    """Return the string value of the given an optional raw bytes tag value.
+
+    If the bytes value is None, return the given default value.
+
+    """
+    if opt_bytes_value is None:
+        return default
+    try:
+        return _bytes_to_string(opt_bytes_value)
```
peterjc: Doesn't this use the default encoding? The NEWS entry says you'd try ASCII first...

bow: Hmm...I did say the first encoding attempt assumes `utf-8`, though I did say the parser can be made more robust by handling non-ascii characters in the beginning of the entry.

I'll revise the `NEWS.rst` entry to be clearer.

peterjc: I may be confused/wrong, but does ``.decode()`` not use the system default encoding anyway - which may not be utf-8?

bow: [Looking at the docs](https://docs.python.org/3/library/codecs.html#codecs.decode), I think it uses `utf-8` by default, regardless of the system encoding.

Initially I thought: is it also possible that the value shown in the docs is actually the result of `sys.getdefaultencoding()`? That seems unlikely however, and I feelthat  they would have said so had it it been the case.

peterjc: OK, you're right. Thanks. I've not looked at the encoding and locales for a while now.

peterjc:
You need to tweak ``NEWS.rst`` to pass validation, but otherwise this looks good to merge.

Bio/SeqIO/AbiIO.py
bow: Great, ok ~ the tests finally pass, so I am marking this as resolved and will merge this soon today :).

bow:
Ok, it seems to be fine now. I'm rebasing this onto master now ~

Issue #2015, opened 2019-04-18
Fix use of assert


peterjc:
I'm logging this as a good first issue ideal for a new contributor (but if you are the maintainer of any of the files listed here, feel free to fix them yourself). If you want to work on this, I suggest commenting first saying something like *"I'll work on Bio/SwissProt to start with"* in order to avoid duplicated effort.

If you use the code checking tool flake8 with the bugbear plugin, it finds a number of issues with our code - specifically we have a modest number of assert statements which should ideally be replaced.

You can install the tools as follows:

```
$ pip install --upgrade flake8 flake8-bugbear
```

Run flake8 to look for this problem specifically:

```
$ flake8 --isolated --select B011 Bio/ Tests/
Bio/SwissProt/__init__.py:354:13: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Bio/SwissProt/__init__.py:396:13: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Bio/SeqIO/PirIO.py:185:5: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Bio/SeqIO/QualityIO.py:1372:5: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Bio/AlignIO/EmbossIO.py:61:9: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Bio/AlignIO/EmbossIO.py:207:17: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/test_Nexus.py:687:13: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/test_SffIO.py:564:9: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/seq_tests_common.py:142:17: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/common_BioSQL.py:328:13: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/common_BioSQL.py:356:13: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
Tests/common_BioSQL.py:367:17: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
[...]

PR #2023, opened 2019-04-21
Bugfix/fix use of assert 2015

benjaminr:
This pull request addresses issue #2015 


Bio/AlignIO/EmbossIO.py
```diff
@@ -57,8 +57,6 @@ def write_alignment(self, alignment):
         handle.write("#\n")
         handle.write("#=======================================\n")
         handle.write("\n")
-        # ...
-        assert False
```
peterjc: In general ``assert False`` is something better written as ``raise ValueError`` or similar. In this specific case, as this is a base object which should have these methods redefined, ``raise NotImplementedError`` as in the example here: https://github.com/biopython/biopython/blob/biopython-173/Bio/SeqIO/Interfaces.py#L54

Bio/AlignIO/EmbossIO.py
```diff
@@ -204,7 +202,7 @@ def __next__(self):
                 pass
             else:
                 print(line)
-                assert False
+                raise ValueError("Unrecognised EMBOSS pairwise alignment record format.")
```
peterjc: It would be even better to remove the ``print(line)`` and put that into the message, e.g.

```
raise ValueError("Unrecognised EMBOSS pairwise line: %r\n" % line)
```

Tests/common_BioSQL.py
```diff
@@ -323,11 +323,8 @@ def test_server(self):
         del server["biosql-test"]
         del server["biosql-test2"]
         self.assertEqual(0, len(server))
-        try:
+        with self.assertRaises(KeyError):
```
peterjc: Looks good - If I recall correctly, ``self.assertRaises`` on the older versions of Python we supported when that test was written.

Tests/test_SffIO.py
```diff
@@ -561,9 +561,8 @@ def fileiter(handle):
     try:
         print(ReadRocheXmlManifest(
               open("Roche/E3MFGYR02_alt_index_at_end.sff", "rb")))
-        assert False, "Should fail!"
     except ValueError:
-        pass
+        raise ValueError("Should fail.")
```
peterjc: I think you've accidentally flipped the meaning of the test. This looks like another case were we could use ``self.assertRaises(ValueError)``?

benjaminr: As this wasn't a part of a unittest TestCase, I've called the function directly instead of accessing via a reference to self. Not sure if it was the best way to go about it, or if a new TestCase would be more appropriate.

peterjc: You're absolutely right - I didn't see that from the context of the diff. Does this new form work nicely (I've never used it that way myself).

peterjc:
While we wait for the automated testing: Do you have strong views on a squash-and-merge as a single commit, versus keeping your full commit history, for this pull request?

benjaminr:
> While we wait for the automated testing: Do you have strong views on a squash-and-merge as a single commit, versus keeping your full commit history, for this pull request?

Nope, happy to squash-and-merge, makes everything cleaner 👍

Tests/test_SffIO.py
benjaminr: As far as I'm aware, it works equivalently and it appeared to work okay locally on my machine.

peterjc:
I'll deal with the merge now (and the trivial conflict)...

PR #2044, opened 2019-05-01
E731 issues (don't assign lambda, use a def) in Bio.KEGG and Bio.Phylo

MarkusPiotrowski:
This pull request is related to issues #2014, #883.

### DIFF
--- a/Bio/KEGG/Compound/__init__.py
+++ b/Bio/KEGG/Compound/__init__.py
@@ -13,19 +13,17 @@ Classes:
  - Record - A representation of a KEGG Ligand/Compound.
 """
 
-# other Biopython stuff
 from __future__ import print_function
 
-from Bio.KEGG import _write_kegg
-from Bio.KEGG import _wrap_kegg
+from Bio.KEGG import _default_wrap, _struct_wrap, _wrap_kegg, _write_kegg
 
 
 # Set up line wrapping rules (see Bio.KEGG._wrap_kegg)
 name_wrap = [0, "",
              (" ", "$", 1, 1),
              ("-", "$", 1, 1)]
-id_wrap = lambda indent: [indent, "", (" ", "", 1, 0)]
-struct_wrap = lambda indent: [indent, "", ("  ", "", 1, 1)]
+id_wrap = _default_wrap
+struct_wrap = _struct_wrap
 
 
 class Record(object):
--- a/Bio/KEGG/__init__.py
+++ b/Bio/KEGG/__init__.py
@@ -18,12 +18,23 @@ KEGG_ITEM_LENGTH = 12
 KEGG_LINE_LENGTH = 80
 KEGG_DATA_LENGTH = KEGG_LINE_LENGTH - KEGG_ITEM_LENGTH
 
-# wrap rule = [indent, connect, (splitstr, connect, splitafter, keep), ...]
-_default_wrap = lambda indent: [indent, "", (" ", "", 1, 0)]
+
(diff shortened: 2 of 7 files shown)


Bio/KEGG/Compound/__init__.py
```diff
@@ -13,19 +13,21 @@
...
 # Set up line wrapping rules (see Bio.KEGG._wrap_kegg)
 name_wrap = [0, "",
              (" ", "$", 1, 1),
              ("-", "$", 1, 1)]
-id_wrap = lambda indent: [indent, "", (" ", "", 1, 0)]
-struct_wrap = lambda indent: [indent, "", ("  ", "", 1, 1)]
+id_wrap = _default_wrap
+
+
+def struct_wrap(indent):
```
peterjc: Since it seems to get defined twice, why not put ``struct_wrap`` into the main KEGG file too as ``_struct_wrap``?

Bio/KEGG/__init__.py
```diff
@@ -18,8 +18,14 @@
 KEGG_LINE_LENGTH = 80
 KEGG_DATA_LENGTH = KEGG_LINE_LENGTH - KEGG_ITEM_LENGTH
 
-# wrap rule = [indent, connect, (splitstr, connect, splitafter, keep), ...]
-_default_wrap = lambda indent: [indent, "", (" ", "", 1, 0)]
+
+def _default_wrap(indent):
+    """Return default wrap rule for _wrap_kegg (PRIVAT).
```
peterjc: Typo, PRIVATE.

Bio/Phylo/CDAOIO.py
```diff
@@ -19,13 +19,15 @@
...
+import os
+
 from Bio._py3k import StringIO
 
 from Bio import MissingPythonDependencyError
 
 from Bio.Phylo import CDAO
-from ._cdao_owl import cdao_elements, cdao_namespaces, resolve_uri
-import os
+from ._cdao_owl import cdao_namespaces, resolve_uri  # , cdao_elements
```
peterjc: The apparently extra import looks worth double checking - @bendmorris do you recall why it was done? Is this something useful for users of the public API?

peterjc:
A few minor comments.

There were a few cases where using a lambda seemed reasonable, and adding ``# noqa: E731`` would be equally acceptable to me.

Bio/KEGG/Compound/__init__.py
MarkusPiotrowski: Yes, agree. Have done so.

Bio/KEGG/__init__.py
MarkusPiotrowski: Corrected

MarkusPiotrowski:
I reverted some changes, so some lambda assignments are back, with `# noqa: E731`.

peterjc:
Given the update now only adds the ``# noqa: E731`` lines in ``Bio.Phylo``, I don't think we need to wait for Eric's formal approval.

However, unless @bendmorris replies soon, I would remove the ``cdao_elements`` import change from this pull request, and look at it separately instead.

Bio/Phylo/CDAOIO.py
bendmorris: I don't recall - possibly it was used in an earlier iteration. If it's unused now no reason to keep it there.

MarkusPiotrowski:
After @bendmorris' answer, I removed the unused import from `CDAOIO.py` but left a note when it happened.

Issue #2040, opened 2019-04-30
ZeroDivisionError in test_GraphicsGeneral.py


peterjc:
There is a long standing problem with the scatter plot test in ``test_GraphicsGeneral.py`` in that because it uses random points, sometimes they are co-linear or something, and we get a divide by zero error. This periodically triggers a false alarm in our continuous integration testing, often on a pull request for something completely unrelated which is confusing.

The test should either set a random seed (to ensure the same pseudo random numbers are used), or more simply use a fixed set of data points.

Example failure:

```
======================================================================
ERROR: test_simple_scatter_plot_1 (test_GraphicsGeneral.ComparativeTest)
Test creation of a simple ScatterPlot with one list.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/test_GraphicsGeneral.py", line 71, in test_simple_scatter_plot_1
    compare_plot.draw_to_file(output_file, "Testing Scatter Plots")
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.5/Bio/Graphics/Comparative.py", line 85, in draw_to_file
    return _write(cur_drawing, output_file, self.output_format)
  File "/home/travis/build/biopython/biopython/build/lib.linux-x86_64-3.5/Bio/Graphics/__init__.py", line 84, in _write
    return drawmethod.drawToFile(drawing, output_file)
  File "/home/travis/build/biopython/biopython/.tox/py35-cover/lib/python3.5/site-packages/reportlab/graphics/renderPDF.py", line 302, in drawToFile
    draw(d, c, 0, 0, showBoundary=showBoundary)
  File "/home/travis/build/biopython/biopython/.tox/py35-cover/lib/python3.5/site-packages/reportlab/graphics/renderPDF.py", line 29, in draw
    R.draw(renderScaledDrawing(drawing), canvas, x, y, showBoundary=showBoundary)
  File "/home/travis/build/biopython/biopython/.tox/py35-cover/lib/python3.5/site-packages/reportlab/graphics/renderbase.py", line 205, in draw
    self.drawNode(drawing)
  File "/home/travis/build/biopython/biopython/.tox/py35-cover/lib/python3.5/site-packages/reportlab/graphics/renderPDF.py", line 54, in drawNode
    self.drawNodeDispatcher(node)
[...]

PR #2053, opened 2019-05-10
Add SHORT_QUERY_ADJUST support in Bio.Blast.NCBIWWW

### LINKED ISSUE #2049: ADJUSTED_FOR_SHORT_QUERY
### Setup

system as follows:

3.6.3 (v3.6.3:2c5fed86e0, Oct  3 2017, 00:32:08) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
CPython
Darwin-17.7.0-x86_64-i386-64bit
1.70

### Expected behaviour

I am running a search for a peptide sequence using blastp on the PDB database. I would like to set the parameter to adjust for short queries, but that parameter is not available.  As a result, the output returned by the python query does not match the output produced by the website when that parameter is activated

### Steps to reproduce

result_handle=NCBIWWW.qblast('blastp', 'pdb_v5', 'QRFSQPTFKLPQGRLTLSRKF')

run above query and notice that no results are returned with the is query, while a large number or results are returned when run with the Adjust for short query parameter set to on (located in algorithm parameters section) in the website.

MarkusPiotrowski:
This pull request addresses issue #2049. It allows to use the SHORT_QUERY_ADJUST parameter of BLAST's common URL API in `Bio.Blast.NCBIWWW`via a new keyword parameter `short_query`. 
Unfortunately, SHORT_QUERY_ADJUST is faulty for blastn, thus a workaround for blastn was implemented.

Also, `test_NCBI_qblast.py` was corrected (did fail on two tests) and updated.

This will close #2049.


MarkusPiotrowski:
Need to implement sequence length dependency for blastn workaround.

MarkusPiotrowski:
Hey, how comes a D400 in my commit from a file I didn't touch? :-( 


<pre>
Tests/test_PDB_parse_pdb_header.py:53:1: D400 First line should end with a period
ERROR: InvocationError for command /home/travis/build/biopython/biopython/.tox/style/bin/flake8 Tests/ (exited with code 1)
</pre>

I'm going to correct this...

MarkusPiotrowski:
Now I got caught by:
```
ERROR: test_simple_scatter_plot_1 (test_GraphicsGeneral.ComparativeTest)
Test creation of a simple ScatterPlot with one list.
...
ZeroDivisionError: float division by zero
```

Sigh...

peterjc:
It seems #2040 is causing a disproportionate amount of pain, we should fix it ASAP.

peterjc:
I would not include the workaround.

Have you reported the blastn issue to the BLAST team? They can likely fix this on their server (relatively quickly), which seem far better in the long run that having a workaround in (some versions of) biopython.

MarkusPiotrowski:
It seems that they are not interested in fixing this issue...
Yes, I have contacted them and also give them the likely reason why it fails: It seems that they are just using the blastp SHORT_QUERY_ADJUST values in blastn (where the word size is to small for blastn and thus throws an error).

> Dear Sir or Madame,
> 
> I want to report an error with the common URL API.
> 
> The parameter SHORT_QUERY_ADJUST does not work for blastn, obviously the parameter adjustment procedure uses the parameters for blastp, which results in a blastn error (WORD_SIZE must be >=4).
> 
> A working blastp search with SHORT_QUERY_ADJUST:
> 
> https://blast.ncbi.nlm.nih.gov/Blast.cgi?PROGRAM=blastp&DATABASE=nr&QUERY=ICWENK&HITLIST_SIZE=5&SHORT_QUERY_ADJUST=on&CMD=Put
> 
> Now a failing blastn search with SHORT_QUERY_ADJUST:
> 
> https://blast.ncbi.nlm.nih.gov/Blast.cgi?PROGRAM=blastn&DATABASE=nt&QUERY=ATGCTGATCGGATCAACG&HITLIST_SIZE=5&SHORT_QUERY_ADJUST=on&CMD=Put
> 
> Message ID#36 Error: Cannot validate the Blast options: Word-size must be 4 or greater for nucleotide comparison
> 
> Taking a look at the html source code on the Format Request page, you see that the parameters were wrongly set for blastn:
>  
> 
> \<input name="EXPECT" type="hidden" value="200000"/>
> ...
> \<input name="WORD_SIZE" type="hidden" value="2"/>
> ...
> \<input name="MATCH_SCORES" type="hidden" value=""/>
> 
> 
> EXPECT should be 1000, WORD_SIZE 7, and MATCH_SCORES 1, -3. There may be other parameters which require changes (e.g. filtering and masking.)
> 
> Sincerely yours,
> Markus Piotrowski

The reply was that they are not officially supporting this 'SHORT_QUERY_ADJUST' and they recommend to set the respective values manually (what I did in the workaround).

> Hi Markus,
> 
> The fact is, we only officially support the parameters outlined here,
> 
> https://ncbi.github.io/blast-cloud/dev/api.html
> 
> For short queries, I suggest you explicitly set EXPECT, WORD_SIZE, etc.
> Best regards,
> Wayne

What is the problem with the workaround? If it is ommitted people will not be able to use `SHORT_QUERY_ADJUST` with `blastn`.

peterjc:
> What is the problem with the workaround? If it is ommitted people will not be able to use SHORT_QUERY_ADJUST with blastn.

Exactly 👍 

It sounds like the NCBI would prefer we didn't expose the ``SHORT_QUERY_ADJUST`` parameter at all. The problem is once we do add it to Biopython, people will expect us to support it *and the workaround*.

If you do want to add the undocumented parameter, I would say explicitly in our documentation that at the time of writing it is not supported by the NCBI, and furthermore, note  it does not (currently) work with BLASTN, and document the workaround they suggest.

But perhaps I'm being too cautious?-By all means ask another Biopython developer or two for their perspective.

MarkusPiotrowski:
I'm sorry, Peter, I didn't thought that adding another 'undocumented' parameter would harm, since the 'old' NCBI's [qblast API](https://web.archive.org/web/20070701014147/http://www.ncbi.nlm.nih.gov/BLAST/Doc/node5.html) (on which `Bio.Blast.NCBIWWW` is based) did support more or other parameters than are supported in the 'new' [common URL API](https://ncbi.github.io/blast-cloud/dev/api.html). 

So we may think about a new module which is fully compliant with NCBI's Common URL API.

peterjc:
That's a good counter argument. OK, stick a cautionary note in the docstring, and go ahead with the code as is.

MarkusPiotrowski:
@bow I fear I don't understand what you have in mind with the toggle-able switch.
Let me explain what I did and why:
1. I added the possibility to use the parameter `SHORT_QUERY_ADJUST`. 
`SHORT_QUERY_ADJUST` is an (undocumented) parameter for NCBI's remote web searches. It works nicely for **blastp, tblastn and blastx** and it (automatically) relaxes some search parameters like word size and e-value if the sequence is shorter than 31 characters.
2. However, NCBI has made an error in the implementation for **blastn**. They use the wrong parameters and thus a BLAST error is invoked (word-size to small). To bypass this problem, I set the respective parameters manually, but only for this case, so that users of `Bio.Blast.NCBIWWW` didn't get caught by the problem. But we don't set the parameters manually in case of **blastp** etc.

`SHORT_QUERY_ADJUST` is off by default when doing remote web searches.

MarkusPiotrowski:
@bow BTW the cautionary note that @peterjc is requesting is not this `BiopythonWarning` but about using undocumented parameters of NCBIs Common URL API. So I added the following paragraph:

> Please note that NCBI uses the new Common URL API for BLAST searches
> on the internet (http://ncbi.github.io/blast-cloud/dev/api.html). Thus,
>  some of the parameters used by this function are not or not longer
>  officially supported by NCBI. Although they are still functioning, this
>  may change in the future.

However, I corrected the `BiopythonWarning` message as you have requested.

Bio/Blast/NCBIWWW.py
```diff
@@ -42,13 +46,18 @@ def qblast(program, database, sequence, url_base=NCBI_BLAST_URL,
...
+    Supports all parameters of the old qblast API for Put and Get.
 
-    Please note that BLAST on the cloud supports the NCBI-BLAST Common
-    URL API (http://ncbi.github.io/blast-cloud/dev/api.html). To
-    use this feature, please set url_base to
-    'http://host.my.cloud.service.provider.com/cgi-bin/blast.cgi' and
-    format_object='Alignment'. For more details, please see
+    Please note that NCBI uses the new Common URL API for BLAST searches
+    on the internet (http://ncbi.github.io/blast-cloud/dev/api.html). Thus,
+    some of the parameters used by this function are not or not longer
```
peterjc: Something odd with the wording here, "are not are not longer officially supported".

Maybe just "are no longer officially supported" is close to your intent?

Bio/Blast/NCBIWWW.py
```diff
@@ -80,6 +91,23 @@ def qblast(program, database, sequence, url_base=NCBI_BLAST_URL,
...
+    if short_query and program == 'blastn':
+        short_query = None
+        # We only use the 'short-query' parameters for short sequences:
+        if len(sequence) < 31:
+            expect = 1000
+            word_size = 7
+            nucl_reward = 1
+            filter = None
+            lcase_mask = None
+            warnings.warn('"short_query" for blastn is wronlgy implemented by '
```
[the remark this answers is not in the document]
peterjc: Maybe "is not currently implemented by the NCBI"?

Also, is the rest of this warning clear enough that the workaround will ignore some values the user might have set (expect, word size, etc)?

Bio/Blast/NCBIWWW.py
MarkusPiotrowski: "Are not *or* not longer" ("are not/not longer supported") to say that some of them were supported in the past (old qblast) but some of them (e.g. SHORT_QUERY_ADJUST) have never been officially supported (or documented)

Tests/test_NCBI_qblast.py
```diff
@@ -1,7 +1,9 @@
 # Copyright 2008-2016 by Peter Cock.  All rights reserved.
-# This code is part of the Biopython distribution and governed by its
-# license.  Please see the LICENSE file that should have been included
-# as part of this package.
+#
+# This file is part of the Biopython distribution and governed by your
+# choice of the "Biopython License Agreement" or the "BSD 3-Clause License".
+# Please see the LICENSE file that should have been included as part of this
+# package.
```
peterjc: Can we do the dual licensing separately, with a more explicit audit trail?

MarkusPiotrowski: OK, it thought because you are the author we can do this in one go.

Bio/Blast/NCBIWWW.py
peterjc: Sorry, I missed the "or" in the earlier quote.

How about "... are not (or are no longer) officially supported." here?

Tests/test_NCBI_qblast.py
peterjc: The main author yes, but there are half a dozen or so contributors to this file according to GitHub, so I'd prefer to do it a little more formally. I've got this fairly automated now ;)

Bio/Blast/NCBIWWW.py
MarkusPiotrowski: OK.

Bio/Blast/NCBIWWW.py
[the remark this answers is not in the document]
MarkusPiotrowski: I think that is the same what is happening when 'SHORT_QUERY_ADJUST' is used in **blastp* etc. I can put this in the docstring if you like?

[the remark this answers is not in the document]
MarkusPiotrowski: Actually, it is implemented, but wrongly. If you use SHORT_QUERY_ADJUST with **blastn** you get a BLAST error. 

This is the warning as have it now:

> '"SHORT_QUERY_ADJUST" is incorrectly implemented '
>                           '(by NCBI) for blastn. We bypass the problem by '
>                           'manually adjusting the search parameters. Thus, '
>                           'results may slightly differ from web page '
>                           'searches.'

MarkusPiotrowski:
I implemented the requested and suggested changes (mostly...).

peterjc:
Looks good, over to you @bow

bow:
@MarkusPiotrowski Aha, thanks for the explanation. I did miss some parts of it; now I see why having the parameter I suggested earlier does not really make sense.

Anyway, it looks ok for merging now. I'll do a rebase very soon :).

PR #2054, opened 2019-05-11
E731 issues (don't assign a lambda expression) in Bio.SearchIO

MarkusPiotrowski:
This pull request, together with PR #2047, will remove or silence all remaining E731 issues in `Bio.Search`.

### DIFF
--- a/Bio/SearchIO/BlastIO/blast_xml.py
+++ b/Bio/SearchIO/BlastIO/blast_xml.py
@@ -214,8 +214,9 @@ def _extract_ids_and_descs(raw_id, raw_desc):
                      for x in re.split(_RE_ID_DESC_PAIRS_PATTERN, id_desc_line)]
     # make sure empty descriptions are added as empty strings
     # also, we return lists for compatibility reasons between Py2 and Py3
-    add_descs = lambda x: x if len(x) == 2 else x + [""]
-    for pair in (add_descs(p) for p in id_desc_pairs):
+    for pair in id_desc_pairs:
+        if len(pair) != 2:
+            pair.append("")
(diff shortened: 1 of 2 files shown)


peterjc:
Looks good, over to @bow

Bio/SearchIO/BlastIO/blast_xml.py
```diff
@@ -214,8 +214,9 @@ def _extract_ids_and_descs(raw_id, raw_desc):
                      for x in re.split(_RE_ID_DESC_PAIRS_PATTERN, id_desc_line)]
     # make sure empty descriptions are added as empty strings
     # also, we return lists for compatibility reasons between Py2 and Py3
-    add_descs = lambda x: x if len(x) == 2 else x + [""]
-    for pair in (add_descs(p) for p in id_desc_pairs):
+    for pair in id_desc_pairs:
+        if len(pair) != 2:
+            pair += [""]
```
bow: If I recall correctly, I was using `x + [""]` just so that I can have the append operation become an expression. Had Python had something like `list.get(index, value_if_missing)`, I would use it directly in `descs.append`.

Anyway, if we are doing the check like this, I think `pair.append("")` is more suitable.

MarkusPiotrowski:
OK, changed that line and rebased.

Bio/SearchIO/BlastIO/blast_xml.py
MarkusPiotrowski: Yes, no problem. I saw this but was not sure about the implications when changing to `append()`, so I kept it.

bow:
It's rebased & merged now ~ thank you.

PR #2065, opened 2019-05-16
E731 issues (Don't assign lambda...) in /Bio

MarkusPiotrowski:
This pull request is related to  issue #2014.

It removes or silences the last E731 issues (don't assign a lambda expression, use def) in `/Bio`, so we can stop ignoring it in our style checks.

### DIFF
--- a/Bio/SeqIO/NibIO.py
+++ b/Bio/SeqIO/NibIO.py
@@ -43,30 +43,33 @@ import sys
 try:
     hex2bytes = bytes.fromhex  # python3
 except AttributeError:
-    hex2bytes = lambda s: s.decode('hex')  # python2
+    # python 2
+    hex2bytes = lambda s: s.decode('hex')  # noqa: E731
 
 if sys.version_info < (3, ):
     # python2
     import binascii
     bytes2hex = binascii.hexlify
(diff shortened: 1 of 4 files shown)


Bio/SeqIO/NibIO.py
```diff
@@ -43,30 +43,38 @@
 try:
     hex2bytes = bytes.fromhex  # python3
 except AttributeError:
-    hex2bytes = lambda s: s.decode('hex')  # python2
+    # python 2
+    hex2bytes = lambda s: s.decode('hex')  # noqa: E731
+
+
+def bytes2hex(b):
```
peterjc: Can you put the version check outside the definition (as before), so it happens once at import time, rather than for every function call? (Not sure how much difference this would actually make to performance)

MarkusPiotrowski: This is called only one or two times so I thought it would be OK. Otherwise, I have to define two methods with the same name (or noqa it).

peterjc: In this case the ``# noqa`` would make sense since the two functions implementations are for different platforms. But if the function is only called once or twice per file, it should be fine as is.

PR #2068, opened 2019-05-17
Syntax coloring in Tutorial (PDF only, using minted + Pygments)

peterjc:
The short term benefit of this change is syntax colouring in the PDF output from ``pdflatex``, i.e. our file ``Tutorial.pdf``, currently using  ``python``, ``pycon``, ``bash`` only - may be scope for more, e.g. XML fragments.

Sadly this does not work in the ``Tutorial.html`` generated from the LaTeX source with ``hevea`` (output should be unchanged).

This pull request is linked to #907, replacing LaTeX, in that with this change we get much better reStructuredText or (GitHub Flavour) Markdown output from conversion using pandoc which also understands the syntax markup being used.

I have tried to check that we preserve nice copy-and-paste behaviour with the PDF (at least on macOS), which can be a problem (e.g. curly left/right quote marks, missing indentation). That failed with using ``\usepackage{listings}`` for syntax colouring (although could be made to look quite nice, and pandoc understood this).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ Doc/*/*/*.out
 Doc/*/*/*.toc
 Doc/*/*/*.haux
 Doc/*/*/*.htoc
+Doc/_minted-Tutorial/
 Doc/Tutorial.txt
 Doc/Tutorial.pdf
 Doc/Tutorial.html
--- a/Doc/Makefile
+++ b/Doc/Makefile
@@ -5,9 +5,9 @@ all:  Tutorial.html Tutorial.txt pdf $(subdirs)
 pdf:  Tutorial.pdf biopdb_faq.pdf
 
 Tutorial.pdf: Tutorial.tex Tutorial/chapter_*.tex
-	pdflatex Tutorial.tex
-	pdflatex Tutorial.tex
-	pdflatex Tutorial.tex
+	pdflatex --shell-escape Tutorial.tex
+	pdflatex --shell-escape Tutorial.tex
+	pdflatex --shell-escape Tutorial.tex
 
 biopdb_faq.pdf: biopdb_faq.tex
 	pdflatex biopdb_faq.tex
--- a/Doc/Tutorial.tex
+++ b/Doc/Tutorial.tex
@@ -47,7 +47,7 @@
 \usepackage{fullpage}
 \usepackage{hevea}
 \usepackage{graphicx}
-\usepackage{listings}
+\usepackage{minted}
 
 % make everything have section numbers
 \setcounter{secnumdepth}{4}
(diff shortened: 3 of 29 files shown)


peterjc:
Here's the PDF output from this pull request:

[Tutorial.pdf](https://github.com/biopython/biopython/files/3192136/Tutorial.pdf)

Compare to the current version from our last release,

https://github.com/biopython/DIST/blob/bd541f649747ea72846235e3a71766c4757e38af/docs/tutorial/Tutorial.pdf

Latest version currently available here:

http://biopython.org/DIST/docs/tutorial/Tutorial.pdf

smsaladi:
on Windows 10:
- using Chrome or Adobe Acrobat IX as pdf viewers, indentation doesn't work upon copy-paste 
- using MS Edge, indentation as well as new lines also don't come through upon copy-paste

fwiw, this is the same both with the version in this PR as well as the older one.

Used the snippet on page 17 (here with correct indentation):

```python
from Bio import SeqIO
for seq_record in SeqIO.parse("ls_orchid.fasta", "fasta"):
    print(seq_record.id)
    print(repr(seq_record.seq))
    print(len(seq_record))
```

as well as on page 21:

```python
>>> from Bio.Seq import Seq
>>> my_seq = Seq("GATCG", IUPAC.unambiguous_dna)
>>> for index, letter in enumerate(my_seq):
...     print("%i %s" % (index, letter))
````

peterjc:
Thanks @smsaladi - that's a shame, but on the bright side there is no new loss of functionality, only an improvement in now having syntax coloring, so this shouldn't stop us applying the changes.

peterjc:
Looks fine and indentation seems fine after copy-paste on iOS.

peterjc:
@deepaklorkhatri007 if you wanted to test this, I just need you to load the old and new PDF files, confirm them look fine (without syntax colouring, with syntax colouring), and then try copy-and-paste of a Python example with indentation into a text editor.

If you have access to more than one PDF viewer, please try them all - especially on Windows.

deepaklorkhatri007:
@peterjc I will try it on windows and Linux and update you here.

peterjc:
@deepaklorkhatri007 that's impressive work - thank you! 🥇 

Just to clarify, the old and new PDF looks fine in all of them, and the tick/cross is if indentation worked on a copy-and-paste, or not?

If so, this seems pretty clear that the changes are an improvement (adding syntax colours) without breaking copy-and-paste in viewers where it was already working.

deepaklorkhatri007:
@peterjc Affirmative, this is indeed an improvement they behave exactly the same in all readers and sincerely your work makes the tutorial more enjoyable with coloured syntax.

peterjc:
Branch rebased and merge conflicts resolved by hand. New PDF as of this commit (just in case there was a side effect, not seeing anything right now):

[Tutorial.pdf](https://github.com/biopython/biopython/files/3260830/Tutorial.pdf)

peterjc:
Applied, thank you everyone who helped with testing, especially @deepaklorkhatri007 👍

PR #2079, opened 2019-05-24
Apply black style formatting to GenomeDiagram

peterjc:
This pull request is a test case for issue #2008, chosen since the GenomeDiagram author @widdowquinn is a fan of the black tool and it's output, and we historically have had some variation in style between modules.

Note that I reverted two black edits (white space with slicing) due to a known disagreement between black and flake8 over the PEP8 ruling (looks like a flake8 false positive).

I also manually reviewed and fixed some multi-line comments (previously vertically aligned with spaces), in general making them into one-line comments.

### DIFF
--- a/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
+++ b/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
@@ -57,8 +57,8 @@ def page_sizes(size):
      - size - A string representing a standard page size, eg 'A4' or 'LETTER'
 
     """
-    sizes = {
-        "A0": pagesizes.A0,  # ReportLab pagesizes, keyed by ISO string
+    sizes = {  # ReportLab pagesizes, keyed by ISO string
+        "A0": pagesizes.A0,
         "A1": pagesizes.A1,
         "A2": pagesizes.A2,
         "A3": pagesizes.A3,
@@ -87,9 +87,10 @@ def _stroke_and_fill_colors(color, border):
     if not isinstance(color, colors.Color):
         raise ValueError("Invalid color %r" % color)
 
-    if color == colors.white and border is None:  # Force black border on
-        strokecolor = colors.black  # white boxes with
-    elif border is None:  # undefined border, else
+    if color == colors.white and border is None:
+        # Force black border on white boxes with undefined border
+        strokecolor = colors.black
+    elif border is None:
         strokecolor = color  # use fill color
     elif border:
         if not isinstance(border, colors.Color):
--- a/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
+++ b/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
@@ -436,8 +436,6 @@ class AbstractDrawer(object):
            margins to the page
          - xl        Float (0->1) describing the relative size of the left X
            margin to the page (overrides x)
-         - xl        Float (0->1) describing the relative size of the left X
-           margin to the page (overrides x)
          - xr        Float (0->1) describing the relative size of the right X
            margin to the page (overrides x)
          - yt        Float (0->1) describing the relative size of the top Y
--- a/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
+++ b/Bio/Graphics/GenomeDiagram/_AbstractDrawer.py
@@ -57,24 +57,25 @@ def page_sizes(size):
      - size - A string representing a standard page size, eg 'A4' or 'LETTER'
 
     """
-    sizes = {'A0': pagesizes.A0,    # ReportLab pagesizes, keyed by ISO string
-             'A1': pagesizes.A1,
-             'A2': pagesizes.A2,
-             'A3': pagesizes.A3,
-             'A4': pagesizes.A4,
-             'A5': pagesizes.A5,
-             'A6': pagesizes.A6,
-             'B0': pagesizes.B0,
-             'B1': pagesizes.B1,
-             'B2': pagesizes.B2,
-             'B3': pagesizes.B3,
-             'B4': pagesizes.B4,
-             'B5': pagesizes.B5,
-             'B6': pagesizes.B6,
-             'ELEVENSEVENTEEN': pagesizes.ELEVENSEVENTEEN,
-             'LEGAL': pagesizes.LEGAL,
-             'LETTER': pagesizes.LETTER
-             }
+    sizes = {
+        "A0": pagesizes.A0,  # ReportLab pagesizes, keyed by ISO string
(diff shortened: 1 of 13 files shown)


widdowquinn:
One of the things I love about `black` is that it makes my code look like it's written by a better coder than me.

peterjc:
I suppose by discarding any idiosyncratic personal style quirks (in line breaking etc), it makes the code more standard, and thus leaves a better impression? Or maybe you just like the black style more than your style from ten years back ;)

widdowquinn:
😂I like the black style more than my style, now!

MarkusPiotrowski:
Wow, this looks even worse than I have feared... Even if it's not breaking PEP8 literally, I have the feeling it's breaking the spirit of PEP8. It looks so un-pythonic. 
Will it really be easier to work with method headers that fill up the whole screen? You will need much more scrolling when working on the code, I guess it will be harder to keep an overview?

In this examples, you probably want to keep the x,y values in the point lists on one line?
Indeed, it takes away much of the freedom to sort/arange things in a meaningful manner.

peterjc:
I disagree with that it's breaking the spirit of PEP8 or that it is un-pythonic - but it is different and seems to generate polarising views.

Quoting a specific example from ``Bio/Graphics/GenomeDiagram/_AbstractDrawer.py``, before:

```python
    def __init__(self, parent, pagesize='A3', orientation='landscape',
                 x=0.05, y=0.05, xl=None, xr=None, yt=None, yb=None,
                 start=None, end=None, tracklines=0, cross_track_links=None):
```

after black:

```python
    def __init__(
        self,
        parent,
        pagesize="A3",
        orientation="landscape",
        x=0.05,
        y=0.05,
        xl=None,
        xr=None,
        yt=None,
        yb=None,
        start=None,
        end=None,
        tracklines=0,
        cross_track_links=None,
    ):
```

The multi-line layout does make it hard to fit on a screen - as you say more scrolling. I would even argue this is a good thing, as functions/classes shouldn't have this many arguments (sorry Leighton!). Adopting black will discourage this in new code.

In hindsight, given there are lots of (x, y) pairs in GenomeDiagram, perhaps passing these points a 2-tuples would have been neater for the API (but we're stuck with this now for backward compatibility).

As an aside, there is probably a flake8 plugin for this kind of complexity check?

As you say, "*it takes away much of the freedom to sort/arrange things in a meaningful manner*". I highlighted the example or horizontal placement of some multi-line comments in this pull request - do you have another example in GenomeDiagram?

widdowquinn:
> "I would even argue this is a good thing, as functions/classes shouldn't have this many arguments (sorry Leighton!). Adopting black will discourage this in new code."

No apology necessary :D It's bad design and I should have used something `struct`-like, instead.

One of the things I like about `black` is that where I think it makes my code look unwieldy or ugly, what it's often telling me is that I've made a poor design choice.

MarkusPiotrowski:
When reading PEP8 and looking at all the examples in the Python documentation, there is a clear preference for using line continuation using parentheses (brackets, etc.), *together* with an indentation to the parentheses level.
On the other hand, there is no indication that argument lists should preferentially be split to have one argument per line (again, most what I have seen from official Python documents are multiple arguments per line). This is what I mean when saying "the spirit of PEP8". Sorry for starting this argument, this should be discussed (and has been or is being discussed) elsewhere.

One example of "taking away the freedom to sort thing in a meaningful way":
```python
# before black
def draw_box(point1, point2,
             color=colors.lightgreen, border=None, colour=None,
             **kwargs):
```
Here the author has obviously organized the arguments in a meaningful manner; the two necessary and important arguments on the first line, optional formatting parameters on the second, optional other parameters last. 
This is easier/faster to read and understand (for me) than the black version:
```python
def draw_box(
    point1, point2, color=colors.lightgreen, border=None, colour=None, **kwargs
):
```

I don't write much new code, here in Biopython I'm mostly working on existing code. Often, I don't have a clue what a class/method/function is doing exactly after I read the code for the first time, so I need to become acquainted with the code. And this is easier when seeing more of the code at the same time, without too much scrolling up and down. Also, a meaningful and *visible* sorting of arguments will help. 
I fear that `black` will make the reading of code more difficult  (Our newspaper still publishes Charles M Schulz comics. In the last weeks they published a story where Snoopy was reading Tolstoy's War and Peace, but only one word each day. Isn't this strange?).

peterjc:
The ``draw_box`` example is a good one.

Black will squeeze things down to one line if it fits (as here). It will respect comments though, so black is happy with this:

```python
def draw_box(
    point1,
    point2,
    # Optional color arguments,
    color=colors.lightgreen,
    border=None,
    colour=None,  # historical UK spelling
    **kwargs  # Accept arbitrary arguments to pass to ReportLab Polygon
):
```

i.e. If there is a real reason for the original grouping (beyond cosmetic line breaking), or other things, you can make it explicit. Granted this is still _much_ longer than the tree line original.

Personally I am OK with the black one line version here. The meaning is still there in the Python mandatory vs optional arguments, but accept it is less visually emphasised.

widdowquinn:
> "Here the author has obviously organized the arguments in a meaningful manner; the two necessary and important arguments on the first line, optional formatting parameters on the second, optional other parameters last."

As that original author, my view is that the `black` formatting is fine, and it highlights that I should have used a struct-like object to group formatting parameters to make the code more readable, rather than the cosmetic linebreaks ;)

I appreciate that there are a diverse, strong and heartfelt opinions about code style snd formatting, and preferences for what is most readable. I guess this code sat in the "not brilliantly written, but forced to be readable" category, and after `black` it sits in the "this code now looks as ugly as it actually is" category. I see why that could be taken as not being an improvement.

MarkusPiotrowski:
I found it interesting, that without any knowledge of the `GenomeDiagram` code, I immediately was able to understand that there was an intention in sorting these three "categories" of parameters. Really, I didn't search for this example, it sprang to my eye (that's likely not an English idiom, but you understand what I mean) when I went through the diffs.

peterjc:
I've applied the minor improvements noticed via the black changes in #2127 (which is why there are now loads of merge conflicts).

I don't mind redoing this branch (it should be as simple as re-running black, and cherry-picking the commit to undo the slice changes which would otherwise trigger flake8 E203), but it seems it would not be universally welcome at this time.

I'm going to leave this open until at least the BOSC 2019 CoFest meeting next month, when some of the Biopython team will be able to have an in person discussion, and perhaps experiment further with black.

sbliven:
After discussing at CoFest I am in favor of using black.

peterjc:
Closing in favour of #2251 (done as a new pull request to preserve this version and comments as a medium term record of the discussion).

PR #2126, opened 2019-06-13
Fixed Doctest for Bio.phenotype

svalqui:
This pull request addresses issue #1902 


Tests/run_tests.py
```diff
@@ -97,9 +97,6 @@ def is_pypy():
     'Bio.PDB.PDBIO',
     'Bio.PDB.ResidueDepth',
     'Bio.PDB.vectors',
-    'Bio.phenotype',
-    'Bio.phenotype.parse',
-    'Bio.phenotype.phen_micro',
```
peterjc: Rather than removing them from the general exclude list, you need to move them to the NumPy specific list - they are failing on the TravisCI target which deliberately does not have numpy:

```
...
Bio.pairwise2 docstring test ... ok
Bio.phenotype docstring test ... FAIL, ImportError
Bio.phenotype.phen_micro docstring test ... FAIL, ImportError
Bio.triefind docstring test ... ok
BioSQL docstring test ... ok
BioSQL.BioSeq docstring test ... ok
BioSQL.DBUtils docstring test ... ok
BioSQL.Loader docstring test ... ok
ERROR while importing Bio.phenotype: No module named 'numpy'
ERROR while importing Bio.phenotype.phen_micro: No module named 'numpy'
```

Bio/phenotype/__init__.py
```diff
@@ -201,29 +201,29 @@ def read(handle, format):
...
     >>> print("Number of wells %i" % len(record))
     Number of wells 96
 
     If the handle contains no records, or more than one record,
     an exception is raised.  For example:
 
-    >>> from Bio import phenotype
-    >>> record = phenotype.read("plates.csv", "pm-csv")
-    Traceback (most recent call last):
+        from Bio import phenotype
```
peterjc: You need a double colon on the preceding paragraph to make this an RST literal block.

Bio/phenotype/phen_micro.py
```diff
@@ -528,32 +565,32 @@ class WellRecord(object):
...
-    WellRecord or using lists indeces or slices:
+    WellRecord or using lists indexes or slices:
 
     >>> from Bio import phenotype
-    >>> plate = phenotype.read("plate.csv", "pm-csv")
+    >>> plate = phenotype.read("phenotype/Plate.json", "pm-json")
     >>> well = plate['A05']
     >>> for time, signal in well:
-    ...    print("Time: %f, Signal: %f" % (time, signal))
+    ...    print("Time: %f, Signal: %f" % (time, signal)) # doctest:+ELLIPSIS
```
peterjc: Personally I would use 2dp here, ``print("Time: %0.2f, Signal: %0.2f" % (time, signal))``

This is easier to read in the output, but also makes the test less likely to break on different platforms due to floating point differences (although that seems to be less of a problem that it used to be with older Python 2 on Windows in particular IIRC).

On the other hand, this old style string formatting in Python is going out of fashion, and the test works on all our automated platforms, so it is probably fine as is.

peterjc:
I spotted one missing colon, but otherwise this looks ready.

Bio/phenotype/phen_micro.py
MarkusPiotrowski: I wonder if the doctest directive (`# doctest:+ELLIPSIS`) is necessary here. First, ELLIPSIS is already turned on in `run_tests.py`, second the three dots in the following line are not an ELLIPSIS but are related to the indentation (here, in fact, the end of the indented block).

svalqui: I have corrected the typo. should be fine now.

@MarkusPiotrowski , if I remove the doctest directive, the test fails as if it expects the ellipsis as literal rather than the continuation of output.

Failed example:
    for time, signal in well:
       print("Time: %f, Signal: %f" % (time, signal))
Expected:
    Time: 0.000000, Signal: 14.000000
    Time: 0.250000, Signal: 13.000000
    Time: 0.500000, Signal: 15.000000
    Time: 0.750000, Signal: 15.000000
    ...
Got:
    Time: 0.000000, Signal: 14.000000
    Time: 0.250000, Signal: 13.000000
    Time: 0.500000, Signal: 15.000000
    Time: 0.750000, Signal: 15.000000
    Time: 1.000000, Signal: 16.000000
    Time: 1.250000, Signal: 19.000000
(This continues)

I think this relates to the indentation on loops.

MarkusPiotrowski: If I look at the doctest in `TogoWS` it seems that you don't need/should not include the three dots that closes the loop (line 575). Can you try to remove them and the doctest directive and see if this works?

peterjc: Normally including the ``...`` for closing the for loop is fine, and preferable for better matching what the user would see if following the example themselves. Perhaps ``# doctest:+ELLIPSIS`` interacts with this badly?

Also beware that the TogoWS doctest is not run on TravisCI/AppVeyor as they are done in offline mode.

peterjc: Actually the TogoWS doctest is broken - it needs a blank line above the ">>>" otherwise the RST is rendered as a single paragraph.

MarkusPiotrowski: @svalqui It's merged, I'm just curious: Did you run this test via the `run_tests.py` module in test or differently, e.g. with `python doctest ...`?

peterjc: Good point - it ought to have picked up ``optionflags=doctest.ELLIPSIS`` via ``run_tests.py``,or via this trick used at the end of the source code files:

```python
if __name__ == "__main__":
    run_doctest()
```

So if that is working, we _should_ be able to remove the special comment from the doctest itself...

Issue #455, opened 2015-01-07
PDB Entity.copy() mangles disordered atoms.


macraild:
Disordered atoms dont inherit `disordered_get_list()` when a parent entity is copied. Eg:

```
>>> pdbl = Bio.PDB.PDBList(server='http://www.rcsb.org/pdb/files')
>>> pdbParser = Bio.PDB.PDBParser()
>>> struct_4LGR = pdbParser.get_structure('4LGR', pdbl.retrieve_pdb_file('4LGR'))
Downloading PDB structure '4LGR'...
>>> Bio.PDB.PPBuilder().build_peptides(struct_4LGR)
[<Polypeptide start=5 end=33>, <Polypeptide start=36 end=259>, <Polypeptide start=2 end=123>]
>>> Bio.PDB.PPBuilder().build_peptides(struct_4LGR.copy())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.6/site-packages/Bio/PDB/Polypeptide.py", line 357, in build_peptides
    and is_connected(prev_res, next_res):
  File "/usr/lib/python2.6/site-packages/Bio/PDB/Polypeptide.py", line 417, in _is_connected
    nlist=n.disordered_get_list()
AttributeError: 'Atom' object has no attribute 'disordered_get_list'
```

peterjc:
I suspect (from just browsing the code) that the problem is neither ``Bio.PDB.Atom.DisorderedAtom`` nor the underlying base class ``Bio.PDB.Entity.DisorderedEntityWrapper`` define a ``.copy()`` method?

macraild:
That looks right to me. I'm not sure I completely understand how disorder is handled here, but from my reading ``.copy()`` should be defined at the ``DisorderedAtom`` level, and a minimal implimentation might be:

```
def copy(self):
    shallow = copy.copy(self)
    for child in self.child_dict.values():
        shallow.disordered_add(child.copy())
    return shallow
```

This fixes my immediate problem (ie. disordered ``Atom`` objects in copies are now appropriately contained in ``DisorderedAtom``, and ``Bio.PDB.PPBuilder`` now works). I may have missed something important in the above, and don't know what other behaviour needs to be tested here.

peterjc:
[...]

PR #2134, opened 2019-06-14
Fix PDB Entity.copy() for disordered atoms

ghost:
This pull request addresses issue #455

Implemented fix suggested in comments.

### DIFF
--- a/Bio/PDB/Atom.py
+++ b/Bio/PDB/Atom.py
@@ -434,6 +434,12 @@ class DisorderedAtom(DisorderedEntityWrapper):
         """Return disordered atom identifier."""
         return "<Disordered Atom %s>" % self.get_id()
 
+    def __copy__(self):
+        shallow = copy.copy(self)
+        for child in self.child_dict.values():
+            shallow.disordered_add(child.copy())
+        return shallow
+
     def disordered_add(self, atom):
         """Add a disordered atom."""
         # Add atom to dict, use altloc as key


peterjc:
I've retitled the pull request to something more meaningful. Note we will not consider at this for potential inclusion unless you agree to the licensing terms. i.e. Please read and tick the three boxes from the issue template. Thanks!

JoaoRodrigues:
Will look into it this weekend! Thanks for the patch, this is a long-standing issue.

peterjc:
@Yuval-Shechter Thanks for agreeing to the licensing 👍 

@JoaoRodrigues Thanks for agreeing to look at this

peterjc:
We're agreed that our early 2020 release will be the last Biopython to support Python 2, so if there is an elegant/simple Python 3 only solution that isn't very far away... even doing this conditionally in the short term might be acceptable if a Python 2 solution is judged too much work?

JoaoRodrigues:
That's a good suggestion @ChrisMoth. I implemented a `DisorderedEntity.copy()` method that gets your example to work. I'm testing it in other situations to see if it holds properly.

```
    def copy(self):
        """Returns a shallow copy of the DisorderedEntity, including children"""

        shallow = copy(self)

        shallow.child_list = []
        shallow.child_dict = {}
        shallow.xtra = copy(self.xtra)

        for child in self.disordered_get_list():
            shallow.disordered_add(copy(child))

        shallow.detach_parent()
```

JoaoRodrigues:
Ran into an interesting behavior (ie bug) while checking this and checking #455.
```from Bio import PDB

inputPdb = "mini.pdb"
output  = "out.pdb"

parser = PDB.PDBParser()
struct = parser.get_structure("myStruct", inputPdb)

idMat=( (1,0,0), (0,1,0), (0,0,1) )
trVect=(20.,20.,20.)

s = struct.copy()
s.transform(idMat, trVect)

pdbio = PDB.PDBIO()
pdbio.set_structure(s)

# Print correct, translated coordinates
for atom in pdbio.structure.get_atoms():
    print(atom, atom.coord)

# Prints original, unchanged coordinates
for res in pdbio.structure.get_residues():
    for atom in res.get_unpacked_list():
        print(atom, atom.coord)

with open(output, "w") as fd :
  pdbio.save(fd)
```

The problem seems to come from `Entity.get_list`, which somehow does not let transform operate properly on all children. We could add a transform method to DisorderedEntities as well, to make sure they call `.get_unpacked_list` instead, which leads to the correct behavior.

I'm afraid I might be missing a more obvious fix here. Would anyone else have some time to have a look? @jgreener64 @peterjc ?

jgreener64:
Don't think I'll get a chance to dive into this soon, sorry, but maybe in a week or two!

PR #2137, opened 2019-06-17
Speed up index loading from SQLite for SeqIO, SearchIO, etc

pkerpedjiev:
The current `SELECT COUNT(key) FROM offset_data;` call to obtain the number of sequences in the database is slow for substantial fasta files.

Because the index is read-only, this can be replaced with `SELECT MAX(_ROWID_) FROM offset_data;` for orders of magnitude better performance.

For example, on a fasta with 17M sequences, loading the index with the current `COUNT(key)` call takes about 4 seconds. With `MAX(_ROWID_)` it takes 2.64ms.

On a fasta file with 307M sequences, loading the index with `COUNT(key)` takes 1 minute and 20 seconds. With `MAX(_ROWID_)` it takes 10ms. 

### DIFF
--- a/Bio/File.py
+++ b/Bio/File.py
@@ -587,7 +587,7 @@ class _SQLiteManySeqFilesDict(_IndexedSeqFileDict):
                 con.close()
                 raise ValueError("Unfinished/partial database")
             count, = con.execute(
-                "SELECT COUNT(key) FROM offset_data;").fetchone()
+                "SELECT MAX(_ROWID_) FROM offset_data;").fetchone()
             if self._length != int(count):
                 con.close()
                 raise ValueError("Corrupt database? %i entries not %i"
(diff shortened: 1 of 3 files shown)


peterjc:
Discussion on eg https://stackoverflow.com/questions/8988915/sqlite-count-slow-on-big-tables suggests this is a known SQLite weakness, and common work around.

The only reason we're doing this check is for data consistency, so I think your change is fine - even given the possibility of other tools/scripts building or modifying our SQLite index format.

However, I do think the code change does need some explanation as a comment, e.g.

```python
# Using "SELECT MAX(key) FROM offset_data;" is too slow
```

@bow Do you agree?

@pkerpedjiev Do you want to put your name in ``NEWS.rst`` too?

pkerpedjiev:
@peterjc Thanks for the quick response! I added a comment with a link to the StackOverflow question and a short explanation and also added my name to NEWS.rst.

peterjc:
Merged, thank you 👍

Issue #2139, opened 2019-06-18
SearchIO does not support nhmmscan outout from HMMER3


peterjc:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```pycon
>>> import sys; print(sys.version)
3.7.2 (default, Dec 29 2018, 06:19:36) 
[GCC 7.3.0]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-3.10.0-957.12.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
>>> import Bio; print(Bio.__version__)
1.74.dev0
```

(*Please copy and run the above in your Python, and copy-and-paste the output*)

### Expected behaviour

The pair of tools ``nhmmer`` / ``nhmmscan`` are new to HMMER3.1.

Much as we can parser the output from protein based HMMER3 tool ``hmmscan``, I would like to parse output from the DNA based HMMER3 tool ``nhmmscan`` as well.

### Actual behaviour

Parsing fails.

### Steps to reproduce

Using HMMER 3.2.1, tested under both Linux and macOS:

Get sample data,

```
$ curl -L -O http://eddylab.org/software/hmmer/hmmer-3.2.1.tar.gz
$ tar -zxvf hmmer-3.2.1.tar.gz
$ cd hmmer-3.2.1/tutorial/
$ ls MADE1.hmm dna_target.fa 
```

Compile the DNA HMM:

```
$ hmmpress MADE1.hmm 
Working...    done.
Pressed and indexed 1 HMMs (1 names and 1 accessions).
Models pressed into binary file:   MADE1.hmm.h3m
SSI index for binary model file:   MADE1.hmm.h3i
Profiles (MSV part) pressed into:  MADE1.hmm.h3f
Profiles (remainder) pressed into: MADE1.hmm.h3p
```

Run test search with ``nhmmscan``, getting multiple output formats:

```
$ nhmmscan -o n.txt --tblout n.table --dfamtblout n.dfam MADE1.hmm dna_target.fa
```

Run similar search but with ``hmmscan`` (intended for proteins):
```
$ hmmscan -o p.txt --tblout p.table --domtblout p.domtable --pfamtblout p.pfam MADE1.hmm dna_target.fa
```

First, the ``hmmscan`` output parses two of the four possible outputs:

```pycon
>>> from Bio import SearchIO
>>> print(len(list(SearchIO.parse("p.txt", "hmmer3-text"))))
1
[...]

PR #2140, opened 2019-06-18
Replace assert in HMMER3 table parser

peterjc:
This pull request fixes a corner case found while looking at #2139 (trying to parse ``nhmmscan`` output which has different columns). It replaces an assert statement with an explicit ``ValueError`` instead.

### DIFF
--- a/Bio/SearchIO/HmmerIO/hmmer3_tab.py
+++ b/Bio/SearchIO/HmmerIO/hmmer3_tab.py
@@ -38,15 +38,11 @@ class Hmmer3TabParser(object):
     def _parse_row(self):
         """Return a dictionary of parsed row values (PRIVATE)."""
         cols = [x for x in self.line.strip().split(' ') if x]
+        if len(cols) < 18:
+            raise ValueError("Less columns than expected, only %i" % len(cols))
         # if len(cols) > 19, we have extra description columns
         # combine them all into one string in the 19th column
-        if len(cols) > 19:
-            cols[18] = ' '.join(cols[18:])
-        # if it's < 19, we have no description columns, so use an empty string
-        # instead
-        elif len(cols) < 19:
-            cols.append('')
-            assert len(cols) == 19
+        cols[18] = ' '.join(cols[18:])
 
         # assign parsed column data into qresult, hit, and hsp dicts
         qresult = {}


Bio/SearchIO/HmmerIO/hmmer3_tab.py
```diff
@@ -42,11 +42,12 @@ def _parse_row(self):
...
         if len(cols) > 19:
             cols[18] = ' '.join(cols[18:])
-        # if it's < 19, we have no description columns, so use an empty string
+        # if it's 18, we have no description columns, so use an empty string
         # instead
-        elif len(cols) < 19:
+        elif len(cols) == 18:
             cols.append('')
-            assert len(cols) == 19
+        else:
```
bow: This is slightly different than the previous check, which does nothing when `len(cols) == 19`. Perhaps this can be `elif len(cols) < 18` instead of `else`. The tests seem to detect this as well.

On a related note, it seems ok if the entire loop is re-written also. Something like:

```
cols = [x for x in self.line.strip().split(' ') if x]
if len(cols) < 18:
    raise ValueError("Less columns than expected, only %i" % len(cols)
cols[18] = '  '.join(cols[18:])    # which has the same effect as the two, previous if-clauses.
```

peterjc: You're right, I missed the case when there already where 19 columns.

peterjc:
If you want to use the test cases I logged on #2139, please do so.

Bio/SearchIO/HmmerIO/hmmer3_tab.py
peterjc: Applied your alternative which is much shorter :)

bow: Thank you :). I made a slight mistake in the `.join` call, I now realize. It should be `' '` (single space), instead of `'  '` (double space). That's why the tests is still failing now. After this is fixed, it should be good to merge ~

peterjc: Thanks - I didn't see that when I copy-and-pasted it.

peterjc:
AppVeyor broke with a conda outage (getting ready to upload test coverage).

PR #2141, opened 2019-06-18
Fix RST markup; doctest ELLIPSIS is redundant

peterjc:
Cross reference discussion on #2126 with @MarkusPiotrowski and @svalqui 

This was missing a blank line which on http://rst.ninjs.org at least was needed before the ``>>>`` to render the code snippet.

Unfortunately, with that changed, the doctest fails due to the incomplete output. We could try using ``# doctest:+ELLIPSIS`` but the nature of this example is going to be highly unstable over time.

### DIFF
--- a/Bio/TogoWS/__init__.py
+++ b/Bio/TogoWS/__init__.py
@@ -189,6 +189,7 @@ def search_iter(db, query, limit=None, batch=100):
        TogoWS (currently limited to 100).
 
     You would use this function within a for loop, e.g.
+
     >>> from Bio import TogoWS
     >>> for id in TogoWS.search_iter("pubmed", "diabetes+human", limit=10):
     ...     print("PubMed ID: %s" %id) # maybe fetch data with entry?
--- a/Bio/TogoWS/__init__.py
+++ b/Bio/TogoWS/__init__.py
@@ -349,3 +349,8 @@ def _open(url, post=None):
 
 
 _open.previous = 0
+
+
+if __name__ == "__main__":
+    from Bio._utils import run_doctest
+    run_doctest(verbose=0)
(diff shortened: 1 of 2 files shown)


MarkusPiotrowski:
@peterjc I don't think it's necessary to convert the doctest to a literal string. It is working (run it in verbose mode and you see that it is passing):
```python
"""Run this as doctest.

>>> from Bio import TogoWS
>>> for id in TogoWS.search_iter("pubmed", "diabetes+human", limit=10):
...     print("PubMed ID: %s" %id) # maybe fetch data with entry? # doctest:+ELLIPSIS 
...
PubMed ID: ...

"""

if __name__ == '__main__':
    import doctest
    doctest.testmod(verbose=True)
```
Here I had to use the docstring directive  # doctest:+ELLIPSIS since I was not running this from our `run_tests`.
I guess this was also the same problem with `Bio.phenotype`. Possibly @svalqui tested it not using `run_tests.'

peterjc:
I should re-test then.

There could be something not quite right in how ``run_tests.py`` calls the doctests, perhaps dependent on how Biopython is installed (e.g. develop mode?), see also #531 which was never fully explained.

svalqui:
run_test.py calls optionflags = doctest.ELLIPSIS

https://github.com/biopython/biopython/blob/c82a86cff5560c5f0ebabd128a50d1a129c4a3a1/Tests/run_tests.py#L376-L377

This doctest only expects the text "PubMed ID: ", which is included in the print statement, for this to fail the search would need to return nothing, unlikely for the search.
I think adding the missing line should be good.

peterjc:
Yes, you seem to be right. Updating...

Issue #2136, opened 2019-06-17
Unexpected output of pairwise2 localms


juheon:
### Setup

I am reporting a problem with Biopython version: 1.73, Python version, and operating
system as follows: 

```
Python 3.5.4 |Anaconda custom (64-bit)| (default, Nov 20 2017, 18:44:38)
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys; print(sys.version)
3.5.4 |Anaconda custom (64-bit)| (default, Nov 20 2017, 18:44:38)
[GCC 7.2.0]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-3.19.0-61-generic-x86_64-with-debian-stretch-sid
>>> import Bio; print(Bio.__version__)
1.73
```

### Expected behaviour

```
GCCATTAAAAGCATTGTCCATAAAAAAAAAAAAAAAAAAA-
                     |||||||||||||||||||
---------------------AAAAAAAAAAAAAAAAAAAA
    score=33

...(18 cases with the same score but with different gap locations)...
GCCATTAAAAGCATTGTCCAT-AAAAAAAAAAAAAAAAAAA
                      |||||||||||||||||||
---------------------AAAAAAAAAAAAAAAAAAAA
    score=33
```

### Actual behaviour
According to the https://biopython.org/DIST/docs/api/Bio.pairwise2-module.html, "Since Biopython 1.71 identical matches are shown with a pipe character, mismatches as a dot, and gaps as a space."
In the first result, the 21st base of sequence A is 'T' and the corresponding base of sequence B is 'A'. Since there is neither pipe nor dot character, the 21tst base seems to regarded as the gap. In that case, I think the score should be 33 (=2*19-5). However, the score is 38.
If the 21st base is a mismatch, the score should be 37 (=2*19-1). 
The unexpected behavior of the second result is basically the same as the first result. 
```
GCCATTAAAAGCATTGTCCATAAAAAAAAAAAAAAAAAAA
                     |||||||||||||||||||
--------------------AAAAAAAAAAAAAAAAAAAA
  Score=38

GCCATTAAAAGCATTGTCCATAAAAAAAAAAAAAAAAAAA-
                     |||||||||||||||||||
---------------------AAAAAAAAAAAAAAAAAAAA
  Score=38
```

### Steps to reproduce

```
from Bio import pairwise2
from Bio.pairwise2 import format_alignment
[...]

PR #2143, opened 2019-06-18
Pairwise2: Add keyword argument "show_full_sequences" to "format_alignment"

MarkusPiotrowski:
This pull request results from the discussion in issue #2136.

With PR #1505 the output of local alignments in `pairwise2` with `format_alignment` was optimized to only show the aligned parts of both sequences. However, there may be situations where it would be beneficial to see also the un-aligned leading and trailing parts of both sequences (https://github.com/biopython/biopython/issues/2136#issuecomment-502736721).

Thus, with this PR, a new keyword argument `show_full_sequences` is added to `format_alignment` to restore the old output (prior Biopython 1.74) of local alignments. The new argument defaults to `False`.

### DIFF
--- a/Bio/pairwise2.py
+++ b/Bio/pairwise2.py
@@ -1100,7 +1100,7 @@ def print_matrix(matrix):
 
 
 def format_alignment(align1, align2, score, begin, end,
-                     show_full_sequences=False):
+                     full_sequences=False):
     """Format the alignment prettily into a string.
 
     IMPORTANT: Gap symbol must be "-" (or ['-'] for lists)!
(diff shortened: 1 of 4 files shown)


Bio/pairwise2.py
```diff
@@ -1099,7 +1099,8 @@ def print_matrix(matrix):
                        for j in range(len(matrix[i]))))
 
 
-def format_alignment(align1, align2, score, begin, end):
+def format_alignment(align1, align2, score, begin, end,
+                     show_full_sequences=False):
```
peterjc: Would ``full_seq`` be enough as the variable name? That this is about displaying or showing seems clear to me from the context.

peterjc:
Looks fine to me (minor comment on argument name), @mdehoon ?

Bio/pairwise2.py
MarkusPiotrowski: OK. I changed that.

peterjc:
In light of #2216, could the new option be documented more explicitly for how to reproduce the historic behaviour?

Issue #2101, opened 2019-06-03
Update rebase_update.py for Python 3


peterjc:
See discussion on #2098, it would be good to update ``Scripts/Restriction/rebase_update.py`` to work on Python 3. This script still works on Python 2.7 as is used as part of updating ``Bio.Restriction`` to the latest version of the enzymes in the REBASE database.

(See also #972 for background, and and also the work on #268 to replace ``Bio.Restriction`` entirely)

PR #2144, opened 2019-06-19
Rebase update

Andrey-Raspopov:
This pull request addresses issue #2101 

This is a skeleton of another take on this task, using FTPlib instead of deprecated url.request mechanisms. Any feedback is appreciated.

- [ X] I have read the ``CONTRIBUTING.rst`` file, have run ``flake8`` locally, and

### DIFF
--- a/Bio/Restriction/RanaConfig.py
+++ b/Bio/Restriction/RanaConfig.py
@@ -58,10 +58,10 @@ ftp_proxy = ''
 #
 #   Do not modify the addresses.
 #
-ftp_Rebase = 'ftp://ftp.neb.com/'
-ftp_emb_e = ftp_Rebase + 'pub/rebase/emboss_e.###'
-ftp_emb_s = ftp_Rebase + 'pub/rebase/emboss_s.###'
-ftp_emb_r = ftp_Rebase + 'pub/rebase/emboss_r.###'
+ftp_Rebase = 'ftp.neb.com'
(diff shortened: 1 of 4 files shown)


peterjc:
Will this conflict with #2130?

MarkusPiotrowski:
@Andrey-Raspopov I have the same error under Python 2 and Python 3 on two different Windows computers. A quik Google search let suggest that the error is related to setting of system variables dealing with proxies. 

While I don't object in general using `ftplib` and having `rebase_update` less drastically changed as I did in PR #2130, I'd prefer to keep most oft these simplifications (e.g. removing `RanaConfig.py`) and then build upon that.

@peterjc Yes, this will conflict #2130.

peterjc:
So unless @Andrey-Raspopov can improve this, we should go with #2130 instead?

MarkusPiotrowski:
@peterjc If you can confirm that #2130 is working for you, I'd suggest going with #2130 for now. Then, if @Andrey-Raspopov is willing to do that, he can try to change the updated `rebase_update.py`, e.g. using `ftplib` (I used `urllib`'s `urlretrieve`, which is deprecated but nicely working...)

Andrey-Raspopov:
Sorry, was out of reach last two weeks. I've checked out #2130 and it is more simple and robust in my opinion, so I think we should go with it.

PR #2164, opened 2019-07-06
Make Entrez module retry requests on failure automatically

ExplodingCabbage:

---

I figured it was dumb and user-unfriendly for the end-user to have to write this logic all the time, especially given that we can see [evidence in the wild](https://www.biostars.org/p/121287/) of people hitting these transient errors and not knowing what to do about them, so I thought it'd make sense to chuck it into biopython itself. Here's a proposal that does that.

It turns out the tutorial already proposed using basically the same retry-on-error logic as I've used here, but the tutorial's approach is slightly inferior because it only catches `HTTPError`s. It's possible to get a transient `URLError` that is not a `HTTPError` but rather a connection timeout due to NCBI dropping a connection. It makes sense to catch those too.

At this point I think this is a good change but have a few concerns that you may want to consider before deciding whether to merge:

* I haven't built the docs and checked my doc changes look good; I'm not sure how to do so and don't see instructions on it in CONTRIBUTING.rst

* I don't have any unit tests for this functionality. I figure we probably don't want them since we don't want tests that hit NCBI when run. I'm currently using the version of the code from this PR to run a script that downloads ~20000 sequence ranges from NCBI with `efetch` and reliably failed at some point with a 5XX or connection timeout prior to me making these changes; assuming that still runs to completion fine, I guess that's at least a reasonable level of one-off testing. Update: this ran to completion fine.

* People who've already implemented their own retry-on-failure logic around these functions in their programs will end up with an unwantedly large maximum number of retries (e.g. 3*3 = 9 retries) if they don't see this change in the release notes and delete their own retry boilerplate. Probably not really a problem, but maybe annoying to some people.

* If for some reason somebody wants to use a different retry strategy altogether (e.g. exponential backoff instead of a fixed sleep) this change gets in their way instead of making their life simpler.

Despite all that I think this is a sensible default that's more user-friendly than what was there before, and therefore a desirable change. But, up to you!

### DIFF
--- a/Bio/Entrez/__init__.py
+++ b/Bio/Entrez/__init__.py
@@ -103,7 +103,7 @@ import warnings
 # Importing these functions with leading underscore as not intended for reuse
 from Bio._py3k import urlopen as _urlopen
 from Bio._py3k import urlencode as _urlencode
-from Bio._py3k import HTTPError as _HTTPError
+from Bio._py3k import URLError as _URLError, HTTPError as _HTTPError
 
 from Bio._py3k import _binary_to_string_handle, _as_bytes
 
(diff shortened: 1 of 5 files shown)


peterjc:
Looks like some glitch in the flake8 dependencies is breaking the style checks on Python 2, probably harmless for this pull request:

```
$ flake8
"flake8-docstrings" failed during execution due to "module 'pydocstyle' has no attribute 'tokenize_open'"
```

peterjc:
This looks like a very nice improvement - thank you!

Thanks also  @mdehoon for reviewing it.

Issue #2142, opened 2019-06-18
Appending to bgzf files misses sequences


pkerpedjiev:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
>>> import sys; print(sys.version)
3.6.7 |Anaconda, Inc.| (default, Oct 23 2018, 14:01:38)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Darwin-17.7.0-x86_64-i386-64bit
>>> import Bio; print(Bio.__version__)
1.73
```

### Expected behaviour

The script in `Steps to reproduce` outputs both of the sequences added to the bgzf file:

```
>hello
aaaaaaaaaaaaaaaaaa
>there
cccccccccccccccccc
```

### Actual behaviour

It only outputs the first sequence:

```
>hello
aaaaaaaaaaaaaaaaaa
```

### Steps to reproduce

Run the following script:

```
import Bio.bgzf as bgzf
from Bio import SeqIO
import gzip
import os
import os.path as op

# file to be used for testing bgzf fasta
bgz_filename = '/tmp/test.bgz'
if op.exists(bgz_filename):
    os.remove(bgz_filename)

# file to be used for testing text fasta
txt_filename = '/tmp/test.fa'
if op.exists(txt_filename):
    os.remove(txt_filename)

# write the first sequence
with bgzf.open(bgz_filename, 'ab') as f, open(txt_filename, 'ab') as f1:
    f.write('>hello\n'.encode('utf8'))
    f.write('aaaaaaaaaaaaaaaaaa\n'.encode('utf8'))

    f1.write('>hello\n'.encode('utf8'))
    f1.write('aaaaaaaaaaaaaaaaaa\n'.encode('utf8'))

# write the second sequence
with bgzf.open(bgz_filename, 'ab') as f, open(txt_filename, 'ab') as f1:
    f.write('>there\n'.encode('utf8'))
    f.write('cccccccccccccccccc\n'.encode('utf8'))

    f1.write('>there\n'.encode('utf8'))
    f1.write('cccccccccccccccccc\n'.encode('utf8'))

print("bgzf")
# the following should print all the lines but it only prints the
# lines for the first sequence

with bgzf.open(bgz_filename, 'rb') as f:
    bgzf_txt = f.read(1000000)
    f.seek(0)
    for line in f:
        print("line:", line)

[...]

PR #2166, opened 2019-07-08
bgzf: Make append mode work

tipabu:
Previously, trying to open a BGZF file in append mode would start writing data after the previous EOF marker. As a result, the appended data would not be available when the file was re-opened for reading.

Now when opening in append mode:

  - read the last 28 bytes of the file,
  - verify that it matches the EOF marker, and
  - truncate the file so it no longer includes the EOF

before re-opening the file in append mode. This ensures that only one EOF marker is written when the `BgzfWriter` is closed. At the same time, it (mostly) preserves callers' append-only expectations.

Note that a series of small appends will create a series of small segments; this is similar to what would happen if the caller did a flush rather than a close and re-open.

This pull request addresses issue #2142

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -247,6 +247,7 @@ If your data is in UTF-8 or any other incompatible encoding, you must use
 binary mode, and decode the appropriate fragments yourself.
 """
 
+import os
 import struct
 import sys
 import zlib
@@ -806,8 +807,10 @@ class BgzfWriter:
             if "w" not in mode.lower() and "a" not in mode.lower():
                 raise ValueError(f"Must use write or append mode, not {mode!r}")
             if "a" in mode.lower():
-                raise NotImplementedError("Append mode is not implemented yet")
-                # handle = _open(filename, "ab")
+                handle = _open(filename, "ab")
+                # py27 on Windows will get confused about our current position
+                # immediately after opening; explicitly seek to end to fix it
+                handle.seek(0, os.SEEK_END)
             else:
                 handle = _open(filename, "wb")
         self._text = "b" not in mode.lower()
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -247,7 +247,6 @@ If your data is in UTF-8 or any other incompatible encoding, you must use
 binary mode, and decode the appropriate fragments yourself.
 """
 
-import os
 import struct
 import sys
 import zlib
@@ -808,9 +807,6 @@ class BgzfWriter:
                 raise ValueError(f"Must use write or append mode, not {mode!r}")
             if "a" in mode.lower():
                 handle = _open(filename, "ab")
-                # py27 on Windows will get confused about our current position
-                # immediately after opening; explicitly seek to end to fix it
-                handle.seek(0, os.SEEK_END)
             else:
                 handle = _open(filename, "wb")
         self._text = "b" not in mode.lower()
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -25,6 +25,9 @@ The Local Composition Complexity functions from ``Bio.SeqUtils`` now uses
 base 4 log instead of 2 as stated in the original reference Konopka (2005),
 Sequence Complexity and Composition. https://doi.org/10.1038/npg.els.0005260
 
+Append mode is now supported in ``Bio.bgzf`` (and a bug parsing blocked GZIP
+files with an internal empty block fixed).
+
 The experimental warning was dropped from ``Bio.phenotype`` (which was new in
 Biopython 1.67).
 
@@ -39,6 +42,7 @@ possible, especially the following contributors:
 - Damien Goutte-Gattat
 - Fabian Egli
 - Sebastian Bassi
+- Tim Burke
 - Michiel de Hoon
 - Peter Cock
 
--- a/Tests/test_bgzf.py
+++ b/Tests/test_bgzf.py
@@ -384,8 +384,36 @@ class BgzfTests(unittest.TestCase):
             self.assertEqual(h.read(5), "Magic")
 
     def test_append_mode(self):
-        with self.assertRaises(NotImplementedError):
-            bgzf.open(self.temp_file, "ab")
+        with bgzf.open(self.temp_file, "wb") as h:
+            h.write(b">hello\n")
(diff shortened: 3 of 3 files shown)


peterjc:
I'm unsure what would be the best behaviour for ``.tell`` in append mode... right now with your changes I think you would get virtual offsets relative to the start of the new data, which is almost certainly not desired.

I've not checked what happens in a non-compresses situation though...

peterjc:
It looks like Windows testing may be painful...

As an aside, we should probably fix ``setUp`` to use a random filename each time via ``tempfile`` instead (or the fixed filename but in a random temp folder).

tipabu:
Huh. I guess `seek()` doesn't return the new absolute offset on Windows? Oh! No, that's a py3-ism!

I'm kind of glad of the subsequent failures because of the already-open file, though; I should fix it so `rw_handle` always gets closed. I'll still shove a UUID in the filename though.

`tell()` behavior seems reasonable to me -- I'll add some more assertions to one of the tests to explore it a little. But since
``` python
with bgzf.open(file_name, 'wb') as h:
    h.write(data1)
with bgzf.open(file_name, 'ab') as h:
    posA = h.tell()
    h.write(data2)
```
is (as the patch stands now) equivalent to
``` py
with bgzf.open(file_name, 'wb') as h:
    h.write(data1)
    h.flush()
    posB = h.tell()
    h.write(data2)
```
I think it makes sense that `posA` and `posB` would be equal, too.

FWIW, I considered reading the data from the final non-EOF chunk into the buffer and truncating the file earlier, but
- I couldn't think of a reliable way to *find* that second-to-last chunk short of walking the entire file (which would be... unfortunate) and
- it wasn't clear to me that users would *expect* that behavior after closing and re-opening the file, particularly after thinking about the commentary in `BgzfBlocks`'s docstring about giving the header its own block.

Still need #2167 though :wink:

peterjc:
You shouldn't need to truncate anything - assuming we are appending to an existing valid BGZF file, it should contain zero or more complete gzip blocks, ideally ending with the 28 byte empty EOF block.

Before, say we had:

``(old block 1)(old block 2)(old block 3)(empty EOF block)``

After opening in append mode and writing some mode data, I would expect:

``(old block 1)(old block 2)(old block 3)(empty EOF block)(new block A)(new block B)(empty EOF block)``

I can see why you might want to remove the old EOF marker, giving the following - but this breaks the expectation that append mode should not touch what was already there on disk:

``(old block 1)(old block 2)(old block 3)(new block A)(new block B)(empty EOF block)``

tipabu:
Ah -- fair enough. So we'll mainly be fixing the reader, with an explicit `seek` to fix up that initial `tell`. Should simplify the patch a good bit.

tipabu:
Anything else I should do to help move this forward?

peterjc:
There's a switch from ``self._buffer`` to  ``self._block_raw_length`` (micro optimisation?)

I don't see anything which would make ``.tell()`` work? i.e. ``handle = bgzf.open("existing.bgz", "a")`` then ``print(handle.tell())`` should give the virtual offset of the end of the file.

tipabu:
From our earlier conversation, we expect an appended bgzf to look like

```
(old block 1)(old block 2)(old block 3)(empty EOF block)(new block A)(new block B)(empty EOF block)
```

Prior to this patch, the issue was in the _reader_ -- when it got to the end of `(old block 3)` and loaded another block, the EOF block would mean `self._buffer` was empty, while `self._block_raw_length` was not. As a result, appended files previously could appear truncated. Now, `.read()` and `.readline()` will continue looping until the underlying file object has run out of data.

(Side note: just realized I probably should have removed the comment about the `corner case of an empty block followed by a non-empty block` -- that's exactly what needed fixing and we now have tests that hit it.)

The `.tell()` behavior seems reasonable to me -- immediately after `.open()`, we've got a virtual offset pointing us at the end of the file (i.e., start of new data):

```
(old block 1)(old block 2)(old block 3)(empty EOF block)
                                                        ^
```

very similar to the how normal files usually behave when opened for appending. As a result, you can do things like

```
with bgzf.open(file_name, 'wb') as h:
    h.write(data1)
with bgzf.open(file_name, 'ab') as h:
    append_pos = h.tell()
    h.write(data2)
with bgzf.open(file_name, 'rb') as h:
    h.seek(append_pos)
    assert b''.join(h) == data2
```

There was some kind of issue with `.tell()` immediately after `open()` on Windows, though I don't recall the exact details now. I want to say it would return `None` until you either wrote data or did a seek? That's what the explicit `.seek(0, os.SEEK_END)` fixes.

Bio/bgzf.py
```diff
@@ -770,8 +769,10 @@ def __init__(self, filename=None, mode="w", fileobj=None, compresslevel=6):
             if "w" not in mode.lower() and "a" not in mode.lower():
                 raise ValueError("Must use write or append mode, not %r" % mode)
             if "a" in mode.lower():
-                raise NotImplementedError("Append mode is not implemented yet")
-                # handle = _open(filename, "ab")
+                handle = _open(filename, "ab")
+                # py27 on Windows will get confused about our current position
```
fabianegli, line 773: Python 2.7 is not supported any more.

fabianegli (changes requested):

Bio/bgzf.py
peterjc: Yes, sadly this pull request has languished - I've have less time available for Biopython that I used to.

peterjc:
Not sure what I was doing in late April 2020, but I've just worked through that parser bug with an empty block mid file, and extracted your test and fix as #3755. Your explanation was very clear, thank you. That should leave this PR as a much smaller change...

Bio/bgzf.py
peterjc: So, the question now is do we still need this ``handle.seek(0, os.SEEK_END)`` on Windows under Python 3.6 onwards?

Guess we should try it and see... will need another test

peterjc:
Rebased on top of #3755 leaving just the append work.

I think we need a seek/tell test here too - and that should give the reassurance needed that we can remove that Windows Python 2.7 workaround.

peterjc:
All green except the coverage numbers, see #3752 - so the Windows workaround appears not to be needed on Python 3.6 onwards.

Merging, thank you!

PR #2186, opened 2019-07-27
mmCIF reader always reads in lists

jgreener64:
This pull request addresses issue #1533 and #1884.

I *keep* running into the following gotcha when reading in mmCIF files: single values are read in as a string rather than a list, so when you iterate over them you get characters rather than the desired list items. This leads to cryptic errors such as `ValueError: could not convert string to float: '-'`, see the discussion on the issues above for more. The silent version of the error is even worse as you don't see it messing up your data.

This PR makes everything a list, even if it only contains one item. The advantages are that everything is consistent and that you won't get the above error. The disadvantages are that things with only a single value (e.g. title) will always be in a list, and that it will likely break people's code. The change would need to be communicated clearly for this reason.

Does anybody who uses the mmCIF parser have input here?


peterjc:
The proposed behaviour seems appropriate and matches a similar example with the GenBank parser where for feature annotation we consistently use lists even when for some cases one value is only ever expected.

This would need some kind of entry in the ``NEWS.rst`` file (along with ideally including your name too).

jgreener64:
Great, have made those changes.

peterjc:
Oh good, you updated the Tutorial too 👍

peterjc:
Other than @JoaoRodrigues (review requested), are there any other structure folk who might want to comment on this mmCIF reader change - @lennax maybe?

StefansM:
I got pinged by a linked ticket, so I thought I'd offer my 2p.

I unfortunately no longer have the pleasure of working with mmCIF files, but I remember this being a particularly annoying pain point. In the absence of a proper validating grammar (STAR doesn't count) and authors that stick to it (ha ha), I think this is the best that can be done.

Big :+1: for Joe's changes.

jgreener64:
Thanks for the input @StefansM, and hope you are getting on okay!

peterjc:
Hopefully Joao will get to look at this soon (not sure what his travel plans are after BOSC 2019 and the CoFest), but this looks ready to merge 👍

peterjc:
Rebasing and fixing quotes conflict...

peterjc:
Thanks Joe!

jgreener64:
Thanks for the merge Peter.

PR #2200, opened 2019-08-02
Add a bit more error checking for MutableSeq initiation

chris-rands:
Previously it was possible to initiate `MutableSeq` with integers or floats:

```
>>> from Bio.Seq import MutableSeq
>>> s = MutableSeq(1)
```

With only later errors being raised:

```
>>> s
TypeError: object of type 'int' has no len()
```

Now an error is raised at initiation:

```
>>> s = MutableSeq(1)
TypeError: The sequence data given to a MutableSeq object should be a string or an array (not a Seq object etc)
```


peterjc:
Seems sensible - thanks!

PR #2204, opened 2019-08-06
Support for the Xdna, SnapGene, and GCK formats

gouttegd:
This pull request is the follow-up to the [discussion initiated on the mailing list](https://lists.open-bio.org/pipermail/biopython/2019-July/016647.html) on 30 July.

It adds three new parsers to the Bio.SeqIO framework for binary formats used by some molecular biology programs:

* "xdna" format, used by DNA Strider and Serial Cloner;
* "snapgene" format, used by GSL Biotech's SnapGene;
* "gck" format, used by Textco Biosoftware's Gene Construction Kit.

It also adds write support for the "xdna" format.


Bio/SeqIO/GckIO.py
```diff
@@ -0,0 +1,208 @@
...
+    Similar to a Pascal string but length is encoded on 4 bytes.
+    """
+    length = _read(handle, 4)
+    length = unpack('>I', length)[0]
+    data = _read(handle, length).decode('ASCII')
+    return data
+
+
+def GckIterator(handle):
+    """Parse a GCK file and return a SeqRecord object."""
```
peterjc: Is there only ever one record per file?

Bio/SeqIO/SnapGeneIO.py
```diff
@@ -0,0 +1,255 @@
...
+    n = 0
+
+    for type, length, data in _PacketIterator(handle):
+        if n == 0 and type != 0x09:
+            raise ValueError("The file does not start with a SnapGene cookie packet")
+
+        if type in _packet_handlers:
+            _packet_handlers[type](length, data, record)
+
+        n = n + 1
```
peterjc: You could use ``n += 1`` here, but perhaps this is more elegant:

```python
for n, (type, length, data) in enumerate(_PacketIterator(handle)):
```

Bio/SeqIO/SnapGeneIO.py
```diff
@@ -0,0 +1,255 @@
...
+def SnapGeneIterator(handle):
+    """Parse a SnapGene file and return a SeqRecord object."""
+    record = SeqRecord(None)
+    n = 0
+
+    for type, length, data in _PacketIterator(handle):
+        if n == 0 and type != 0x09:
+            raise ValueError("The file does not start with a SnapGene cookie packet")
+
+        if type in _packet_handlers:
```
peterjc: Are unknown packets comment or rare? If rare, does a ``BiopythonParserWarning`` seem appropriate?

Bio/SeqIO/GckIO.py
gouttegd: In all the GCK, SnapGenes, or Xdna files that I have ever seen, yes. Those formats seem strictly mono-sequence.

Bio/SeqIO/XdnaIO.py
```diff
@@ -0,0 +1,296 @@
...
+        record as the Xdna format is a mono-record format.
+        """
+        if not records:
+            raise ValueError("Must have one sequence")
+        if len(records) > 1:
+            raise ValueError("More than one sequence found")
+
+        record = records[0]
+
+        alptype = type(record.seq.alphabet)
```
peterjc: This ought to work, but the alphabet system is too complicated for its own good. See this example for how it has to be done:

https://github.com/biopython/biopython/blob/biopython-174/Bio/SeqIO/SeqXmlIO.py#L348

This is one of the reasons we're looking at a massive simplification - see #2046.

Bio/SeqIO/XdnaIO.py
```diff
@@ -0,0 +1,296 @@
...
+        self._write_pstring('0')            # left-side overhand
+
+        # Write features
+        # We must skip features with fuzzy locations as they cannot be
+        # represented in the Xdna format
+        features = [f for f in record.features if type(f.location.start) == ExactPosition and type(f.location.end) == ExactPosition]
+        # We also cannot store more than 255 features as the number of
+        # features is stored on a single byte...
+        if len(features) > 255:
+            features = features[:255]
```
peterjc: I think that data loss deserves a warning (as done in some of the other writers)

Bio/SeqIO/SnapGeneIO.py
gouttegd: No, they are a common occurrence. In fact all the SnapGene files I have contain at least two or three unknown packets. Therefore I don’t think a warning would be appropriate here.

NEWS.rst
```diff
@@ -23,10 +23,16 @@ making an error, but will require user code to be updated accordingly.
 ``Bio.PDB`` has been updated to support parsing REMARK 99 header entries from
 PDB-style Astral files.
 
+``Bio.SeqIO`` now supports reading and writing files in the native format of
+Christian Marck's DNA Strider program ("xdna" format, also used by Serial
+Cloner), as well as reading files in the native formats of GSL Biotech's
+SnapGene and Textco Biosoftware's Gene Construction Kit.
```
peterjc: Probably helpful to spell out all the new format names we're using here?

Tests/test_SeqIO_Gck.py
```diff
@@ -0,0 +1,339 @@
...
+
+# The Gck parser has been developed using the files of the
+# Drosophila Gateway Vector Collection from Carnegie Science
+# (<https://emb.carnegiescience.edu/drosophila-gateway-vector-collection>)
+# as sample Gck files. We cannot redistribute those files along with
+# Biopython, therefore we skip the test case by default.
+#
+# To test the Gck parser, download the GCK.zip archive from the URL
+# above and place the files it contains under a 'Tests/Gck' directory.
+@unittest.skipUnless(os.path.exists('Gck'), 'Gck sample files not available')
```
peterjc: That's an elegant solution.

Bio/SeqIO/GckIO.py
peterjc: Could you put a comment about that in the docstring then (as done for one of the other new formats). Thanks

Bio/SeqIO/SnapGeneIO.py
peterjc: Agreed.

peterjc:
Given the size of the changes, I'm impressed how few things I noticed on reading the diff. Nice work :)

Also the fact you've been using the standalone version of some of this code for quite some time helps give confidence in it's ability to handle real world data.

gouttegd:
Thanks @peterjc for the review. I agree with all your comments and suggestions above and will update my branch accordingly.

Tests/test_SeqIO_Gck.py
gouttegd: Actually maybe we could *automatically* download the archive during the setup phase of the test, if it is not already there? Unless the `--offline` flag has been set of course.

I am testing that right now and it seems to work fine. I can push an update to my branch, but before I do that:

* Do you think it's a good idea?
* If we do download the archive when the tests start, should we remove it when the tests are over? I am inclined to keep it (otherwise it would be downloaded again everytime the tests are run), but then we have to make sure it never finds its way into the repository or a release tarball (easily done by adding an entry to `.gitignore` and `MANIFEST.in`).

peterjc: How big is the download?

If we do this automatically in the test suite, I would also cache the file.

Even if we don't automatically download it, we should add it to the ``.gitignore`` as as a precaution exclude it in ``MANIFEST.in`` as well.

gouttegd: The archive is 439 kb.

I agree about the need to prevent the inclusion of any of those files in any case, independently of whether they have been downloaded manually or automatically.

peterjc:
Looks good to me - @bow could you give a quick second opinion?

peterjc:
I'll just rebase this to fix the merge conflict in the NEWS file...

peterjc:
Rather than holding this up any longer, I've merged it. We can make tweaks as needed.

Thank you @gouttegd 👍 

Do you want to update the table on https://biopython.org/wiki/SeqIO too? It is just a markdown file - edit on github link top right.

gouttegd:
> Rather than holding this up any longer, I've merged it. We can make tweaks as needed.
> 
> Thank you @gouttegd 👍

Thank you for your reviews of the pull request and for merging it.

> Do you want to update the table on https://biopython.org/wiki/SeqIO too? It is just a markdown file - edit on github link top right.

I've just done that, the request is ready to be merged.

Issue #2179, opened 2019-07-22
NCBIXML does not set "bits" in 1.74


SchwarzMarek:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)

3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
CPython
Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
1.74
```

### Expected behaviour

The `bits` attribute in the HSP object having a value (float).

### Actual behaviour

The `bits`attribute is set to `None`.

### Steps to reproduce

Obtain XML file from `biopython/Tests/Blast/xml_2212L_blastx_001.xml`.

Run attached code:
```
import sys
import Bio
from Bio.Blast import NCBIXML

print('Biopython version {}'.format(Bio.__version__))

with open('xml_2212L_blastx_001.xml', 'r') as f:
    for out in NCBIXML.parse(f):
        for hit in out.alignments:
            for hsp in hit.hsps:
                print(hsp.bits)
                sys.exit(0)
```

For biopython 1.74 output is `None`, while for 1.72 and 1.73, there is `247.284`.

peterjc:
From the timing, this is most likely a side effect of adding BLAST XML v2 support in #1997, could you take a quick look @rtf-const?

peterjc:
Fixed by #2180

alffdtu:
This bug does not seem fixed in the Biopython that is installed with pip. I get the same error with the biopython 1.74 upgraded with pip

peterjc:
@alffdtu Correct, this bug was reported against Biopython 1.74.

The fix ce2a57d1b9ec53ec5814d1de83bf89438bdfe747 was merged on 23 July, so will be included in Biopython 1.75.

PR #2208, opened 2019-08-12
Fix TypeError raised by str(HSP())

ecederstrand:
`TypeError` is raised if you try to print an uninitialized `HSP` instance:

```python
>>> from Bio.Blast.Record import HSP
>>> str(HSP())
  File ".../site-packages/Bio/Blast/Record.py", line 227, in __str__
    % (self.score, self.bits, self.expect, self.align_length)]
TypeError: %i format: a number is required, not NoneType
```

### DIFF
--- a/Bio/Blast/Record.py
+++ b/Bio/Blast/Record.py
@@ -26,6 +26,13 @@ from Bio.SeqRecord import SeqRecord
 from Bio.Align import MultipleSeqAlignment
 
 
+def fmt_(value, format_spec="%s", default_str="<unknown>"):
+    """Ensure the given value formats to a string correctly"""
+    if value is None:
+        return default_str
+    return format_spec % value
+
+
 class Header(object):
     """Saves information from a blast header.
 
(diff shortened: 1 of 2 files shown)


peterjc:
The idea seems sound, but I feel there ought to be a more elegant solution to this... but other than a helper function for int or None to string (and float or None to string) which isn't ideal either, I'm not immediately seeing it. @bow?

Does defaulting to zero rather than None have any merit?

Cross reference #2179

ecederstrand:
I didn't find any obvious workarounds in the stdlib (except defaulting to `%s`). And everything mentioned in https://stackoverflow.com/questions/20248355/how-to-get-python-to-gracefully-format-none-and-non-existing-fields seemed to exceedingly complicate a simple class like this.

You could of course format each argument separately:

```python
# Instead of
'%i %i %i' % (maybe_none_1, maybe_none_2, maybe_none_3)
# Do this
'%s %s %s' % (
    str(None) if maybe_none_1 is None else '%i' % maybe_none_1, 
    str(None) if maybe_none_2 is None else '%i' % maybe_none_2, 
    str(None) if maybe_none_3 is None else '%i' % maybe_none_3, 
)
```

but I'm not sure it's less ugly.

peterjc:
I prefer this idea, but will leave it to @bow as module owner to judge.

bow:
@ecederstrand, thanks for opening the PR.

With regards to formatting optional values, having them formatted separately gives us more flexibility actually. The repeating `if` expressions can be made into a helper function, which should make the code less-ugly.

Something like:
```
def fmt_(value, default_spec="%s", default_str="<unknown>"):
    """Ensure the given value formats to a string correctly"""
    if value is None:
        return default_str
    return default_spec % value

...

"Score %s (%s bits), expectation %0.1e, alignment length %i" % (
    fmt_(v1), fmt_(v2), fmt_(v3, "%0.1e"), fmt_(v4)
)
```

Does this look sensible?

ecederstrand:
I don't have strong opinions on the matter.I just want our debug statements to not blow up on empty records :-)

bow:
That is a good concern :).

Anyway, the updated `__str__` looks better now ~ just waiting for all the tests to pass.

peterjc:
Looks good - any objection to a squash-and-merge?

ecederstrand:
Go ahead :-)

peterjc:
Thanks. I'm taking Bow's earlier comment as approval.

PR #2209, opened 2019-08-14
Fix #2207 - make feature position instances pickable

### LINKED ISSUE #2207: SeqFeature.WithinPosition instances fail to unpickle
### Setup

system as follows:

3.6.6 (default, Jan 26 2019, 16:53:05) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
CPython
Linux-3.10.0-693.21.1.el7.x86_64-x86_64-with-centos-7.2.1511-Core
1.74

### Expected behaviour

Instances of WithinPosition should pickle and unpickle

### Actual behaviour

Unpickling a WithinPosition instance fails with:

TypeError                                 Traceback (most recent call last)
<ipython-input-8-7e6a36b7b135> in <module>
      4 wp = SeqFeature.WithinPosition(10,10,13)
      5 p = pickle.dumps(wp)
----> 6 unpickled_wp = pickle.loads(p)

TypeError: \_\_new\_\_() missing 2 required positional arguments: 'left' and 'right'

### Steps to reproduce

I ran into this unpickling an object containing a SeqRecord derived from UniProt acc P08514, but some digging narrowed it down to:
 
import pickle
from Bio import SeqFeature

wp = SeqFeature.WithinPosition(10,10,13)
p = pickle.dumps(wp)
unpickled_wp = pickle.loads(p)

Pickle expects \_\_getnewargs\_\_() or \_\_getnewargs\_ex\_\_() to be defined for instances of classes that alter the call signature of \_\_new\_\_(cls). I assume WithinPosition inherits \_\_getnewargs\_\_() from int, but then overrides the \_\_new\_\_ signature so the two are not compatible. It looks as if several other *Position classes in this module will fail in the same way

macraild:
This adds \_\_getnewargs\_\_ to SeqFeatures.*Positions classes where required, so that instances are (un)picklable

This pull request addresses issue #2207 

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -1695,22 +1695,6 @@ class WithinPosition(int, AbstractPosition):
     >>> p2 == 13
     True
 
-
-    To allow pickling and unpickling of class instances, the arguments
-    to __new__ must be defined by __getnewargs__:
-
-    >>> p.__getnewargs__() == (10, 10, 13)
-    True
-
-    >>> import pickle
-    >>> p3 = pickle.loads(pickle.dumps(p))
-    >>> p3 == p
-    True
-    >>> p3._left == p._left
-    True
-    >>> p3._right == p._right
-    True
-
     """
 
     def __new__(cls, position, left, right):
@@ -1723,8 +1707,11 @@ class WithinPosition(int, AbstractPosition):
         obj._right = right
         return obj
 
-    # Must define this to allow instances to be unpickled
     def __getnewargs__(self):
+        """Return the arguments accepted by __new__.
+        
+        Necessary to allow pickling and unpickling of class instances.
+        """
         return (int(self), self._left, self._right)
 
     def __repr__(self):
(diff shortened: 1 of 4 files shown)


Bio/SeqFeature.py
```diff
@@ -1695,6 +1695,22 @@ class WithinPosition(int, AbstractPosition):
     >>> p2 == 13
     True
 
+
```
peterjc: Could you move this to the ``__getnewargs__`` docstring instead?

Bio/SeqFeature.py
```diff
@@ -1804,6 +1824,22 @@ class BetweenPosition(int, AbstractPosition):
 
     i.e. For equality (and sorting) the position objects behave like
     integers.
+
```
peterjc: Again, could you move this to the ``__getnewargs__`` docstring please.

Bio/SeqFeature.py
```diff
@@ -2039,6 +2079,20 @@ class OneOfPosition(int, AbstractPosition):
     >>> p2 == 1901
     True
 
+    To allow pickling and unpickling of class instances, the arguments
```
peterjc: Same comment, this isn't of general interest but ``__getnewargs__`` lacks any docstring at all - this would work better there I think.

peterjc:
I'd like a cosmetic change in the docstring placement, but otherwise the change looks good.

Are you comfortable adding something to ``Tests/test_SeqFeature.py`` or do you think the doctests are enough?

macraild:
I agree, the docstrings were out of place, so I've rewritten as method-level docstrings. But doctests don't seem to run from here, so I've converted the original doctests  to unittests in Test/SeqFeature.py. Note that the scope of the tests is unchanged. To the extent I understand what is required here, these tests cover it, but happy to add more if there is something specific I have missed.

peterjc:
Doctests are fragile, so having explicit unittests seems better all round. Are you OK with a squash-merge? Do you want your name in the ``NEWS.rst`` and ``CONTRIB.rst``?

macraild:
Yes, squash-merge is fine. I'm happy to have my name in NEWS and CONTRIB if you think it is warranted.

peterjc:
Thank you, applied.

PR #1192, opened 2017-04-25
rule D413 corrections

morrme:
Ref. #1170

PR #2230, opened 2019-08-26
Remove pinning of pydocstyle

MarkusPiotrowski:
With the recent update of `flake8-docstring`, pinning of `pydocstyle` to versions < 4.0.0 should no longer be necessary (https://gitlab.com/pycqa/flake8-docstrings/issues/36).

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -108,10 +108,6 @@ deps =
     # https://bugs.launchpad.net/doc8/+bug/1837515
     # https://sourceforge.net/p/docutils/bugs/366/
     docutils==0.14
-    # flake8-docstrings uses a function removed in pydocstyle 4.0.0; once a fix
-    # is released in flake8-docstrings we can remove the following constraint:
-    pydocstyle<4.0.0
-    # See https://gitlab.com/pycqa/flake8-docstrings/issues/36
 commands =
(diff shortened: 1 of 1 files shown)


MarkusPiotrowski:
There are now numerous D413 (missing blank line after last section) and some D412 (no blank lines allowed between a section header and its content) issues now. Need to investigate.
A quick inspection let suggest that D413 is reported if the docstring ends with single line 'section'.

peterjc:
D413 is strange, we fixed them back on #1192 - but evidentally pydocstyle have tweaked the rules.

MarkusPiotrowski:
Reported as issue at https://github.com/PyCQA/pydocstyle/issues/401#issue-485979204. 

If we update to `pydocstyle` version 4.0.0 we need to (temporarily) ignore D413 and D412, either until `pydocstyle` changes it rules or until we have solved all these 'issues'.

peterjc:
I think sticking with the older pydocstyle makes sense, pending some comments on that bug you kindly filed.

peterjc:
It seems the D413 false positives are fixed in pydocstyle v5 released earlier this month.

We could remove the pinning (as in the original change on #2230), but we now know that v4 does not work for us. Therefore #2414 changes this to require at least v5.

peterjc:
Superseded by #2414

PR #2233, opened 2019-08-28
Added support for check_same_thread

pkerpedjiev:
This pull request addresses issue https://github.com/biopython/biopython/issues/2234

### DIFF
--- a/Tests/test_SeqIO_index.py
+++ b/Tests/test_SeqIO_index.py
@@ -112,24 +112,9 @@ if sqlite3:
             self.assertEqual(54, len(d))
             self.assertRaises(FileNotFoundError, d.get_raw, "alpha")
 
-        def test_old_no_check_same_thread(self):
-            """Using check_same_thread=True (the default) raises an exception
-            when index_db is accessed from a thread"""
-            d = SeqIO.index_db("Roche/triple_sff_rel_paths.idx")
-            self.assertEqual(d._check_same_thread, True)
-
-            def reader_thread():
-                self.assertRaises(sqlite3.ProgrammingError, d.__getitem__, 'alpha')
-
-            reader = threading.Thread(target=reader_thread)
-            reader.start()
-            reader.join()
-
         def test_old_check_same_thread(self):
             """Setting check_same_thread to False doesn't raise an exception"""
-            d = SeqIO.index_db("Roche/triple_sff_rel_paths.idx",
-                               check_same_thread=False)
-            self.assertEqual(d._check_same_thread, False)
+            d = SeqIO.index_db("Roche/triple_sff_rel_paths.idx")
 
             def reader_thread():
                 try:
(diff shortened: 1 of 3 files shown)


peterjc:
Is there a downside to doing ``check_same_thread=False`` as the default? i.e. Don't change the ``Bio.SeqIO`` or ``Bio.SearchIO`` (etc) interface at all.

CC @bow

pkerpedjiev:
Afaict, it just puts the onus on the user to make sure write operations don't conflict with each other. In the case of sequence indexes that shouldn't be an issue since the index creation can't be done in a multithreaded fashion. 

> By default, check_same_thread is True and only the creating thread may use the connection. If set False, the returned connection may be shared across multiple threads. When using multiple threads with the same connection writing operations should be serialized by the user to avoid data corruption.

If you're ok with making check_same_thread=false the default then I'm happy to modify the PR.

peterjc:
For our read only interface in ``Bio/File.py`` class ``_SQLiteManySeqFilesDict``, I am inclined to just change the internal connection setup:

```python
        con = _sqlite.connect(index_filename,
                              check_same_thread=self._check_same_thread)
```

i.e. Don't change our user facing API at all.

Does that seem sensible?

pkerpedjiev:
> Does that seem sensible?

Yes. I'll modify the PR.

peterjc:
You probably understood what I meant, but I think I should have written:

```python
con = _sqlite.connect(index_filename, check_same_thread=False)
```

pkerpedjiev-zymergen:
Yup, I figured you just took the line from the PR to point where you meant for that to go.

Tests/test_SeqIO_index.py
```diff
@@ -111,6 +112,20 @@ def test_old(self):
             self.assertEqual(54, len(d))
             self.assertRaises(FileNotFoundError, d.get_raw, "alpha")
 
+        def test_old_check_same_thread(self):
+            """Setting check_same_thread to False doesn't raise an exception"""
```
peterjc: Needs a trailing full stop,

```
$ flake8
./Tests/test_SeqIO_index.py:116:1: D400 First line should end with a period
```

peterjc: Strange - you're already in the top level Biopython folder so https://gitlab.com/pycqa/flake8/issues/517 shouldn't be the problem. Maybe the version of flake8 is important...

peterjc: As to the task at hand, we currenlty tell flake8 to ignore the following under ``Tests/``:

F401,F841,D101,D102,D103,B009,B010,B011,C812

That is on top of the things we ignore everywhere like E501 - hopefully the ``.flake8`` file comments are clear.

i.e. D400 is the only one you should be seeing here.

peterjc:
Any comments @bow?

peterjc:
See also fc8ac0ea0048e152e409d229fa85aee205c95c01 as I missed the use of single quotes when merging this (wasn't enforced when the PR was tested).

Issue #2247, opened 2019-09-01
Problem with compound locations wrapping origin in SeqIO


mmcguffi:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

3.7.4 (default, Jul  9 2019, 18:13:23) 
[Clang 10.0.1 (clang-1001.0.46.4)]
CPython
Darwin-18.2.0-x86_64-i386-64bit
1.74

### Expected behaviour
When opening a gbk file that has a feature which contains a(n incorrectly) specified location which wraps the origin, but also initially contains a compound location, this compound location should be "repaired" and also throw a warning.

### Actual behaviour

When trying to open a .gbk that has this issue, this error is thrown:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-4e2d1f4e6ad8> in <module>()
     12         file.write("%s\n" % line)
     13 with open(tmp.name,'r') as file_handle:
---> 14     record_dict = SeqIO.to_dict(SeqIO.parse(file_handle, 'gb'))
     15 tmp.close()

/usr/local/lib/python3.7/site-packages/Bio/SeqIO/__init__.py in to_dict(sequences, key_function)
    808 
    809     d = _dict()
--> 810     for record in sequences:
    811         key = key_function(record)
    812         if key in d:

/usr/local/lib/python3.7/site-packages/Bio/SeqIO/__init__.py in parse(handle, format, alphabet)
    659             raise ValueError("Unknown format '%s'" % format)
    660         # This imposes some overhead... wait until we drop Python 2.4 to fix it
--> 661         for r in i:
    662             yield r
    663 

/usr/local/lib/python3.7/site-packages/Bio/GenBank/Scanner.py in parse_records(self, handle, do_features)
    491         # This is a generator function
    492         while True:
--> 493             record = self.parse(handle, do_features)
    494             if record is None:
    495                 break

/usr/local/lib/python3.7/site-packages/Bio/GenBank/Scanner.py in parse(self, handle, do_features)
    475                                     feature_cleaner=FeatureValueCleaner())
    476 
--> 477         if self.feed(handle, consumer, do_features):
    478             return consumer.data
    479         else:

/usr/local/lib/python3.7/site-packages/Bio/GenBank/Scanner.py in feed(self, handle, consumer, do_features)
    447         # Features (common to both EMBL and GenBank):
    448         if do_features:
[...]

PR #2254, opened 2019-09-03
Origin wrap bad formatting of compound locations

nimne:
This pull request addresses issue #2247 

Last time I fixed this, I didn't consider genes with introns (compound locations) spanning the origin. It looks like most current tools don't generate files that are incompatible with the specification, but there are some old files floating around.

Because of the way this was fixed last time, it was simple to add the fix to other location string types. This is beginning to feel a little hack-y, but it is consistent with previous changes. 

I think I said this last time, but I don't think there are other types of location strings that won't be fixed.

I modified one of the test files to capture the problem raised in the issue.

### DIFF
--- a/Bio/GenBank/__init__.py
+++ b/Bio/GenBank/__init__.py
@@ -1122,6 +1122,11 @@ class _FeatureConsumer(_BaseGenBankConsumer):
                                                            strand))
                 except ValueError:
                     # Could be non-integers, more likely bad origin wrapping
+
+                    # In the case of bad origin wrapping, _loc will return
+                    # a CompoundLocation. CompoundLocation.parts returns a
+                    # list of the FeatureLocation objects inside the
+                    # CompoundLocation.
                     locs.extend(_loc(part,
                                      self._expected_size,
                                      strand,
@@ -1164,6 +1169,10 @@ class _FeatureConsumer(_BaseGenBankConsumer):
                 else:
                     part_strand = strand
                 try:
+                    # There is likely a problem with origin wrapping.
+                    # Using _loc to return a CompoundLocation of the
+                    # wrapped feature and returning the two FeatureLocation
+                    # objects to extend to the list of feature locations.
                     loc = _loc(part, self._expected_size, part_strand,
                                seq_type=self._seq_type.lower()).parts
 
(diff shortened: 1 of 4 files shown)


Bio/GenBank/__init__.py
```diff
@@ -1155,12 +1164,14 @@ def location(self, content):
...
                 try:
-                    loc = _loc(part, self._expected_size, part_strand)
+                    loc = _loc(part, self._expected_size, part_strand,
+                               seq_type=self._seq_type.lower()).parts
+
                 except ValueError as err:
                     print(location_line)
                     print(part)
                     raise err
-                locs.append(loc)
```
peterjc: Why append vs extend? The ``_loc(...)`` function might return either a ``FeatureLocation`` or a ``CompoundLocation``,  yet we want ``locs`` to be a list of ``FeatureLocation`` objects to wrap up as a ``CompoundLocation``? I may be missing something as I've not run the new code yet,

nimne: There was a small issue with the way this worked. When the improperly wrapped origin was split into two FeatureLocations, it returned a CompoundLocation containing two FeatureLocations. This could result in passing a CompoundLocation to a CompoundLocation object, causing a problem. Essentially it was creating this:

`CompoundLocation(FeatureLocation(x,y,z), CompoundLocation([FeatureLocation(x,y,z), FeatureLocation(x,y,z)]))`

I've used `extend()` for the location strings that were processed in a loop, and collected the FeatureLocation objects in a list that would ultimately go to a new `CompoundLocation` object. You'll note that I've grabbed a list of the FeatureLocation objects returned from `_loc()` using `.parts`. `extend()` is used to iterate over the list of two `FeatureLocation` objects returned when the origin wrapped string is split in two. I'm happy to commend the code to make this more clear.

I realize my explanation isn't very straight forward. Let me know if this doesn't make sense!

peterjc:
An extra comment or two might help (your explanation makes sense from my memory of the code, replying to this by phone)

Bio/GenBank/__init__.py
peterjc: It was the ``.parts`` that I missed, the updated comments do help I think.

peterjc:
Looks good - do you want to add your name to the next release in ``NEWS.rst``?

nimne:
Thanks for the reminder :)

peterjc:
Applied, thank you.

Issue #1838, opened 2018-10-30
Review empty file behaviour for SeqIO parsers


peterjc:
In some sequence formats (e.g. FASTA, FASTQ, GenBank, EMBL, Swiss) an empty file is a legitimate corner case meaning no sequences. In order formats (e.g. SFF, XML) an empty file is invalid as a minimal header is expected.

This issue is to review all the SeqIO file formats and their behaviour when asked to parse an empty file - and document on this issue what they currently do.

They should either return no sequences, or raise a clear exception indicating an invalid empty file. We may want to standardise the wording of the exception and ensure this is always the same exception class, e.g. ``ValueError``.

See discussion on #1835 which touches on this, and white listing some of the parsers in the tests. Those test would ideally validate the exception follows any agreed convention.

DevangThakkar:
I've tried to document the results that are printed out when SeqIO encounters an empty file in each of the formats mentioned in the [documentation](https://biopython.org/wiki/SeqIO). I've uploaded the test script and the results [here](https://github.com/DevangThakkar/Misc_scripts). 

Most formats do not print anything (25/31). cif-atom and cif-seqres give `Unknown format 'cif-atom'` and `Unknown format 'cif-seqres'` errors. pdb-atom raises an exception the message for which is simply 0. sff and sff-trim raise more meaningful `Empty file` errors. Finally, uniport-xml raises a `no element found: line 1, column 0` error. 

This is my first attempt at contributing to Biopython, so if you could help me get started with what comes next, that'll be really great.

peterjc:
It sounds like a pull request for cif-atom and cif-seqres, and/or a pull request for uniprot-xml, would be a good idea. Rather than the current message I think these should return ``ValueError("Empty file")`` like SFF?

For pdb-atom, I'm not sure if return no records is sensible, or again, the Empty File message is better.

vinisalazar:
@peterjc is there any work left to do on this?

peterjc:
Current results, using this ``empty.py`` script:

```python
from Bio import SeqIO

for format in SeqIO._FormatToIterator:
    try:
        records = list(SeqIO.parse("empty.ext", format))
    except Exception as err:
        print("%s: %s" % (format, err))
    else:
        print("%s: %i records" % (format, len(records)))
```

Gives:

```
$ touch empty.ext
$ python empty.py
fasta: 0 records
fasta-2line: 0 records
gb: 0 records
genbank: 0 records
genbank-cds: 0 records
[...]

PR #2255, opened 2019-09-04
Consistent error messages for empty files in SeqIO (PdbSeqres, CifSeqres, Abi, AbiTrim)

DevangThakkar:
This pull request addresses issue #1838 

### DIFF
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -372,7 +372,8 @@ def AbiIterator(handle, alphabet=None, trim=False):
     marker = handle.read(4)
     if not marker:
         # handle empty file gracefully
-        return
+        raise ValueError("Empty file.")
+
     if marker != b"ABIF":
         raise IOError("File should start ABIF, not %r" % marker)
 
--- a/Bio/SeqIO/GckIO.py
+++ b/Bio/SeqIO/GckIO.py
@@ -14,6 +14,7 @@ from Textco BioSoftware, Inc.
 from struct import unpack
 
 from Bio import Alphabet
+from Bio._utils import _read_header
 from Bio.Seq import Seq
 from Bio.SeqFeature import SeqFeature, FeatureLocation
 from Bio.SeqRecord import SeqRecord
@@ -75,7 +76,8 @@ def GckIterator(handle):
     # GCK files start with a 24-bytes header. Bytes 4 and 8 seem to
     # always be 12, maybe this could act as a magic cookie. Bytes
     # 17-20 and 21-24 contain variable values of unknown meaning.
-    _read(handle, 24)
+    # check if file is empty
+    _read_header(handle, 24)
 
     # Read the actual sequence data
     packet, length = _read_packet(handle)
--- a/Bio/SeqIO/NibIO.py
+++ b/Bio/SeqIO/NibIO.py
@@ -117,6 +117,11 @@ def NibIterator(handle, alphabet=None):
     if alphabet is not None:
         raise ValueError("Alphabets are ignored.")
     word = handle.read(4)
+
+    # check if file is empty
+    if not word:
+        raise ValueError("Empty file.")
+
     signature = bytes2hex(word)
     if signature == "3a3de96b":
         byteorder = "little"  # little-endian
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -139,9 +139,13 @@ def PdbSeqresIterator(handle):
     # Late-binding import to avoid circular dependency on SeqIO in Bio.SeqUtils
     from Bio.SeqUtils import seq1
 
+    # raise exception if file is empty
+    empty = True
+
     chains = collections.defaultdict(list)
     metadata = collections.defaultdict(list)
     for line in handle:
+        empty = False
         rec_name = line[0:6].strip()
         if rec_name == "SEQRES":
             # NB: We only actually need chain ID and the residues here;
@@ -197,6 +201,9 @@ def PdbSeqresIterator(handle):
                 }
             )
         # ENH: 'SEQADV' 'MODRES'
+
+    if empty:
+        raise ValueError("Empty file.")
 
     for chn_id, residues in sorted(chains.items()):
         record = SeqRecord(Seq("".join(residues), generic_protein))
@@ -270,17 +277,16 @@ def PdbAtomIterator(handle):
     # Only import PDB when needed, to avoid/delay NumPy dependency in SeqIO
     from Bio.PDB import PDBParser
 
-    # Deduce the PDB ID from the PDB header
-    # ENH: or filename?
     from Bio.File import UndoHandle
 
     undo_handle = UndoHandle(handle)
     firstline = undo_handle.peekline()
-
     # check if file is empty
-    if firstline == "":
+    if not firstline:
         raise ValueError("Empty file.")
 
+    # Deduce the PDB ID from the PDB header
+    # ENH: or filename?
     if firstline.startswith("HEADER"):
         pdb_id = firstline[62:66]
     else:
@@ -367,9 +373,16 @@ def CifSeqresIterator(handle):
     # Only import PDB when needed, to avoid/delay NumPy dependency in SeqIO
     from Bio.PDB.MMCIF2Dict import MMCIF2Dict
 
+    from Bio.File import UndoHandle
+    undo_handle = UndoHandle(handle)
+
+    # check if file is empty
+    if not undo_handle.peekline():
+        raise ValueError("Empty file.")
+
     chains = collections.defaultdict(list)
     metadata = collections.defaultdict(list)
-    records = MMCIF2Dict(handle)
+    records = MMCIF2Dict(undo_handle)
 
     # Explicitly convert records to list (See #1533).
     # If an item is not present, use an empty list
@@ -500,7 +513,7 @@ def CifAtomIterator(handle):
     shutil.copyfileobj(handle, buffer)
 
     # check if file is empty
-    if len(buffer.getvalue()) == 0:
+    if not buffer.getvalue():
         raise ValueError("Empty file.")
 
     buffer.seek(0)
--- a/Bio/SeqIO/SeqXmlIO.py
+++ b/Bio/SeqIO/SeqXmlIO.py
@@ -57,7 +57,12 @@ class XMLRecordIterator(object):
         """Iterate over the records in the XML file."""
         record = None
         try:
+            empty = True
             for event, node in self._events:
+
+                # the for loop is entered only if there is some content in self._events
+                # if the empty flag is still True, the file is empty
+                empty = False
 
                 if event == "START_ELEMENT" and node.namespaceURI == self._namespace:
 
@@ -85,6 +90,9 @@ class XMLRecordIterator(object):
                     and node.localName == self._recordTag
                 ):
                     yield record
+
+            if empty:
+                raise ValueError("Empty file.")
 
         except SAXParseException as e:
 
--- a/Bio/SeqIO/XdnaIO.py
+++ b/Bio/SeqIO/XdnaIO.py
@@ -16,6 +16,7 @@ from struct import pack, unpack
 import warnings
 
 from Bio import Alphabet, BiopythonWarning
+from Bio._utils import _read_header
 from Bio.Seq import Seq
 from Bio.SeqIO.Interfaces import SequenceWriter
 from Bio.SeqFeature import SeqFeature, FeatureLocation, ExactPosition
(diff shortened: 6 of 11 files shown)


Bio/SeqIO/AbiIO.py
```diff
@@ -353,6 +353,10 @@ def _get_string_tag(opt_bytes_value, default=None):
 
 def AbiIterator(handle, alphabet=None, trim=False):
     """Return an iterator for the Abi file format."""
+    # raise exception if file is empty
+    if not handle.readline():
+        raise ValueError("Empty file.")
+
```
peterjc: Can you move the raise to line below after the comment "handle empty file gracefully" please?

Bio/SeqIO/PdbIO.py
```diff
@@ -134,6 +134,10 @@ def PdbSeqresIterator(handle):
     # Late-binding import to avoid circular dependency on SeqIO in Bio.SeqUtils
     from Bio.SeqUtils import seq1
 
+    # raise exception if file is empty
+    if not handle.readline():
```
peterjc: This will read a line an discard it - so I would expect it to break the parser

Bio/SeqIO/PdbIO.py
```diff
@@ -183,6 +187,8 @@ def PdbSeqresIterator(handle):
             # db_icode_end = line[67]
             metadata[chn_id].append({"pdb_id": pdb_id, "database": database,
                                     "db_acc": db_acc, "db_id_code": db_id_code})
+        # elif rec_name == "":
+        #     raise ValueError("Empty file.")
```
peterjc: You should remove these new comment lines

Bio/SeqIO/PdbIO.py
```diff
@@ -346,6 +352,10 @@ def CifSeqresIterator(handle):
     metadata = collections.defaultdict(list)
     records = MMCIF2Dict(handle)
 
+    # raise exception if file is empty
+    if len(records) == 0:
+        raise ValueError("Empty file.")
```
peterjc: Is this the only way to get an empty dictionary back? Should the exception happen inside MMCIF2Dict instead?

Tests/test_SeqIO.py
```diff
@@ -4422,7 +4422,7 @@ def test_empty_file(self):
         """Check parsers can cope with an empty file."""
         for t_format in SeqIO._FormatToIterator:
             if t_format in SeqIO._BinaryFormats or \
-               t_format in ("uniprot-xml", "pdb-seqres", "pdb-atom", "cif-atom"):
+               t_format in ("uniprot-xml", "pdb-seqres", "pdb-atom", "cif-atom", "cif-seqres"):
```
peterjc: Rather than a simple continue, the test ought to be for the expected exception.

peterjc:
CodeCov has highlighted only 50% of the changes got tested - see my comment in the test framework about checking for the exception (use assertRaises) rather than skipping these formats.

Bio/SeqIO/PdbIO.py
DevangThakkar: Can we use UndoHandle(handle).peekline() here instead?

Tests/test_SeqIO.py
DevangThakkar: True, that makes sense. I see that there are more formats in SeqIO._BinaryFormats ("gck", "seqxml", "snapgene", "nib", "xdna") that currently do not have a consistent error message for empty files. Do we make them raise the same `ValueError("Empty file")` too?

Bio/SeqIO/PdbIO.py
DevangThakkar: I'm now using the buffer logic used in `CifAtomIterator()` to check for emptiness.

```
buffer = StringIO()
shutil.copyfileobj(handle, buffer)

# check if file is empty
if len(buffer.getvalue()) == 0:
```

Bio/SeqIO/PdbIO.py
peterjc: As a last resort - the fact this parser is a big for loop makes other approaches more complicated.

Tests/test_SeqIO.py
peterjc: That would expected for #1838, yes.

Bio/SeqIO/PdbIO.py
peterjc: That does not look elegant just to catch an empty handle. There was good reason for CIF to use a buffer as internally it calls two parsers.

Bio/SeqIO/PdbIO.py
DevangThakkar: One other way to do it is by making use of the fact that if the handle is empty, the loop `for line in handle:` would not be entered into. We can have something of the sort:
```
flag = False
for line in handle:
    flag = True
    ...

if not flag:
```

I concede it's not the most elegant way to do it, but it's simple and standalone.

peterjc: That'd work. Maybe start with ``empty = True`` and inside the loop set ``empty = False``, but I like this approach more.

DevangThakkar: Great! Thank you.

Bio/SeqIO/PdbIO.py
DevangThakkar: Using `UndoHandle(handle).peekline()` instead.

Bio/SeqIO/AbiIO.py
```diff
@@ -373,6 +373,7 @@ def AbiIterator(handle, alphabet=None, trim=False):
     marker = handle.read(4)
     if not marker:
         # handle empty file gracefully
+        raise ValueError("Empty file.")
```
peterjc: If you raise, you can remove the next line (return).

Bio/SeqIO/GckIO.py
```diff
@@ -71,11 +71,15 @@ def GckIterator(handle):
...
+    # check if file is empty
+    if not handle.read(1):
+        raise ValueError("Empty file.")
+
     # Skip file header
     # GCK files start with a 24-bytes header. Bytes 4 and 8 seem to
     # always be 12, maybe this could act as a magic cookie. Bytes
     # 17-20 and 21-24 contain variable values of unknown meaning.
-    _read(handle, 24)
+    _read(handle, 23)
```
peterjc: Performance is not going to be an issue, but I think doing a single 24 bytes read is better:

```python
temp = handle.read(24)
if not temp:
elif len(temp) < 24:
    raise ValueError("Cannot read 24 bytes from handle")
    # This preserves the current exception, perhaps should be more specific?
    # raise ValueError("Truncated file? 24 byte header incomplete.")
del temp
```

Bio/SeqIO/NibIO.py
```diff
@@ -111,6 +111,11 @@ def NibIterator(handle, alphabet=None):
     if alphabet is not None:
         raise ValueError("Alphabets are ignored.")
     word = handle.read(4)
+
+    # check if file is empty
+    if len(word) == 0:
```
peterjc: Does this work?

```python
if not word:
```

See https://www.python.org/dev/peps/pep-0008/#id51

Bio/SeqIO/PdbIO.py
```diff
@@ -342,9 +349,17 @@ def CifSeqresIterator(handle):
     # Only import PDB when needed, to avoid/delay NumPy dependency in SeqIO
     from Bio.PDB.MMCIF2Dict import MMCIF2Dict
 
+    from Bio.File import UndoHandle
+    undo_handle = UndoHandle(handle)
+    firstline = undo_handle.peekline()
+
+    # check if file is empty
+    if firstline == "":
```
peterjc: We don't need to define ``firstline``, how about:

```python
if not undo_handle.peekline():
```

Bio/SeqIO/SeqXmlIO.py
```diff
@@ -57,8 +57,13 @@ def __iter__(self):
         """Iterate over the records in the XML file."""
         record = None
         try:
+            flag = False
```
peterjc: ``flag`` is too generic - I suggest ``empty`` as you've done in ``Bio/SeqIO/PdbIO.py`` (with the inverted values)

Bio/SeqIO/XdnaIO.py
```diff
@@ -36,6 +36,22 @@
...
+def _read_header(handle, length):
+    """Read the specified number of bytes from the given handle.
+
+    Raise a ValueError("Empty file.") if the length of data read is zero. The
+    reason for having a separate function for the header is it enables raising
+    an empty file error if the length of the data read is zero. This might
+    not always be the case later in the file.
+    """
+    data = handle.read(length)
+    if len(data) == 0:
```
peterjc: Style wise,

```python
if not data:
```

i.e. PEP8 style, see earlier comment - where you could use the same helper. I wonder if this function is worth putting somewhere central like ``Bio/_utils.py``?

Bio/SeqIO/GckIO.py
peterjc: See also the ``_read_header()`` helper function you defined in one of the other files... might be worth sharing that?

Tests/test_SeqIO.py
```diff
@@ -4422,12 +4422,15 @@ def test_empty_file(self):
...
                 # Not allowed empty SFF files.
-                continue
-            handle = StringIO()
-            records = list(SeqIO.parse(handle, t_format))
-            self.assertEqual(len(records), 0)
+                handle = StringIO()
```
peterjc: You shouldn't use ``StringIO`` on binary files, only plain text files. There's ``BytesIO`` for them. However, the test structure looks good.

i.e. If binary, use BytesIO empty test, elif non-empty file type, use StringIO for empty test, else parse file

Bio/SeqIO/XdnaIO.py
DevangThakkar: Yeah, that would make sense. Have added it to Bio/_utils.py!

Tests/test_SeqIO_Gck.py
```diff
@@ -380,6 +380,20 @@ def test_conflicting_lengths(self):
...
 
+class TestGckWithImproperHeader(unittest.TestCase):
+
+    def setUp(self):
+        with open('Gck/improperHeader.gck', 'rb') as f:
+            self.buffer = f.read()
+
+    def test_read(self):
+        """Read a file with an incomplete header."""
+        h = BytesIO(self.buffer)
```
peterjc: I would remove the ``setUp`` here and just do the open directly in this test method.

peterjc: (This also avoids the old fashioned open and close - just use the with statement)

Bio/_utils.py
```diff
@@ -46,6 +46,22 @@ def read_forward(handle):
             return line
 
 
+def read_header(handle, length):
+    """Read the specified number of bytes from the given handle.
```
peterjc: Maybe bytes -> characters?

As written, this should work just fine on both raw handles (bytes) and text handles (unicode strings).

DevangThakkar:
I think we should be good to go!

Tests/Gck/improperHeader.gck
```diff
@@ -0,0 +1 @@
+foo
```
peterjc: This is not a helpful test file - we can remove it, see comment on test.

Tests/test_SeqIO_Gck.py
```diff
@@ -380,6 +380,17 @@ def test_conflicting_lengths(self):
...
+
+    def test_read(self):
+        """Read a file with an incomplete header."""
+        with open('Gck/improperHeader.gck', 'rb') as f:
+            h = BytesIO(f.read())
+            with self.assertRaisesRegexp(ValueError, "Improper header, cannot read 24 bytes from handle"):
+                SeqIO.read(h, 'gck')
+            h.close()
+
+
```
peterjc: There is no reason to use ``BytesIO`` if you have a file on disk, so initially I was going to suggest removing ``BytesIO`` and passing the handle directly to ``SeqIO.read()``.

However, given ``Tests/Gck/improperHeader.gck`` is just a tiny irrelevant file, why not remove it and use ``BytesIO`` as follows:

```python
    def test_read(self):
        """Read a file with an incomplete header."""
        handle = BytesIO(b"tiny")
        with self.assertRaisesRegexp(ValueError, "Improper header, cannot read 24 bytes from handle"):
            SeqIO.read(handle, 'gck')
        handle.close()
```

Bio/SeqIO/GckIO.py
```diff
@@ -14,6 +14,7 @@
 from struct import unpack
 
 from Bio import Alphabet
+from Bio._utils import read_header
```
peterjc: Either:

```python
from Bio._utils import read_header as _read_header
```

Or, rename the function and just:

```python
from Bio._utils import _read_header
```

Otherwise you are adding ``read_header`` as a public part of the ``GckIO`` API.

(Same applies to ``XdnaIO.py``)

peterjc:
Rebased and squashed, dealing with the merge conflict.

peterjc:
Missed a quote change, rebased again.

peterjc:
Thank you!

DevangThakkar:
Thank you for taking the time for all your reviews/suggestions! I'd like it if you could tag me in more SeqIO issues. I don't mind other modules either but I think it might be better to be more involved in one.

PR #2286, opened 2019-09-26
Address future D412 cases.

MarkusPiotrowski:
This pull request is related to PR #2230. Recent releases of `pydocstyle` do recognize more section headers by using additional keywords. This will bring up some new `D412` issues (no blank lines allowed between a section header and its content) and many new `D413` issues (no blank lines allowed between a section header and its content. However, `D413` was just removed from the `pep257` convention that we are using (https://github.com/PyCQA/pydocstyle/pull/404).

This PR addresses the future `D412` issues which will come up when we remove the pinning of `pydocstyle` to versions < 4.0.0

### DIFF
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -6,41 +6,40 @@
 """Simple protein analysis.
 
 Example:
+>>> from Bio.SeqUtils.ProtParam import ProteinAnalysis
+>>> X = ProteinAnalysis("MAEGEITTFTALTEKFNLPPGNYKKPKLLYCSNGGHFLRILPDGTVDGT"
+...                     "RDRSDQHIQLQLSAESVGEVYIKSTETGQYLAMDTSGLLYGSQTPSEEC"
+...                     "LFLERLEENHYNTYTSKKHAEKNWFVGLKKNGSCKRGPRTHYGQKAILF"
+...                     "LPLPV")
+>>> print(X.count_amino_acids()['A'])
+6
+>>> print(X.count_amino_acids()['E'])
+12
+>>> print("%0.2f" % X.get_amino_acids_percent()['A'])
+0.04
+>>> print("%0.2f" % X.get_amino_acids_percent()['L'])
+0.12
+>>> print("%0.2f" % X.molecular_weight())
+17103.16
+>>> print("%0.2f" % X.aromaticity())
+0.10
+>>> print("%0.2f" % X.instability_index())
+41.98
+>>> print("%0.2f" % X.isoelectric_point())
+7.72
+>>> sec_struc = X.secondary_structure_fraction()  # [helix, turn, sheet]
+>>> print("%0.2f" % sec_struc[0])  # helix
+0.28
+>>> epsilon_prot = X.molar_extinction_coefficient()  # [reduced, oxidized]
+>>> print(epsilon_prot[0])  # with reduced cysteines
+17420
+>>> print(epsilon_prot[1])  # with disulfid bridges
+17545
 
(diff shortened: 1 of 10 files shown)


Bio/SeqUtils/ProtParam.py
```diff
@@ -6,41 +6,40 @@
...
-    >>> print(epsilon_prot[0])  # with reduced cysteines
-    17420
-    >>> print(epsilon_prot[1])  # with disulfid bridges
-    17545
-
-    Other public methods are:
-     - gravy
-     - protein_scale
-     - flexibility
+>>> from Bio.SeqUtils.ProtParam import ProteinAnalysis
```
peterjc: Does this render correctly? Do we need something like:

```rst
Examples
--------
>>> print(123)
123
```

Note this is deliberately with a "s" and no colon as in many other places in Biopython - to match the numpydoc standard, which seems the most likely standard we might want to adopt.

Scripts/Restriction/ranacompiler.py
```diff
@@ -103,9 +103,8 @@ def regex(site):
     """Construct a regular expression (string) from a DNA sequence.
 
     Example:
-
-        >>> regex('ABCGN')
-        'A[CGT]CG.'
+    >>> regex('ABCGN')
```
peterjc: Again, does this render OK. Does this work better?:

```rst
Examples
--------
>>> regex('ABCGN')
'A[CGT]CG.'
```

peterjc:
Can you check the RST with a doctest example?

Bio/SeqUtils/ProtParam.py
MarkusPiotrowski: I thought that the `>>>` should be enough to mark this as code block, unfortunately, you seem to be right that it is not rendered correctly (at least when relying on https://livesphinx.herokuapp.com/ or http://rst.aaroniles.net/, although I had the feeling in the past that even rst.ninjs.org was not doing everything correctly. Unfortunately, I don't have an actual example)

Do you have an example how 
```
Examples
----------
>>> print(123)
123
```
is rendered in Sphinx? Regarding the size of the header? If it is rendered as larger header, I'd prefer to mark the code block by a simple one-space indentation.

peterjc: Looks OK here, really need an example with more than one section...

https://github.com/biopython/biopython/blob/biopython-174/Bio/File.py#L78
https://biopython.org/docs/1.74/api/Bio.File.html#Bio.File.as_handle
https://biopython.org/DIST/docs/api/Bio.File-module.html#as_handle

MarkusPiotrowski: Okay, I go with this. 

Can you clarify what you mean with 'check the RST with a doctest example' below?

peterjc: It is the same situation - rather than this:

```
Example:
>>> regex('ABCGN')
'A[CGT]CG.'
```

How about:

```
Examples
--------
>>> regex('ABCGN')
'A[CGT]CG.'
```

Scripts/Restriction/ranacompiler.py
MarkusPiotrowski: OK

Bio/motifs/minimal.py
```diff
@@ -17,7 +17,6 @@ def read(handle):
     """Parse the text output of the MEME program into a meme.Record object.
 
     Example:
-
```
peterjc: Does this (and the other case in this file) also need an RST tweak?

MarkusPiotrowski: Yep

PR #2287, opened 2019-09-27
Define Seq object .index() and .rindex()

peterjc:
This gives Seq (and subclasses) a string-like .index() and .rindex() method.

Note that MutableSeq has just an array or list-like .index(), and like a list, no .rindex() method or .find() and .rfind() - adding these would clash with the partial support for things like three-letter-alphabet sequences using this object.


peterjc:
Looks like pypy3 has a difference string exception message, which breaks the doctest

peterjc:
This might be the right fix in PyPy to have their exception be consistent...
https://bitbucket.org/pypy/pypy/pull-requests/669/use-same-valueerror-as-cpython-for-string/diff

Issue #2281, opened 2019-09-24
"make -C Doc/api/ html" exits with code 1 unless you create _static folder first


ExplodingCabbage:
In .travis-tox.ini, this command is run before `make -C Doc/api/ html`:

```
mkdir -p Doc/api/_templates Doc/api/_static Doc/api/_build
```

If I try to build the docs locally without doing this first, I get a warning at the start...

> WARNING: html_static_path entry '_static' does not exist

and ultimately the build reports a failure as a consequence:

```
build finished with problems, 1 warning.
make: *** [html] Error 1
```

This is fixed if I run the `mkdir` command before building. This seems like a minor but easily fixable source of confusion for someone trying to build the docs locally for the first time.

Some possible fixes:

- Commit `Doc/api/_static/.gitkeep` and `Doc/api/_templates/.gitkeep` to the repo; remove the line creating those folders from the `.travis-tox.ini`
- Move the `mkdir` command that creates them into the makefile instead of it being in `.travis-tox.ini`
- Change some config somewhere to remove the need for these folders to exist (the build never puts anything in them!) and then remove the line creating them from `.travis-tox.ini`

peterjc:
Moving the ``mkdir`` in to ``Makefile`` sounds simplest - do you want to make a pull request on that?

peterjc:
Fixed, thank you!

PR #2289, opened 2019-09-28
Make make not complain about lack of _build dir when run locally

ExplodingCabbage:
This pull request addresses issue #2281 

---

In the end I decided to use `.gitkeep` files instead of adding the `mkdir` command to the Makefile, contrary to @peterjc's suggestion in #2281, because:

* I wasn't sure that running that command for *every* Make operation made sense, but since they're all deferred to Sphinx, I can't easily make it run only on build.
* I wasn't sure how to cleanly get the right path in the makefile to pass to the command; if I just copy it verbatim, then the paths are relative to the current working directory and so wrong if you run make from within the `api` folder.

I think this solution is also fine, @peterjc?

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -167,5 +167,4 @@ deps =
     sphinx_rtd_theme
 commands =
     bash -c \'python setup.py install > /dev/null\'
-    bash -c \'mkdir -p Doc/api/_templates Doc/api/_static Doc/api/_build\'
     make -C Doc/api/ html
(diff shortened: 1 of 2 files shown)


peterjc:
Hmm, not immediately obvious why this breaks from the log:

```
Makefile:20: recipe for target 'html' failed
make: *** [html] Error 1
make: Leaving directory '/home/travis/build/biopython/biopython/Doc/api'
ERROR: InvocationError for command /usr/bin/make -C Doc/api/ html (exited with code 2)
```

However, it looks likely to be down to ``Doc/api/_static/`` not existing?

ExplodingCabbage:
D'oh, sorry. You're right, I screwed up. Will fix tonight.

peterjc:
That's what automated testing is for - we're all human.

ExplodingCabbage:
Fix force-pushed.

Issue #2270, opened 2019-09-16
Default treatment of .description in SeqIO FASTA format


rsharris:
### Setup

I am reporting a problem with Biopython 1.74, Python 3.7.3, and linux as follows:

Python 3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys; print(sys.version)
3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-4.19.0-6-amd64-x86_64-with-debian-10.1
>>> import Bio; print(Bio.__version__)
1.73

for r in SeqIO.parse(infile, "fasta"):
	d=str(len(r.seq))
	SeqIO.write(SeqRecord(r.seq,id=r.id,description=d),outfile,"fasta")

### Expected behaviour

I expect the fasta headers to contain the sequence name and the sequence length.

### Actual behaviour

If the sequence length happens to equal the sequence name, as_fasta() incorrectly assumes the "description includes the id at the start" and discards the id.

This is a serious problem. I am using a third party package that uses BioPython. the package reads a fasta file, does some computation on each sequence, then outputs the same sequences with several statistics annotated in the fasta header lines. The first stat happens to be the sequence length.

Meanwhile, the input sequence I am using is the output of a 3rd party sequence assembler. The sequence assembler happens to number its output sequences.

I can't change either of the 3rd party tools. (And why should I need to -- both are correct).

As SeqIO is currently implemented, it seems like for any program using SeqIO.write() and using a description, that calling program has to check to make sure the description's prefix doesn't match the sequence's name. And then I guess they can work around that by doubling the prefix. But that really seems nutty.

Maybe you could add an argument to SeqIO.write() that says 'use all of my description.'

### Steps to reproduce

Create a fasta file with a length 5 sequence named "5". Then feed it to this snippet.

for r in SeqIO.parse(infile, "fasta"):
	d=str(len(r.seq))
	SeqIO.write(SeqRecord(r.seq,id=r.id,description=d),outfile,"fasta")

The output will contain a fasta header ">5" rather than the correct ">5 5".

chris-rands:
I believe this behaviour is [by design
](https://github.com/biopython/biopython/blob/master/Bio/SeqIO/FastaIO.py#L359)

Would this kind of workaround be okay for you?

```
[...]

PR #2296, opened 2019-10-10
Improve docs by illustrating how to modify a SeqRecord 

chris-rands:
This pull request addresses issue #2270 (following #2295).

### DIFF
--- a/Bio/AlignIO/Interfaces.py
+++ b/Bio/AlignIO/Interfaces.py
@@ -116,7 +116,7 @@ class AlignmentWriter(object):
 
     def clean(self, text):
         """Use this to avoid getting newlines in the output."""
-        return text.replace("\n", " ").replace("\r", " ").replace("  ", " ")
+        return text.replace("\n", " ").replace("\r", " ")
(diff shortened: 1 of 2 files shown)


peterjc:
Would it make sense to you as a reader to present this in doctest style? i.e. leading ``>>> `` markers, and the output from Python?

If so, we can turn it into an actually tested doctest - see "Writing doctests in the Tutorial" in http://biopython.org/DIST/docs/tutorial/Tutorial.html and https://github.com/biopython/biopython/blob/master/Tests/test_Tutorial.py for the implementation.

peterjc:
Yep, looks like the minor path issue I suspected is to blame.

chris-rands:
Naive question- when are the HTML and PDF tutorials updated?

peterjc:
As part of each release. Building the LaTeX as part of the continuous integration testing would be nice but the dependencies are a pain -  see #1311 .

Separately, we're looking at moving it all to RST and using Sphinx to combine this and the API documentation - see #907.

Issue #2259, opened 2019-09-10
MMCIFParser. Non integer _atom_site.auth_seq_id


andrei-istrate:
I get an error when the _atom_site.auth_seq_id is not an integer number.

According to http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Items/_atom_site.auth_seq_id.html
this item does not have to be an integer.

### Setup
I am reporting a problem with Biopython version, Python version, and operating
system as follows:

import sys; print(sys.version)
3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Darwin-18.5.0-x86_64-i386-64bit
import Bio; print(Bio.__version__)
1.74

### Actual behaviour

Traceback (most recent call last):
  File "/Users/andrei/PycharmProjects/3D-STRUDEL/playground/test.py", line 15, in <module>
    model = parser.get_structure(in_path.split('/')[-1].split('.')[0], in_path)
  File "/Users/andrei/PycharmProjects/3D-STRUDEL/venv/lib/python3.6/site-packages/Bio/PDB/MMCIFParser.py", line 64, in get_structure
    self._build_structure(structure_id)
  File "/Users/andrei/PycharmProjects/3D-STRUDEL/venv/lib/python3.6/site-packages/Bio/PDB/MMCIFParser.py", line 144, in _build_structure
    int_resseq = int(seq_id_list[i])
ValueError: invalid literal for int() with base 10: '20A'

### Steps to reproduce
parser = MMCIFParser()
model = parser.get_structure('structure_id', 'test.cif')

data_zzz
loop_
  _atom_site.group_PDB
  _atom_site.id
  _atom_site.type_symbol
  _atom_site.label_atom_id
  _atom_site.label_alt_id
  _atom_site.label_comp_id
  _atom_site.label_asym_id
  _atom_site.label_entity_id
  _atom_site.label_seq_id
  _atom_site.pdbx_PDB_atom_name
  _atom_site.pdbx_PDB_ins_code
  _atom_site.pdbx_PDB_residue_name
  _atom_site.pdbx_PDB_residue_no
  _atom_site.Cartn_x
  _atom_site.Cartn_y
  _atom_site.Cartn_z
  _atom_site.occupancy
  _atom_site.B_iso_or_equiv
[...]

PR #2307, opened 2019-10-20
Added argument to the mmCIF parser to predefine the use of label_seq_id

svalqui:
This pull request addresses issue #2259 

This should address the issue, I tested it on Python3.6; it would be good is someone else review and test it.

### DIFF
--- a/Bio/PDB/MMCIFParser.py
+++ b/Bio/PDB/MMCIFParser.py
@@ -49,25 +49,27 @@ class MMCIFParser(object):
 
     # Public methods
 
-    def get_structure(self, structure_id, filename):
+    def get_structure(self, structure_id, filename, use_label_seq_id=False):
         """Return the structure.
 
         Arguments:
          - structure_id - string, the id that will be used for the structure
          - filename - name of mmCIF file, OR an open text mode file handle
+         - use_label_seq_id - boolean, to use _atom_site.label_seq_id over
+             _atom_site.auth_seq_id.
 
         """
         with warnings.catch_warnings():
             if self.QUIET:
                 warnings.filterwarnings("ignore", category=PDBConstructionWarning)
             self._mmcif_dict = MMCIF2Dict(filename)
-            self._build_structure(structure_id)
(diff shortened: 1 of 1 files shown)


jgreener64:
This more or less looks good to me, I have made the odd comment on the code.

Could you add a test case, ideally re-using an existing mmCIF file in the test suite?

Could you also add a sentence to the docs on this, perhaps at the end of section 11.1.1?

Bio/PDB/MMCIFParser.py
```diff
@@ -141,7 +143,12 @@ def _build_structure(self, structure_id):
             altloc = alt_list[i]
             if altloc in _unassigned:
                 altloc = " "
-            int_resseq = int(seq_id_list[i])
+            try:
+                int_resseq = int(seq_id_list[i])
+            except ValueError:
+                raise PDBConstructionException(
+                    "try parser.get_structure('structure_id', 'file_name', use_label_seq_id=True)"
```
jgreener64: Perhaps clarify this error, something like "conversion of sequence ID to integer failed, consider using...".

svalqui:
I'm afraid this fails, on files that have HETATM information, the field _atom_site.label_seq_id is sometimes a dot ".".

from Tests/PDB/2XHE.cif

ATOM   6267 C CG2 . VAL B 2 261 ? 17.741  -74.246 1.436   1.00 209.01 ? 261  VAL B CG2 1
HETATM 6268 O O   . HOH C 3 .   ? -17.130 -57.437 4.024   1.00 70.68  ? 2001 HOH A O   1

I do not seem to find another way than changing
https://github.com/biopython/biopython/blob/dbdfd1977d48c503c64a6669f8a8785e4f18346f/Bio/PDB/MMCIFParser.py#L144

which would expand to StructureBuilder.py, and that further expands to other files.
Is a significant change to the module.

I'll think about this for a couple of weeks to see if I can afford the time.

peterjc:
Closed as per https://github.com/biopython/biopython/issues/2259#issuecomment-546780021 which describes the larger body of work needed.

PR #2324, opened 2019-11-05
Apply black style to Bio.SubsMat.

svalqui:

These are about 65 changes, the number of lines on MatrixInfo.py  increases significantly as it contains matrices, it would be good to be reviewed before merge.


Bio/SubsMat/MatrixInfo.py
```diff
@@ -15,17 +15,48 @@
...
+    "pam180",
+    "pam250",
+    "pam30",
+    "pam300",
+    "pam60",
+    "pam90",
+    "rao",
+    "risler",
+    "structure",
+]
```
peterjc: The black version of this list is very long - this could be an example where locally disabling the black formatting, and instead having manual line wrapping at 88 characters could be preferable. What do you think?

peterjc:
The changes to ``Bio/SubsMat/MatrixInfo.py`` are examples where black's tendency to use a lot of vertical space for data structures is very very obvious. One might argue for disabling black on the entire file?

If we did that, manual reformatting to use the full 88 character line width might help with a more pleasing layout?

What do you think - this is exactly the sort of special case we anticipated debating on #2008.

svalqui:
I think it makes it difficult to read and to support the code, at the moment the file is showing up to 4 items per list per line, looks neat and passes flake8, So I'll leave this file as it was.
After the first run through all modules we can run black --diff to identify these files, and find a way to exclude them from an automated black.

Bio/SubsMat/__init__.py
```diff
@@ -601,7 +618,7 @@ def two_mat_relative_entropy(mat_1, mat_2, logbase=2, diag=diagALL):
         if mat_1[key] > EPSILON and mat_2[key] > EPSILON:
             val_1 = mat_1[key] / sum_ent_1
             val_2 = mat_2[key] / sum_ent_2
-#            rel_ent += mat_1[key] * log(mat_1[key]/mat_2[key])/log(logbase)
+            #            rel_ent += mat_1[key] * log(mat_1[key]/mat_2[key])/log(logbase)
```
peterjc: The indentation of this comment looks strange now - do you think we can just remove it?

svalqui: I have removed it, is been commented for more than 10 Years, I think it didn't calculate the relative entropy properly so got commented.

peterjc:
Applied, thank you.

PR #2339, opened 2019-11-11
Apply black style to Bio.SeqUtils

svalqui:

Merging this PR should be fine but these changes are a mix, as follow:

Applied black:
Bio/SeqUtils/__init__.py
Bio/SeqUtils/CheckSum.py
Bio/SeqUtils/lcc.py
Bio/SeqUtils/ProtParam.py

Partially changed, when lists have been presented one item per line, and looked long, the section has been kept the same as it was; other black changes to the rest of file have been applied manually.
Bio/SeqUtils/CodonUsage.py
Bio/SeqUtils/IsoelectricPoint.py
Bio/SeqUtils/MeltingTemp.py

No changed, the only changes applying black style relate to list items presented one item per line, so no changes to these files.
Bio/SeqUtils/CodonUsageIndices.py
Bio/SeqUtils/ProtParamData.py

Just to note, partially changed or no changed files will need manual attention.

### DIFF
--- a/Bio/SeqUtils/CodonUsage.py
+++ b/Bio/SeqUtils/CodonUsage.py
@@ -51,7 +51,8 @@ SynonymousCodons = {
     "TRP": ["TGG"],
     "VAL": ["GTA", "GTC", "GTG", "GTT"],
     "GLU": ["GAG", "GAA"],
-    "TYR": ["TAT", "TAC"]}
+    "TYR": ["TAT", "TAC"],
+}
 
 
 class CodonAdaptationIndex(object):
@@ -89,8 +90,10 @@ class CodonAdaptationIndex(object):
         """
         # first make sure we're not overwriting an existing index:
         if self.index != {} or self.codon_count != {}:
-            raise ValueError("an index has already been set or a codon count "
-                             "has been done. Cannot overwrite either.")
+            raise ValueError(
+                "an index has already been set or a codon count "
+                "has been done. Cannot overwrite either."
+            )
 
         # count codon occurrences in the file.
         self._count_codons(fasta_file)
@@ -134,7 +137,7 @@ class CodonAdaptationIndex(object):
             dna_sequence = dna_sequence.upper()
 
         for i in range(0, len(dna_sequence), 3):
-            codon = dna_sequence[i:i + 3]
+            codon = dna_sequence[i : i + 3]
             if codon in self.index:
                 # these two codons are always one, exclude them:
                 if codon not in ["ATG", "TGG"]:
@@ -142,8 +145,9 @@ class CodonAdaptationIndex(object):
                     cai_length += 1
             # some indices may not include stop codons:
             elif codon not in ["TGA", "TAA", "TAG"]:
-                raise TypeError("illegal codon in sequence: %s.\n%s"
-                                % (codon, self.index))
+                raise TypeError(
+                    "illegal codon in sequence: %s.\n%s" % (codon, self.index)
+                )
 
         return math.exp(cai_value / (cai_length - 1.0))
 
@@ -161,12 +165,13 @@ class CodonAdaptationIndex(object):
                 else:
                     dna_sequence = str(cur_record.seq)
                 for i in range(0, len(dna_sequence), 3):
-                    codon = dna_sequence[i:i + 3]
+                    codon = dna_sequence[i : i + 3]
                     if codon in self.codon_count:
                         self.codon_count[codon] += 1
                     else:
-                        raise TypeError("illegal codon %s in gene: %s"
-                                        % (codon, cur_record.id))
+                        raise TypeError(
+                            "illegal codon %s in gene: %s" % (codon, cur_record.id)
+                        )
 
     def print_index(self):
         """Print out the index used.
--- a/Bio/SeqUtils/CodonUsage.py
+++ b/Bio/SeqUtils/CodonUsage.py
@@ -12,21 +12,29 @@ import math
 from .CodonUsageIndices import SharpEcoliIndex
 from Bio import SeqIO  # To parse a FASTA file
 
+# Turn black code style off
+# fmt: off
 
 CodonsDict = {
-    "TTT": 0, "TTC": 0, "TTA": 0, "TTG": 0, "CTT": 0,
-    "CTC": 0, "CTA": 0, "CTG": 0, "ATT": 0, "ATC": 0,
-    "ATA": 0, "ATG": 0, "GTT": 0, "GTC": 0, "GTA": 0,
-    "GTG": 0, "TAT": 0, "TAC": 0, "TAA": 0, "TAG": 0,
-    "CAT": 0, "CAC": 0, "CAA": 0, "CAG": 0, "AAT": 0,
-    "AAC": 0, "AAA": 0, "AAG": 0, "GAT": 0, "GAC": 0,
-    "GAA": 0, "GAG": 0, "TCT": 0, "TCC": 0, "TCA": 0,
-    "TCG": 0, "CCT": 0, "CCC": 0, "CCA": 0, "CCG": 0,
-    "ACT": 0, "ACC": 0, "ACA": 0, "ACG": 0, "GCT": 0,
-    "GCC": 0, "GCA": 0, "GCG": 0, "TGT": 0, "TGC": 0,
-    "TGA": 0, "TGG": 0, "CGT": 0, "CGC": 0, "CGA": 0,
-    "CGG": 0, "AGT": 0, "AGC": 0, "AGA": 0, "AGG": 0,
+    "TTT": 0, "TTC": 0, "TTA": 0, "TTG": 0,
+    "CTT": 0, "CTC": 0, "CTA": 0, "CTG": 0,
+    "ATT": 0, "ATC": 0, "ATA": 0, "ATG": 0,
+    "GTT": 0, "GTC": 0, "GTA": 0, "GTG": 0,
+    "TAT": 0, "TAC": 0, "TAA": 0, "TAG": 0,
+    "CAT": 0, "CAC": 0, "CAA": 0, "CAG": 0,
+    "AAT": 0, "AAC": 0, "AAA": 0, "AAG": 0,
+    "GAT": 0, "GAC": 0, "GAA": 0, "GAG": 0,
+    "TCT": 0, "TCC": 0, "TCA": 0, "TCG": 0,
+    "CCT": 0, "CCC": 0, "CCA": 0, "CCG": 0,
+    "ACT": 0, "ACC": 0, "ACA": 0, "ACG": 0,
+    "GCT": 0, "GCC": 0, "GCA": 0, "GCG": 0,
+    "TGT": 0, "TGC": 0, "TGA": 0, "TGG": 0,
+    "CGT": 0, "CGC": 0, "CGA": 0, "CGG": 0,
+    "AGT": 0, "AGC": 0, "AGA": 0, "AGG": 0,
     "GGT": 0, "GGC": 0, "GGA": 0, "GGG": 0}
+
+# Turn black code style on
+# fmt: on
 
 
 # this dictionary shows which codons encode the same AA
--- a/Bio/SeqUtils/IsoelectricPoint.py
+++ b/Bio/SeqUtils/IsoelectricPoint.py
@@ -80,6 +80,7 @@ class IsoelectricPoint(object):
         self.sequence = str(protein_sequence).upper()
         if not aa_content:
             from Bio.SeqUtils.ProtParam import ProteinAnalysis as _PA
+
             aa_content = _PA(self.sequence).count_amino_acids()
         self.charged_aas_content = self._select_charged(aa_content)
 
--- a/Bio/SeqUtils/IsoelectricPoint.py
+++ b/Bio/SeqUtils/IsoelectricPoint.py
@@ -25,8 +25,15 @@ http://fields.scripps.edu/DTASelect/20010710-pI-Algorithm.pdf
 positive_pKs = {"Nterm": 7.5, "K": 10.0, "R": 12.0, "H": 5.98}
 negative_pKs = {"Cterm": 3.55, "D": 4.05, "E": 4.45, "C": 9.0, "Y": 10.0}
 pKcterminal = {"D": 4.55, "E": 4.75}
-pKnterminal = {"A": 7.59, "M": 7.0, "S": 6.93, "P": 8.36, "T": 6.82, "V": 7.44,
-               "E": 7.7}
+pKnterminal = {
+    "A": 7.59,
+    "M": 7.0,
+    "S": 6.93,
+    "P": 8.36,
+    "T": 6.82,
+    "V": 7.44,
+    "E": 7.7,
+}
 charged_aas = ("K", "R", "H", "D", "E", "C", "Y")
 
 
--- a/Bio/SeqUtils/MeltingTemp.py
+++ b/Bio/SeqUtils/MeltingTemp.py
@@ -565,11 +565,12 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
         corr = 11.7 * math.log10(mon)
     if method == 5:
         corr = 0.368 * (len(seq) - 1) * math.log(mon)
+    if method == 6:
+        corr = (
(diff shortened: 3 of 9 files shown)


peterjc:
For the long lists, do you think adding the special format off/on comments is worthwhile to temporarily disable black for that part of the file?

svalqui:
Yes, I think there would be more files that I had initially though, more than 10; so is better start commenting them now, I will update this.

Bio/SeqUtils/IsoelectricPoint.py
```diff
@@ -25,8 +25,15 @@
...
-               "E": 7.7}
+pKnterminal = {
+    "A": 7.59,
+    "M": 7.0,
+    "S": 6.93,
+    "P": 8.36,
+    "T": 6.82,
+    "V": 7.44,
+    "E": 7.7,
+}
```
MarkusPiotrowski: How about 
```python
pKnterminal = {
    "A": 7.59, "M": 7.0, "S": 6.93, "P": 8.36, "T": 6.82, "V": 7.44, "E": 7.7
}
```

Bio/SeqUtils/IsoelectricPoint.py
```diff
@@ -80,6 +87,7 @@ def __init__(self, protein_sequence, aa_content=None):
         self.sequence = str(protein_sequence).upper()
         if not aa_content:
             from Bio.SeqUtils.ProtParam import ProteinAnalysis as _PA
+
```
MarkusPiotrowski: Is this blank line added by black? If not, I'd prefer to omit it.

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -498,15 +534,17 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
 
     """
     if method in (5, 6, 7) and not seq:
-        raise ValueError("sequence is missing (is needed to calculate " +
-                         "GC content or sequence length).")
+        raise ValueError(
+            "sequence is missing (is needed to calculate "
+            + "GC content or sequence length)."
```
MarkusPiotrowski: I think the `+` in the second line can be ommited

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -515,8 +553,9 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
     mon = Mon * 1e-3
     # Note: math.log = ln(), math.log10 = log()
     if method in range(1, 7) and not mon:
-        raise ValueError("Total ion concentration of zero is not allowed in " +
-                         "this method.")
+        raise ValueError(
+            "Total ion concentration of zero is not allowed in " + "this method."
```
MarkusPiotrowski: Write as one string

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -528,40 +567,50 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
     if method == 5:
         corr = 0.368 * (len(seq) - 1) * math.log(mon)
     if method == 6:
-        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(mon) +\
-            9.40e-6 * math.log(mon) ** 2
+        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(
+            mon
+        ) + 9.40e-6 * math.log(mon) ** 2
```
MarkusPiotrowski: Noooo! There must be a different type of reformatting this formula. But this is really ugly and nonsense.

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -528,40 +567,50 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
...
+                + math.sqrt((ka * dntps - ka * mg + 1.0) ** 2 + 4.0 * ka * mg)
+            ) / (2.0 * ka)
         if Mon > 0:
             R = math.sqrt(mg) / mon
             if R < 0.22:
-                corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * \
-                    1e-5 * math.log(mon) + 9.40e-6 * math.log(mon) ** 2
+                corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(
+                    mon
+                ) + 9.40e-6 * math.log(mon) ** 2
```
MarkusPiotrowski: Same as above. It doesn't make sense to separate `math.log(mon)` in this way.

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -712,19 +777,21 @@ def Tm_GC(seq, check=True, strict=True, valueset=7, userset=None, Na=50, K=0,
 
     """
     if saltcorr == 5:
-        raise ValueError("salt-correction method 5 not applicable "
-                         "to Tm_GC")
+        raise ValueError("salt-correction method 5 not applicable " "to Tm_GC")
```
MarkusPiotrowski: Write as one string

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -1035,17 +1124,19 @@ def Tm_staluc(s, dnac=50, saltc=50, rna=0):
     # Original method was by Sebastian Bassi <sbassi@genesdigitales.com>. It is
     # now superseded by Tm_NN.
 
-    warnings.warn("Tm_staluc may be depreciated in the future. Use Tm_NN " +
-                  "instead.", PendingDeprecationWarning)
+    warnings.warn(
+        "Tm_staluc may be depreciated in the future. Use Tm_NN " + "instead.",
```
MarkusPiotrowski: Write as one string

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -404,12 +410,25 @@ def make_table(oldtable=None, values=None):
...
+            "AT/TA": (0, 0),
+            "TA/AT": (0, 0),
+            "CA/GT": (0, 0),
+            "GT/CA": (0, 0),
+            "CT/GA": (0, 0),
+            "GA/CT": (0, 0),
+            "CG/GC": (0, 0),
+            "GC/CG": (0, 0),
+            "GG/CC": (0, 0),
+        }
```
MarkusPiotrowski: Since you turned off black formatting for the tables above, you might think about turning it off here, too. However, it's your decision.

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -1035,17 +1124,19 @@ def Tm_staluc(s, dnac=50, saltc=50, rna=0):
...
-        return Tm_NN(s, dnac1=dnac / 2.0, dnac2=dnac / 2.0, Na=saltc,
-                     nn_table=RNA_NN2)
+        return Tm_NN(s, dnac1=dnac / 2.0, dnac2=dnac / 2.0, Na=saltc, nn_table=RNA_NN2)
     else:
         raise ValueError("rna={0} not supported".format(rna))
 
 
 if __name__ == "__main__":
     from Bio._utils import run_doctest
+
```
MarkusPiotrowski: Again, is this blank line introduce by black? If not, I'd removed it, otherwise OK

Bio/SeqUtils/IsoelectricPoint.py
peterjc: Black puts a blank line after any import - fine for top level imports, but a bit annoying for lazy imports.

Bio/SeqUtils/MeltingTemp.py
peterjc: Yes - that would be nicer.

Bio/SeqUtils/IsoelectricPoint.py
peterjc: If we want to disable black in this region, that looks fine. But if that is the only case in this file, I would just let black do what it wants.

Bio/SeqUtils/MeltingTemp.py
peterjc: That is rather a mess, I wonder if additional brackets might make us and black happy?

peterjc:
Markus read this more carefully than me - lots of good points.

Bio/SeqUtils/IsoelectricPoint.py
MarkusPiotrowski: Does this require disabling black? I thought that would be black compatible?

svalqui: This would need disabling black, since is a list of 7 items, short,  I will leave black take care of it.

Bio/SeqUtils/IsoelectricPoint.py
svalqui: Yes added by black.

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -528,40 +567,50 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
...
+                g = 8.31 * (
+                    0.486 - 0.258 * math.log(mon) + 5.25e-3 * math.log(mon) ** 3
+                )
+        corr = (
+            a
+            + b * math.log(mg)
+            + (SeqUtils.GC(seq) / 100) * (c + d * math.log(mg))
+            + (1 / (2.0 * (len(seq) - 1)))
+            * (e + f * math.log(mg) + g * math.log(mg) ** 2)
+        ) * 1e-5
```
svalqui: This doesn't read well neither.

svalqui:
@MarkusPiotrowski , @peterjc , Thank you for the feedback. I have update it, I think is better now.
So far there is a couple of considerations:
- When lists are presented one item per line and has many items, is hard to read; how long is long for a list presented one item per line? I think about 20 lines, then is worth considering switching black off.
- When black makes it hard to read, whether because it rewrites the style differently to just part of a  section due to length or splits a formula making it hard to follow, then is worth considering switching black off.
Let's see what else we found.
Cheers,

Bio/SeqUtils/IsoelectricPoint.py
MarkusPiotrowski: From what I read [here](https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines) I would think that my suggestion would be black compatible. It's something like the second rule: if it doesn't fit in the same line with the opening bracket, try it in an extra line. So, did you do this manually, or was it done by black? If it was done by black, I would file a ticket to them.

svalqui: black does this, It doesn't take your suggestion, it puts one item per line, not even with an extra line.

MarkusPiotrowski: Okay, thank you. Actually, I don't understand the difference in the black examples between a function call with parameters and a function definition with arguments, and why you could do this with a list but not with a dic, but that's not our problem. So, OK.

Bio/SeqUtils/MeltingTemp.py
peterjc: Added by black (see above)

Bio/SeqUtils/MeltingTemp.py
```diff
@@ -527,10 +565,16 @@ def salt_correction(Na=0, K=0, Tris=0, Mg=0, dNTPs=0, method=1, seq=None):
...
     if method == 5:
         corr = 0.368 * (len(seq) - 1) * math.log(mon)
+    # Turn black code style off
+    # fmt: off
     if method == 6:
         corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(mon) +\
             9.40e-6 * math.log(mon) ** 2
+    # Turn black code style on
+    # fmt: on
     if method == 7:
```
MarkusPiotrowski: Wouldn't it be less noisy to let black turned off until line 603?

svalqui: Yes, updated now.

Bio/SeqUtils/MeltingTemp.py
svalqui: done.

Bio/SeqUtils/MeltingTemp.py
svalqui: done.

Bio/SeqUtils/MeltingTemp.py
svalqui: black switched off for this.

Bio/SeqUtils/MeltingTemp.py
svalqui: black switched off for this.

Bio/SeqUtils/MeltingTemp.py
svalqui: done.

Bio/SeqUtils/MeltingTemp.py
svalqui: done.

Bio/SeqUtils/CodonUsage.py
```diff
@@ -12,22 +12,30 @@
...
-    "GTG": 0, "TAT": 0, "TAC": 0, "TAA": 0, "TAG": 0,
-    "CAT": 0, "CAC": 0, "CAA": 0, "CAG": 0, "AAT": 0,
-    "AAC": 0, "AAA": 0, "AAG": 0, "GAT": 0, "GAC": 0,
-    "GAA": 0, "GAG": 0, "TCT": 0, "TCC": 0, "TCA": 0,
-    "TCG": 0, "CCT": 0, "CCC": 0, "CCA": 0, "CCG": 0,
-    "ACT": 0, "ACC": 0, "ACA": 0, "ACG": 0, "GCT": 0,
-    "GCC": 0, "GCA": 0, "GCG": 0, "TGT": 0, "TGC": 0,
-    "TGA": 0, "TGG": 0, "CGT": 0, "CGC": 0, "CGA": 0,
-    "CGG": 0, "AGT": 0, "AGC": 0, "AGA": 0, "AGG": 0,
+    "TTT": 0, "TTC": 0, "TTA": 0, "TTG": 0,
```
peterjc: Given the nature of the genetic code, four entries per row is much nicer than five.

Bio/SeqUtils/MeltingTemp.py
peterjc: Did adding brackets not help?

svalqui: - For the current  style
```
    if method == 6:
        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(mon) +\
            9.40e-6 * math.log(mon) ** 2
```
- Black produces
```
    if method == 6:
        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(
            mon
        ) + 9.40e-6 * math.log(mon) ** 2
```
- Adding brackets around math.log
```
    if method == 6:
        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * (math.log(mon)) +\
            9.40e-6 * math.log(mon) ** 2
```
- Black produces
```
    if method == 6:
        corr = (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * (
            math.log(mon)
        ) + 9.40e-6 * math.log(mon) ** 2
```
- Adding brackets around the first addend
```
    if method == 6:
        corr = ((4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(mon)) +\
            9.40e-6 * math.log(mon) ** 2
```
- Black produces
```
    if method == 6:
        corr = (
            (4.29 * SeqUtils.GC(seq) / 100 - 3.95) * 1e-5 * math.log(mon)
        ) + 9.40e-6 * math.log(mon) ** 2
```

Making this 2 lined formula 3 lines and removing the addition operator from the end of the first line, makes it a bit odd to read but if it needs to change let me know your choice.

peterjc: I prefer the final version, Markus?

MarkusPiotrowski: Agree

svalqui: updated.

peterjc:
Applied, thank you both 👍

PR #2353, opened 2019-11-22
Apply black style to Bio.SCOP

svalqui:

Minor changes should be fine to merge.

### DIFF
--- a/Bio/SCOP/__init__.py
+++ b/Bio/SCOP/__init__.py
@@ -58,88 +58,33 @@ from . import Residues
 from Bio import SeqIO
 from Bio.Seq import Seq
 
-
-nodeCodeDict = {
-    "cl": "class",
-    "cf": "fold",
-    "sf": "superfamily",
-    "fa": "family",
-    "dm": "protein",
-    "sp": "species",
-    "px": "domain",
-}
+# Turn black code style off
+# fmt: off
+nodeCodeDict = {"cl": "class", "cf": "fold", "sf": "superfamily",
+                "fa": "family", "dm": "protein", "sp": "species", "px": "domain"}
 
 
-_nodetype_to_code = {
-    "class": "cl",
-    "fold": "cf",
-    "superfamily": "sf",
-    "family": "fa",
-    "protein": "dm",
-    "species": "sp",
-    "domain": "px",
-}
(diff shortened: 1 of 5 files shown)


Bio/SCOP/__init__.py
```diff
@@ -467,31 +502,48 @@ def write_cla_sql(self, handle):
         cur = handle.cursor()
 
         cur.execute("DROP TABLE IF EXISTS cla")
-        cur.execute("CREATE TABLE cla (sunid INT, sid CHAR(8), pdbid CHAR(4),\
+        cur.execute(
+            "CREATE TABLE cla (sunid INT, sid CHAR(8), pdbid CHAR(4),\
         residues VARCHAR(50), sccs CHAR(10), cl INT, cf INT, sf INT, fa INT,\
-        dm INT, sp INT, px INT, PRIMARY KEY (sunid), INDEX (SID) )")
+        dm INT, sp INT, px INT, PRIMARY KEY (sunid), INDEX (SID) )"
+        )
```
peterjc: This is odd - I don't think we need the slashes inside the SQL command string - SQL should cope just fine with the command as one long string. In that case, Python implicit string concatenation would be neatest, so something like this:

```python
        cur.execute(
            "CREATE TABLE cla (sunid INT, sid CHAR(8), pdbid CHAR(4), "
            "residues VARCHAR(50), sccs CHAR(10), cl INT, cf INT, sf INT, fa INT, "
            "dm INT, sp INT, px INT, PRIMARY KEY (sunid), INDEX (SID))"
        )
```

Bio/SCOP/__init__.py
```diff
@@ -467,31 +502,48 @@ def write_cla_sql(self, handle):
...
         cur = handle.cursor()
 
         cur.execute("DROP TABLE IF EXISTS des")
-        cur.execute("CREATE TABLE des (sunid INT, type CHAR(2), sccs CHAR(10),\
+        cur.execute(
+            "CREATE TABLE des (sunid INT, type CHAR(2), sccs CHAR(10),\
         description VARCHAR(255),\
-        PRIMARY KEY (sunid) )")
+        PRIMARY KEY (sunid) )"
+        )
```
peterjc: Again, same point - I don't see why the original code had internal line continuation slashes in the SQL command.

svalqui: Done.

PR #2356, opened 2019-11-25
Apply black style to Bio.Restriction.

svalqui:

The style has changed some lists appearance, making it long but it reads better; should be fine to merge.

### DIFF
--- a/Bio/Restriction/PrintFormat.py
+++ b/Bio/Restriction/PrintFormat.py
@@ -168,7 +168,7 @@ class PrintFormat(object):
         """
         return self._make_list(cut, title, nc, s1)
 
-# _make_* methods to be used with the virtual method make_format
+    # _make_* methods to be used with the virtual method make_format
 
     def _make_list(self, ls, title, nc, s1):
         """Summarise a list of positions by enzyme (PRIVATE).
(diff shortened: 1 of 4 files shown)


MarkusPiotrowski:
@svalqui `Bio.Restriction_Dictionary` is an autogenerated file. So it does not make sense to change the file itself but the script (`ranacompiler.py` under Scripts) that generates it.

svalqui:
@MarkusPiotrowski , Thank you for that, I missed it; I'll have a think about this one, the lines are too long horizontally as is, after black they are too long vertically, needs a middle ground.

peterjc:
If the black version is too long (and it often is), I would just mark the file a black format off - which needs a tiny change to the autogeneration script.

svalqui:
I have updated ranacompiler.py to switch off black style, and run it to check Restriction_Dictionary.py; seems fine now; however it had been producing long lines already, about 2290 characters long
https://github.com/biopython/biopython/blob/7844bcaa558b7922af3601a5b82173a76bc51c69/Bio/Restriction/Restriction_Dictionary.py#L24354-L24359

That would need some sort of styling, I do not think that is part of this PR, added a TODO note for that.

peterjc:
I would leave the automatically generated file as it was other than adding a _single_ set of black format off/on lines at the top and bottom (not adding them for every enzyme).

svalqui:
There is only one black off and one black on for the entire file, not per each enzyme;  the first after the header, the last after "del _temp\n". ranacompiler.py produces the file fine with a single set of black off/on.

peterjc:
Sorry - you're right. I see you've also updated the REBASE database as part of this - which I would prefer to see as a separate pull request, or at least separate commit. It usually gets a line in the NEWS file too.

svalqui:
I understand now, I have left Restriction_Dictionary the same version as is now, just added the black statement manually.

peterjc:
Thank you!

PR #2358, opened 2019-11-27
Fix same line implicit string concatenation from black

peterjc:
There is an open issue on black for fixing this:
https://github.com/psf/black/issues/26#issuecomment-558983613

These specific cases were spotted while testing a new flake8 plugin:
https://github.com/keisheiled/flake8-implicit-str-concat/

Once that plugin as matured a little, we should start using it on our continuous integration testing (at least until the black issue is addressed).


peterjc:
Now covered in #2359 along with using a new flake8 plugin to enforce this.

PR #2362, opened 2019-11-27
Cif header

rob-miller:
This pull request adds code to PDB/MMCIFParser.py to populate the `<structure>.header` dict with fields `idcode`, `head`, `deposition_date`, `structure_method` and `resolution` if they can be easily found in the CIF dictionary as loaded. A `test_header()` routine is added to test_PDB_MMCIFParser.py to confirm the functionality, and the test_ file is black formatted.

The use case is to provide similar metadata regardless of loading from a mmCIF or a PDB file.

### DIFF
--- a/Bio/PDB/MMCIFParser.py
+++ b/Bio/PDB/MMCIFParser.py
@@ -67,6 +67,49 @@ class MMCIFParser(object):
 
     # Private methods
 
+    def _mmcif_get(self, key, dict, deflt):
+        if key in dict:
+            rslt = dict[key][0]
+            if "?" != rslt:
+                return rslt
+        return deflt
+
+    def _get_mmcif_header_dict(self, md):
+        dict = {
+            "name": "",
+            "head": "",
+            "idcode": "",
+            "deposition_date": "",
+            "structure_method": "",
+            "resolution": 0.0,
+        }
+
+        dict["idcode"] = self._mmcif_get("_entry.id", md, None)
+        if not dict["idcode"]:
+            dict["idcode"] = self._mmcif_get("_exptl.entry_id", md, None)
+        if not dict["idcode"]:
+            dict["idcode"] = self._mmcif_get("_struct.entry_id", md, None)
+
+        dict["name"] = self._mmcif_get("_struct.title", md, None)
+
+        dict["head"] = self._mmcif_get("_struct_keywords.pdbx_keywords", md, None)
+        if not dict["head"]:
+            dict["head"] = self._mmcif_get("_struct_keywords.text", md, None)
+
+        dict["deposition_date"] = self._mmcif_get(
+            "_pdbx_database_status.recvd_initial_deposition_date", md, None
+        )
+
+        dict["structure_method"] = self._mmcif_get("_exptl.method", md, None)
+
+        dict["resolution"] = float(self._mmcif_get("_refine.ls_d_res_high", md, 0.0))
+        if 0.0 == dict["resolution"]:
+            dict["resolution"] = float(
+                self._mmcif_get("_refine_hist.d_res_high", md, 0.0)
+            )
+
+        return dict
+
     def _build_structure(self, structure_id):
 
         # two special chars as placeholders in the mmCIF format
@@ -122,6 +165,7 @@ class MMCIFParser(object):
         structure_builder = self._structure_builder
         structure_builder.init_structure(structure_id)
         structure_builder.init_seg(" ")
+        structure_builder.set_header(self._get_mmcif_header_dict(mmcif_dict))
         # Historically, Biopython PDB parser uses model_id to mean array index
         # so serial_id means the Model ID specified in the file
         current_model_id = -1
--- a/Bio/PDB/MMCIFParser.py
+++ b/Bio/PDB/MMCIFParser.py
@@ -41,7 +41,7 @@ class MMCIFParser(object):
             self._structure_builder = structure_builder
         else:
             self._structure_builder = StructureBuilder()
-        # self.header = None
+        self.header = None
         # self.trailer = None
         self.line_counter = 0
         self.build_structure = None
@@ -62,6 +62,7 @@ class MMCIFParser(object):
                 warnings.filterwarnings("ignore", category=PDBConstructionWarning)
             self._mmcif_dict = MMCIF2Dict(filename)
             self._build_structure(structure_id)
+            self._structure_builder.set_header(self._get_header())
 
         return self._structure_builder.get_structure()
 
@@ -74,8 +75,20 @@ class MMCIFParser(object):
                 return rslt
         return deflt
 
-    def _get_mmcif_header_dict(self, md):
-        dict = {
+    def _update_header_entry(self, target_key, keys, is_float=False):
+        md = self._mmcif_dict
+        for key in keys:
+            val = md.get(key)
+            try:
+                item = val[0]
+            except (TypeError, IndexError):
+                continue
+            if item != "?":
+                self.header[target_key] = item if not is_float else float(item)
+                return
+
+    def _get_header(self):
+        self.header = {
             "name": "",
             "head": "",
             "idcode": "",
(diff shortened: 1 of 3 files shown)


Bio/PDB/MMCIFParser.py
```diff
@@ -67,6 +67,49 @@ def get_structure(self, structure_id, filename):
 
     # Private methods
 
+    def _mmcif_get(self, key, dict, deflt):
```
JoaoRodrigues: I'd refactor the function to use a default parameter for default, and avoid using `dict` as a variable name and accessing the key twice:

```
def __mmcif_get(self, key, mmcif_dict, default=None):
    val = mmcif_dict.get(key, default)
    try:
        item = val[0]
    except (TypeError, IndexError):
        pass
    else:
        if item != '?':
            return item
    return val
```

Bio/PDB/MMCIFParser.py
```diff
@@ -67,6 +67,49 @@ def get_structure(self, structure_id, filename):
...
+        if key in dict:
+            rslt = dict[key][0]
+            if "?" != rslt:
+                return rslt
+        return deflt
+
+    def _get_mmcif_header_dict(self, md):
+        dict = {
```
JoaoRodrigues: Avoid using `dict` as a variable name.

Bio/PDB/MMCIFParser.py
```diff
@@ -67,6 +67,49 @@ def get_structure(self, structure_id, filename):
...
+            "name": "",
+            "head": "",
+            "idcode": "",
+            "deposition_date": "",
+            "structure_method": "",
+            "resolution": 0.0,
+        }
+
+        dict["idcode"] = self._mmcif_get("_entry.id", md, None)
```
JoaoRodrigues: Would it be reasonable to refactor the `__mmcif_get` function to take a list of keys, in order of preference, and keep trying until finding a value (or returning the default at the end)? It would avoid all these `if` statements here and more extensible?

Bio/PDB/MMCIFParser.py
```diff
@@ -67,6 +67,49 @@ def get_structure(self, structure_id, filename):
...
+            dict["head"] = self._mmcif_get("_struct_keywords.text", md, None)
+
+        dict["deposition_date"] = self._mmcif_get(
+            "_pdbx_database_status.recvd_initial_deposition_date", md, None
+        )
+
+        dict["structure_method"] = self._mmcif_get("_exptl.method", md, None)
+
+        dict["resolution"] = float(self._mmcif_get("_refine.ls_d_res_high", md, 0.0))
+        if 0.0 == dict["resolution"]:
```
JoaoRodrigues: I'd rather swap the order in if-statements to be more readable: `if x == 0.0` instead of `if 0.0 == x`.

Bio/PDB/MMCIFParser.py
```diff
@@ -122,6 +165,7 @@ def _build_structure(self, structure_id):
         structure_builder = self._structure_builder
         structure_builder.init_structure(structure_id)
         structure_builder.init_seg(" ")
+        structure_builder.set_header(self._get_mmcif_header_dict(mmcif_dict))
```
JoaoRodrigues: The `PDBParser` class implements a `get_header` function that handles this step. To be coherent with that implementation, I'd do the same. The other reason being that the header is a separate piece of (meta)data than the coordinates so logically the two parsing steps should be clearly delineated and separated.

Tests/test_PDB_MMCIFParser.py
```diff
@@ -81,8 +82,11 @@ def test_parsers(self):
             self.assertEqual(s.alphabet, generic_protein)
 
             # Here non-standard MSE are shown as M
-            self.assertEqual("MDIRQGPKEPFRDYVDRFYKTLRAEQASQEVKNWMTETLLVQ"
-                             "NANPDCKTILKALGPGATLEEMMTACQG", str(s))
+            self.assertEqual(
```
JoaoRodrigues: These changes seem to reflect more formatting than the addition of the header code. Not an issue, but it'd be better to justify them than bundle them together in this PR.

Tests/test_PDB_MMCIFParser.py
```diff
@@ -278,30 +293,42 @@ def _run_point_mutation_tests(self, parser):
...
+        )
 
         res_1.disordered_select("SER")
         self.assertAlmostEqual(
-            res_1["CA"].get_occupancy(),
-            0.17, 2, "Residue 1 serine occupancy correcy")
+            res_1["CA"].get_occupancy(), 0.17, 2, "Residue 1 serine occupancy correcy"
+        )
+
+    def test_header(self):
```
JoaoRodrigues: Perfect. Could you add a 'negative' control test? A mmCIF file without headers to make sure the default values apply correctly?

Bio/PDB/MMCIFParser.py
rob-miller: done, slightly different approach but my previous effort was to match dictionary .get() syntax.

Bio/PDB/MMCIFParser.py
rob-miller: done.

Bio/PDB/MMCIFParser.py
rob-miller: removed with other changes.

Bio/PDB/MMCIFParser.py
rob-miller: I think this is done, if I understand your intent correctly.

Tests/test_PDB_MMCIFParser.py
rob-miller: done.

JoaoRodrigues: Thanks!

Bio/PDB/MMCIFParser.py
JoaoRodrigues: Thanks!

Bio/PDB/MMCIFParser.py
```diff
@@ -74,8 +75,20 @@ def _mmcif_get(self, key, dict, deflt):
...
+    def _update_header_entry(self, target_key, keys, is_float=False):
+        md = self._mmcif_dict
+        for key in keys:
+            val = md.get(key)
+            try:
+                item = val[0]
+            except (TypeError, IndexError):
+                continue
+            if item != "?":
+                self.header[target_key] = item if not is_float else float(item)
```
JoaoRodrigues: I'm not entirely sure about this line here. The default value will now be `None` for all cases, while the previous version you had (with the default in the function call) was more flexible. The ternary operator call also makes is slower since this comparison is called for every token in the header (right?). I would be more comfortable if this function acted more like a `.get` (like you had before) and then outside we handled the conversion to float explicitly depending on the value. More verbose, but more readable in my opinion.

Tests/test_PDB_MMCIFParser.py
```diff
@@ -323,10 +323,18 @@ def _run_point_mutation_tests(self, parser):
...
+        self.assertEqual("", structure.header["head"])
+        self.assertEqual("", structure.header["deposition_date"])
+        self.assertEqual("", structure.header["structure_method"])
+        self.assertEqual(0.0, structure.header["resolution"])
+
         structure = parser.get_structure("example", "PDB/1A8O.cif")
         self.assertEqual("1A8O", structure.header["idcode"])
         self.assertEqual("Viral protein", structure.header["head"])
-        self.assertIsNone(structure.header["deposition_date"])
+        self.assertEqual("", structure.header["deposition_date"])
```
JoaoRodrigues: Should this be None instead?

Bio/PDB/MMCIFParser.py
rob-miller: Yes, adding and handling the 'float' option was flakey in retrospect, and better handled outside as you suggest.  The default though is not to update - if the key is not found in the cif dictionary, None is returned and triggers the TypeError in the try/except.  The default is the initialization value in _get_header(), and we only update if there is data that is not "?".

Tests/test_PDB_MMCIFParser.py
rob-miller: I don't think it should be None here because the other entries default to "" to be consistent with PDB parsing, but PDBParser generates a default date of 1909-01-08.  
- arguably both PDB and CIF parsers should set things that are missing to None, but that ship has probably sailed 
- I note there's no 'test we handle no header data sensibly' entry in the Tests/ set BTW
- I do not know what the relevance of that date is, but does not seem beneficial that it is definitely 'wrong' for any data here

So I don't know what the right thing is  :-)

Bio/PDB/MMCIFParser.py
```diff
@@ -84,7 +84,7 @@ def _update_header_entry(self, target_key, keys, is_float=False):
             except (TypeError, IndexError):
                 continue
             if item != "?":
-                self.header[target_key] = item if not is_float else float(item)
+                self.header[target_key] = item
                 return
```
JoaoRodrigues: I'd remove this `return` and replace it with `break`, since the function will anyway return `None` and the goal of the return is to break the for loop.

rob-miller: done.

peterjc:
Restarting the PyPy TravisCI test which timed out on a URL.

Assuming that passes, do you want to go ahead and merge this @JoaoRodrigues?

peterjc:
Joao said he'd be travelling the few days, given his approval I'll probably merge this myself once the test finished.

peterjc:
Thanks both, rebased and merged

PR #2364, opened 2019-11-28
No space pdb title

rob-miller:
This pull request fixes an apparent bug in parse_pdb_header where a `space` is pre-pended to TITLE record data when read from a PDB file.  The original code adds the space to separate subsequent appended lines when processing multi-line titles, but no check is made to avoid appending the space to the initial `""` string.

I modified test_PDB to remove the initial space from the comparison lines in the header checks, and added a `test_parse_title()` check in test_PDB_parse_pdb_header which confirms that the three line title from 1LCD.pdb (already in Tests/PDB) is read in correctly.

This work also had the side effect of applying black formatting to the two test_PDB files mentioned.

### DIFF
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -214,7 +214,7 @@ def _parse_pdb_header_list(header):
         # From here, all the keys from the header are being parsed
         if key == "TITLE":
             name = _chop_end_codes(tail).lower()
-            if "name" in dict:
+            if "name" in dict and "" != dict["name"]:
                 dict["name"] += " " + name
             else:
                 dict["name"] = name
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -180,7 +180,7 @@ def _parse_remark_465(line):
 
 def _parse_pdb_header_list(header):
     # database fields
-    dict = {
+    pdbh_dict = {
         "name": "",
         "head": "",
         "idcode": "",
@@ -197,8 +197,8 @@ def _parse_pdb_header_list(header):
         "missing_residues": [],
     }
 
-    dict["structure_reference"] = _get_references(header)
-    dict["journal_reference"] = _get_journal(header)
+    pdbh_dict["structure_reference"] = _get_references(header)
+    pdbh_dict["journal_reference"] = _get_journal(header)
     comp_molid = "1"
     last_comp_key = "misc"
     last_src_key = "misc"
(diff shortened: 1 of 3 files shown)


peterjc:
Have you run black on the test code as part of this? That would explain the large diff which makes reviewing your actual changes a little harder.

Tests/test_PDB.py
```diff
@@ -741,89 +820,96 @@ def test_strict(self):
...
+            "O C O C N C C O C N C C O C S N C C O C C C O N "
+            "N C C O O O O O O O O O O O O O O O O O O O O O "
+            "O O O O O O O O O O O O O O O O O O O O O O O O "
+            "O O O O O O O O O O O O O O O O O O O O O O O O "
+            "O O O O O O O O O O O O O O O O O O O O O",
+        )
 
     def test_model_numbering(self):
         """Preserve model serial numbers during I/O."""
+
```
peterjc: This blank line is a problem:

```
$ flake8
./Tests/test_PDB.py:911:1: D202 No blank lines allowed after function docstring
```

For now we've putting a comment line here to avoid the ``D202`` versus ``black`` disagreement.

rob-miller:
> Have you run black on the test code as part of this? That would explain the large diff which makes reviewing your actual changes a little harder.

Yes, sorry - I've set black as the formatter for my development and thought this would be a net positive.  Also sorry I overlooked the flake8 issue, it was highlighted at my end.

I've fixed the D202 problem, would you prefer I remove the black formatting from the test_ files or have your already gone through them now?  The only lines changed outside the black formatting in test_PDB are 158 and 176:

`"name": "hiv capsid c-terminal domain",`
`"name": "3d structure of alzheimer's abeta(1-42) fibrils",`

rob-miller:
I've updated this PR to separate out the black formatting commits from the actual code changes.

Bio/PDB/parse_pdb_header.py
```diff
@@ -214,7 +214,7 @@ def _parse_pdb_header_list(header):
         # From here, all the keys from the header are being parsed
         if key == "TITLE":
             name = _chop_end_codes(tail).lower()
```
JoaoRodrigues: Ignoring the fact that we are inheriting a `dict`-named dictionary... I'd make use of the `dict.get` method here:
```
dict_name = dict.get('name', '')  # returns empty string if not found
dict_name = ' '.join([dict_name, name]).strip()  # faster than string concatenation
```

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -55,6 +87,16 @@ def test_parse_header_line(self):
         self.assertEqual(header["idcode"], "3EFG")
         self.assertEqual(header["deposition_date"], "2008-09-08")
 
+    def test_parse_title_line(self):
```
JoaoRodrigues: I'd add a second test with an empty `header['name']` just to make sure it still returns something sensible.

Bio/PDB/parse_pdb_header.py
rob-miller: > Ignoring the fact that we are inheriting a `dict`-named dictionary... 

As we are in agreement here I have renamed to pdbh_dict

> I'd make use of the `dict.get` method here:
> 
> ```
> dict_name = dict.get('name', '') + ' ' + name
> dict_name = dict_name.strip()
> ```

just pushed a modified version incorporating this - with addition that pdbh_dict{"name"] is defined at the beginning (legacy) to be "" so can just do the (.join()).strip().

Tests/test_PDB_parse_pdb_header.py
rob-miller: done - I think this is your intent, unless you mean adding an additional test file with 'TITLE' in the header but nothing after it.

rob-miller:
On the formatting, that's the result of black added to my development pipeline (so it runs every time I write a file out).   I could take it out but I think the goal is to black-ify all the code anyway?  My effort at separating the formatting out was to put it as separate commits at the beginning so you can see the effect separately.  I admit I have not scrutinized black results at all :-)

peterjc:
@rob-miller would submitting a separate pull request just running black on ``Bio/PDB/`` first make life easier for you (#2008)? This PR should then merge and/or rebase nicely.

rob-miller:
Thanks @peterjc, #2264 blackened Bio/PDB - these changes are in Tests/.  I can't find the comment I thought I saw about why not to apply black to Tests/ as you suggested initially, but went with the 'black changes in one commit' to try to separate them out here.

peterjc:
@JoaoRodrigues are you able to merge pull requests? If not, do you want to be able to?

peterjc:
Thanks Rob, merged.

JoaoRodrigues:
Hi @peterjc , I am able to, but I wasn't that confident about it since I wasn't active for a long time. If that's OK with you, whenever we all agree that some contribution is good enough, I can merge it if it's assigned to me.

Thanks for merging and thanks @rob-miller for the contribution and the patience! Happy new year to all of you.

peterjc:
@JoaoRodrigues that should be fine - better than us waiting for each other ;)

PR #2366, opened 2019-11-29
Apply black to the Scripts/ directory

peterjc:
This pull request addresses issue #2008.

@svalqui would you like to review this?

### DIFF
--- a/Scripts/debug/debug_blast_parser.py
+++ b/Scripts/debug/debug_blast_parser.py
@@ -84,7 +84,7 @@ def choose_parser(outfile):
     data = open(outfile).read()
     ldata = data.lower()
     if "<html>" in ldata or "<pre>" in ldata:
-        raise NotImplementedError("Biopython no longer has an HTML BLAST " "parser.")
+        raise NotImplementedError("Biopython no longer has an HTML BLAST parser.")
     if "results from round)" in ldata or "converged!" in ldata:
         return NCBIStandalone.PSIBlastParser
     return NCBIStandalone.BlastParser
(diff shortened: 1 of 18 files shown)


peterjc:
Grr. I need to switch back to Python 2 locally to work out what I broke under flake8...

peterjc:
My attempts to modernise this line failed, flake8 under Python 2 was not happy (unsure why, perhaps a plugin fails):

https://github.com/biopython/biopython/blob/biopython-175/Scripts/Restriction/ranacompiler.py#L440

Scripts/debug/debug_blast_parser.py
```diff
@@ -41,7 +42,9 @@
 
 -o    <testfile> is a BLAST output file.
 
-""" % sys.argv[0]
+"""
```
svalqui: This closing triple quote doesn't align with the opening one. not sure if we want them aligned as for instance the parentheses do not align either, black style does that; I have mostly aligned DocStrings but assignments I might have left them like this.

Scripts/xbbtools/nextorf.py
```diff
@@ -171,17 +178,18 @@ def GetCDS(self, seq, strand=1):
                 elif codon_type == STOP:
                     if start_site == 0:
                         continue
-#                    if codon == 'XXX': print('do something')
+                    #                    if codon == 'XXX': print('do something')
                     stop = pos + 2
-#                    print("stop")
+                    #                    print("stop")
                     length = stop - start_site + 1
```
svalqui: I think the comments should be closer to the hash as "# comment..."

Scripts/debug/debug_blast_parser.py
peterjc: I haven't explored the rules of the syntax to properly understand how Python interprets the indentation vs which leading spaces are part of the string values.  However, this does look like a style judgement - in which case do whatever black does is the easy solution.  This is worth discussing for #2008 as a whole.

Scripts/xbbtools/nextorf.py
peterjc: This does look odd - and yes, worth a manual fix.

I think black assumed the lines would be: indentation, hash, space, code - but here they were hash, indentation, code.

peterjc:
Thank you @svalqui for looking at this 👍

PR #2373, opened 2019-12-03
Apply Black to Doc/examples/ & start checking in CI

peterjc:
This pull request addresses issue #2008 in part.


peterjc:
I had a silly typo in the tox configuration, but also some of the other files need black run on them...

peterjc:
Looks like I didn't have black version 19.10b0 installed locally, but the previous release. This introduces a couple of further changes - something to beware of in future.

BioSQL/BioSeq.py
```diff
@@ -191,7 +191,7 @@ def _retrieve_seq_len(adaptor, primary_id):
     if not seqs:
         return None
     assert len(seqs) == 1
-    given_length, = seqs[0]
+    (given_length,) = seqs[0]
```
MarkusPiotrowski: Why using a one-element tuple here? (If I understand this correctly).
Can't we use just `return int(seqs[0])`?

Doc/examples/Proux_et_al_2002_Figure_6.py
```diff
@@ -31,23 +31,55 @@
...
+)
+C_colors = (
+    [grey] * 33
+    + [green] * 5
+    + [brown] * 4
+    + [blue] * 2
+    + [grey, blue]
+    + [lightblue] * 2
+    + [grey] * 8
+)
```
MarkusPiotrowski: Personally, I would use the parentheses to get rid of the line continuation backshlashes, but would turn black off here. On my 17'' laptop, these three lists fill the whole screen.

Scripts/debug/debug_blast_parser.py
```diff
@@ -246,7 +246,7 @@ def test_blast_output(outfile):
     if len(args) != 1:
         sys.stderr.write(USAGE)
         sys.exit(-1)
-    TESTFILE, = args
+    (TESTFILE,) = args
```
MarkusPiotrowski: `TESTFILE = args[0]`?

BioSQL/BioSeq.py
peterjc: I don't recall why the code was written that way in the first place, but it was me: 877a5f2dee468a160549f76b5fd91d65eea7ba29

There must be something unintuitive about the DB results object, the obvious fix fails:

```
TypeError: int() argument must be a string, a bytes-like object or a number, not 'tuple'
```

Doc/examples/Proux_et_al_2002_Figure_6.py
peterjc: I agree on parentheses over slash line continuation, but actually don't mind black taking a screen's worth of vertical space for defining these lists - they are a key part of the graphical example.

Scripts/debug/debug_blast_parser.py
peterjc: This dates back to the original check in, bbe8f5c94839aeed927ab65afca485240c33d304 - but I agree with the suggestion.

peterjc: However, the script no longer works - ``Bio.ParserSupport`` was removed in Release 1.72, while ``Bio.Blast.NCBIStandalone`` was moved away to ``Bio.SearchIO._legacy`` as an interim measure. We can just remove the entire script.

BioSQL/BioSeq.py
peterjc: There are more of these in ``Bio/File.py``, see e.g.

https://github.com/biopython/biopython/blob/biopython-175/Bio/File.py#L585

It seems to be an (anti) pattern with DB access.

MarkusPiotrowski: OK, now I got it, the contents of `seqs[0]` is a one-element tuple.

MarkusPiotrowski: Actually, I was irritated in the first place that tuple unpacking now also requires parenthesis. I don't like it, but OK.

peterjc: Ah - I was more bothered on why there was tuple unpacking (I wrote the code years ago, but this was non-obvious). I'm on the fence about the recent black style change - adding the brackets makes it difficult to overlook the tuple unpacking - previously that trailing comma was easy to miss.

peterjc:
Thanks Markus.

peterjc:
Thanks Markus.

PR #2393, opened 2019-12-09
Apply black style to Bio.Nexus

svalqui:

This is a bit of a change it would be good to review it.

### DIFF
--- a/Bio/Nexus/StandardData.py
+++ b/Bio/Nexus/StandardData.py
@@ -74,11 +74,8 @@ class StandardData(object):
                     continue
                 elif coding in [")", "}"]:
                     raise NexusError(
-                        'Improper character "'
-                        + coding
-                        + '" at position '
-                        + pos
-                        + " of a coding sequence."
+                        "Improper character %s at position %i of a coding sequence."
+                        % (coding, pos)
                     )
                 else:
                     coding_list["d"].append(coding)
--- a/Bio/Nexus/StandardData.py
+++ b/Bio/Nexus/StandardData.py
@@ -35,7 +35,9 @@ class StandardData(object):
 
         # Enforce string data requirement
         if not isinstance(data, basestring):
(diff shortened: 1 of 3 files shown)


Bio/Nexus/StandardData.py
```diff
@@ -71,9 +73,13 @@ def __init__(self, data):
...
-                    raise NexusError('Improper character "' + coding +
-                                     '" at position ' + pos +
-                                     " of a coding sequence.")
+                    raise NexusError(
+                        'Improper character "'
+                        + coding
+                        + '" at position '
+                        + pos
+                        + " of a coding sequence."
+                    )
```
peterjc: Not a black issue, but this could be rewritten to avoid the string concatenation, e.g. use ``%s`` formatting.

Bio/Nexus/Trees.py
```diff
@@ -392,24 +414,54 @@ def is_compatible(self, tree2, threshold, strict=True):
...
+        ]
+        t2 = [
+            (set(tree2.get_taxa(n)), tree2.node(n).data.support)
+            for n in tree2.all_ids()
+            if tree2.node(n).succ
+            and (
+                tree2.node(n).data
+                and tree2.node(n).data.support
+                and tree2.node(n).data.support >= threshold
+            )
```
peterjc: Would this look better without the redundant brackets? ``A and (B and C and D)`` is equivalent to ``A and B and C and D`` isn't it?

svalqui: Both amended, Thank you.

Bio/Nexus/StandardData.py
peterjc: The exact output of the message has changed (missing quotes around the problem character), but I don't think that is critical.

Bio/Nexus/Trees.py
peterjc: That does look better :)

PR #2399, opened 2019-12-14
Internal coords submit

rob-miller:
This pull request offers infrastructure to work with internal coordinates of protein structures: dihedral angles, bond angles and bond lengths. The transform is bidirectional, in that PDB file coordinates for a complete structure can be regenerated from the computed internal coordinates.

See #2346 for background

- [] I have added my name to the alphabetical contributors listings in the files

### DIFF
--- a/Bio/PDB/PICIO.py
+++ b/Bio/PDB/PICIO.py
@@ -243,7 +243,7 @@ def read_PIC(file: TextIO, verbose: bool = False) -> Structure:
         chnp = chn.internal_coord = IC_Chain(chn)
         # done in IC_Chain init : chnp.set_residues()
         chnp.link_residues()
-        chnp.render_dihedra()
+        chnp.init_edra()
 
     # print(report_PIC(struct_builder.get_structure()))
     return struct
--- a/Bio/PDB/PICIO.py
+++ b/Bio/PDB/PICIO.py
@@ -281,7 +281,6 @@ def _enumerate_entity_atoms(entity):
                 for atm in res.get_unpacked_list():
                     atm.set_serial_number(anum)
                     anum += 1
-    pass
 
 
 def enumerate_atoms(entity):
--- a/Bio/PDB/PICIO.py
+++ b/Bio/PDB/PICIO.py
@@ -46,6 +46,17 @@ def read_PIC(file: TextIO, verbose: bool = False) -> Structure:
 
     An improvement would define relative positions for HOH (water) entries.
 
+    N.B. dihedron (i-1)C-N-CA-CB is ignored in assembly if O exists.
+
+    C-beta is by default placed using O-C-CA-CB, but O is missing
+    in some PDB file residues, which means the sidechain cannot be
+    placed.  The alternate CB path (i-1)C-N-CA-CB is provided to
+    circumvent this, but if this is needed then it must be adjusted in
+    conjunction with PHI ((i-1)C-N-CA-C) as they overlap.  (i-1)C-N-CA-CB is
+    included by default in .pic files for consistency and informational
+    (e.g. statistics gathering) purposes, as otherwise the dihedron would only
+    appear in the few cases it is needed for.
+
     :param Bio.File file: file name or handle
     :param bool verbose: complain when lines not as expected
     :returns: Biopython Structure object, Residues with .internal_coord attributes
--- a/Bio/PDB/internal_coords.py
+++ b/Bio/PDB/internal_coords.py
@@ -307,6 +307,7 @@ class IC_Chain:
         # ndx = 0
         last_res: List["IC_Residue"] = []
         last_ord_res: List["IC_Residue"] = []
+
         for res in self.chain.get_residues():
             # select only not hetero or accepted hetero
             if res.id[0] == " " or res.id[0] in IC_Residue.accept_resnames:
@@ -356,6 +357,9 @@ class IC_Chain:
             process
 
         """
+        for ric in self.ordered_aa_ic_list:
+            ric.clear_transforms()
+
         for ric in self.ordered_aa_ic_list:
             if not hasattr(ric, "NCaCKey"):
                 if verbose:
@@ -469,7 +473,7 @@ class IC_Chain:
 
     @staticmethod
     def _writeSCAD_dihed(
-        fp: TextIO, d: "Dihedron", transformations, hedraNdx: Dict, hedraSet: Set[EKT]
+        fp: TextIO, d: "Dihedron", hedraNdx: Dict, hedraSet: Set[EKT]
     ) -> None:
         fp.write(
             "[ {:9.5f}, {}, {}, {}, ".format(
@@ -487,8 +491,7 @@ class IC_Chain:
             )
         )
         fp.write("        ")
-        mtx = transformations[d.id3]
-        IC_Chain._write_mtx(fp, mtx)
+        IC_Chain._write_mtx(fp, d.rcst)
         fp.write(" ]")  # close residue array of dihedra entry
 
     def write_SCAD(self, fp: TextIO, backboneOnly: bool) -> None:
@@ -566,7 +569,8 @@ class IC_Chain:
             )
             ndx += 1
             # assemble with no start position, return transform matrices
-            transformations = ric.assemble(transforms=True, resetLocation=True)
+            ric.clear_transforms()
+            ric.assemble(resetLocation=True)
             ndx2 = 0
             started = False
             for i in range(1 if backboneOnly else 2):
@@ -580,8 +584,17 @@ class IC_Chain:
                     if d.h2key in hedraNdx and (
                         (i == 0 and d.is_backbone()) or (i == 1 and not d.is_backbone())
                     ):
-                        if started:
-                            fp.write(",\n")
+                        if d.rcst is not None:
+                            if started:
+                                fp.write(",\n")
+                            else:
+                                started = True
+                            fp.write("      ")
+                            IC_Chain._writeSCAD_dihed(fp, d, hedraNdx, hedraSet)
+                            dihedraNdx[dk] = ndx2
+                            hedraSet.add(d.h1key)
+                            hedraSet.add(d.h2key)
+                            ndx2 += 1
                         else:
                             print(
                                 f"Atom missing for {d.id3}-{d.id32}, OpenSCAD chain may be discontiguous"
@@ -803,7 +816,7 @@ class IC_Residue(object):
     -------
     applyMtx()
         multiply all IC_Residue atom_cords by passed matrix
-    assemble(atomCoordsIn, transforms, verbose)
+    assemble(atomCoordsIn, resetLocation, verbose)
         Compute atom coordinates for this residue from internal coordinates
     atm241(coord)
         Convert 1x3 cartesian coords to 4x1 homogeneous coords
@@ -1236,27 +1249,27 @@ class IC_Residue(object):
                 startPos = cast(Dict["AtomKey", numpy.array], sp)
 
         if startPos == {}:
-            # fallback: use N-CA-C initial coords from creating dihedral
-            startPos = {}
-            dlist0 = [self.id3_dh_index[akl2] for akl2 in sorted(self.NCaCKey)]
-            # https://stackoverflow.com/questions/11264684/flatten-list-of-lists
-            dlist = [val for sublist in dlist0 for val in sublist]
-            # dlist = self.id3_dh_index[NCaCKey]
-            for d in dlist:
-                for i, a in enumerate(d.aks):
-                    startPos[a] = d.initial_coords[i]
+            startPos = self.default_startpos()
+
         return startPos
 
+    def clear_transforms(self):
+        """Set cst and rcst attributes to none before assemble()."""
+        for d in self.dihedra.values():
+            d.cst = None
+            d.rcst = None
+
     def assemble(
-        self,
-        transforms: bool = False,
-        resetLocation: bool = False,
-        verbose: bool = False,
+        self, resetLocation: bool = False, verbose: bool = False,
     ) -> Union[Dict["AtomKey", numpy.array], Dict[HKT, numpy.array], None]:
         """Compute atom coordinates for this residue from internal coordinates.
 
         Join dihedrons starting from N-CA-C and N-CA-CB hedrons, computing protein
         space coordinates for backbone and sidechain atoms
+
+        Sets forward and reverse transforms on each Dihedron to convert from
+        protein coordinates to dihedron space coordinates for first three
+        atoms (see coord_space())
 
         **Algorithm**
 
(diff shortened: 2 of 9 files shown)


JoaoRodrigues:
Thanks a __lot__ for the contribution @rob-miller . I will be traveling in the next few days but will happily have a look at this whenever I have time over the next couple of weeks. Hope that's OK with you!

Bio/PDB/internal_coords.py
```diff
@@ -0,0 +1,2765 @@
...
+        """Get list of hedra containing specified atom pair.
+
+        :param ak_spec:
+            - tuple of two AtomKeys
+            - string: two atom names separated by ':', e.g. 'N:CA' with
+              optional position specifier relative to self, e.g. '-1C:N' for
+              preceding peptide bond.
+
+        """
+
```
JoaoRodrigues: Maybe the `__repr__` method would be a better choice? I cannot print a dihedron:
```
  File "average_torsions.py", line 24, in <module>
    print(dihe)
TypeError: __repr__ returned non-string (type tuple)
```

Bio/PDB/internal_coords.py
```diff
@@ -0,0 +1,2765 @@
...
+        Set bond length in all relevant hedra for specified pair
+    write_PIC(pdbid, chainId, s)
+        Generate PIC format strings for this residue
+
+    """
+
+    # add 3-letter residue name here for non-standard residues with
+    # normal backbone.  CYG for test case 4LGY (1305 residue contiguous
+    # chain)
+    accept_resnames = ("CYG", "YCM", "UNK")
```
JoaoRodrigues: Maybe have this as a parameter the user can pass, otherwise throw an error or ignore unknown residues?

JoaoRodrigues:
Hi @rob-miller,

This is definitely a lot of work and a great contribution. I pulled this branch to test locally on some of my files and have a couple of issues that I think would be worth addressing before merging. I'll leave some comments directly in the code, and other more general ones here. I can update this comment as we go through the code.

- [ ] `model.atom_to_internal_coordinates()` should return a friendlier error when it cannot handle a particular residue. In my case, I had ACE/NME caps on my protein and the code failed with the following message: `AttributeError: 'IC_Residue' object has no attribute 'NCaCKey'`.
- [] Maybe I couldn't find it but I think it would be valuable to be able to get a list of all the angles you can edit (by default), for a given residue. Essentially, a list of values we could pass to `get_angle` by default.
- [] I can't seem to change dihedrals using `set_angle`. The snippet below does not seem to update the cartesian coordinates. Am I doing something wrong?

```python

# imports ...

s = parser.get_structure('1ctf', '1ctf.pdb')  # PDB1CTF
s.atom_to_internal_coordinates()

for r in s.get_residues():
    ric = r.internal_coord
    if ric and ric.rprev:
        chi1_val = ric.get_angle('chi1')
        ric.set_angle('chi1', chi1_val + 90)
        break

s.internal_to_atom_coordinates()

io.set_structure(s)
io.save('flipped.pdb')  # exactly the same coordinates as the original file.
```

Bio/PDB/internal_coords.py
rob-miller: Could not work out the code you were commenting on, but changed __str__s to __repr__s which does seem a better choice.

Bio/PDB/internal_coords.py
rob-miller: This one I have not addressed, easy to do but want to understand more about the reasoning.  

I wanted it somewhat hard to access without thought, as plugging any HETATM is likely to break things, and what's *really* needed for a new residue is to go and edit the ic_data tables to put in a set of di/hedra to build the thing.

I agree the list is pretty arbitrary - UNK seems reasonable because I think that will just be backbone but could not resolve the sidechain (and for some reason the sequence is unknown?), while CYG and YCM are ones I made the effort to check fit into the peptide backbone and they helped me get bigger proteins going through the cullpdb lists.  Clearly though I did not make the effort to go and work out the angle sets.

Without the entry you just get a chain break and extra noise about atoms that don't match doing the rebuild test.  See your ACE/NME caps with current code.

rob-miller:
Ok, a bit more work done and [gist](https://gist.github.com/rob-miller/4da6a4bab5fa91635f5b748d4e60e875) updated.
In addition to issues you raised, I did a lot of work type hinting in the hopes that I could get cython to pick up the info and speed it up better, but to no avail.  Nonetheless it gets through mypy now (pytype takes too long on it), and the process did identify three bugs I think.
There was some general cleanup on the API - hid the allBonds stuff because no one cares, more verbosity about structure issues if you ask for it.  Added code to fill in partial residues at a chain terminus, that seems to have helped many more cases pass the rebuild test.

> model.atom_to_internal_coordinates() should return a friendlier error when it cannot handle a particular residue. In my case, I had ACE/NME caps on my protein and the code failed with the following message: AttributeError: 'IC_Residue' object has no attribute 'NCaCKey'.

Done, in the particular case of the sample you provided:  (1) There was an issue having a filename with non-alphanumeric ('_') that propagated in due to not having a pdbid either, now handled; and (2) don't know why I had not come across a test file with modified C-terminal ('twas the NME that broke it), also now handled.

>Maybe I couldn't find it but I think it would be valuable to be able to get a list of all the angles you can edit (by default), for a given residue. Essentially, a list of values we could pass to get_angle by default.

Have a look at the improved pydoc header for IC_Residue.pick_angle(), I think that should cover it.  Probably this could be placed somewhere else more prominent as well?  Basically you have psi, phi, omega, chiN and tau (N-Ca-C), otherwise it is specifying the sequence of bonded atoms you're interested in and some are not available due to the defined construction process - and to get this, you have to look through ic_data.py (I think more than is wanted in documentation)

> I can't seem to change dihedrals using set_angle. The snippet below does not seem to update the cartesian coordinates. Am I doing something wrong?

Ah, sorry, I forgot to finish implementing the 'atoms_updated' flag, done that now.  To make this work before you needed to use IC_duplicate() to generate a structure with internal coords and no atoms, and then run internal_coords_to_atoms().  This is now fixed so your code above should work, however it is not as efficient as it should be - changing one chi angle causes reassembling everything, and obviously it should not.  I can address this later as it should not any APIs ore external code.  See the gist or the test_model_change_internal_coords() unittest I added for code, note that you can't get back the exact angle you asked for because it passes through the 3 decimal place precision PDB ATOM coordinates.

Sorry I didn't squash down the new commits: expect no one cares, this helps me to find what broke something if I find a problem, and and someone will just squash it all anyway at the end.

peterjc:
We're generally running the latest version of flake8 and its plugins on TravisCI, which occasionally causes the continuous integration to fail when a plugin adds a new warning or similar. In this case the messages are all about docstring formatting:

```
$ flake8
./Bio/PDB/internal_coords.py:1935:1: RST301 Unexpected indentation.
./Bio/PDB/internal_coords.py:1941:1: RST201 Block quote ends without a blank line; unexpected unindent.
./Bio/PDB/internal_coords.py:1943:1: RST301 Unexpected indentation.
./Bio/PDB/internal_coords.py:1949:1: RST201 Block quote ends without a blank line; unexpected unindent.
./Bio/PDB/internal_coords.py:2050:1: RST301 Unexpected indentation.
./Bio/PDB/internal_coords.py:2052:1: RST201 Block quote ends without a blank line; unexpected unindent.
./Bio/PDB/internal_coords.py:2054:1: RST301 Unexpected indentation.
./Bio/PDB/internal_coords.py:2058:1: RST201 Block quote ends without a blank line; unexpected unindent.
```

Bio/PDB/internal_coords.py
```diff
@@ -1933,23 +1932,23 @@ def pick_angle(
         The following are equivalent (except for sidechains with non-carbon
         atoms for chi2):
```
peterjc: This probably needs a double colon if you want the next block to be an RST literal block.

Bio/PDB/internal_coords.py
```diff
@@ -1933,23 +1932,23 @@ def pick_angle(
...
+            r,
+            ric.get_angle("N:CA:C:1N"),
+            ric.get_angle("-1C:N:CA:C"),
+            ric.get_angle("-1CA:-1C:N:CA"),
+            ric.get_angle("N:CA:C"),
+            ric.get_angle("CA:CB:CG:CD"),
+        )
 
         See ic_data.py for detail of atoms in the enumerated sidechain angles
         and the backbone angles which do not span the peptide bond. Using 's'
```
peterjc: I think this indentation was wrongly removed - RST would have used it to define a block, a literal block if the preceding paragraph ended with a double colon. Certainly the flake8 complaints fall here.

Bio/PDB/internal_coords.py
```diff
@@ -1933,23 +1932,23 @@ def pick_angle(
...
-                    ric.get_angle("chi2"),
-                )
-                print(
-                    r,
-                    ric.get_angle("N:CA:C:1N"),
-                    ric.get_angle("-1C:N:CA:C"),
-                    ric.get_angle("-1CA:-1C:N:CA"),
-                    ric.get_angle("N:CA:C"),
-                    ric.get_angle("CA:CB:CG:CD"),
-                )
```
peterjc: I think this indentation was wrongly removed - RST would have used it to define a block, a literal block if the preceding paragraph ended with a double colon. Certainly the flake8 complaints fall here.

Bio/PDB/internal_coords.py
rob-miller: Thanks, fixed and I knew I needed to check through the html for the API docs :-)

What's bothering me is that I'm not seeing the same flake8 issues:

    rob@water:03 19:06:51 biopython $ flake8 --version
    3.7.9 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.7.6 on Darwin
    rob@water:03 19:07:00 biopython $ flake8 Bio/PDB/internal_coords.py 
    rob@water:03 19:07:05 biopython $ 

I went through the list for the travis output of 'style installdeps' and ensured I had them all.

peterjc: You are missing ``flake8-rst-docstrings`` which gives the ``RST`` prefix codes (the point of this one is to catch things which will upset Sphinx and the HTML API docs), and a bunch more, see:

https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

rob-miller:
I'm currently working more on the typing to make this code work better with cython

rob-miller:
still working on speed-ups

rob-miller:
Once the latest push gets through the build tests this should be OK for inclusion in 1.77 if approved (@JoaoRodrigues?).  Nonetheless I'm currently working on vectorizing the numpy processing and should finish in the next few (less than 4) days; I'm not expecting any API changes from this work.

Bio/PDB/PICIO.py
```diff
@@ -0,0 +1,377 @@
...
+            if 2 == res.is_disordered():
+                for r in res.child_dict.values():
+                    for atm in r.get_unpacked_list():
+                        atm.set_serial_number(anum)
+                        anum += 1
+            else:
+                for atm in res.get_unpacked_list():
+                    atm.set_serial_number(anum)
+                    anum += 1
+    pass
```
peterjc, line 284: Don't need the ``pass`` line?

peterjc:
Reviewing this is outside my expertise - if @JoaoRodrigues is unavailable, perhaps @etal / @lennax / @jgreener64 might be able to comment?

jgreener64:
I won't be able to review this fully but from a quick look it seems okay.

Could you add a paragraph to the tutorial with a couple of basic examples?

If Joao's comments are addressed then I think it's basically ready to be merged.

rob-miller:
The documentation is in #2400 .  There are still two more (at least :-)) optimizations I'm thinking of, but I'm not sure how complicated they will be.  They can be added later if they work.

rob-miller:
> If Joao's comments are addressed then I think it's basically ready to be merged.
The issues he raised have been fixed / addressed.

peterjc:
GitHub says there are merge conflicts, not sure where :(

peterjc:
To clarify - GitHub says there are conflicts blocking a rebase-and-merge, but it looks like a squash-merge as a single commit would work.

If you want to keep the (long) commit history, a manual rebase and update of this branch would be needed.

Do you have a preference on keeping the full commit history here, or not?

rob-miller:
No, the commit history is just for me.  
Sorry, if you can wait one more day I've found a small bug - so can fix it and get the next optimization in.

JoaoRodrigues:
The example code I ran before works now, looking forward to actually using this in production! This is a massive contribution, sorry if my handling of such a big PR wasn't the best.

peterjc:
Seems it still needs a ``NEWS.rst`` entry?

rob-miller:
OK I think that's it.  Just over 15 months since I decided to port to Python.

> Seems it still needs a `NEWS.rst` entry?

I would prefer if @JoaoRodrigues might write that?  I think it's a big deal (but I would, and don't really want to 'blow my own horn'), but I think Joao understands better how it fits into the wider field of relevant software.

JoaoRodrigues:
I can take a stab at it, how does this (very short summary) sound? Do list what you think is the most important!

>A major contribution by Rob Miller to Bio.PDB provides new methods to handle molecular transformations in internal coordinates. Among other uses, these new features simplify the analysis and manipulation of coordinates of large macromolecular structures.

rob-miller:
A major contribution by Rob Miller to Bio.PDB provides new methods to
handle molecular transformations in internal coordinates. Among other uses,
these new features simplify the analysis and manipulation of coordinates of
large macromolecular structures.  'Internal coordinates' are protein
structure phi, psi and omega dihedral angles, and all bond angles and
lengths; the new framework supports lossless interconversion between
internal coordinates and the normal X, Y, Z coordinates in PDB files.

> Do list what you think is the most important!

The wording I added is not significant to me, but the points I want to get
across are:
- 'internal coordinates' means/includes 'phi / psi dihedral angles'
- lossless interconversion (can exactly re-create XYZ coordinates in PDB
file)

r.

JoaoRodrigues:
A major contribution by Rob Miller to Bio.PDB provides new methods to handle protein structure transformations using dihedral angles (internal coordinates). The new framework supports lossless interconversion between internal and cartesian coordinates, which, among other uses, simplifies the analysis and manipulation of coordinates of proteins structures.

Shorter?

rob-miller:
Perfect, I'm happy.
Thanks João!
Rob.

peterjc:
Over to you @JoaoRodrigues 👍

JoaoRodrigues:
Thanks @rob-miller !

@peterjc I can merge this, but we should also merge the documentation at the same time? @rob-miller , is that also up to date and complete? There is a PR at #2400 .

rob-miller:
Yes, just confirming my other message, #2400 is up to date with this work.

JoaoRodrigues:
Sounds good, I will merge this and #2400 then. Thanks again!

PR #2439, opened 2019-12-26
Remove Python 2 only __nonzero__ and __cmp__

peterjc:
Given we have dropped Python 2 support, this pull request follows from #2434 tackling ``__cmp__`` (replaced by rich object comparison methods) and ``__nonzero__`` (essentially replaced by ``__bool__``). Quoting  https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons

> The ``cmp()`` function should be treated as gone, and the ``__cmp__()`` special method is no longer supported. Use ``__lt__()`` for sorting, ``__eq__()`` with ``__hash__()``, and other rich comparisons as needed.

### DIFF
--- a/Tests/test_SearchIO_model.py
+++ b/Tests/test_SearchIO_model.py
@@ -165,7 +165,7 @@ class QueryResultCases(unittest.TestCase):
 
     def test_bool(self):
         """Test QueryResult.__bool__."""
-        # nonzero should return true only if the qresult has hits
+        # should return true only if the qresult has hits
         self.assertTrue(self.qresult)
         blank_qresult = QueryResult()
         self.assertFalse(blank_qresult)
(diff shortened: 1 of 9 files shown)


Tests/test_SearchIO_model.py
```diff
@@ -163,8 +163,8 @@ def test_len(self):
         # len() should return the number of hits contained
         self.assertEqual(3, len(self.qresult))
 
-    def test_nonzero(self):
-        """Test QueryResult.__nonzero__."""
+    def test_bool(self):
+        """Test QueryResult.__bool__."""
         # nonzero should return true only if the qresult has hits
```
bow: The comment looks suitable for an update as well, considering the function is not called `test_nonzero` anymore.

Also, should [a similar test function for the Hit object](https://github.com/peterjc/biopython/blob/drop_cmp/Tests/test_SearchIO_model.py#L814) also be renamed, if this one is?

peterjc: Good catches, I'll update this - thanks!

peterjc:
I think Michiel wanted to review this too, so I won't merge it yet.

PR #2450, opened 2019-12-28
Test pyupgrade on all files

chris-rands:
Just testing running `pyupgrade --py36-plus` to see if the builds fail. This might save a lot of time over individual PRs but it might be too sweeping and unpleasant to review in one go? Would people be in favour of the `pyupgrade` approach?

### DIFF
--- a/Bio/Affy/CelFile.py
+++ b/Bio/Affy/CelFile.py
@@ -218,7 +218,7 @@ def read_v4(f):
     # data.
     def raiseBadHeader(field, expected):
         actual = int(headersMap[field])
-        message = "The header {field} is expected to be 0, not {value}".format(value=actual, field=field)
+        message = f"The header {field} is expected to be 0, not {actual}"
         if actual != expected:
             raise ParserError(message)
 
--- a/Bio/Align/__init__.py
+++ b/Bio/Align/__init__.py
@@ -224,13 +224,13 @@ class MultipleSeqAlignment:
         """
         if record.seq.__class__.__name__ == "CodonSeq":
             if len(record.seq) <= length:
-                return "%s %s" % (record.seq, record.id)
+                return f"{record.seq} {record.id}"
             else:
                 return "%s...%s %s" \
                     % (record.seq[:length - 3], record.seq[-3:], record.id)
         else:
             if len(record.seq) <= length:
-                return "%s %s" % (record.seq, record.id)
+                return f"{record.seq} {record.id}"
             else:
                 return "%s...%s %s" \
                     % (record.seq[:length - 6], record.seq[-3:], record.id)
@@ -1031,7 +1031,7 @@ class PairwiseAlignment:
         aligned_seq1 += seq1[end1:] + " " * (n - n1)
         aligned_seq2 += seq2[end2:] + " " * (n - n2)
         pattern += " " * n
-        return "%s\n%s\n%s\n" % (aligned_seq1, pattern, aligned_seq2)
+        return f"{aligned_seq1}\n{pattern}\n{aligned_seq2}\n"
 
     def _format_generalized(self):
         seq1 = self.target
@@ -1104,7 +1104,7 @@ class PairwiseAlignment:
         aligned_seq1 = " ".join(aligned_seq1)
         aligned_seq2 = " ".join(aligned_seq2)
         pattern = " ".join(pattern)
-        return "%s\n%s\n%s\n" % (aligned_seq1, pattern, aligned_seq2)
+        return f"{aligned_seq1}\n{pattern}\n{aligned_seq2}\n"
 
     def _format_psl(self):
         query = self.query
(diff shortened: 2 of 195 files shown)


bow:
Thanks for trying this out @chris-rands. I would be in favor of at least splitting this into two PRs: one for everything under `Bio` + `BioSQL` and another for everything under `Tests`, in that order.

Seeing the updated code, most of it falls under these categories:
* string formatting (`str.format` and `%`-style formatting replaced with f-strings)
* class definitions without `(object)`
* new-style `super()` statements
* use of `yield from` where appropriate

Aside from the `super()` statements, most of the changes can be evaluated locally by looking at the local context of the statement / expression, which makes them quick to review individually.

Separating the PR for `Tests`, in addition to reducing the load, also ensures that we do not change the tests inadvertently ~ the chance of this happening seems higher with a big, somewhat repetitive changeset ;).

peterjc:
Like Bow, I'd split this up - the ``class Xxxx(object):`` to ``class Xxx:`` for example would make a big but easily reviewed change.

peterjc:
The object change can be done with sed, so maybe start with that?

As to the rest, it like the only way to do this in batches would be to hack https://github.com/asottile/pyupgrade/blob/master/pyupgrade.py to remove lots of fixers...

peterjc:
e.g.

```
$ sed -i.bak -e "s/(object):$/:/g" */*.py Bio/*/*.py Bio/*/*/*.py
```

That seems to work...

chris-rands:
Thanks all- it's possible to limit the string formatting changes of `pyupgrade` via the `--keep-percent-format` flag if preferred?

I can carry on with this since i've started but if we split it up into many PRs it might make good first issues instead?

peterjc:
I suggest you do three pull requests (one of which can be this one updated if you like), in this order:

- objects
- pyupgrade except f-strings
- f-strings

That ought to make 3 manageable ones for review.

peterjc:
Rebased onto master with the object changes via #2453.

Bio/Align/Applications/_ClustalOmega.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
peterjc: Nice that the tool catches there - Python 3 defaults to UTF8, which is great for using unicode directly in string literals.

peterjc:
With the object changes done, I am hopeful that splitting out the f-string changes will leave a manageable next batch 👍

Bio/Align/Applications/_ClustalOmega.py
peterjc: I pull out the five cases of this as #2457 to help break down this monster pull request into more manageable chunks.

peterjc:
Rebasing to current master...

peterjc:
As @bow noted on #2455, some of these exceptions should be ``FileNotFoundError`` which is new in Python 3.

peterjc:
I've ended up doing some ``FileNotFoundError`` fixes in the test on #2469 as part of switching to ``subprocess.getoutput``.

chris-rands:
I propose to close this. After #2480, what is left is the f-strings. If we want to use f-strings instead of `str.format` and decimal formatting etc. across the entire code-base, this will be quite a lot of changes. For example, I ran `flynt` (an alternative to `pyupgrade` focused on f-strings only) on the entire code-case  and it resulted in "255 files changed, 1521 insertions(+), 1672 deletions(-)". So, this might require a module by module approach like the onging efforts with black, if it is worth the errort at all. f-strings are more performant, use fewer characters and can (arguably) be more readable- those are the arguments- thoughts on this?

If we agree to proceed with the goal of adopting f-strings, I will create a new issue and perhaps there will be volunteers to work on it

bow:
> I propose to close this. After #2480, what is left is the f-strings. If we want to use f-strings instead of str.format and decimal formatting etc. across the entire code-base, this will be quite a lot of changes. For example, I ran flynt (an alternative to pyupgrade focused on f-strings only) on the entire code-case and it resulted in "255 files changed, 1521 insertions(+), 1672 deletions(-)". So, this might require a module by module approach like the onging efforts with black, if it is worth the effort at all. f-strings are more performant, use fewer characters and can (arguably) be more readable- those are the arguments- thoughts on this?

> If we agree to proceed with the goal of adopting f-strings, I will create a new issue and perhaps there will be volunteers to work on it

This is a +1 for f-strings from me.

I have been using f-strings almost exclusively these days, falling back to `.format` for some rare cases (mostly when I need a more fine-grained control over when I want to instantiate a template string). In addition to the improved readability, I also find that it can sometimes reduce the need to define variables used only in displayed messages.

peterjc:
OK then, closing as suggested (this is the kind of pull request best merged quickly - thank fully as it is automated, the changes can be regenerated easily).

Let's open a debate about using f-strings - and starting with ``pyupgrade --py36-plus`` would be a sensible first code change.

Issue #2369, opened 2019-12-02
Over-indented multiline triple quote strings


peterjc:
On #2366 and also on some of the application wrappers recently updated to follow the black style (#2008), we had some code assigning a multi-line string defined with triple quotes to a variable (which is different in appearance to triple quotes as used in docstrings). How should these be indented?

This issue is to review all the existing use of multiline triple quote strings as variables (not the special case of docstrings).

# Background

This may look nice, but is not doing what the author might have expected:

```python
def usage():
    tool_name = "example"
    help_text = """This is the first line for help for %s.

    This is the third line (second line blank).
    And 4th line with trailing new line character.
    """ % tool_name
    print(help_text)

usage()
```

Output:

```
$ python /tmp/bad.py
This is the first line for help for example.

    This is the third line (second line blank).
    And 4th line with trailing new line character.
    
```

The following to me looks wrong/ugly, but gives the desired output:

```python
def usage():
    tool_name = "example"
    help_text = """This is the first line for help for %s.

This is the third line (second line blank).
And 4th line with trailing new line character.
""" % tool_name
    print(help_text)

usage()
```

```
$ python /tmp/good.py
This is the first line for help for example.

This is the third line (second line blank).
And 4th line with trailing new line character.
```

To improve on this, @mdehoon has often used the following style which is nice in that it explicitly puts all the lines of the multi-line string flush left (including the first line):

https://github.com/biopython/biopython/blob/biopython-175/Bio/motifs/transfac.py#L208

```python
def write(motifs):
    """Write the representation of a motif in TRANSFAC format."""
    blocks = []
    try:
        version = motifs.version
    except AttributeError:
        pass
    else:
        if version is not None:
            block = """\
VV  %s
XX
//
""" % version
```

Doing this and running black, my example becomes:

```python
def usage():
    tool_name = "example"
    help_text = (
        """\
This is the first line for help for %s.

This is the third line (second line blank).
And 4th line with trailing new line character.
"""
        % tool_name
    )
    print(help_text)


usage()
```

# Application wrappers

For example, https://github.com/biopython/biopython/blob/biopython-175/Bio/Phylo/Applications/_Fasttree.py#L68 (and still currently over-indented on the master):

```python
    def __init__(self, cmd="fasttree", **kwargs):
        """Initialize the class."""
        self.parameters = [
            _Switch(["-nt", "nt"],
                    "By default FastTree expects protein alignments, use -nt for nucleotides",
                    ),
[...]

PR #2498, opened 2020-01-06
Black on several miscellaneous files

peterjc:
This pull request addresses in part issue #2008, targeting small modules without an contentious changes.


peterjc:
@svalqui would you like to review this?

peterjc:
Realise this overlaps with my own older pull request #2466

peterjc:
The changes made on #2466 should be identical, so with that applied, git ought not to find any merge conflicts...

Bio/motifs/__init__.py
```diff
@@ -601,11 +601,14 @@ def format(self, format_spec):
         This method is deprecated; instead of motif.format(format_spec),
         please use format(motif, format_spec).
         """
-        warnings.warn("""\
+        warnings.warn(
+            """\
```
svalqui: I think this slash is not needed.

Bio/Data/IUPACData.py
```diff
@@ -148,39 +164,30 @@ def _make_ranges(mydict):
...
-}
+unambiguous_rna_weights = {"A": 347.2212, "C": 323.1965, "G": 363.2206, "U": 324.1813}
 
 monoisotopic_unambiguous_rna_weights = {
     "A": 347.063084,
     "C": 323.051851,
     "G": 363.057999,
-    "U": 324.035867
+    "U": 324.035867,
 }
```
svalqui: The current style for weights is even, after black monoisotopics don't fit in one line, given that is minor I think we just leave black do the styling.

svalqui:
Only found an slash that could be removed, the rest seems fine to me.

Bio/motifs/__init__.py
peterjc: Actually it is (surprisingly) important if you want to avoid a leading newline in the string, see discussion on #2369.

However, in this case there is no strong reason to use triple quotes - I'll tweak this, so I'm glad you queried this.

Bio/Data/IUPACData.py
peterjc: Black added the trailing comma, which other people recommend anyway in lists and dictionaries as it reduced the diff if you later add more entries.

peterjc: See also https://github.com/PyCQA/flake8-commas/

peterjc:
Thanks you for looking at this so carefully.

PR #2503, opened 2020-01-07
Black 19.10b on misc files

peterjc:
This pull request addresses issue #2008, largely avoiding files already updated with black 19.03b which is slightly different, some more problematic cases (e.g. dictionaries of data), and a few files likely to be updated in separate work.

### DIFF
--- a/Bio/Alphabet/__init__.py
+++ b/Bio/Alphabet/__init__.py
@@ -19,11 +19,13 @@ https://github.com/biopython/biopython/issues/2046
 
 import warnings
 
-warnings.warn("We intend to remove or replace Bio.Alphabet in 2020, "
-              "ideally avoid using it explicitly in your code. Please "
-              "get in touch if you will be adversely affected by this. "
-              "https://github.com/biopython/biopython/issues/2046",
-              PendingDeprecationWarning)
+warnings.warn(
+    "We intend to remove or replace Bio.Alphabet in 2020, "
+    "ideally avoid using it explicitly in your code. Please "
+    "get in touch if you will be adversely affected by this. "
+    "https://github.com/biopython/biopython/issues/2046",
+    PendingDeprecationWarning,
+)
 
 
 class Alphabet:
@@ -39,7 +41,7 @@ class Alphabet:
 
     """
 
-    size = None     # default to no fixed size for words
+    size = None  # default to no fixed size for words
     letters = None  # default to no fixed alphabet
     # In general, a list-like object. However,
     # assuming letters are single characters, use a
@@ -102,7 +104,7 @@ class SingleLetterAlphabet(Alphabet):
     """Generic alphabet with letters of size one."""
 
     size = 1
-    letters = None   # string of all letters in the alphabet
+    letters = None  # string of all letters in the alphabet
 
 
 single_letter_alphabet = SingleLetterAlphabet()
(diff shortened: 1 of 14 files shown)


peterjc:
Strange, GitHub shows TravisCI as pending, but it finished OK (apart from ARM64 which is allowed to fail).

Bio/SeqIO/SeqXmlIO.py
```diff
@@ -262,9 +283,13 @@ def startDBRefElement(self, attrs):
                 elif localname == "id":
                     ID = value
                 else:
-                    raise ValueError("Unexpected attribute '%s' found for DBRef element", key)
+                    raise ValueError(
+                        "Unexpected attribute '%s' found for DBRef element", key
```
MarkusPiotrowski: Shouldn't this be `% key` instead of `, key`?

Bio/SeqIO/SeqXmlIO.py
```diff
@@ -305,9 +336,13 @@ def startPropertyElement(self, attrs):
                 elif localname == "value":
                     property_value = value
                 else:
-                    raise ValueError("Unexpected attribute '%s' found for property element", key)
+                    raise ValueError(
+                        "Unexpected attribute '%s' found for property element", key
```
MarkusPiotrowski: Also here, `% key`?

peterjc: Good catch, yes - a lurking error.

Bio/SeqIO/SeqXmlIO.py
peterjc: Yes, good catch

Bio/Alphabet/__init__.py
```diff
@@ -242,8 +265,9 @@ def contains(self, other):
         hierarchy, and attempts to check the gap character.  This fails
         if the other alphabet does not have a gap character!
         """
-        return (other.gap_char == self.gap_char and
-                self.alphabet.contains(other.alphabet))
+        return other.gap_char == self.gap_char and self.alphabet.contains(
+            other.alphabet
+        )
```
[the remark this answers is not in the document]
peterjc: I agree with you on these examples, but don't care enough to disable black here. We're planning to remove/replace the alphabet code this year anyway, and I especially dislike the ``AlphabetEncoder`` part, so we'll deal with it one way or another.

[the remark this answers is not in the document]
MarkusPiotrowski: I just wondered if there is a simple way to force `black` to line-break between the conditions.
I agree that switching `black` off for these small pieces would be cumbersome.

MarkusPiotrowski:
Strange, appveyor fails with a test that should be fine (it passed on Travis). It seems that  `test_CodonTable.py` was not updated in the appveyor environment? Maybe a problem with simultaneously merging #2530?

peterjc:
Let's try re-starting the AppVeyor in case that sorts itself out...

peterjc:
Good news: AppVeyor is happy now

Bad news: merge conflict, it looks like I got the Codon Table stuff on this branch too somehow. I'll take another look.

peterjc:
Will have to rebase again if #2538 gets merged first as that also changes the BLAST files.

Issue #2512, opened 2020-01-08
Unicode encoding in BGZF


peterjc:
The work and discussion in #2468 leaves the BGZF code in a state where it uses a mix of latin1 (used in the ``Bio._py3k`` conversion functions) and the default encoding.

Properly when used in text mode, the BGZF code should follow the gzip library and take a user specified encoding.

Self-assigning issue.

peterjc:
This turns out to be complicated. Python 2 text mode didn't do universal new lines mode by default, but Python 3 does. This means on examples like ``Tests/SamBam/ex1.bam`` which has Windows line lines we run into complications in the test as the BGZF code does not do universal new lines mode, but gzip.open does in text mode.

```
$ python
Python 3.7.4 (default, Aug 13 2019, 15:17:50) 
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import gzip
>>> with gzip.open("ex1.bam", "rb") as handle: raw_bytes = handle.read()
... 
>>> with gzip.open("ex1.bam", "rt") as handle: open_as_text = handle.read()
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/xxx/miniconda3/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x84 in position 100: invalid start byte
>>> with gzip.open("ex1.bam", "rt", encoding="latin1") as handle: open_as_latin1 = handle.read()
... 
>>> len(raw_bytes)
456614
>>> len(raw_bytes.decode("latin1"))
456614
>>> len(open_as_latin1)
456610
>>> raw_bytes.decode("latin1").replace("\r\n", "\n").replace("\r", "\n") == open_as_latin1
True
>>> 
```

Sadly attempting to implement universal new lines mode in BGZF text mode would be a major piece of work due to the special file offsets (the entire point of BGZF). However, we do need to document this limitation.

peterjc:
Closed by #2517 - sticking with hard coded ``latin1`` due to the nature of the blocked compression.

peterjc:
Almost closed by #2517, I spotted an oversight - fixed in #2532

peterjc:
I should have closed this back in Jan 2020.

It may not be impossible, but it would certainly be non-trivial to implement.

PR #2517, opened 2020-01-09
Clarify BGZF text mode encoding limitations

peterjc:
This pull request addresses issue #2512, and documents a limitation (very reasonably under Python 2, but likely a surprise under Python 3) that BGZF in text mode does not do universal new lines mode.

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -214,6 +214,37 @@ you want to index BGZF compressed sequence files:
 >>> print(record.id)
 NC_000932.1
 
+Text Mode
+---------
+
+Like the standard library gzip.open(...), the BGZF code defaults to opening
+files in binary mode.
+
+You can request the file be opened in text mode, but beware that this is hard
+coded to the simple "latin1" (aka "iso-8859-1") encoding (which includes all
+the ASCII characters), which works well with most Western European languages.
+However, it is not fully compatible with the more widely used UTF-8 encoding.
+
+In variable width encodings like UTF-8, some single characters in the unicode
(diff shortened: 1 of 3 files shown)


peterjc:
This also partly addresses #2490 by fixing that for BGZF.

peterjc:
Thinking more, universal new line mode has a corner case with CR at the end of one compression block, and LF at the start of the next.

Moreover, that same problem likely applies to most unicode encodings - especially variable length encodings like UTF8.

The BGZF code might only be reliable on single-byte encodings like latin1 (which is what it currently is hard coded to use). https://en.wikipedia.org/wiki/Single-byte_encoding

Perhaps I should deprecate the text mode support...

peterjc:
Reading https://docs.python.org/3.6/library/codecs.html I still think only single-byte encodings will work, i.e. ``latin1`` aka ``iso-8859-1`` or perhaps other charmaps like the Windows centric ``cp1252`` which again only defines 256 characters.

peterjc:
@mdehoon could you look at this for comment?

Who else has a good grasp of encodings?

peterjc:
Thanks!

peterjc:
Hah - skimmed it again when closing #2512, and I realise I left the latin1/utf8 mismatch in place. The tests are not broad enough... probably an omission on the writing side.

peterjc:
So right now it insists on latin1 on reading (because we can't deal with multi-byte characters split between blocks), but the code still takes the default encoding on output.

We could probably support any codec on output... but for symmetry putting it back to the previous release's behaviour and using latin1 again is safer.

Issue #2510, opened 2020-01-08
Adopt f-strings


chris-rands:
Proposal for discussion and testing to adopt f-strings, new in Python 3.6 (see python docs and PEP 498), instead of `str.format()` and decimal style string formatting. See this and subsequent comments https://github.com/biopython/biopython/pull/2450#issuecomment-570091366

`pyupgrade --py36-plus` is one tool for automatically converting to f-strings, but after some minor testing, I think that [`flynt`](https://github.com/ikamensh/flynt) may be more comprehensive (and will only work on f-strings). If we proceed, I suggest PRs are submitted one module (or directory) at a time

peterjc:
It still feels a bit alien right now, but for the simple ``"Hello %s, how are you?" % name`` style examples with just ``%s`` default string formatting, ``f"Hello {name}, how are you?"`` is clear and concise. I think I could get used to that pretty quickly.

Also, it finally makes the ``__format__`` methods far more useful (via the colon syntax), e.g.

```python
>>> from Bio import SeqIO
>>> record = SeqIO.read("lavender.nu", "fasta")
>>> print(f"Our record ID {record.id} in FASTA format:\n\n{record:fasta}")
```
```
Our record ID gi|5690369|gb|AF158246.1|AF158246 in FASTA format:

>gi|5690369|gb|AF158246.1|AF158246 Cricetulus griseus glucose phosphate isomerase (GPI) gene, partial intron sequence
GGCTCTTAAGTCATGTCTAGGCAGGTGTGCACAAGTTTAGGTCTTGTTCCGCCAGCATCT
GAGGAGATTGGGTTGGGAACTGGTATGCATGTCACCATCTGAGTTTAGCCCAGATCGGGA
GTCCTCTTCATCAGCATTGAGAAAGTAGAGACTGGCACAGGTCATAATGAGGTGGGTCAA
GCATGGATATGAGGCCCTGGTTCTGCCATCACCGAGACGGCAAAGGCTAGCATAGGTATG
AGACTTCCTTCCTGCCAGTGCTGAGGAACTGGGAGCCTACAGAGGTTTGAGGCCTTGGGT
CTGCCAGAAGAAAGGCCAGCTGCGGTAACATCTTTGACCGTCCCTCACGTAGTCTGCCTT
GCTGGCCAGCATGAAGAGTCACTAGTTCCTGGATCTGCTGTCTGAACTGGGAAGAGTCGG
[...]

PR #2524, opened 2020-01-09
Convert string format method usage to f-strings

peterjc:
This pull request is mainly for discussion/illustration for issue #2510, the question of if we should adopt f-strings.

This shows what it would be like to replace the string format method (the second major string formatting style in Python, used in a minority of Biopython) with f-strings (the third major method added in Python 3.6, not yet used in Biopython), while leaving the C-style percent operator formatting alone (the original string formatting style, used in most of Biopython).

This might be a happy compromise over moving everything to f-strings (as showcased in #2510)? i.e. Allow legacy percent operator _or_ the new f-strings?

This was done with a modified version of flynt v0.38, where I disabled method ``visit_BinOp`` in the class ``flynt.transform.FstringifyTransformer.FstringifyTransformer`` - and would be easy to repeat later if required.

### DIFF
--- a/Bio/Affy/CelFile.py
+++ b/Bio/Affy/CelFile.py
@@ -224,9 +224,7 @@ def read_v4(f):
     # data.
     def raiseBadHeader(field, expected):
         actual = int(headersMap[field])
-        message = "The header {field} is expected to be 0, not {value}".format(
-            value=actual, field=field
-        )
+        message = f"The header {field} is expected to be 0, not {actual}"
         if actual != expected:
             raise ParserError(message)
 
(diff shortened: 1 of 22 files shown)


peterjc:
The discussion on #2524 is on going, and while it is unclear if we will prioritise f-strings exclusively, no one is objecting to using them.

This pull request does not tackle _all_ the format method usage - the rest need manual review (as done for ``Bio/Seq.py`` - run flynt allowing a longer line length and then fixing that with black and/or by hand).

Anyway, I think it is still an improvement, and with two approvals I'm merging this.

PR #2567, opened 2020-01-19
Drop redundant open(file, "rU") `mode` argument

chris-rands:
As noted in https://github.com/biopython/biopython/pull/2564#discussion_r368020173, "U" (universal newlines) mode was deprecated in Python 3.4 and the `newline=None` default argument handles it instead. I did the replacement with `sed`

### DIFF
--- a/Bio/AlignIO/__init__.py
+++ b/Bio/AlignIO/__init__.py
@@ -364,7 +364,7 @@ def parse(handle, format, seq_count=None, alphabet=None):
     if seq_count is not None and not isinstance(seq_count, int):
         raise TypeError("Need integer for seq_count (sequences per alignment)")
 
-    with as_handle(handle, "rU") as fp:
+    with as_handle(handle) as fp:
         # Map the file format to a sequence iterator:
         if format in _FormatToIterator:
             iterator_generator = _FormatToIterator[format]
@@ -471,7 +471,7 @@ def convert(in_file, in_format, out_file, out_format, alphabet=None):
     """
     # TODO - Add optimised versions of important conversions
     # For now just off load the work to SeqIO parse/write
-    with as_handle(in_file, "rU") as in_handle:
+    with as_handle(in_file) as in_handle:
         # Don't open the output file until we've checked the input is OK:
         alignments = parse(in_handle, in_format, None, alphabet)
 
--- a/Bio/Nexus/Nexus.py
+++ b/Bio/Nexus/Nexus.py
@@ -688,7 +688,7 @@ class Nexus:
         # file-like object.
         # Note we need to add parsing of the path to dir/filename
         try:
-            with File.as_handle(input, "rU") as fp:
+            with File.as_handle(input) as fp:
                 file_contents = fp.read()
                 self.filename = getattr(fp, "name", "Unknown_nexus_file")
         except (TypeError, OSError, AttributeError):
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -89,7 +89,7 @@ class PDBParser:
             # Make a StructureBuilder instance (pass id of structure as parameter)
             self.structure_builder.init_structure(id)
 
-            with as_handle(file, mode="rU") as handle:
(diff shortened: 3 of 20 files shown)


peterjc:
With most of these the default mode is "r" anyway, so we can remove the entire mode argument.

The catch is there are corner cases like ``gzip.open(...)`` where the default is ``rb`` and this would break.

Do want to work on that, or do you think we should commit this as is (still an improvement as universal read lines is implicit with Python 3 text mode)?

peterjc:
Actually, many of these are using ``Bio.File.as_handle(..., "r")`` which is an easy case - that does default to ``mode="r"``

chris-rands:
Good point Peter, actually in all these example the `"r"` arg can be omitted I think? I agree it's better (although less explicit) to omit them, so, I could drop these for this PR, but if you want to check across the whole codebase, might be worth a separate PR/issue?

peterjc:
I think you're right - we can drop the mode argument for all the cases already being updated in this pull request 👍

chris-rands:
@peterjc  should I just also remove the other redundant "r" flags here also? It would change another ~35 files. I think it's easy, although I'm not sure on the `bgzf` stuff

`open(file, "r")` >>> `open(file)` ("r" is the default)
`as_handle(handle, "r")` >>> `as_handle(handle)` ("r" is the default)
`gzip.open(file, "r")` >>> `gzip.open(file)` ("r" defaults to "rb", which is the default)
`bgzf.BgzfReader(file, "r"`) >>> `bgzf.BgzfReader(file)` ("r" is the default)

The exception
`bgzf.open(file, "r")` >>> `bgzf.open(file, "r")` (leave it the same, the default is "rb")

peterjc:
I'd leave out the gzip and bgzf ones for now as they're non-trivial (are you sure about that summary - it doesn't match my recollection from memory, but I've not checked the code and docs to confirm).

Anyway, the plain ``open`` and ``as_handle`` cases should be fine - I'd be tempted to split that into a second followup pull request as this one already touches 20 files, which is plenty for reviewing (and ought already to have another set of eyes review it).

chris-rands:
Can we merge this as it is then? Note, that removing "U" will be essential for compatibility with Python 3.9: https://docs.python.org/3.9/whatsnew/3.9.html#changes-in-the-python-api

which part of the summary contradicts your memory?

peterjc:
I thought gzip and bgzf behaved the same (by default)

chris-rands:
`gzip.open` and `bgzf.open` both have the same default `mode`:

```
>>> gzip.open.__defaults__
('rb', 9, None, None, None)
>>> bgzf.open.__defaults__
('rb',)
```

However, I think (correct me if I'm wrong), `gzip.open(file, "r")` is equivalent to `gzip.open(file, "rb")`, to read as text you need `"rt"`: https://github.com/python/cpython/blob/master/Lib/gzip.py#L64

 By contrast, `bgzf.open(file, "r")`, the "r" arg is fed to `BgzfReader`, where "r" is not equivalent to "rb" since the code logic explicitly checks for "b": https://github.com/biopython/biopython/blob/master/Bio/bgzf.py#L577

Therefore, replacing "r" with no argument is okay for `gzip.open` (defaults to binary reading, the default) but not for `bgzf.open` (defaults to text reading, not the default)

peterjc:
You are right - for ``gzip.open`` you must use ``'rt'`` for text, ``'r'`` alone like default ``'rb'`` gives binary.

I personally find that ``gzip.open(file)`` and ``gzip.open(file, "r")`` give binary mode surprising, and would lean to using ``gzip.open(file, "rb")`` for clarity.

Meanwhile for ``bzgf.open`` you can use ``'rt'`` or ``'r'`` for text. I don't recall why, perhaps the git history might remind me - this was likely guided by the Python 2/3 situation at the time, but could just have been an error. Probably I should document the mode more explicitly here, probably too late to align with the gzip behaviour?

Also, I wonder why I put the ``BgzfReader`` and ``BgzfWriter `` class defaults as just ``'r'``, most likely a simple oversight.

PR #2568, opened 2020-01-19
Convert `os.system()` to `subprocess.call()`

chris-rands:
The `subprocess` module is preferred to `os.system` these days, see [PEP 324](https://www.python.org/dev/peps/pep-0324/#rationale) and [the docs](https://docs.python.org/3/library/subprocess.html). Also a couple of minor `black` changes post the previous f-string work.

### DIFF
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -16,7 +16,7 @@ Comput Appl Biosci 1997 , 13:291-295
 ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
 """
 
-import os
+import subprocess
 
 from Bio.PDB.Polypeptide import is_aa
 
@@ -32,7 +32,7 @@ def run_psea(fname):
 
     Note that P-SEA will write output to the terminal while run.
     """
-    os.system("psea " + fname)
+    subprocess.call(["psea", fname])
     last = fname.split("/")[-1]
     base = last.split(".")[0]
     return base + ".sea"
@@ -84,7 +84,7 @@ def annotate(m, ss_seq):
         raise ValueError("Length mismatch %i %i" % (L, len(ss_seq)))
     for i in range(0, L):
         residues[i].xtra["SS_PSEA"] = ss_seq[i]
-    # os.system("rm "+fname)
+    # subprocess.call(["rm", fname])
 
 
 class PSEA:
--- a/Bio/PDB/ResidueDepth.py
+++ b/Bio/PDB/ResidueDepth.py
@@ -54,6 +54,7 @@ in a residue)::
 import os
 import tempfile
 import warnings
+import subprocess
 
 import numpy
 
@@ -499,9 +500,7 @@ def _get_atom_radius(atom, rtype="united"):
     elif resname in {"FAD", "NAD", "AMX", "APU"} and at_name.startswith("H"):
         return _atomic_radii[15][typekey]
     else:
-        warnings.warn(
-            f"{at_name}:{resname} not in radii library.", BiopythonWarning
-        )
+        warnings.warn(f"{at_name}:{resname} not in radii library.", BiopythonWarning)
         return 0.01
 
 
@@ -527,7 +526,7 @@ def get_surface(model, PDB_TO_XYZR=None, MSMS="msms"):
 
     Arguments:
      - PDB_TO_XYZR - deprecated, ignore this.
-     - MSMS - msms executable (used as argument to os.system)
+     - MSMS - msms executable (used as argument to subprocess.call)
 
     """
     # Issue warning if PDB_TO_XYZR is given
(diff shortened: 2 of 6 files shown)


Bio/PDB/ResidueDepth.py
```diff
@@ -547,15 +546,13 @@ def get_surface(model, PDB_TO_XYZR=None, MSMS="msms"):
...
-                f"{x:6.3f}\t{y:6.3f}\t{z:6.3f}\t{radius:1.2f}\n"
-            )
+            pdb_to_xyzr.write(f"{x:6.3f}\t{y:6.3f}\t{z:6.3f}\t{radius:1.2f}\n")
 
     # make surface
     surface_tmp = tempfile.mktemp()
     MSMS = MSMS + " -probe_radius 1.5 -if %s -of %s > " + tempfile.mktemp()
     make_surface = MSMS % (xyz_tmp, surface_tmp)
-    os.system(make_surface)
+    subprocess.call(make_surface, shell=True)
```
peterjc: subprocess.check_call?

Bio/PDB/PSEA.py
```diff
@@ -32,7 +32,7 @@ def run_psea(fname):
 
     Note that P-SEA will write output to the terminal while run.
     """
-    os.system("psea " + fname)
+    subprocess.call(["psea", fname])
```
peterjc: subprocess.check_call?

Bio/Wise/__init__.py
```diff
@@ -105,7 +106,7 @@ def align(
     if debug:
         sys.stderr.write("%s\n" % cmdline_str)
 
-    status = os.system(cmdline_str) >> 8
+    status = subprocess.call(cmdline_str, shell=True) >> 8
```
peterjc: Does subprocess definitely return the same value as os.system? There are weird and wonderful corner cases like minus values for system errors etc, and I am unclear what the ``>> 8`` was for.

peterjc:
The changes look good in principle (see comments about ``subprocess.check_call`` in a couple of places where it looks like it should abort on error), but doing this kind of change without matching command line testing makes me a little nervous - were you able to test any of them locally?

Bio/Wise/__init__.py
chris-rands: I'm not sure. `>> 8` is checking for error codes `>=256?`

peterjc: CC @mdehoon - would a return code outside 0-255 indicate it was the OS not the tool?

chris-rands:
I can change those examples to `subprocess.check_call`, but that will change the behaviour (`os.system` does not lead to an abort)- is that desired?

Otherwise, I expect the behaviour to be the same based on the documentation, e.g.
```
>>> import os
>>> os.system("blastn -version")
blastn: 2.2.31+
Package: blast 2.2.31, build Jun  2 2015 10:47:46
0
>>> import subprocess
>>> subprocess.call(["blastn", "-version"])
blastn: 2.2.31+
Package: blast 2.2.31, build Jun  2 2015 10:47:46
0
```

However, if this feels risky I can strip back the PR to make fewer changes? The changes to `Bio/PDB/DSSP.py` are merely documentation changes, since `subprocess` is already used. The formatting changes are also minor.

I just spotted another change, maybe `python xbb_blast.py` should be `python3 xbb_blast.py` now, since `python` often points to `python2` still

Bio/Wise/__init__.py
chris-rands: if someone can clarify the desired behaviour, perhaps the logic can be written and commented to improve readability, otherwise I suggest maybe we just don't alter this?

peterjc: OK, leave as it is pending clarification of intent.

peterjc:
The examples I flagged to use ``check_call`` all looked to me like they ought to have aborted if the system call failed, but yes - it would be a behaviour change. If you are unsure on any of those, then just switch to ``subprocess.call`` to preserve the existing behaviour more closely?

peterjc:
I'm not sure if switching python to python3 is needed, would need to look at a few recent operating systems to be sure...

JoaoRodrigues:
`python3` is _not_ supported in some OSes (namely Win10). Careful with that!

chris-rands:
Thanks both, I didn't change to `python3` and I left the original `os.system()` call with a comment for the case where the intent is unclear

peterjc:
Do you think we should just stick with ``subprocess.call`` or change some to ``subprocess.check_call``?

chris-rands:
> Do you think we should just stick with `subprocess.call` or change some to `subprocess.check_call`?

I don't have an strong opinion, I'm not so familiar with these parts of the codebase, maybe @JoaoRodrigues has a view?

peterjc:
Well, I think this is an improvement as it stands, and if Joao or others likes the check_call idea, we can do that on a followup pull request. Thank you all, merging....

JoaoRodrigues:
Sorry, this slipped under all the other notifications .. It's fine by me, I just raised the `python3` issue because hardcoding the interpreter call might be problematic in the future. Something to look at later I guess!

peterjc:
Good call - it seems sticking with ``python`` is going to work most of the time, and that will only improve as more systems switch to python 3 by default.

PR #2571, opened 2020-01-21
Apply black to Test/test_PDB*

svalqui:

Applied black style to all test_PDB* files, also switched off black on some section that generated long lists.


Tests/test_PDBList.py
```diff
@@ -24,21 +25,27 @@ class TestPBDListGetList(unittest.TestCase):
 
     def test_get_recent_changes(self):
         """Tests the Bio.PDB.PDBList.get_recent_changes method."""
-        pdblist = PDBList(obsolete_pdb="unimportant")  # obsolete_pdb declared to prevent from creating the "obsolete" directory
+        pdblist = PDBList(
+            obsolete_pdb="unimportant"
+        )  # obsolete_pdb declared to prevent from creating the "obsolete" directory
```
peterjc: Here (and the next few cases), I would put the comment on the line above - then the code does not need to split over three lines.

Tests/test_PDB_MMCIF2Dict.py
```diff
@@ -26,37 +26,305 @@
...
+        # Turn black code style off
+        # fmt: off
+        self.assertEqual(mmcif["_entity_poly_seq.mon_id"], ["MSE", "ASP", "ILE", "ARG",
```
peterjc: Maybe use a more black-like style here, e.g. line break after the ``[`` and put  ``"MSE", "ASP", "ILE", "ARG", "GLN", "GLY", "PRO", "LYS", ...`` on the second line (only four more spaces indented than the self line). Maybe 10 resides per line would fit comfortably?

Tests/test_PDB_MMCIF2Dict.py
```diff
@@ -26,37 +26,305 @@
...
             IAVLYCVHQRIDVKDTKEALDKIEEEQNKSKKKAQQAAADTGNNSQVSQNYPIVQNLQGQMVHQAISPRTLNAWVKVVEE
             KAFSPEVIPMFSALSEGATPQDLNTMLNTVGGHQAAMQMLKETINEEAAEWDRLHPVHAGPIAPGQMREPRGSDIAGTTS
             TLQEQIGWMTHNPPIPVGEIYKRWIILGLNKIVRMYSPTSILDIRQGPKEPFRDYVDRFYKTLRAEQASQEVKNWMTETL
             LVQNANPDCKTILKALGPGATLEEMMTACQGVGGPGHKARVLAEAMSQVTNPATIMIQKGNFRNQRKTVKCFNCGKEGHI
             AKNCRAPRKKGCWKCGKEGHQMKDCTERQANFLGKIWPSHKGRPGNFLQSRPEPTAPPEESFRFGEETTTPSQKQEPIDK
-            ELYPLASLRSLFGSDPSSQ""")])
+            ELYPLASLRSLFGSDPSSQ"""
+                )
+            ],
+        )
```
peterjc: Perhaps implicit string concatenation would be better than tripled quoted with dedent?

Tests/test_PDB_MMCIF2Dict.py
```diff
@@ -26,37 +26,305 @@
...
+                "G",
+                "H",
+                "D'",
+                "E'",
+                "A'\"",
+                "BC",
+                "CD",
+                "DE",
+            ],
+        )
```
peterjc: This is an example where I don't like black's output - but it doesn't seem bad enough to disable black for this method.

Tests/test_PDB_MMCIF2Dict.py
```diff
@@ -90,21 +360,30 @@ def test_verbatim_block(self):
...
             _test_value
             ;First line
                 Second line
             Third line
             ;
-        """)))
+        """
+                )
+            )
+        )
```
peterjc: Maybe try implicit string concatenation with explicit line breaks?

```python
mmcif_dict = MMCIF2Dict(io.StringIO(
    "data_verbatim_test\n"
    "_test_value\n"
    ...
)
```

Tests/test_PDB_ResidueDepth.py
```diff
@@ -45,39 +47,51 @@ def check_msms(self, prot_file, first_100_residues):
...
+            "PDB/2XHE.pdb",
+            "HISMETSERLEULYSSERALAVALLYSTHRVALLEUTH"
+            "RASNSERLEUARGSERVALALAASPGLYGLYASPTRPL"
+            "YSVALLEUVALVALASPLYSPROALALEUARGMETILE"
+            "SERGLUCYSALAARGMETSERGLUILELEUASPLEUGL"
+            "YVALTHRVALVALGLUASPVALSERLYSGLNARGLYSV"
+            "ALLEUPROGLNPHEHISGLYVALTYRPHEILEGLUPRO"
+            "THRGLUGLUASNLEUASPTYRVALILEARGASPPHEAL"
+            "AASPARGTHRPROTHRTYRGLUALAALAHISLEU",
+        )
```
peterjc: Is there any reason for these line breaks at 38 residues? Could use fewer longer lines otherwise.

Tests/test_PDB_KDTree.py
```diff
@@ -73,7 +73,10 @@ def test_KDTree_exceptions(self):
         coords = random((nr_points, 3)) * 100000000000000
         with self.assertRaises(Exception) as context:
             kdt = kdtrees.KDTree(coords, bucket_size)
-        self.assertTrue("coordinate values should lie between -1e6 and 1e6" in str(context.exception))
+        self.assertTrue(
+            "coordinate values should lie between -1e6 and 1e6"
+            in str(context.exception)
+        )
```
peterjc: Not really black, but I think there is an assert contains method we should be using here.

svalqui:
@peterjc  Thank you for the review I think all suggestions are incorporated now.
@JoaoRodrigues Would you have some time to comment on this?

Tests/test_PDB_ResidueDepth.py
peterjc: OK, now breaks at 74 residues - maximum to fit black's line length 88.

(I would have used 60 or 80 if that fit as those are relatively common in FASTA files - but this is fine)

Tests/test_PDB_MMCIF2Dict.py
peterjc: Hmm - I hadn't appreciated that this way the lines of content would need two lines each - if we ignore that rule we'd have 99 character lines.

What do you think, was the dedent version better - even with 92 character lines? Or any other ideas?

(This makes me wonder why does ``pdbx_seq_one_letter_code`` have the embedded line breaks in the first place?)

Tests/test_PDB_MMCIF2Dict.py
peterjc: Either works - I have no strong preference having seen this result after black.

peterjc:
Thanks - that is better.

I'm still not sure if and how the ``pdbx_seq_one_letter_code`` test layout could be improved, but I'm now happy for this to be merged once a PDB person has had a look.

JoaoRodrigues:
Looks good to me, thanks @svalqui !

peterjc:
Thank you both.

PR #2576, opened 2020-01-22
Drop "r" mode arg with open() and as_handle()

chris-rands:
This follows on from #2567 

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -116,7 +116,7 @@ This is an ordinary GenBank file compressed using BGZF, so it can
 be decompressed using gzip,
 
 >>> import gzip
->>> handle = gzip.open("GenBank/NC_000932.gb.bgz", "rb")
+>>> handle = gzip.open("GenBank/NC_000932.gb.bgz", "r")
 >>> assert 0 == handle.tell()
 >>> line = handle.readline()
 >>> assert 80 == handle.tell()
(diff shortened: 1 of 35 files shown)


Bio/bgzf.py
```diff
@@ -217,8 +217,11 @@
 Text Mode
 ---------
 
-Like the standard library gzip.open(...), the BGZF code defaults to opening
-files in binary mode.
+Like the standard library gzip.open(...), the BGZF code, with bgzf.open(...),
+defaults to opening files in binary mode. However, the BgzfReader and
+BgzfWriter classes currently default to "r", intupreted as text mode. Note
```
peterjc: Typo: intupreted

Bio/bgzf.py
```diff
@@ -217,8 +217,11 @@
...
+that for legacy reasons, gzip.open(..., "r") reads in binary mode, while
+bgzf.open(.., "r") reads in text mode.
```
peterjc: Does it really? :sob:

chris-rands: I don't know if it would break things to change the behaviour, making the default always "rb" and interpreting "r" as "rb" (or banning "r" all-together and forcing specifying "rt")

peterjc: This is a bit separate from the other open calls, which is why I was thinking a separate pull request for BGZF would be simpler for reviewing and merging. You could revert the BGZF edits here maybe?

What do you think about a warning in BGZF for just "r" suggesting "rb" or "rt" (better to be explicit; likewise for "w")? And also change any BGZF default of just "r" or "w" to the more explicit form (without actually changing the behaviour)?

chris-rands: Ok I reverted the changes for `bgzf.py`-- that can be handled in a separate PR

peterjc:
With the BGZF bits excluded, I don't think there is anything objectionable in this stylistic change. Other comments would be welcome, but if there are none I propose to merge this at the end of the week.

peterjc:
We've since set s390x to an allowed fail on TravisCI, otherwise the tests are all fine.

peterjc:
Thanks Chris.

peterjc:
I meant to squash merge that for a simpler history (without the BGZF changes and revert), oh well.

PR #2596, opened 2020-01-29
Convert remaining string literal .format to f-strings

peterjc:
This pull request addresses issue #2510, building on #2524 which only did the easiest cases with flynt.

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -445,10 +445,10 @@ class SeqFeature:
 
         if start_offset not in [0, 1, 2]:
             raise ValueError(
-                "The start_offset must be 0, 1, or 2. "
-                "The supplied value is {}. Check the value "
-                "of either the codon_start qualifier or "
-                "the start_offset argument".format(start_offset)
+                f"The start_offset must be 0, 1, or 2. "
+                f"The supplied value is {start_offset}. "
+                "Check the value of either the codon_start qualifier "
+                "or the start_offset argument"
             )
 
         feat_seq = self.extract(parent_sequence)[start_offset:]
@@ -798,8 +798,8 @@ class FeatureLocation:
             and self.start > self.end
         ):
             raise ValueError(
-                "End location ({}) must be greater than or equal "
-                "to start location ({})".format(self.end, self.start)
+                f"End location ({self.end}) must be greater than "
+                f"or equal to start location ({self.start})"
             )
         self.strand = strand
         self.ref = ref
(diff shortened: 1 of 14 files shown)


Bio/motifs/transfac.py
```diff
@@ -193,24 +183,20 @@ def read(handle, strict=True):
...
-                    "should start with a "
-                    '"[": "{0:s}"'.format(index, line)
+                    f'The index "{index}" in a TRANSFAC RN line should start'
+                    f' with a "[": "{line}"'
                 )
             if index[-1] != "]":
                 raise ValueError(
-                    'The index "{0:s}" in a TRANSFAC RN line '
-                    "should end with a "
-                    '"]": "{0:s}"'.format(index, line)
```
peterjc: This example had a bug, ``{0:s}`` instead of ``{1:s}`` - a perfect example of a downside of this syntax.

peterjc:
@chris-rands you might like to review this too?

Bio/SeqFeature.py
```diff
@@ -445,10 +445,10 @@ def translate(
 
         if start_offset not in [0, 1, 2]:
             raise ValueError(
-                "The start_offset must be 0, 1, or 2. "
-                "The supplied value is {}. Check the value "
-                "of either the codon_start qualifier or "
-                "the start_offset argument".format(start_offset)
+                f"The start_offset must be 0, 1, or 2. "
```
chris-rands: `f"` is redundant on this line. Also a minor style point, should it be aligned like this:
```
f"a"
"b"
```
or
```
f"a"
 "b"
```

Bio/__init__.py
```diff
@@ -125,7 +125,7 @@ class BiopythonExperimentalWarning(BiopythonWarning):
         " missing compiled C extensions. We recommend that you"
         " try running your code from outside the source tree."
         " If you are outside the source tree then you have a"
-        " setup.py file in an unexpected directory: {}.".format(_parent_dir),
+        " setup.py file in an unexpected directory: " + _parent_dir,
```
chris-rands: could (optionally) be an f-string

Bio/SeqFeature.py
peterjc: Good catch, thank you - there was a flake8 plugin which caught this kind of pointless f-string, but it had problems with false positives.

The black alignment is the first of the two.

Bio/__init__.py
peterjc: It could - honestly unsure in a case like this which is faster, or which is clearer.

PR #2626, opened 2020-02-04
xbbtools: Remove subprocess call to Python interpreter

MarkusPiotrowski:
This pull request addresses an issue discussed in #2568. It will remove a direct call of the Python interpreter (which may be different on different OS).

### DIFF
--- a/Scripts/xbbtools/xbb_blastbg.py
+++ b/Scripts/xbbtools/xbb_blastbg.py
@@ -15,9 +15,8 @@
 import os
 import tempfile
 import threading
-import subprocess
 
-from tkinter import messagebox
+from tkinter import Tk, messagebox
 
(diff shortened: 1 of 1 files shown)


peterjc:
Do you want to review this @chris-rands ?

chris-rands:
looks good to me,  minor suggestion,  `xbb_blast.py` could define this logic in a function `def test(): win = Tk()` etc., which could be called in `if __name__ == "__main__": test()`, then in `xbb_blastbg.py` you could just call `BlastIt.test()`. This would save duplicating the logic

...but maybe it's not worth it, I've never actually used this, so I don't really feel qualified to review it

peterjc:
Cheers - I've only looked at this GUI tool briefly, will leave that suggestion to Markus to consider.

PR #2681, opened 2020-02-10
fix: padding search in CELv4 files

wwydmanski:
Before this fix, the parser expected \x04 EOT byte before the
padding. However, not all CEL files provide such flag, which causes
reading errors. Replacing this check with a regular search of
15 bytes of zero padding fixes this issue and allows for reading CEL
files regardless of whether they provide EOT or not.


peterjc:
Can you include a small real work example where this is needed as a test case?

Does this new check work fine on files which do have the ``b"\x04"`` EOF marker?

wwydmanski:
@peterjc I've found some data from a [Drug Safety CAMDA challenge](http://camda2018.bioinf.jku.at/doku.php/contest_dataset) that caused problems in the parser.
[CEL example.zip](https://github.com/biopython/biopython/files/4181562/CEL.example.zip)

Also, the checks are failing because of spacing errors in outputs. Can i somehow circumvent this problem?
```
 Failed example:
    c.stdevs
Expected:
    array([[  24. ,   34.5, 2669. ,   19.7, 3661.2],
           [  29.8,   29.8, 2795.9,   67.9, 2792.4],
           [  29.8,   88.7, 2976.5,   62. , 2914.5],
           [  29.8,   76.2, 2759.5,   49.2, 2762. ],
           [  38.8, 2611.8,   26.6, 2810.7,   24.1]])
Got:
    array([[   24. ,    34.5,  2669. ,    19.7,  3661.2],
           [   29.8,    29.8,  2795.9,    67.9,  2792.4],
           [   29.8,    88.7,  2976.5,    62. ,  2914.5],
           [   29.8,    76.2,  2759.5,    49.2,  2762. ],
           [   38.8,  2611.8,    26.6,  2810.7,    24.1]])
```

peterjc:
The docstrings have to match exactly - which is a pain. Note we do this in ``run_tests.py``,

```
numpy.set_printoptions(legacy="1.13")
```

This tells recent numpy to use their older style and was necessary to avoid breaking loads of pre-existing doctests.

peterjc:
I'm not clear that http://camda2018.bioinf.jku.at/doku.php/agreement makes sense for using one of their examples as a test case. It might be better to get in touch with CAMDA directly to ask permission?

Before you do that, how big is the smallest such test file (if it is too large to bundle with Biopython there is no point asking them).

peterjc:
Back to the doctest, the fragility might have been why the old version was using ``repr(...)`` whereas by adding ``print(...)`` you were implicitly calling ``str(...)`` instead. They call the ``__repr__`` and ``__str__`` methods respectively, with different behaviours in terms of suitability for doctest use..

wwydmanski:
[CEL example.zip](https://github.com/biopython/biopython/files/4181685/CEL.example.zip)
I've made a 2KB artificial CEL file with header from the CAMDA one (except for length) and random values as data. It does not have the EOT character before padding, so it can't be parsed properly without this fix

Issue #2725, opened 2020-02-17
Errors running test_Tutorial, chapter Entrez


svalqui:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version) 3.6.9 (default, Nov  7 2019, 10:44:02) [GCC 8.3.0]
import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-4.15.0-76-generic-x86_64-with-Ubuntu-18.04-bionic
import Bio; print(Bio.__version__)
1.77.dev0
```

### Expected behaviour

Run the test and pass it

### Actual behaviour

```
Python version: 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
Operating system: posix linux
test_Tutorial ... FAIL
======================================================================
ERROR: test_doctests (test_Tutorial.TutorialTestCase)
Run tutorial doctests.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/sergio/PythonProjects/biopython/Tests/test_Tutorial.py", line 256, in test_doctests
    (len(failures), ", ".join(failures)))
ValueError: 4 Tutorial doctests failed: test_chapter_entrez_line_00202, test_chapter_entrez_line_00325, test_chapter_entrez_line_00419, test_chapter_entrez_line_00466

----------------------------------------------------------------------
Ran 1 test in 48.020 seconds

FAILED (failures = 1)
```
### Steps to reproduce

Build, install and run the test:
```
~/PythonProjects/biopython$ sudo python3 setup.py build
~/PythonProjects/biopython$ sudo python3 setup.py install
~/PythonProjects/biopython$ cd Tests/
~/PythonProjects/biopython/Tests$ python3 run_tests.py test_Tutorial.py
```

peterjc:
Links pointing at the current master branch tip, which should be the same as what you used in terms of ``Bio.Entrez`` and the tutorial,
 
- test_chapter_entrez_line_00202 -https://github.com/biopython/biopython/blob/bbfefc134fff6ae9cf2838b1e4dbdba560e1bc51/Doc/Tutorial/chapter_entrez.tex#L202
- test_chapter_entrez_line_00325 - https://github.com/biopython/biopython/blob/bbfefc134fff6ae9cf2838b1e4dbdba560e1bc51/Doc/Tutorial/chapter_entrez.tex#L325
[...]

PR #2734, opened 2020-02-19
Added retmax value to Entrez Tutorial for error test_chapter_entrez_line_00202.

svalqui:
This pull request addresses issue #2725 

This address the first issue test_chapter_entrez_line_00202., by extending the number of items of the list, the ID is included.

### DIFF
--- a/Doc/Tutorial/chapter_entrez.tex
+++ b/Doc/Tutorial/chapter_entrez.tex
@@ -197,13 +197,13 @@ familiar with a particular database.
 
 \section{ESearch: Searching the Entrez databases}
 \label{sec:entrez-esearch}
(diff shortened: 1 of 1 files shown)


peterjc:
Assuming Markus was right, and they list the newest entries first, at some point the test would need updating again once enough new results are added.

We could narrow the search to reduce the chances? How about using ``biopython[title]`` instead which currently only has two hits? This also shows how the NCBI do field searching.

peterjc:
Also, could you update the ``print(record["IdList"])`` output a few lines later?

svalqui:
Yes, Markus is right the list comes with the newest entries first; Yes, it would need to be updated when more articles include Biopython in the title. Updated, Thank you.

peterjc:
Lovely, thank you.

PR #2766, opened 2020-02-27
Apply black style to test_GenBank.py, version 19.10b0; after  v19.3b0 was applied.

svalqui:

This PR applies black style to test_GenBank.py; small changes as black v19.3b0 was already applied. Should be fine to merge.

### DIFF
--- a/Tests/test_GenBank.py
+++ b/Tests/test_GenBank.py
@@ -394,7 +394,7 @@ class TestRecordParser(unittest.TestCase):
                         ("/protein_id=", '"CAB39890.1"'),
                         ("/db_xref=", '"GI:4538893"'),
                         ("/translation=", '"DKAKDAAAAAGASAQQAGKNISDAAAGGVNFVKEKTG"'),
-                        ),
+                    ),
                 ),
                 ("intron", "49..142", (("/gene=", '"csp14"'), ("/number=", "2"))),
                 ("exon", "143..206", (("/gene=", '"csp14"'), ("/number=", "3"))),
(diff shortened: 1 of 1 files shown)


Tests/test_GenBank.py
```diff
@@ -1519,9 +1519,7 @@ def test_record_parser_13(self):
             record = next(records)
             length = 2007
             locus = "AB000048"
-            definition = (
-                "Feline panleukopenia virus DNA for nonstructural protein 1, complete cds"
-            )
+            definition = "Feline panleukopenia virus DNA for nonstructural protein 1, complete cds"
```
MarkusPiotrowski: Isn't this strange? This doesn't fit in one line, does it? See the similar but shorter string in line 1600/1596, which is put on an extra line.

Tests/test_GenBank.py
```diff
@@ -1597,7 +1593,9 @@ def test_record_parser_13(self):
             record = next(records)
             length = 1755
             locus = "AB000050"
-            definition = "Feline panleukopenia virus DNA for capsid protein 2, complete cds"
```
MarkusPiotrowski: If this doesn't fit on the same line, how can line 1522 fit in one line?

Tests/test_GenBank.py
peterjc: This gives a line of 99, the old version had three lines the longest of which was 90. So both break black's target of 88, not much to choose between them (and different versions of black differed on this). I prefer the one line version.

We ought to break the string in two, I didn't notice in review,

Tests/test_GenBank.py
peterjc: The old version had one line, length 92 (too long).

The new version has three lines, the middle one is 83 long.

So this does make sense - but I agree the example above is different.

Tests/test_GenBank.py
svalqui: I'm not sure why black has removed the parentheses here and not on 1596.
usually if there is few PEP8 (line too long), I'll break them but this file have many so I have left that, I think that should be part of other Issue (PEP Compliance)?.

MarkusPiotrowski: I was just curious about the 'different' behavior of black, but didn't realize that in this case, both versions produce too long lines. The PR is already merged, so I wouldn't re-open it.

peterjc: Agreed - there is still an overly long line (due to the unbroken string), but there are a *lot* of overly long lines in the test suite. Something for another day.

Issue #2778, opened 2020-03-03
Using unittest.mock for online modules and moduls with external dependencies 


MarkusPiotrowski:
This discussion is related to #384 and the recent #2743. 

Online modules and modules with external dependencies are usually not tested regularly in our tests framework (for good reasons). As a result, these modules and/or their tests may accumulate errors and failures. This may be because of the external dependency or online source are changing their interface or output, but also as 'collateral damage' of changes elsewhere in our codebase (e.g. our recent 'Python-3-only-campaign' introduced some bugs which went under the radar of our testing framework).

For these latter cases, `unittest.mock` may be a useful tool to include such modules in our regular tests to check their general functionality. `mock` allows to mock or patch modules, classes, methods, functions, etc, e.g., to change their return value. As an example, an online call using `urllib.request.urlopen` can be mocked to return the content of a file. 

I'll file a PR soon for mock testing `Bio.Blast.NCBIWWW`.

PR #2779, opened 2020-03-03
Make mocked version of NCBI_qblast tests

MarkusPiotrowski:
This pull request is related to issue #2778

The tests for `Bio.Blast.NCBIWWW` are extended also to run offline with a mocked version of `urllib.request.urlopen`. So the same tests can run online and in our regular PR testing. 


MarkusPiotrowski:
I managed to combine the three new test files `NCBI_qblast_testCases.py`, `test_NCBI_qblast_online,py`, and `test_NCBI_qblast_mock.py` into one file, which, dependend on internet availability and test runner options (`--offline`) will run the tests either online or offline.
Will commit later.

Tests/run_tests.py
```diff
@@ -192,14 +192,14 @@ def main(argv):
...
-            import urllib
+            import urllib.request
 
             def dummy_urlopen(url):
                 raise RuntimeError(
                     "Internal test suite error, attempting to use internet despite --offline setting"
                 )
 
-            urllib.urlopen = dummy_urlopen
+            urllib.request.urlopen = dummy_urlopen
```
peterjc: That looks like a Python 3 fix worth committing in its own right - good catch.

MarkusPiotrowski:
> Do you expect future edits to test_NCBI_qblast.py must maintain this structure (live if offline, mock if offline)? If so can we make that more explicit in the docstring?

Agree. In fact, for each new online test you would need to add another side-effect for the mocked version. Depending on the actual test, this can be a new xml file or one of the existing ones. So it may be worth to design new test cases in a way that one can use one of the many files in the `Blast` folder, if possible.

I'll make this more explicit in the docstring and place some comments at the respective code lines.

PR #2780, opened 2020-03-03
Apply black to Tests/test_raxml_tool.py

martonlanga:

### DIFF
--- a/Tests/test_raxml_tool.py
+++ b/Tests/test_raxml_tool.py
@@ -26,7 +26,7 @@ except FileNotFoundError:
 if not raxml_exe:
     raise MissingExternalDependencyError(
         "Install RAxML (binary raxmlHPC) if you want"
-        "to test the Bio.Phylo.Applications wrapper."
+        " to test the Bio.Phylo.Applications wrapper."
     )
 
 # Example Phylip file with 4 aligned protein sequences
--- a/Tests/test_raxml_tool.py
+++ b/Tests/test_raxml_tool.py
@@ -15,6 +15,7 @@ from Bio import MissingExternalDependencyError
 raxml_exe = None
 try:
     from subprocess import getoutput
+
     output = getoutput("raxmlHPC -v")
     if "not found" not in output and "not recognized" not in output:
         if "This is RAxML" in output:
@@ -24,7 +25,8 @@ except FileNotFoundError:
 
 if not raxml_exe:
     raise MissingExternalDependencyError(
-        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications wrapper.")
+        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications wrapper."
+    )
 
 # Example Phylip file with 4 aligned protein sequences
 EX_PHYLIP = "Phylip/interlaced2.phy"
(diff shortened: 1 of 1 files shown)


peterjc:
It shouldn't make any difference, but do you have RAXML installed to run the test locally?

Tests/test_raxml_tool.py
```diff
@@ -24,7 +25,8 @@
 
 if not raxml_exe:
     raise MissingExternalDependencyError(
-        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications wrapper.")
+        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications wrapper."
```
svalqui: @peterjc  should this line be split in two?, it exceeds 88; otherwise is fine to be merged.
```
        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications"
        " wrapper."
```

peterjc: Black doesn't (yet) deal with breaking/joining strings, but yes it would be nice to fix that by hand here.

martonlanga:
@peterjc Installed RAXML, this is what I got:
```
test_raxml (__main__.AppTests)
Run RAxML using the wrapper. ... ok

Ran 1 test in 0.051s

OK
```

Tests/test_raxml_tool.py
```diff
@@ -8,9 +8,8 @@
 import os
 import unittest
 
-from Bio import Phylo
+from Bio import MissingExternalDependencyError, Phylo
```
martonlanga: Not sure if this is encouraged, can change it back if you want to have only 1 import per line.

MarkusPiotrowski: I have the feeling that the general trend is to have one import per line. I'm often using several imports per line if they are closely related (e.g. `from urllib.request import urlopen, Request`). However, here this is not the case, thus I would put them on different lines.

martonlanga: Reverted, now commit only splits line. Thanks for your input!

peterjc: Agreed - separate lines except perhaps examples like Markus' where it is several functions/classes from a module. Here you are importing an exception and a sub-module, nicer to have them on separate lines.

On this note, I quite like how ``reorder-python-imports`` reformats imports:

https://github.com/asottile/reorder_python_imports

I am somewhat tempted to suggest applying this to Biopython....

Tests/test_raxml_tool.py
```diff
@@ -24,7 +25,9 @@
 
 if not raxml_exe:
     raise MissingExternalDependencyError(
-        "Install RAxML (binary raxmlHPC) if you want to test the Bio.Phylo.Applications wrapper.")
+        "Install RAxML (binary raxmlHPC) if you want"
+        "to test the Bio.Phylo.Applications wrapper."
+    )
```
peterjc: Missing the space at the split point - opinion is divided over a trailing space vs a leading space. Doing the later makes it easy to spot missing spaces.

martonlanga: Added a leading space, thanks for spotting it! Will be more careful

peterjc:
Applied, thank you!

Issue #2825, opened 2020-03-31
.travis.yml: amd64_only_packages is not defined


MartinThoma:
### Setup

I've just checked out the latest version (`e3870f78f7df9a584eabbd15f4d7c50f170cc3b0`) and ran [pre-commit](https://pre-commit.com/) with the following `.pre-commit-config.yaml`:
```
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.3.0
    hooks:
    -   id: check-yaml
```

which gave

```
hookid: check-yaml

found undefined alias
  in ".travis.yml", line 60, column 15
```

The issue is `*amd64_only_packages` - there is no `&amd64_only_packages`.

peterjc:
Can you see how to fix this? See #2401 where this was added.

MartinThoma:
Sorry, I don't know that.

peterjc:
@mr-c if you have a moment, your input would be appreciated.

mr-c:
That warning can be ignored

peterjc:
Can we just define an empty list to match ``&default_packages``?

https://github.com/biopython/biopython/blob/biopython-176/.travis.yml#L139

PR #2827, opened 2020-03-31
Add pre-commit config

MartinThoma:
I have seen that you have agreed in #2008 to use the black formatter - nice!

One tool that helped me in another team to keep code quality high is [pre-commit](https://pre-commit.com/). It is a tool every developer can install locally (but doesn't have to). Before a commit is done, it runs the pre-commit hooks defined in `.pre-commit-config.yaml`.

There are some hooks which I would recommend, but might need to be discussed first:

* [isort](https://github.com/timothycrosley/isort): #2828
* check-yaml: #2825 needs to be fixed first
* [pyupgrade](https://github.com/asottile/pyupgrade): I guess I'll run it and make a PR after this one so that you can see the effect -- see #2510
* mypy: Only useful if you add type annotations ... and I think that is mainly useful for Python 3.6+. I'm not sure if this project still wants to support prior versions.

This PR should only be merged after #2552 is done and #2826 is merged.

## How to use pre-commit

```
# Install it locally
$ pip install pre-commit

# Tell pre-commit to run pre-commit hooks for the current repository
# This only needs to be done once
$ pre-commit install

# Tell pre-commit to run all hooks on all files
pre-commit run --all-files
```

## Checkboxes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,7 @@ repos:
     rev: 19.10b0
     hooks:
     -   id: black
+        args: [--check]
         additional_dependencies: [
             'restructuredtext_lint',
             'doc8',
@@ -49,18 +50,11 @@ repos:
         additional_dependencies: [
             'flake8-blind-except',
             'flake8-bugbear',
-            # 'flake8-coding==1.3.2',
             'flake8-commas',
             'flake8-comprehensions',
-            # 'flake8-debugger==3.2.1',
-            # 'flake8-deprecated==1.3',
             'flake8-docstrings',
             'flake8-implicit-str-concat',
-            # 'flake8-isort==2.7.0',
-            # 'flake8-pep3101==1.2.1',
             'flake8-pie',
-            # 'flake8-polyfill==1.0.2',
-            # 'flake8-print==3.1.4',
             'flake8-quotes',
             'flake8-rst-docstrings',
             'flake8-sfs',
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,51 @@
+# pre-commit run --all-files
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    -   id: check-executables-have-shebangs
+    -   id: check-json
+    # -   id: check-yaml
+    -   id: debug-statements
+    # -   id: trailing-whitespace
+    #     files: \.(py|txt|html|css|js)$
+    #     exclude: ^Tests/
+    -   id: end-of-file-fixer
+        files: \.(py|txt|html|css|js)$
+        exclude: ^Tests/
+    -   id: mixed-line-ending
+        files: \.(py|txt|html|css|js)$
+        exclude: ^Tests/
+# -   repo: https://github.com/pre-commit/mirrors-mypy
+#     rev: v0.730
+#     hooks:
+#     -   id: mypy
+#         args: [--ignore-missing-imports]
+# -   repo: https://github.com/asottile/seed-isort-config
+#     rev: v1.9.3
+#     hooks:
+#     -   id: seed-isort-config
+# -   repo: https://github.com/pre-commit/mirrors-isort
+#     rev: v4.3.21
+#     hooks:
+#     -   id: isort
+-   repo: https://github.com/psf/black
+    rev: 19.10b0
+    hooks:
+    -   id: black
+        exclude: ^Tests/
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.9
+    hooks:
+    -   id: flake8
+        exclude: '^Tests/|hsexpo'
+-   repo: https://github.com/asottile/blacken-docs
+    rev: v1.6.0
+    hooks:
+    -   id: blacken-docs
+        additional_dependencies: [black==19.10b0]
+        exclude: ^.github/
+# -   repo: https://github.com/asottile/pyupgrade
+#     rev: v2.1.0
+#     hooks:
+#     -   id: pyupgrade
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,18 +4,20 @@ repos:
     rev: v2.3.0
     hooks:
     -   id: check-executables-have-shebangs
+        files: \.(py|sh)$
+        exclude: ^Tests/
     -   id: check-json
     # -   id: check-yaml
     -   id: debug-statements
-    # -   id: trailing-whitespace
-    #     files: \.(py|txt|html|css|js)$
+    -   id: trailing-whitespace
+        files: \.py$
     #     exclude: ^Tests/
     -   id: end-of-file-fixer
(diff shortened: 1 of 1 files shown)


MarkusPiotrowski:
You have also seen #2580?

MartinThoma:
@MarkusPiotrowski No, I haven't. But it seems as if this is desired :-)

peterjc:
Can you include flake8 here too?

That would let us do the black tests via ``flake8-black`` (as done now on TravisCI) with rules to ignore the lack of black on the tests for now (#2552).

peterjc:
I see we missed applying black to some Python scripts without the ``.py`` extension - can you submit that as a separate pull request which we can merge almost immediately?

(I wonder if our flake8 setup needs to explicitly include those files... or if we should rename them?)

MartinThoma:
@peterjc I've added Flake8, excluded Tests from the black and flake8 checks. I removed the changes from this PR

peterjc:
What do you think @MarkusPiotrowski, @JoaoRodrigues ?

It will need followup work to update the CONTRIBUTING file.

peterjc:
It would also open the way to follow up work replacing the TravisCI/Tox style checks as outlined on #2580 - that should be at least as strict.

MarkusPiotrowski:
I have to admit that I don't understand enough of  `pre-commit` to make a useful contribution here.

I think it would still be nice to have the possibility for a local set-up so that I can prepare my PRs that they hopefully do not fail during submission.

peterjc:
That is exactly the goal - rather than running _just_ flake8 as a git pre-commit hook (our current recommendation which can only check Python files), pre-commit is a framework for running multiple checks (which would include flake8 and other Python checks, but also can do other things like YAML validation).

My goal as outlined on #2580 is to have most of the testing currently only done on TravisCI done locally via the git hook.

i.e. Increase the chance that a PR prepared locally will pass once submitted to GitHub.

MarkusPiotrowski:
OK, my misunderstanding was that `pre-commit` is only for the GitHub side. Sorry, I'll give it a try.

JoaoRodrigues:
Completely 👍 , pre-commit is very nice (and easy to install too). Thanks!

JoaoRodrigues:
Is there something holding this back?

peterjc:
I've not made time to try this myself yet, no objection in principle. Especially since flake8 have deprecated their built in hook.

As long as one core developer has tested it themselves we should probably merge it, and update the CONTRIBUTING.rst file and potentially tutorial accordingly.

JoaoRodrigues:
I'll do it then. Happy to take the plunge and report back.

JoaoRodrigues:
The installation works fine, but it does take a little bit of time to run the first time. Couple of things we need to figure out:

1. Which hooks do we want in the end? Right now this PR enables black, flake8, blacken-docs, and a few JSON, line ending stuff, but there's support for isort (import sorting), mypy, and some others. Personally, I'm in favor of keeping these import sorting stuff out of the equation, same with typing.

2. This might be my inexperience with these style checkers, but why do we need flake8-black and black? 

3. There are a lot of commented-out settings in the `.flake8` config, maybe we could take the opportunity to clean that?

4. On the same note, if I run flake8/black locally it doesn't seem to pick up on the same things that travisCI is picking up on. Where could the difference in set up come from?

peterjc:
1. There is some interest in enforcing sort order and adding type checking, but for now I agree it is simpler to take those commented lines out.

2. We need just one. As a check which would abort if black style is not followed, the hook could run ``black --check`` OR run ``flake8`` using my ``flake8-black`` plugin. What it appears to do is actually run black and edit your code as needed?

3. I'm not sure what you mean, unless it is removing the comments describing the things we are (temporarily or otherwise) ignoring?

4. Black does have a configuration file, but it is rarely used (and Biopython does not attempt to change it). The most likely cause is the flake8 configuration, perhaps interacting with a flake8 bug with how relative paths are handled. See also the work around with duplicated ``per-file-ignores`` settings in our ``.flake8`` file. As long as running the new pre-commit check (regardless of which directory you are in) matches TravisCI, I'm happy.

JoaoRodrigues:
1. 👍 

2. Yeah, I've been using `black --diff` (black check doesn't output errors, just a Pass/Fail note).

3. I mean that there are a lot of good comments we should keep (describing what we ignore and whatnot) but also the lower half of the file seems to be just commented out. We could clean it...

4. OK. I'll check against TravisCI.

peterjc:
Regarding (3), I see what you mean now - does #3017 resolve that?

JoaoRodrigues:
Oh OK, that was dumb of me. Those are explanations of the codes......... 🤦

peterjc:
Not at all, the old layout was a bit misleading - but we've clarified that now in #3017, so thanks for raising it.

JoaoRodrigues:
Thanks! I'll clean a bit the pre-commit config file to keep it as lean as possible and see if I can get the right set of options to reproduce what we have at the moment.

MartinThoma:
Hi folks, I'm currently not quite sure what to do / if I can support. If you want me to adjust something in this PR, just let me know :-)

JoaoRodrigues:
Hi @MartinThoma , first of all, thanks a ton for this PR. I'll be testing/tweaking things in the next few days and try to get it as stable as possible to merge. Would you prefer I write here my changes and you handle the commits yourself or can I push directly to this PR?

MartinThoma:
@JoaoRodrigues You can push directly to this PR, of course :-) I'm happy that I could help a bit and also that you take it over :-)

JoaoRodrigues:
@MartinThoma - I made minimal tweaks to your config file to match the flake8/black dependencies used in our Travis CI pipeline. I also added an argument to `black` (`--check`) so that the pre-commit hook does not modify files in-place.  Finally, I removed the exclusions to the `Tests` folders, so the checks are enforced there as well if those files are being committed and tweaked a bit the extensions for the executables/trailing whitespace/mixed line ending hooks.

I did not add `flake8-black` since we can call black directly, is this OK @peterjc ?

As for testing: I seem to get the same results locally as an open PR (#3210) on Travis. Anyone has better tests?

peterjc:
Note that for pre-commit, running black directly is better. It will be faster than using flake8-black because black has a cache which speeds up subsequent runs (flake8 and the plugin currently do not - see https://github.com/peterjc/flake8-black/issues/26 for more).

Also I dropped ``flake8-black`` on TravisCI - calling black allowed me to split up the checks into parallel tasks which sped up TravisCI. See #3184.

MartinThoma:
@peterjc Should I add the update of CONTRIBUTING.rst in this PR?

peterjc:
It would probably be easier as a new pull request, so I merged this now. You could also be welcome to add your name to the NEWS and CONRIB files too, if you want.

Thank you both!

JoaoRodrigues:
Thank you for the initiative @MartinThoma !! This is great!

peterjc:
In hindsight I should have tested this locally too. I've removed three flake8 plugins we're not (yet) enforcing in #3214, and removed the overly aggressive trailing-whitespace which broke multiple doctests which have meaningful trailing whitespace.

Issue #2828, opened 2020-03-31
Adopt code formatting tool isort for code style?


MartinThoma:
This goes in the same direction as #2008: [isort](https://github.com/timothycrosley/isort) is a small little tool to make the code style more similar.

## Why import sorting?

I like it, because it makes it easier for me to read a file. isort distinguishes those kinds of imports:

* Core library: Anything that ships with Python directly (struct, math, sys, os, ...)
* Third party: Stuff that needs to be installed. This is helpful to be in one block in case people take single files and execute them, without installing the package
* First party: modules defined in biopython
* Local imports: Anything starting with a dot

Overall, isort sometimes puts multiple import statements into a single one if it imports from the same package, the alphabetical sorting makes it a tiny bit more readable.

### try-except

It doesn't deal with those kinds of imports.

Before:

```
import os
from itertools import islice

try:
    from sqlite3 import dbapi2 as _sqlite
except ImportError:
    # Not present on Jython, but should be included in Python 2.5
    # or later (unless compiled from source without its dependencies)
    # Still want to offer simple parsing/output
    _sqlite = None

from Bio.Alphabet import single_letter_alphabet
from Bio.Seq import Seq
from Bio.SeqRecord import SeqRecord
from Bio.Align import MultipleSeqAlignment
from .Interfaces import SequentialAlignmentWriter
```

After:

```
import os
from itertools import islice

from Bio.Align import MultipleSeqAlignment
from Bio.Alphabet import single_letter_alphabet
from Bio.Seq import Seq
from Bio.SeqRecord import SeqRecord

from .Interfaces import SequentialAlignmentWriter

try:
    from sqlite3 import dbapi2 as _sqlite
except ImportError:
    # Not present on Jython, but should be included in Python 2.5
    # or later (unless compiled from source without its dependencies)
    # Still want to offer simple parsing/output
    _sqlite = None
```

## Examples

I used an `.isort.cfg` which is close to what most files here seem to do already:

```
[settings]
indent='    '
length_sort=0
; import_heading_stdlib=Core Library modules
; import_heading_thirdparty=Third party modules
; import_heading_firstparty=First party modules
; import_heading_localfolder=Local modules
[...]

PR #2833, opened 2020-04-06
Apply black style to test_Restriction.py, version 19.10b0.

svalqui:

This PR applies black style to test_Restriction.py. few changes but should be fine to merge.

### DIFF
--- a/Tests/test_Restriction.py
+++ b/Tests/test_Restriction.py
@@ -8,8 +8,23 @@ import unittest
 
 from Bio.Restriction import Analysis, Restriction, RestrictionBatch
 from Bio.Restriction import CommOnly, NonComm, AllEnzymes
-from Bio.Restriction import (Acc65I, Asp718I, BamHI, EcoRI, EcoRV, KpnI, SmaI,
-                             MluCI, McrI, NdeI, BsmBI, AanI, EarI, SnaI, SphI)
+from Bio.Restriction import (
+    Acc65I,
+    Asp718I,
+    BamHI,
+    EcoRI,
+    EcoRV,
+    KpnI,
+    SmaI,
+    MluCI,
+    McrI,
+    NdeI,
+    BsmBI,
+    AanI,
+    EarI,
+    SnaI,
+    SphI,
+)
 from Bio.Restriction import FormattedSeq
 from Bio.Seq import Seq, MutableSeq
 from Bio.Alphabet.IUPAC import IUPACAmbiguousDNA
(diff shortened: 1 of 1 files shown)


Tests/test_Restriction.py
```diff
@@ -8,8 +8,23 @@
...
+    SmaI,
+    MluCI,
+    McrI,
+    NdeI,
+    BsmBI,
+    AanI,
+    EarI,
+    SnaI,
+    SphI,
+)
```
peterjc, line 27: My personal preference would be:

```python
from Bio.Restriction import Acc65I
...
from Bio.Restriction import SphI
```

See discussion on #2828 - we have yet to agree any particular style, so this is fine for now.

Tests/test_Restriction.py
```diff
@@ -38,13 +53,12 @@ def test_non_iupac_letters(self):
...
     def test_formatted_seq(self):
         """Test several methods of FormattedSeq."""
-        self.assertEqual(str(FormattedSeq(Seq("GATC"))),
-                         "FormattedSeq(Seq('GATC'), linear=True)")
-        self.assertFalse(FormattedSeq(Seq("GATC")) ==
-                         FormattedSeq(Seq("TAGC")))
+        self.assertEqual(
+            str(FormattedSeq(Seq("GATC"))), "FormattedSeq(Seq('GATC'), linear=True)"
+        )
+        self.assertFalse(FormattedSeq(Seq("GATC")) == FormattedSeq(Seq("TAGC")))
```
MarkusPiotrowski, line 59: @svalqui Sorry for this request, although not black-related but still style-related: Can you change all `assertFalse(a==b)` and `assertTrue(a==b)` into `assertNotEqual(a, b)` and `assertEqual(a, b)`?

Tests/test_Restriction.py
MarkusPiotrowski: My personal preference would be to let it stand as it was, but let black do what black wants to do here.

Tests/test_Restriction.py
MarkusPiotrowski: Not only here but in the whole file?

Tests/test_Restriction.py
peterjc: Yes, leave it as is for now - we can settle any preferred import style later on the linked issue.

Tests/test_Restriction.py
peterjc: Logged #2835 for the wider issue, but happy for you to fix this file as part of this pull request, or leave it for later.

svalqui: updated, but left one that doesn't pass the test.

```
    self.assertNotEqual(Acc65I, Asp718I)
AssertionError: Acc65I == Asp718I

```
I think is mean to use assertNotEqual.

MarkusPiotrowski: This is a special thing within `Bio.Restriction`, `__eq__` (`==`) and `__ne__` (`!=`) have been overriden with different meanings. In this case Acc65I and Asp718I are neither equal (they are not the same enzyme) nor not equal (they are isoschizomers, doing exactly the same thing). 
Thus `assertFalse` is the only method which allows to test both. OK.

peterjc: I'd forgotten about that quirk. Best or at least standard practise is ``__eq__`` and ``__ne__`` give the negation of each other, but as here that isn't required.

peterjc:
Thank you both

Tests/test_Restriction.py
ajsmith: @peterjc @MarkusPiotrowski Perhaps I'm new here, but is there any reason why the `__ne__` operator is used to check for isoschizomer-ness? It doesn't seem very convenient if it confuses the linter and tests (and potentially humans too). When I first saw this behavior in the tests, I thought it was a bug, but then realized it was by design.

peterjc: I would agree in hindsight this was a poor design choice. We have been looking at replacing the entirety of the old restriction digest module anyway.

ajsmith: @peterjc Thank you for the clarification, and I look forward to seeing the new design!

PR #2837, opened 2020-04-13
Fixed a bug in calculating the height of clades when building a UPGMA tree

SHZ66:
This pull request addresses the issue that the UPGMA constructor yields a tree with wrong branch lengths. In some cases, it may even yield a tree with unequal distances from tips to the root, which violates the constant-rate assumption.

The problem is that `DistanceTreeConstructor._height_of()` returns the wrong clade height. The recursion doesn't work as intended. Please verify the changes. Here is an example that will reproduce the bug:

```
from matplotlib.pyplot import *
from Bio import Phylo
from Bio.Phylo.TreeConstruction import _DistanceMatrix
from scipy.spatial.distance import squareform, pdist
from numpy import *

n = 20
X = 10*random.random((n, 2))

A = squareform(pdist(X))

B = []
for i, row in enumerate(A):
    B.append([int(i) for i in row[:i+1]])

names = [str(i) for i in range(n)]

# UPGMA
dm = _DistanceMatrix(names, B)
constructor = Phylo.TreeConstruction.DistanceTreeConstructor()
tree = constructor.upgma(dm)

Phylo.draw(tree)
```
In this example, 20 points (2D) are randomly generated and their pairwise distances evaluated. The original implementation produces a tree that violates the constant-rate assumption:
figure1
After the fix, the branch lengths are evaluated correctly:
figure2

### DIFF
--- a/Bio/Phylo/TreeConstruction.py
+++ b/Bio/Phylo/TreeConstruction.py
@@ -824,10 +824,10 @@ class DistanceTreeConstructor(TreeConstructor):
     def _height_of(self, clade):
         """Calculate clade height -- the longest path to any terminal (PRIVATE)."""
         if clade.is_terminal():
-            height = 0 
+            height = 0
         else:
             height = max(self._height_of(c) + c.branch_length for c in clade.clades)
-            
+
         return height
 
 
--- a/Bio/Phylo/TreeConstruction.py
+++ b/Bio/Phylo/TreeConstruction.py
@@ -698,15 +698,11 @@ class DistanceTreeConstructor(TreeConstructor):
             inner_clade.clades.append(clade1)
             inner_clade.clades.append(clade2)
             # assign branch length
-            if clade1.is_terminal():
-                clade1.branch_length = min_dist * 1.0 / 2
-            else:
-                clade1.branch_length = min_dist * 1.0 / 2 - self._height_of(clade1)
+            clade1.branch_length = min_dist * \
+                    1.0 / 2 - self._height_of(clade1)
 
-            if clade2.is_terminal():
-                clade2.branch_length = min_dist * 1.0 / 2
-            else:
-                clade2.branch_length = min_dist * 1.0 / 2 - self._height_of(clade2)
+            clade2.branch_length = min_dist * \
+                    1.0 / 2 - self._height_of(clade2)
 
             # update node list
             clades[min_j] = inner_clade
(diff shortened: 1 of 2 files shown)


peterjc:
Can you add a test for this (ideally self contained with a tree constructed in code, or reusing one of the test files already in Biopython as the input)?

I will leave the code change review to @etal etc, stylistically this seems cleaner but I don't understand what the meaning of the change is without taking more time to study it.

SHZ66:
> Can you add a test for this (ideally self contained with a tree constructed in code, or reusing one of the test files already in Biopython as the input)?
> 
> I will leave the code change review to @etal etc, stylistically this seems cleaner but I don't understand what the meaning of the change is without taking more time to study it.

Yes, sure I will figure out a test data that can reproduce the error. But if you look at the original implementation of the `_height_of` function, the recursion is obviously not going to work:
```
    def _height_of(self, clade):
        """Calculate clade height -- the longest path to any terminal (PRIVATE)."""
        height = 0
        if clade.is_terminal():
            height = clade.branch_length
        else:
            height = height + max(self._height_of(c) for c in clade.clades)

        return height
```
Without adding the branch lengths of inner clades somewhere and reset `height=0` in every function call, `max(self._height_of(c) for c in clade.clades)` will simply look for the branch length of some terminal clade of the descendants of `c` and the branch lengths will not accumulate.

etal:
Yes, this fix makes sense to me. My original implementation was not well tested, but this one looks good. Thanks!

ramti:
Hi, any update?

etal (approved):

etal:
I think we're waiting on a unit test before merging this. The test could be:
1. Set the random seed to a fixed value, e.g. 1
2. Create a distance matrix and build the UPGMA tree (as in the code above)
3. Check that the depths of all terminal nodes are all approximately equal (`[depth for node, depth in tree.depths() if node.is_terminal()]`)

fabianegli:
Not sure - What is the best way to contribute code to this PR?

jlumpe:
Anyone still working on this?

WCutePy:
Hi,

I happened to stumble upon this bug as well using the most recent version of BioPython 1.84 at this time.
@etal  are there currently any plans to resolve merging these changes?

etal:
I found and merged another PR #3951 which includes this PR's commits plus a unit test. So, this fix should be in the main branch now. Apologies for the delay, and thanks for your contributions.

Issue #2849, opened 2020-04-24
New argument to handle compression when fetch PDB files in all formats.


tinchogb:
First of all, thanks for your work !

### Setup

I am reporting an improvement in Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```
Python 3.6.9
( CoDNaS-RNA_py3 ) martingb@Ubu18:~/Projects/2020/CoDNaS-RNA$ python
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys; print(sys.version)
3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
>>> import platform; print(platform.python_implementation()); print(platform.platform())
CPython
Linux-5.3.0-46-generic-x86_64-with-Ubuntu-18.04-bionic
>>> import Bio; print(Bio.__version__)
1.76

### Expected behaviour
Bio.PDB.PDBList.retrieve_pdb_files() anyone should be allow to download *.gz files. This is not present in current function.

### Actual behaviour
Actual function definition:
def retrieve_pdb_file(
        self, pdb_code, obsolete=False, pdir=None, file_format=None, overwrite=False
    ):

One possible improved function definition and code:
def retrieve_pdb_file(
        self, pdb_code, obsolete=False, pdir=None, file_format=None, overwrite=False, **compress**=**False**
    ):
    ...<BLOKE OF CODE>
        # Retrieve the file
        if self._verbose:
            print("Downloading PDB structure '%s'..." % pdb_code)
        try:
            urlcleanup()
            urlretrieve(url, filename)
        except OSError:
            print("Desired structure doesn't exists")
        else:
            **if compress:
                pass
            else:**
                with gzip.open(filename, "rb") as gz:
                    with open(final_file, "wb") as out:
                        out.writelines(gz)
                    os.remove(filename)
        return final_file

### Steps to reproduce

1) Go to Bio.PDB.PDBList.py and modify **retrieve_pdb_files()**
2) Add one more argument and set default value in **retrieve_pdb_files()**, could be "**compress=False**"
3) Modify last part of above function code (see doc **# Retrieve the file**) to evaluate new argument.
4) Modify docstring to explain new argument and default behaviour.

peterjc:
[...]

PR #2850, opened 2020-04-25
tinchogb-PDBList-handle-compression

tinchogb:
Incorporating a new argument that allows controlling compression (by default uncompressed) in retrieve_pdb_file().

This pull request addresses issue #2849

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -328,10 +328,16 @@ class PDBList:
 
         # Skip download if the file already exists
         if not overwrite:
-            if os.path.exists(final_file):
-                if self._verbose:
-                    print("Structure exists: '%s' " % final_file)
-                return final_file
+            if compress:
+                if os.path.exists(filename):
+                    if self._verbose:
+                       print("Structure exists: '%s' " % filename)
+                    return filename
+            else:
+                if os.path.exists(final_file):
+                    if self._verbose:
+                        print("Structure exists: '%s' " % final_file)
+                    return final_file
 
         # Retrieve the file
         if self._verbose:
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -215,7 +215,7 @@ class PDBList:
         return obsolete
 
     def retrieve_pdb_file(
-        self, pdb_code, obsolete=False, pdir=None, file_format=None, overwrite=False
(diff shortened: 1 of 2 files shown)


Bio/PDB/PDBList.py
```diff
@@ -256,6 +256,9 @@ def retrieve_pdb_file(
         :param pdir: put the file in this directory (default: create a PDB-style directory tree)
         :type pdir: string
 
+        :param compress: if set to True, existing structure files will be gzip stored. Default: False
```
jgreener64, line 259: The meaning of this isn't immediately clear to me, maybe something like "downloaded files will be gzip stored"?

Bio/PDB/PDBList.py
```diff
@@ -339,10 +342,13 @@ def retrieve_pdb_file(
...
-                    out.writelines(gz)
-            os.remove(filename)
+            if compress:
+                pass
+            else:
+                with gzip.open(filename, "rb") as gz:
+                    with open(final_file, "wb") as out:
+                        out.writelines(gz)
+                os.remove(filename)
         return final_file
```
jgreener64, line 352: This should probably return `filename` if `compress` is `True`.

jgreener64 (changes requested):
This looks broadly good to me, I have left a couple of comments.

Could you also add a test in test_PDBList.py that downloads a compressed file and checks it exists as expected?

Bio/PDB/PDBList.py
tinchogb: Yes, is much clear like you said.

Bio/PDB/PDBList.py
tinchogb: That's correct. It is Ok fixed now?
```suggestion
        if compress:
            return filename
```

Bio/PDB/PDBList.py
```diff
@@ -339,10 +342,13 @@ def retrieve_pdb_file(
         except OSError:
             print("Desired structure doesn't exists")
         else:
-            with gzip.open(filename, "rb") as gz:
-                with open(final_file, "wb") as out:
-                    out.writelines(gz)
-            os.remove(filename)
+            if compress:
+                pass
```
tinchogb, line 346: like this?
```suggestion
                return filename
```

jgreener64: 👍

tinchogb: Added new commits to PR

PR #2852, opened 2020-04-28
fix bug with restriction site elucidate

EdHarry:

### DIFF
--- a/Bio/Restriction/Restriction.py
+++ b/Bio/Restriction/Restriction.py
@@ -1813,7 +1813,7 @@ class Ambiguous(AbstractCut):
             elif 0 <= f3 + length <= length:
                 re = "N^" + abs(f5) * "N" + site[:f3] + "_" + site[f3:]
             elif f3 + length < 0:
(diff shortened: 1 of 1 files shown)


peterjc:
That looks like a sensible change - do you have an example handy, i.e. an enzyme where this showed up?

EdHarry:
> That looks like a sensible change - do you have an example handy, i.e. an enzyme where this showed up?

Yes and no.
I discovered the bug because I was generating site definitions for some dual-cutters. This is not currently supported by Biopython, so to get around this I was overriding the "cut_twice" method (with return False) and then calling "elucidate" for the first site, and then swapping "fst5" and  "fst3" for "scd5" and "scd5" for the second site.
For example, this works fine for "SdeOSI"

I hit the bug when trying to generate the first site definition for "UcoMSI"

I have no idea if that code path gets used for any of the single-cutters, but regardless, I think the bug is an obvious typo; it throws a TypeError: "can't multiply sequence by non-int of type 'str'"

PR #2891, opened 2020-05-14
PDBParser throws TypeError on duplicated Residues

JoaoRodrigues:
This pull request addresses an upcoming issue communicated personally from a user. I'll link to it as soon as it's created. EDIT: #2892 

The problem here is that Residue.id is a tuple, and that trips the old `%s` formatting when raising an Exception. I guess these particular lines were missed during the f-string overhaul. Converting to f-strings seems to restore the expected behavior. 

I can reproduce the problem and test the fix with a simple test file with the following lines:
```
HETATM 6289  O   HOH     5      28.182  -5.239  31.370  1.00 22.99           O
HETATM 6513  O   HOH     6      21.829   3.361  14.003  1.00 14.25           O
HETATM 6607  O   HOH     5      33.861  40.044  18.022  1.00 18.73           O
END
```

### DIFF
--- a/Bio/PDB/Entity.py
+++ b/Bio/PDB/Entity.py
@@ -211,7 +211,9 @@ class Entity:
         """Add a child to the Entity."""
         entity_id = entity.get_id()
         if self.has_id(entity_id):
-            raise PDBConstructionException("%s defined twice" % entity_id)
+            raise PDBConstructionException(
+                f"{entity_id} defined twice"
+            )
         entity.set_parent(self)
         self.child_list.append(entity)
         self.child_dict[entity_id] = entity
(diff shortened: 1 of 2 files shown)


peterjc:
Good catch, the fix makes sense.

peterjc:
Is it easy to turn that example into a tiny test? e.g. put that string in via StringIO?

peterjc:
```
$ flake8
./Bio/PDB/Entity.py:214:44: BLK100 Black would make changes.
```

Probably one or both changes will fit on one line?

JoaoRodrigues:
I'll add the test and fix the black complaint! I broke the line on purpose because I thought black would prefer it that way..

peterjc:
Excellent - apologies for missing this in the pull request where the bug was accidentally introduced. Very pleased to see the test coverage improved too.

JoaoRodrigues:
No problem! I'm surprised it took this long to find this or did something change in the way `%s` handles tuples?

peterjc:
Seems to have been lurking undetected for a while. If we do move over to f-strings we might fix a few more like it?

JoaoRodrigues:
Probably, another thing in favor of f-strings. I am really very surprised that it was hiding for this long!

PR #2980, opened 2020-06-17
Use chain.from_iterable in query.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.

### DIFF
--- a/Bio/SearchIO/_model/query.py
+++ b/Bio/SearchIO/_model/query.py
@@ -414,9 +414,7 @@ class QueryResult(_BaseSearchObject):
     @property
     def hsps(self):
         """Access the HSP objects contained in the QueryResult."""
-        return sorted(
-            chain.from_iterable(self.hits), key=lambda hsp: hsp.output_index
-        )
+        return sorted(chain.from_iterable(self.hits), key=lambda hsp: hsp.output_index)
 
     @property
     def fragments(self):
--- a/Bio/SearchIO/_model/query.py
+++ b/Bio/SearchIO/_model/query.py
@@ -415,13 +415,13 @@ class QueryResult(_BaseSearchObject):
     def hsps(self):
         """Access the HSP objects contained in the QueryResult."""
         return sorted(
-            (hsp for hsp in chain(*self.hits)), key=lambda hsp: hsp.output_index
+            chain.from_iterable(self.hits), key=lambda hsp: hsp.output_index
         )
 
     @property
     def fragments(self):
         """Access the HSPFragment objects contained in the QueryResult."""
-        return list(chain(*self.hsps))
+        return list(chain.from_iterable(self.hsps))
 
     # public methods #
     def absorb(self, hit):


bow:
Hi @cool-RR ,

Thanks for the merge request :).

I have to say that I am not completely sure if `chain` and `chain.from_iterable` are doing things differently underneath. From [the documentation of `chain.from_iterable`](https://docs.python.org/3.8/library/itertools.html#itertools.chain.from_iterable):

> Alternate constructor for chain(). Gets chained inputs from a single iterable argument that is evaluated lazily.

Which indicates that it is merely another way of calling `chain`.

Furthermore, I tried doing my own tests to see any possible performance improvements:

```python
#!/usr/bin/env python

from random import randint
from typing import Generator, List


def make_lists(n: int, x: int = 100) -> Generator[List[int], None, None]:
    """Create a generator that yields n list of ints, each varying between 0 to
    x elements."""
    return (list(range(randint(0, x))) for _ in range(n))
```

And then using `ipython`:

```
$ ipython -i bench.py

In [1]: from itertools import chain                                                                                                                                                                                                 
```

First, using `chain` on a generator:

```
In [2]: %timeit sorted(chain(*make_lists(10000)))                                                                                                                                                                                   
43.6 ms ± 262 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit sorted(chain(*make_lists(10000)))                                                                                                                                                                                   
43.6 ms ± 308 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit sorted(chain(*make_lists(10000)))                                                                                                                                                                                   
43.9 ms ± 773 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Then using `chain.from_iterable` also on a generator:
```
In [5]: %timeit sorted(chain.from_iterable(make_lists(10000)))                                                                                                                                                                      
44 ms ± 1.44 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit sorted(chain.from_iterable(make_lists(10000)))                                                                                                                                                                      
43.5 ms ± 663 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit sorted(chain.from_iterable(make_lists(10000)))                                                                                                                                                                      
44.2 ms ± 1.45 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Then using `chain` on a list of lists:
```
In [8]: items = list(make_lists(10000))                                                                                                                                                                                             

In [9]: %timeit sorted(chain(*items))                                                                                                                                                                                               
29.8 ms ± 323 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit sorted(chain(*items))                                                                                                                                                                                              
29.8 ms ± 303 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [11]: %timeit sorted(chain(*items))                                                                                                                                                                                              
29.9 ms ± 279 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

And finally `chain.from_iterable` on a list of lists as well:
```
In [12]: %timeit sorted(chain.from_iterable(items))                                                                                                                                                                                 
29.7 ms ± 228 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [13]: %timeit sorted(chain.from_iterable(items))                                                                                                                                                                                 
29.7 ms ± 192 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [14]: %timeit sorted(chain.from_iterable(items))                                                                                                                                                                                 
29.7 ms ± 239 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

It could be that the outer `sorted` call negates any benefit coming from the inner function call change, since [the current implementation](https://en.wikipedia.org/wiki/Timsort) needs to load all data in-memory anyway.

I may have overlooked something here though. Do you have any local benchmarks that you can share and/or any estimates on how much performance improvements we are getting by switching to `chain.from_iterable`?

cool-RR:
> I have to say that I am not completely sure if `chain` and `chain.from_iterable` are doing things differently underneath

The only thing they do different is that `chain(*x)` consumes `x` in advance and puts it in a something like a tuple, while `chain.from_iterable(x)` consume items one-by-one at the last possible moment, never storing the full list.

The performance improvement wouldn't be too relevant to the usecase in Biopython, since you're saving them all in memory anyway. I think it's a best practice though, because you never know when your usage of `chain` would benefit from using `from_iterable`, and it never hurts.

Here is a list of cases of this issue in Biopython:

```
Bio/Phylo/NexusIO.py: 76                        
Bio/SearchIO/_model/hit.py: 344                 
Bio/SearchIO/_model/query.py: 418               
Bio/SearchIO/_model/query.py: 424               
Bio/SearchIO/ExonerateIO/exonerate_text.py: 92  
Bio/SearchIO/ExonerateIO/exonerate_text.py: 95  
```

I looked at all of them, and in all of them, they are consumed immediately and stored in a list. If you'd decide it's too esoteric to merge, I'd understand.

bow:
> The only thing they do different is that chain(*x) consumes x in advance and puts it in a something like a tuple, while chain.from_iterable(x) consume items one-by-one at the last possible moment, never storing the full list.

> The performance improvement wouldn't be too relevant to the usecase in Biopython, since you're saving them all in memory anyway. I think it's a best practice though, because you never know when your usage of chain would benefit from using from_iterable, and it never hurts.

Indeed. Generators are particularly useful where the underlying list / iterable is too large to fit in-memory. This is not always the case, however. The benchmarks I showed earlier ran slower when the input are generators instead of lists, highlighting a use case where generators are probably not the best fit (and where using generators would have visible impact on performance).

> Here is a list of cases of this issue in Biopython:

> I looked at all of them, and in all of them, they are consumed immediately and stored in a list. If you'd decide it's too esoteric to merge, I'd understand.

Thanks for the list. After taking a closer look at them, all the use cases do end up creating a list in memory. So here I would say that changing the `chain` call to `chain.from_iterable` is cosmetic. In fact, there's probably a stronger case to be made here for removing the `chain` calls altogether, as the underlying iterables in most cases are already in-memory lists. But for this PR, I am closing it.

Issue #2968, opened 2020-06-09
Make Seq translate method compatible with string method?


peterjc:
We have discussed making the ``Seq`` object subclass the Python string. Additional methods are fine (e.g. ``.reverse_complete(...)``) but changes in meaning are more tricky. Currently the ``.translate(...)`` method is very different, and does not attempt to match the Python string behaviour went it might be inferred (rather it raises a specific exception):

https://github.com/biopython/biopython/blob/biopython-177/Bio/Seq.py#L1162

This issue is to explore making ``Seq(....).translate(table)`` work like the string translate method as needed by distinguishing our valid table arguments (which include integer NCBI genetic codes, and their various string alias) versus a general string mapping.

This is now harder than I expected due to Python 2/3 changes:

```
$ python2 -c "print('ACGTNX'.translate('wibble'))"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ValueError: translation table must be 256 characters long
```

Under Python 2 it was easy, if ``table`` was a 256 character then we could assume it was the string mapping, not a biological translation.

However, Python 3 changed things - it accepts string which are not 256 long (plus dictionaries of unicode ordinals). Shorter strings (like our table names) are apparently treated as partial mappings, with the missing mapping entries treated as no change.

```
$ python3 -c "print('ACGTNX'.translate('wibble'))"
ACGTNX
python3 -c "print('ACGTNX'.translate('?'*9999))"
??????
$ python3 -c "print('ACGTNX'.translate('?'*256))"
??????
$ python3 -c "print('ACGTNX'.translate('?'*80))"
???T?X
python3 -c "print('ACGTNX'.translate(''))"
ACGTNX
```

I've not worked out yet if that is a Python bug, or something missing in the docstring. Part of this stems from the more complicated use cases with unicode.

On the bright side, there is no optional second argument ``deletechars`` anymore.

chris-rands:
Apparently, i read that python 2 `unicode.translate()` is like python 3 `str.translate()`. And python 2 `str.translate()` is like python 3 `bytes.translate()`

```
$ python3 -c "print(b'ACGTNX'.translate(b'wibble'))"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ValueError: translation table must be 256 characters long
```

peterjc:
Yes, that was the impression I got trying to find a pre-existing Python issue - along with a lot of work behind the scenes to get the unicode translate method faster for ASCII examples.

[...]

PR #2996, opened 2020-06-24
Precompute dna/rna complement tables

chris-rands:
Since the `ambiguous_dna_complement`/`ambiguous_rna_complement` dicts are unlikely to change, why re-compute the complement tables every time `Seq` is imported? My suggestion here does add some imports and equality testing, but not calling `_maketrans` should improve speed I reasoned


chris-rands:
black doesn't like the long import lines i think- I can change this if needed

chris-rands:
On reflection, maybe a more elegant (but more complicated) solution might be to generate the complement tables when `setup.py` is called. Other parts of the codebase could do the same, for example, you could generate the quality score dicts based on the fixed offsets too https://github.com/biopython/biopython/blob/master/Bio/SeqIO/QualityIO.py#L546 This just relies on users not editing the values *after* building the package.

peterjc:
Interestingly Python 3 has a radically different translate method, and will now take a dict mapping ordinals as keys. Cross reference #2968 

This might make a faster (reverse)complement practical?

chris-rands:
I'm not sure what you're suggesting Peter in terms of the translate method, but it seems like a different issue?

My aim here was to avoid the issue that currently doing `from Bio.Seq import Seq` will build the complement tables every time. https://github.com/biopython/biopython/blob/master/Bio/Seq.py#L56-L57

This isn't needed since it's very unlikely that these tables actually change (it would require changes to the DNA/RNA complement dicts). Therefore, my suggestion is to effectively hard code the calculated tables (or compute them when the package is built via `setup.py`)- does it make sense?

peterjc:
These mapping are the output of ``str.maketrans`` which is used as input to the python string translate method. The nature of the mapping changed from Python 2 to 3:

```
$ python2 -c "import string; print(repr(string.maketrans('acgt', 'tgca')))"
'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`tbgdefchijklmnopqrsauvwxyz{|}~\x7f\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f\xa0\xa1\xa2\xa3\xa4\xa5\xa6\xa7\xa8\xa9\xaa\xab\xac\xad\xae\xaf\xb0\xb1\xb2\xb3\xb4\xb5\xb6\xb7\xb8\xb9\xba\xbb\xbc\xbd\xbe\xbf\xc0\xc1\xc2\xc3\xc4\xc5\xc6\xc7\xc8\xc9\xca\xcb\xcc\xcd\xce\xcf\xd0\xd1\xd2\xd3\xd4\xd5\xd6\xd7\xd8\xd9\xda\xdb\xdc\xdd\xde\xdf\xe0\xe1\xe2\xe3\xe4\xe5\xe6\xe7\xe8\xe9\xea\xeb\xec\xed\xee\xef\xf0\xf1\xf2\xf3\xf4\xf5\xf6\xf7\xf8\xf9\xfa\xfb\xfc\xfd\xfe\xff
```

```
$ python3 -c "print(str.maketrans('acgt', 'tgca'))"
{97: 116, 99: 103, 103: 99, 116: 97}
```

It used to be a 256 character string for remapping the ASCII characters, now it is a mapping of byte ordinals.

peterjc:
My hunch is this will have negligible impact. Can you measure a difference?

Still, if each line of the bytes mapping was commented with the letters, this would still be quite clear to the reader...

chris-rands:
> It used to be a 256 character string for remapping the ASCII characters, now it is a mapping of byte ordinals.

Thanks Peter, yes, I know of the py 2 vs. 3 difference, I just don't understand your suggestion in terms of changing the biopython implementation

> My hunch is this will have negligible impact. Can you measure a difference?

you might be correct, i'll do a benchmark at some point

peterjc:
Sorry, I wrote (reverse)complement but was thinking (back)transcribe. The complement code already using the string translate method.

The Seq object's transcribe method essentially boils down to calling ``.replace("T", "U").replace("t", "u")`` and that might be faster doing one call to ``.translate(...)`` with a two element mapping. This probably wouldn't have made a measurable difference on Biopython 1.77 but now that we've cut out all the alphabet management, this is worth trying.

Likewise, the ungap method now boils down to calling ``.replace("-", "")`` where perhaps calling ``.translate({45: None})`` is faster?

chris-rands:
I see, well i doubt that `.translate()` should replace a single `.replace()` call- it provides no benefit I can see:

```
In [10]: s = 'ATGC'*1000                                                                                                                                                                                                                      

In [11]: %timeit s.replace('A', 'a')                                                                                                                                                                                                          
4.88 µs ± 120 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: %timeit s.translate({65:97})                                                                                                                                                                                                         
5.97 µs ± 127 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

 But yes, i guess when they are chained, there is a benefit:
```
In [17]: %timeit s.replace('A', 'a').replace('T', 't')                                                                                                                                                                                        
9.67 µs ± 222 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [18]: %timeit s.translate({65:97, 84:116})                                                                                                                                                                                                 
5.89 µs ± 49.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

from `Bio/`:

```
$ grep -Er "replace\(.*.replace\(" .
./SeqIO/Interfaces.py:    return text.replace("\n", " ").replace("\r", " ")
./SeqIO/Interfaces.py:        return text.replace("\n", " ").replace("\r", " ")
./SeqIO/InsdcIO.py:            mol_type = mol_type.replace("unassigned ", "").replace("genomic ", "")
./SeqIO/FastaIO.py:            yield title, "".join(lines).replace(" ", "").replace("\r", "")
./SeqIO/FastaIO.py:    yield title, "".join(lines).replace(" ", "").replace("\r", "")
./AlignIO/Interfaces.py:        return text.replace("\n", " ").replace("\r", " ")
./AlignIO/MsfIO.py:        seqs = ["".join(s).replace("~", "-").replace(".", "-") for s in seqs]
./AlignIO/MauveIO.py:                id_line = id_line.replace("\n", " ").replace("\r", " ")
./AlignIO/MauveIO.py:            id_line = id_line.replace("\n", " ").replace("\r", " ")
./SearchIO/ExonerateIO/exonerate_text.py:                seq_block[line_name].replace("{", "").replace("}", "")
./SearchIO/ExonerateIO/exonerate_text.py:        + len(hsp[seq_type][0].replace("-", "").replace(">", "").replace("<", ""))
./SearchIO/BlastIO/blast_text.py:            qdesc = qdesc.replace("\n", "").replace("\r", "")
./SearchIO/BlastIO/blast_text.py:                hdesc = hdesc.replace("\n", "").replace("\r", "")
./Nexus/Nexus.py:        line = line.replace("\r\n", "\n").replace("\r", "\n").strip()
./Nexus/Trees.py:            tree = tree.strip().replace("\n", "").replace("\r", "")
./codonalign/__init__.py:                qcodon = qcodon.replace("^", "").replace("$", "")
./codonalign/__init__.py:                qcodon = qcodon.replace("^", "").replace("$", "")
./GenBank/__init__.py:        accession = accession_string.replace("\n", " ").replace(";", " ")
./Data/CodonTable.py:        x.strip() for x in name.replace(" and ", "; ").replace(", ", "; ").split("; ")
./Phylo/NewickIO.py:    return "[%s]" % (text.replace("[", "\\[").replace("]", "\\]"))
./Phylo/NewickIO.py:                    label = "'%s'" % label.replace("\\", "\\\\").replace("'", "\\'")
./Seq.py:        return Seq(str(self).replace("T", "U").replace("t", "u"))
./Seq.py:        return Seq(str(self).replace("U", "T").replace("u", "t"))
./Seq.py:        return dna.replace("T", "U").replace("t", "u")
./Seq.py:        return rna.replace("U", "T").replace("u", "t")
```

chris-rands:
Not sure this one if worth pursuing, thanks for the feedback

peterjc:
Thanks for looking into these.

Issue #2993, opened 2020-06-22
test_codonalign.Test_dn_ds regression on scipy 1.5.0


peterjc:
TravisCI Python 3.7 has started failing on the master branch:

```
certifi==2020.6.20,chardet==3.0.4,codecov==2.1.7,coverage==5.1,cycler==0.10.0,idna==2.9,kiwisolver==1.2.0,matplotlib==3.2.2,mmtf-python==1.1.2,msgpack==1.0.0,mysql-connector-python-rf==2.2.2,mysqlclient==1.4.6,numpy==1.19.0,psycopg2-binary==2.8.5,pyparsing==2.4.7,python-dateutil==2.8.1,requests==2.24.0,scipy==1.5.0,six==1.15.0,urllib3==1.25.9
...
======================================================================
FAIL: test_dn_ds (test_codonalign.Test_dn_ds)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biopython/biopython/Tests/test_codonalign.py", line 250, in test_dn_ds
    self.assertAlmostEqual(dN, 0.0194, places=4)
AssertionError: 0.018551097874995984 != 0.0194 within 4 places (0.00084890212500401616 difference)
----------------------------------------------------------------------
```

See https://travis-ci.org/github/biopython/biopython/jobs/700971840

Previous build worked, but had scipy 1.4.1

```
certifi==2020.6.20,chardet==3.0.4,codecov==2.1.7,coverage==5.1,cycler==0.10.0,idna==2.9,kiwisolver==1.2.0,matplotlib==3.2.2,mmtf-python==1.1.2,msgpack==1.0.0,mysql-connector-python-rf==2.2.2,mysqlclient==1.4.6,numpy==1.19.0,psycopg2-binary==2.8.5,pyparsing==2.4.7,python-dateutil==2.8.1,requests==2.24.0,scipy==1.4.1,six==1.15.0,urllib3==1.25.9
```

See https://travis-ci.org/github/biopython/biopython/jobs/700563827

This is currently the only TravisCI target which has SciPy installed, and thus only Python 3.7 shows the problem.

peterjc:
Failure is here:

```python
        try:
            # New in scipy v0.11
            from scipy.optimize import minimize
            dN, dS = cal_dn_ds(codon_seq1, codon_seq2, method="ML")
            self.assertAlmostEqual(dN, 0.0194, places=4)
[...]

PR #2998, opened 2020-06-25
This patch will fix the optimizer issue of scipy 1.5.0 when using ML method for dn/ds calculation

zruan:

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -72,7 +72,7 @@ deps =
     py37: mysqlclient
     py38: rdflib
     pypy3: mysqlclient
-    py37: scipy<1.5.0
+    py37: scipy
     py38: networkx
     py37: matplotlib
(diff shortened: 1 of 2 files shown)


peterjc:
Thanks! Can you explicitly agree to the BSD dual licensing?

Is anyone else available to review this on the technical side?

zruan:
I checked all three boxes.

The issue arises due to the third parameter is fixed to 1 during minimization. The 1.5.0 scipy.optimize.minimize seems to be very strict and doesn't allow this to happen. The fix is to define another temporary function without the third parameter. Everything else remains unchanged. I think this is okay.

peterjc:
Cross reference #2993

peterjc:
Applying now since the new SciPy appears to be available on Windows now, and AppVeyor stopped working too. Thank you.

PR #3001, opened 2020-06-26
Changed header delineator from space to pipe

rchurt:
There are common utilities for processing FASTA files that only use the string in the header preceding a space, so for this reason pipes are preferred in FASTA header strings


peterjc:
We could not take this change as is. This is a break in historic behaviour which would break lots of existing scripts.

If you want to use NCBI style naming with Biopython unchanged, put this in the record's ``.id`` field directly.

Cross reference #2711 (and successor) for improving the parsing of the NCBI style FASTA files. That could in theory be combined with output in NCBI style? You might have some useful feedback on that proposal.

Issue #2995, opened 2020-06-23
Bug identified at PDBIO.set_structure() and solution for it


alexandrefassio:
Hi, 

In the current BioPython version, there are two bugs (lines 86 and 95) in the function **set_structure**() from the class **PDBIO** (Bio/PDB/PDBIO.py):

           Line 86: sb.structure[0]["A"].add(pdb_object.copy())
           Line 95: sb.structure[0]["A"].child_list[0].add(pdb_object.copy())

As you can see, it always tries to access a chain object whose id is "A". However, if one tries to save a Residue/Atom whose chain id is not "A", it raises a KeyError exception because the recently created chain (line 79) had its id updated to **pdb_object.parent.id** at line 83 (when pdb_object is a residue) or 92 (when pdb_object is an atom).

Thus, to fix this problem, move the parent_id to outside the try/except block and substitute the previous lines by

           Line 86: sb.structure[0][parent_id].add(pdb_object.copy())
           Line 95: sb.structure[0][parent_id].child_list[0].add(pdb_object.copy())

bests

JoaoRodrigues:
Hi @alexandrefassio would you be willing to submit a pull request on this?

alexandrefassio:
Hi @JoaoRodrigues 

Sure! I can do it.

PR #3003, opened 2020-06-26
Fixed bug at PDBIO.set_structure()

alexandrefassio:
It always tries to access a chain object whose id is "A". However, if one tries to save a Residue/Atom whose chain id is not "A", it raises a KeyError exception because the recently created chain (line 79) had its id updated to pdb_object.parent.id at line 83 (when pdb_object is a residue) or 92 (when pdb_object is an atom).

This pull request addresses issue #2995 

- [ ] I have read the ``CONTRIBUTING.rst`` file, have run ``flake8`` locally, and

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -78,21 +78,21 @@ class StructureIO:
                 else:
                     sb.init_chain("A")
                     if pdb_object.level == "R":
+                        parent_id = pdb_object.parent.id
                         try:
-                            parent_id = pdb_object.parent.id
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
Hi @alexandrefassio , unfortunately the fix you provided breaks several tests. The fix is not as simple as accessing the parent's id, since you can have the case where you just created a Residue yourself, where parent is `None`. I pushed a more comprehensive fix here #3006 that addresses this bug and cleans the code a little to make it easier to understand in the future.

Issue #2983, opened 2020-06-18
ResidueDepth


DuanhaoranCC:
Hello,

My biopython version is 1.73, executing `rd= PDB.ResidueDepth (model)` , but I'm sorry to report an error.`The runtime error: failed to generate surface file usingcommand:` I have downloaded the binary file of MSMS. What else should I do? Environment variable? Or something.

peterjc:
Can you manually run the command it tried to use as the terminal, and does that work or give a more helpful error message?

Also in case it is relevant, which version of MSMS do you have, and what OS?

DuanhaoranCC:
I'm sorry. I didn't give you a detailed description,

My computer is windows 10 64bit. I use MSMS's official website (http://mgltools.scripps.edu/downloads),MSMS 2.6.1 - current release downloaded Binaries of the MSMS C library,After that, no configuration was made.

Then I called RD in the python program. `PDB.ResidueDepth (model, 'D:/benchmark5/structures/1DFJ_ R_ b. PDB')`, an error is displayed

`C:~\Programs\Python\Python36\lib\site-packages\Bio\PDB\ResidueDepth.py:533: BiopythonDeprecationWarning: ResidueDepth no longer requires a pdb file. This argument will be removed in a future release of Biopython. "of Biopython.", BiopythonDeprecationWarning)
'msms' �����ڲ����ⲿ���Ҳ���ǿ����еĳ���
���������ļ���
Traceback (most recent call last):
  File "E:/jupyter/GCN/10.py", line 30, in <module>
    rd = PDB.ResidueDepth(model, 'D:/blast-2.10.1+/msms_win32_6.2.1/1crn.pdb')
  File "C:\Users\lenovo\AppData\Local\Programs\Python\Python36\lib\site-packages\Bio\PDB\ResidueDepth.py", line 541, in __init__
    surface = get_surface(model)
  File "C:\Users\lenovo\AppData\Local\Programs\Python\Python36\lib\site-packages\Bio\PDB\ResidueDepth.py", line 486, in get_surface
    "command:\n%s" % make_surface)
RuntimeError: Failed to generate surface file using command:
msms -probe_radius 1.5 -if C:\Users\lenovo\AppData\Local\Temp\tmpwxg9y2o2 -of C:~\AppData\Local\Temp\tmpnap3uxk4 > C:~\AppData\Local\Temp\tmpo6jelb8m

[...]

Issue #2877, opened 2020-05-12
Alternatives to Bio.Application for invoking command line tools?


peterjc:
Issue for the following discussion started on the mailing list:

https://mailman.open-bio.org/pipermail/biopython/2020-May/016769.html

Dear Biopythoneers,

Biopython has a lot of command line tool wrappers, based around the objects
in Bio/Application/__init__.py, for building a command line string and
running it. Some time ago I started to think that we might actually be
better off dropping our in-house command line wrappers, and recommending a
standard or third party library approach for defining and executing command
line strings instead.

Taking an example in our tutorial, running the blastx tool from NCBI
BLAST+. Currently Biopython provides a specific object for the blastx
command, which knows all the expected command arguments, can do some
validation, and even has some basic help text included for each of them:

```pycon
>>> from Bio.Blast.Applications import NcbiblastxCommandline
>>> help(NcbiblastxCommandline)
...
>>> blastx_cline = NcbiblastxCommandline(query="opuntia.fasta", db="nr",
evalue=0.001, outfmt=5, out="opuntia.xml")
>>> blastx_cline
NcbiblastxCommandline(cmd='blastx', out='opuntia.xml', outfmt=5,
query='opuntia.fasta',
db='nr', evalue=0.001)
>>> print(blastx_cline)
blastx -out opuntia.xml -outfmt 5 -query opuntia.fasta -db nr -evalue 0.001
>>> stdout, stderr = blastx_cline()
```

This works quite nicely, but writing a unique class for each command line
tool we wish to support is a lot of quiet tedious work, especially if
including minimal documentation for the arguments or argument validation.
This is also an on-going maintenance problem - one of the issues I think we
should fix before the next Biopython release is updating the NCBI BLAST+
wrappers as new arguments have been added.

Some tools have a rather cryptic command line API, and in those cases
perhaps our efforts are sensible. However, with tools like NCBI BLAST+
where is a clear command line API, and I don't see that our efforts
actually add a great deal over constructing the string in code and calling
subprocess:

```pycon
>>> import subprocess
>>> cmd = "blastx -query opuntia.fasta -db nr -out opuntia.xml -evalue
0.001 -outfmt 5"
>>> subprocess.check_call(cmd, shell=True)
```

There are third party libraries which might be easier? For example, the sh
library supports our our current style with keyword arguments:

```pycon
>>> from sh import blastx
>>> blastx(query="opuntia.fasta", db="nr", out="opuntia.xml",
[...]

PR #3005, opened 2020-06-26
Added keyword to specify msms executable on class instantiation

JoaoRodrigues:
This pull request closes issue #2983.

Tested on Win10 with MSMS 2.6.1:

```python
import pathlib
from Bio.PDB import PDBParser, ResidueDepth

parser = PDBParser(QUIET=1)
struct = parser.get_structure('x', 'Tests/PDB/2XHE.pdb')

msms = '...'  # path to your executable

rd = ResidueDepth(s, msms_exec=msms)
```

Does not break compatibility with existing code/function calls.

### DIFF
--- a/Bio/PDB/ResidueDepth.py
+++ b/Bio/PDB/ResidueDepth.py
@@ -600,7 +600,6 @@ class ResidueDepth(AbstractPropertyMap):
 
     def __init__(self, model, pdb_file=None, msms_exec=None):
         """Initialize the class."""
-
         if msms_exec is None:
             msms_exec = "msms"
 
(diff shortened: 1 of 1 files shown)


peterjc:
Do you think using the clearer "msms_exec" is better than the existing choice of "MSMS" in upper case?

JoaoRodrigues:
Hi @peterjc, I do yes. The reason why I didn't change the argument in the other function was to avoid breaking old code. Would you prefer keeping MSMS?

peterjc:
I'm torn - up to you.

JoaoRodrigues:
In an ideal world, we'd change MSMS to msms_exec everywhere with some deprecation warnings. The lower case argument name is more 'pythonic' I think it's fine like this. We should anyway revisit all these third-party application wrappers and try to standardize them like we did for all the sequence alignments tools. Then we can look at this again.

peterjc:
See also #2877 but this example is more complicated than just running a tool, it parses the output too.

PR #3045, opened 2020-07-12
test numpydoc 1.1.0

carlosp420:

### DIFF
--- a/Bio/Restriction/__init__.py
+++ b/Bio/Restriction/__init__.py
@@ -11,6 +11,8 @@
 
 Examples
 --------
+.. code-block:: python
+
     >>> from Bio.Seq import Seq
     >>> from Bio.Restriction import *
     >>> pBs_mcs = 'GGTACCGGGCCCCCCCTCGAGGTCGACGGTATCGATAAGCTTGATATCGAATTCCTG'
@@ -32,7 +34,7 @@ Examples
     HhaI      Hin1II    Hin6I     HinP1I    HpyCH4IV  HpySE526I Hsp92II   HspAI     
     MaeII     MseI      NlaIII    SaqAI     TaiI      Tru1I     Tru9I...
     <BLANKLINE>
-    >>> b = a.blunt()  # Analysis with blunt enzmyes
+    >>> b = a.blunt()  # Analysis with blunt enzymes
     >>> a.print_that(b)  # Print results for blunt cutters
     AccII      :  98.
     AfaI       :  4.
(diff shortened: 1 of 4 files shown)


Bio/Restriction/__init__.py
```diff
@@ -11,6 +11,8 @@
 
 Examples
 --------
+.. code-block:: python
```
peterjc, line 14: Does this help? I thought it was redundant with doctest markup.

peterjc:
P.S. Have you seen #3024 which is why I downgraded numpydoc to avoid v1.1.0, cross reference the numpydoc issue and PR?

Bio/Restriction/__init__.py
carlosp420: it is necessary to avoid badly formatted output, like this:
Screen Shot 2020-07-12 at 10 15 49

The blank line before BmeDI avoids the whole block to be completely formatted. This is fixed by adding `.. code-block:: python`:
Screen Shot 2020-07-12 at 12 51 22

peterjc: Ah - good reason to avoid blank lines within output. Thanks!

carlosp420:
> P.S. Have you seen #3024 which is why I downgraded numpydoc to avoid v1.1.0, cross reference the numpydoc issue and PR?

yes. I wanted to fix the issue and it appears that numpydoc v1.1.0 does not crashes when the mocking for Bio.Restriction.Restriction is removed. Ie the travis builds passed.

peterjc:
Oh. That's probably good, but how long does it make the autogenerated output - and is the order stable (i.e. sorted)?

carlosp420:
> Oh. That's probably good, but how long does it make the autogenerated output - and is the order stable (i.e. sorted)?

I just noticed that the output is huge! it is documenting all the entries in Restriction_Dictionary and the output is not sorted. Multiple builds of the docs result in different sorting of the documented entries of the dictionary.
Maybe that is why this module was ignored in `autodoc_mock_imports` ?

Then maybe I will wait until your PR in numpydoc gets resolved to clean up this draft pull request.

peterjc:
Yes, that was why I'd excluded the restriction docstrings - it was rather unmanageable.

The other tweaks (typos and formatting) are worth applying now, but let's see what the numpydoc decide to do.

Bio/Restriction/__init__.py
peterjc: Fixed in #4902, thanks for flagging this.

PR #3160, opened 2020-07-27
Don't record alphabet in SnapGeneIO

peterjc:
This pull request is a subset of #3126, part of the general removal of ``Bio.Alphabet`` work (#2046).

As it stands it does not remove this line which was suggested during the review of #3126:

```python
record.annotations["molecule_type"] = "DNA"
```

All the comments in the code are DNA centric, as is the SnapGene public facing documentation.

@gouttegd is there _anything_ to suggest SnapGene binary files might contain RNA (or even protein)?

### DIFF
--- a/Bio/SeqIO/SnapGeneIO.py
+++ b/Bio/SeqIO/SnapGeneIO.py
@@ -16,7 +16,6 @@ from re import sub
 from struct import unpack
 from xml.dom.minidom import parseString
 
-from Bio import Alphabet
 from Bio.Seq import Seq
 from Bio.SeqFeature import SeqFeature, FeatureLocation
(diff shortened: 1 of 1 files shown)


gouttegd:
> @gouttegd is there _anything_ to suggest SnapGene binary files might contain RNA (or even protein)?

I don’t think I’ve ever run across a RNA- or protein-containing SnapGene file. It is certainly theoretically possible (the format is easily extensible, one would just need e.g. to define a new type of packet), but until we see a non-DNA SnapGene file I can’t say for sure whether it happens.

peterjc:
Thanks - sounds like we'd have to make code changes for any non DNA packet, so we should be fine as is.

PR #3184, opened 2020-07-30
Call black directly (not via flake8)

peterjc:
Now that the entire code base uses black this is practical - the flake8 configuration gave a convenient way to manage the gradual adoption.

However, right now my flake8-black plugin has more overhead than it should and slows this down: https://github.com/peterjc/flake8-black/issues/26

### DIFF
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -102,8 +102,8 @@ deps =
     flake8-comprehensions
     flake8-bugbear
     flake8-implicit-str-concat
-    flake8-black
     flake8-quotes
+    black
     restructuredtext_lint
     doc8
     pygments
@@ -116,6 +116,8 @@ deps =
 commands =
     # PEP-8 and PEP-257 style checks:
     flake8
+    # Black style check
+    black --check .
     # Now do various checks on our RST files:
     # Calling via bash to get it to expand the wildcard for us
     bash -c \'rst-lint --level warning *.rst\'
--- a/.travis-tox.ini
+++ b/.travis-tox.ini
@@ -114,10 +114,7 @@ deps =
     # pydocstyle v4 had false positives with D413, need v5
     pydocstyle>=5.0.0
 commands =
-    # PEP-8 and PEP-257 style checks:
-    flake8
-    # Black style check
-    black --check .
+    # Leave flake8 till last as it is the slowest
     # Now do various checks on our RST files:
     # Calling via bash to get it to expand the wildcard for us
     bash -c \'rst-lint --level warning *.rst\'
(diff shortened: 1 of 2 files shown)


peterjc:
Key number for us, previous TravisCI style checks took about 6m30s (looked at last three), this run also took 6m30s. So no real change.

peterjc:
Second run was no faster - after more investigation the speed of a standalone ``black --check .`` is down to caching.

peterjc:
So no real speed up (unless running ``tox -c .travis-tox.ini -e style`` locally), but I think the changes are still worthwhile as it is one less TravisCI dependency, and reordering should give quicker failures.

.travis-tox.ini
```diff
@@ -131,6 +131,10 @@ commands =
     bash -c "if grep --include '*.py' --include '*.rst' --include '*.tex' -rni 'doi:' Bio BioSQL Scripts Doc ; then echo 'Please use https://doi.org/... not the doi: or DOI: style.'; false; fi"
     bash -c "if grep --include '*.py' --include '*.rst' --include '*.tex' -rn 'dx\.doi\.org' Bio BioSQL Tests Scripts Doc ; then echo 'Please use https://doi.org/... not the dx.doi.org style.'; false; fi"
     bash -c "if grep --include '*.py' --include '*.rst' --include '*.tex' -rn 'http://doi\.org' Bio BioSQL Tests Scripts Doc ; then echo 'Please use https://doi.org/... not http://doi.org/...'; false; fi"
+    # Black style check:
+    black --check .
```
chris-rands, line 135: One option here would be to provide the full diff if black fails to help people spot the non-conforming lines? e.g. something like `! black --check . && black --diff .`

peterjc: I did wonder about ``black --diff .`` instead of ``black --check .`` but why run both?

Now that all the existing code is black compliant, the diff ought to be modest (unless trying to add entire new files which are not black compliant). Might be worthwhile?

Note if running locally with a cache, I understand ``black --diff`` currently has a performance penalty over ``black --check``, but with TravisCI we don't have the benefit of a black cache anyway.

peterjc:
We do get a win by splitting flake8 up from black and the rest of the style checks:

https://travis-ci.org/github/peterjc/biopython/builds/713439646

style: 2m 51s (most of which is running black)
flake8: 4m 19s (excluding black)
sdist,bdist_wheel: 56s

Could probably split this as two TravisCI jobs each taking about 4ms (flake8, and then everything else). Or perhaps we move flake8 to the main test stage in order to speed up the initial checks as far as possible. Options option options.

.travis-tox.ini
chris-rands: Ah yes of course, running `black --diff .` only would be better- I think it might help new contributors especially (if there is no substantial performance loss)

peterjc:
That worked: flake8 without black took 4m21s, black and the other style checks took 3m8s.

An improvement on flake8 including black taking 6m30s and the other style checks about 1m.

i.e. Total time is about the same, but the split is much more even so the main TravisCI stage will start that much sooner.

peterjc:
I'd suggested @MarkusPiotrowski for review, but happy to have a second opinion from @chris-rands or anyone else available instead.

.travis-tox.ini
peterjc: Note that ``black --diff .`` does not set an error code if changes are needed.

I later had to change this to ``black --check --diff .`` instead, see c470fa2b4d14d7f67f597550d334061c664f1c90

PR #3214, opened 2020-08-14
TravisCI / precommit fixes

peterjc:
This pull request addresses issue #2827, and fixes an oversight in changing how we called black in #3184 (it was not actually erroring when there was a black difference)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,12 +17,6 @@ repos:
     hooks:
     -   id: black
         args: [--check]
-        additional_dependencies: [
-            'restructuredtext_lint',
-            'doc8',
-            'pygments',
-            'docutils==0.14'
-        ]
 -   repo: https://gitlab.com/pycqa/flake8
     rev: 3.7.9
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,14 +32,11 @@ repos:
         additional_dependencies: [
             'flake8-blind-except',
             'flake8-bugbear',
-            'flake8-commas',
             'flake8-comprehensions',
             'flake8-docstrings',
             'flake8-implicit-str-concat',
-            'flake8-pie',
             'flake8-quotes',
             'flake8-rst-docstrings',
-            'flake8-sfs',
             'pydocstyle>=5.0.0',
         ]
 -   repo: https://github.com/asottile/blacken-docs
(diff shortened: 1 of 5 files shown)


JoaoRodrigues:
I'd added those plugins that were mentioned in `.flake8`, thought they were all active - sorry!

This looks good. Do you want to leave this open and add more fixes/tweaks?

Bio/PDB/PDBParser.py
```diff
@@ -142,8 +142,14 @@ def _get_header(self, header_coords_trailer):
...
-            "ATOM  ", "HETATM", "MODEL ", "ENDMDL", "TER   ", "ANISOU",
-            "SIGATM", "SIGUIJ",  # older 2.3 format specs
+            "ATOM  ",
+            "HETATM",
+            "MODEL ",
+            "ENDMDL",
+            "TER   ",
+            "ANISOU",
+            "SIGATM",
+            "SIGUIJ",  # older 2.3 format specs
```
JoaoRodrigues, line 152: Argh. I disagree with black sometimes... I'd move this comment to the line above them, since both SIGATM and SIGUIJ are from the previous format spec.

.pre-commit-config.yaml
```diff
@@ -19,27 +17,18 @@ repos:
     hooks:
     -   id: black
         args: [--check]
```
JoaoRodrigues, line 19: Shouldn't you add --diff here too?

peterjc: Can do - depends what we thing would be most helpful to the user. Personally I'd find it too verbose.

Bio/PDB/PDBParser.py
peterjc: Comment placement is something best tweaked after running black. Sorry, my oversight in the recent TravisCI changes let this slip by (``black --diff`` only shows the differences but does not report an error, thus the change here to ``black --diff --check`` instead).

.pre-commit-config.yaml
```diff
@@ -19,27 +17,18 @@ repos:
...
-        additional_dependencies: [
```
JoaoRodrigues, line 22: Got it! Documentation is still a magical world for me..

.pre-commit-config.yaml
JoaoRodrigues: OK! I'd be happy with less verbose indeed. It should just point us to where to run black.

Bio/PDB/PDBParser.py
peterjc: Is this OK now (moved the comments)?

Issue #2580, opened 2020-01-24
Replace flake8 git pre-commit hook with pre-commit


peterjc:
Currently in ``CONTRIBUTING.rst`` we recommend using the flake8 pre-commit hook as follows:

```
$ flake8 --install-hook git
$ git config --bool flake8.strict true
```

There are at least two drawbacks. First, this only checks our Python code - it does not cover things like our RST files, or the sort order in the contributor listing (currently done via tox).

Also, the flake8 documentation http://flake8.pycqa.org/en/latest/user/using-hooks.html recommends moving to the pre-commit package instead:

https://pre-commit.com/

So, it would be a good idea to move all the style checks defined in ``.travis-tox.in`` (via tox) into ``.pre-commit-config.yaml`` instead:

https://github.com/biopython/biopython/blob/biopython-176/.travis-tox.ini#L117

```
commands =
    # PEP-8 and PEP-257 style checks:
    flake8
    # Now do various checks on our RST files:
    # Calling via bash to get it to expand the wildcard for us
    bash -c \'rst-lint --level warning *.rst\'
    # Check sort order (bash call work around for pipe character)
    bash -c \'grep "^- " CONTRIB.rst | LC_ALL=C sort -u -c -f\'
    # Check copyright date
    bash -c \'grep "1999-`date +'%Y'`" LICENSE.rst\'
    # Would like to tell doc8 to just check *.rst but does *.txt too:
    bash -c "doc8 --ignore-path 'Doc/examples/ec_*.txt' *.rst Doc/"
    # Check no __docformat__ lines
    bash -c "if grep --include '*.py' -rn '^__docformat__ ' Bio BioSQL Tests Scripts Doc ; then echo 'Remove __docformat__ line(s), we assume restructuredtext.'; false; fi"
    # Check DOI link style, see https://www.crossref.org/display-guidelines/
    bash -c "if grep --include '*.py' --include '*.rst' --include '*.tex' -rni 'doi:' Bio BioSQL Scripts Doc ; then echo 'Please use https://doi.org/... not the doi: or DOI: style.'; false; fi"
    bash -c "if grep --include '*.py' --include '*.rst' --include '*.tex' -rn 'dx\.doi\.org' Bio BioSQL Tests Scripts Doc ; then echo 'Please use https://doi.org/... not the dx.doi.org style.'; false; fi"
[...]

PR #3218, opened 2020-08-15
Do RST checks via pre-commit

peterjc:
This pull request addresses issue #2580, using ``rstcheck`` and ``doc8`` via ``pre-commit`` instead of via ``tox``.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,3 +40,9 @@ repos:
     hooks:
     -   id: rstcheck
         args: [--report=warning]
+-   repo: https://github.com/PyCQA/doc8
+    rev: 0.8.1
+    hooks:
+    -   id: doc8
+        additional_dependencies: [pygments]
+        args: [--quiet,--ignore-path=Doc/examples/ec_*.txt]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,3 +37,8 @@ repos:
     -   id: blacken-docs
         additional_dependencies: [black==19.10b0]
         exclude: ^.github/
+-   repo: https://github.com/myint/rstcheck
+    rev: ''
+    hooks:
+    -   id: rstcheck
+        args: [--report=warning]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,12 +5,10 @@ repos:
     hooks:
     -   id: check-executables-have-shebangs
         files: \.(py|sh)$
-        exclude: ^Tests/
     -   id: check-json
     -   id: debug-statements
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -37,3 +35,14 @@ repos:
     -   id: blacken-docs
         additional_dependencies: [black==19.10b0]
         exclude: ^.github/
+-   repo: https://github.com/myint/rstcheck
+    rev: ''
```
JoaoRodrigues, line 39: If there is no preferred revision, why not remove the argument? Otherwise, maybe they support a 'stable' argument?

.travis-tox.ini
```diff
@@ -92,28 +92,15 @@ commands =
 skip_install = True
 whitelist_externals =
     black
```
JoaoRodrigues, line 94: So much cleaner - really happy this is progressing! Thanks @MartinThoma !

.travis.yml
```diff
@@ -33,7 +33,7 @@ matrix:
       before_install: echo "Going to run style checks using pre-commit"
       install:
         - pip install pre-commit
-        - pre-commit install
+        - pre-commit install --install-hooks
```
JoaoRodrigues, line 36: More curiosity than anything else but what's the relevance of this option? The documentation isn't very clear: "Whether to install hook environment for all environments in the config file;"

MartinThoma:
This one is interesting. Personally, I only have used pre-commit on the developers side (locally) and always tox on the server-side. I'm not saying it's bad (it probably is good to be consistent), I'm just saying it's new to me.

.travis.yml
peterjc: Having watched this with and without a pre-existing cache, it controls when the environments are created.

```
pip install pre-commit
pre-commit install
pre-commit run -a # environments created now
```

versus:

```
pip install pre-commit
pre-commit install --install-hooks  # environments created now
pre-commit run -a
```

With the way travis splits things up and times them, it seemed nicer to have this as part of the install step, not the main script. Also it makes the main output less verbose (unless the environments are cached already).

.pre-commit-config.yaml
peterjc: Sadly https://github.com/myint/rstcheck doesn't yet have a release with the hook YAML file present, this seemed the best option. See https://github.com/myint/rstcheck/issues/69

peterjc:
@MartinThoma you may be right. We're on track for having a clear separation that pre-commit does linting and style (both locally and on the cloud), while Tox does the Linux CI work (primarily on the cloud, but can be used locally). The Windows CI setup currently is different again though.

PR #3220, opened 2020-08-15
Break test_PDB.py in multiple smaller, modular, tests.

JoaoRodrigues:
Right now, `test_PDB.py` has nearly 1900 lines of code and tests nearly all parts of the `Bio.PDB` module. To make it easier to maintain the existing tests and extend the coverage of the code I propose to split the code in several small test files all named `test_PDB_XXX.py`. I also think it is prudent to merge some of the tool tests (DSSP, NACCESS) with the code to parse their output/integrate it with Structure since they all live under the same module.

I will update the PR as I follow along breaking `test_PDB.py`. Current status: done! 🎉 
- [x] test_PDB_MMCIFIO.py
- [x] test_PDB_Exposure.py
- [x] test_PDB_DSSP.py
- [x] test_PDB_PDBIO.py
- [x] test_PDB_NACCESS.py
- [x] test_PDB_ResidueDepth.py  # extended
- [x] test_PDB_Polypeptide.py  # extended
- [x] test_PDB_SMCRA.py
- [x] test_PDB_PDBParser.py
- [x] test_PDB_parse_pdb_header.py  # extended

In addition, I am replacing some parts of the code (namely dependency checking) by `unittest.skip` and friends for readability, e.g.: 

```python
        # Check if MSMS is installed
        try:
            v = subprocess.check_output(
                ["msms", "-h"], universal_newlines=True, stderr=subprocess.STDOUT
            )
        except OSError:
            self.skipTest("Install MSMS if you want to use it in Biopython.")
```


JoaoRodrigues:
PS. Tests will always fail because of #3219

peterjc:
So we need to merge #3219 first?

I agree in principle with splitting up the PDB test file.

JoaoRodrigues:
Yup - #3219 should go in first. I'll rebase this branch once it's accepted. 

On this PR: I'm using `setUpClass` to try and speed up some things. I noticed that for instance, `PDBParser` is instantiated for every test, or that certain structures are parsed again and again unnecessarily. It's a bit hard to gauge how much time this might save with all these changes, so I'll only be sure about it at the end. Surely it won't add time.

JoaoRodrigues:
All done and all tests passing locally (with fix #3219 applied).

I installed DSSP and MSMS to check if those pass too, but not NACCESS. I also commented one test for MSMS/ResidueDepth (2XHE) to reduce the testing time from ~45 to ~4 seconds. It's just a large protein dimer and two of the other tests have multiple chains as well.

peterjc:
The use of ``self.skipTest`` looks sensible.

Who would you suggest to review this?

JoaoRodrigues:
I didn't change the tests themselves (only added new ones actually). I noticed a few modules have low coverage so I might also work on that, but maybe in another PR.

I'd tag @jgreener64 @lennax @etal and @rob-miller since they are either somewhat familiar with the code or have had a hand in writing some of the tests themselves.

jgreener64:
Looks good to me.

I don't see the value of a commented out test though - perhaps remove it if the relevant code is tested elsewhere?

rob-miller:
Have not had a look yet, just getting back into this work now.

On a general note I think that we could improve our selection of PDB/mmCIF
test files, and if you're rationalizing all the tests this might be a good
time for it.  At least one of the PDB files (2XHE?) has issues that have
been rectified in the current RCSB version (though not to say that we
should not check for the problems), but also looking to explicitly select
cases exhibiting 'Hamelryck' disorder - alternate atoms/residues/locations
- as well as 'Dunker' disorder - regions of chain with insufficient density
to get any coordinates at all - and to pick these on the basis of problem
coverage and hopefully small size.

Just to mention, I intend to have a go at running the parsers through all
of mmCIF and PDB as suggested in #3915.  Will see if I can find some better
test cases along the way.

> I don't see the value of a commented out test though - perhaps remove it
if the relevant code is tested elsewhere?

Always nice to have code examples, but I agree commented code loses
relevance/accuracy quickly and hurts readability.  Would be really cool if
we could link in the docs to the relevant tests as examples.

jgreener64:
I have some scripts and results for parsing the whole PDB at https://github.com/jgreener64/pdb-benchmarks that might be of use.

One "fun" test case is 1EN2, which has disordered atoms inside a disordered residue. We mention it in the docs but don't use it for testing.

JoaoRodrigues:
Thank you both for the feedback!

Rob, revisiting the files is a good idea. If you want I can merge this PR
and then you can work on it?

@jgreener64, I should have looked first before writing my own code to
download / parse the PDB... :) I have both PDB and mmCIF datasets locally
now, probably the PDB parser test already finished overnight but I won't
have time to check until next weekend (holidays).

peterjc:
Did you us https://github.com/biopython/biopython/blob/master/Bio/PDB/PDBList.py ? (half joking)

jgreener64:
Yes, the script in pdb-benchmarks uses `PDBList` :smile:.

rob-miller:
I added mmCIF to my local mirror the other night as well :-)
It's GMT+3 here so night time now, but I'm happy to work on it this week.
@JoaoRodrigues <https://github.com/JoaoRodrigues> the diffs look fine to
me, I'm happy if you would like to merge and then I will review the test
structures

JoaoRodrigues:
Thank you all! I'll merge then and then we keep working on the tests!

PR #3224, opened 2020-08-17
Cope with missing quality scores in ABI parser

### LINKED ISSUE #3221: SeqIO error: local variable 'qual' referenced before assignment
### Setup

system as follows:

```python
3.7.0 (default, Jun 28 2018, 08:04:48) [MSC v.1912 64 bit (AMD64)]

CPython
Windows-10-10.0.18362-SP0

1.76
```

### Expected behaviour

I send some samples for Sanger sequencing to a commercial facility. I'm able to read the files they send using the command
```
from Bio import SeqIO
from Bio import Seq
rec = SeqIO.read("isolation-round4/3dr23_Forward.ab1",'abi-trim').seq
```

This worked fine until now. 
### Actual behaviour
But recently, due to a move, we had to send the samples elsewhere for sequencing. Now, if I try to run the same command on the files they send to me, I get an error:
``UnboundLocalError: local variable 'qual' referenced before assignment    in
  File "C:\Users\Anaconda3\lib\site-packages\Bio\SeqIO\AbiIO.py", line 462, in AbiIterator
    letter_annotations={"phred_quality": qual}``

### Steps to reproduce

I have uploaded two files, one that works with the above code and one that does not.  They can be downloaded here: [https://easyupload.io/m/k2f60x](url) 

I posted this problem on Biostars ([https://www.biostars.org/p/455577/#455598](url)) and was told to open an issue here. I had also posted it on Stackoverflow ([https://stackoverflow.com/questions/63420626/biopython-seqio-error-local-variable-qual-referenced-before-assignment](url))

This is my first time opening an issue here, so please let me know if you need any more information from me. 
I would appreciate any help with this!
Thank you.

peterjc:

Closes #3221

### DIFF
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -446,6 +446,11 @@ class AbiIterator(SequenceIterator):
         if qual:
             # Expect this to be missing for FSA files.
             record.letter_annotations["phred_quality"] = qual
+        elif self.trim:
+            raise ValueError(
+                "The 'abi-trim' format can not be used for files without"
+                " quality values."
+            )
 
         if self.trim and not is_fsa_file:
             record = _abi_trim(record)
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -382,7 +382,7 @@ class AbiIterator(SequenceIterator):
         sample_id = "<unknown id>"
 
         raw = {}
-        seq = qual = sample_id = None
+        seq = qual = None
         for tag_name, tag_number, tag_data in _abi_parse_header(header, handle):
             key = tag_name + str(tag_number)
 
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -446,7 +446,7 @@ class AbiIterator(SequenceIterator):
         if qual:
             # Expect this to be missing for FSA files.
             record.letter_annotations["phred_quality"] = qual
-        elif self.trim:
+        elif not is_fsa_file and not qual and self.trim:
             raise ValueError(
                 "The 'abi-trim' format can not be used for files without"
                 " quality values."
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -382,6 +382,7 @@ class AbiIterator(SequenceIterator):
         sample_id = "<unknown id>"
 
         raw = {}
+        seq = qual = sample_id = None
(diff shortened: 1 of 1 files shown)


peterjc:
I'd noticed but not explored the failure. From an end user perspective, if there are trimming values but no quality, then it should just trim the sequence. Do you want to work on this or should I?

bow:
> I'd noticed but not explored the failure. From an end user perspective, if there are trimming values but no quality, then it should just trim the sequence. Do you want to work on this or should I?

Hmm..does it make sense instead to raise a ValueError as soon as `qual` is known to be missing and the requested format is `abi-trim`? The reason is because trimming is quality-based (i.e. it's based on a sliding window over the qualities). I thought about just letting it pass silently (i.e. making it functionally behave the same as `abi`), but then decided against it.

I can make the remaining change :).

peterjc:
Oh. I was thinking of SFF which does things differently.

In that case you're right, an exception is appropriate.

bow:
@peterjc Ah, I realized this is on your private branch so I can't commit to it.

The last changeset which raises the exception is quite minimal:

```
commit 09e1f132cf2761fc06f320ea545442e6f58e77fa (HEAD -> abi_no_qual)
Author: Wibowo Arindrarto <contact@arindrarto.dev>
Date:   Wed Aug 19 11:07:12 2020 +0200

    Raise an exception when the abi-trim format is used for files without quals

diff --git a/Bio/SeqIO/AbiIO.py b/Bio/SeqIO/AbiIO.py
index 1a0c3e90a..970f99dcf 100644
--- a/Bio/SeqIO/AbiIO.py
+++ b/Bio/SeqIO/AbiIO.py
@@ -446,6 +446,11 @@ class AbiIterator(SequenceIterator):
         if qual:
             # Expect this to be missing for FSA files.
             record.letter_annotations["phred_quality"] = qual
+        elif not qual and self.trim:
+                "The 'abi-trim' format can not be used for files without"
+                " quality values."
+            )
 
         if self.trim and not is_fsa_file:
             record = _abi_trim(record)

```

Would it be ok for you to push this from your end?

peterjc:
I've ticked the "Allow edits by maintainers" option, so in principle you would be able to edit the branch for this PR on my repository. Anyway, I'll apply your change - one moment.

peterjc:
Now to update the tests... but I didn't expect any of the existing examples to be affected.

bow:
@peterjc yes :) ~ I just saw this message (got back from out of town with no laptops), but indeed I see this is now merged already :).

peterjc:
Good - I assumed you'd be OK once the automated tests had passed.

PR #3232, opened 2020-08-24
DisorderedAtom returns children sorted by altloc

### LINKED ISSUE #3231: StructureBuilder cannot handle non-canonical altloc ordering written by PDBIO
Parsing a PDB file with disordered atoms where the first instance of the disordered atom has a non-blank altloc but other instances are blank fails - second instance is ignored.  See this example:

 - PDB 5j2u, line 10598: No altloc in first instance of disordered atom H (goes from blank to `B`):
```
ATOM   9132  N   SER B 155      30.962  53.255  26.071  1.00 48.84           N
ATOM   9133  CA  SER B 155      32.293  52.794  26.445  1.00 56.65           C
ATOM   9134  C   SER B 155      33.375  53.714  25.883  1.00 50.84           C
ATOM   9135  O   SER B 155      34.387  53.970  26.534  1.00 53.74           O
ATOM   9136  CB  SER B 155      32.510  51.365  25.951  1.00 57.00           C
ATOM   9137  OG  SER B 155      33.470  50.690  26.748  1.00 59.12           O
ATOM   9138  H   SER B 155      30.485  52.658  25.676  1.00 58.61           H
ATOM   9139  H  BSER B 155      30.485  52.658  25.676  0.00 58.61           H
ATOM   9140  HA  SER B 155      32.369  52.794  27.412  1.00 67.98           H
ATOM   9141  HB2 SER B 155      31.669  50.885  26.004  1.00 68.40           H
ATOM   9142  HB3 SER B 155      32.815  51.390  25.032  1.00 68.40           H
ATOM   9143  HG  SER B 155      33.579  49.907  26.465  1.00 70.94           H
```

PDBParser will do its job right, but then PDBIO will place atom 9139 before 9138, which mangles the round-trip parsing. Looking into `StructureBuilder`, it seems the logic behind the disordered atoms could use some improvement since it checks directly for a non-blank altloc code. Although this is not 'canonical' PDB format, it's a tiny improvement to our parser to make it more robust. We should also maybe add some logic to PDBIO to write blank altlocs first, and then named ones - fixing the problem from both ends.

JoaoRodrigues:

Closes #3231. This fix was the simplest and the existing tests already cover this case so no extra test necessary (just a tweak to the test data file).


JoaoRodrigues:
Pinging @rob-miller @jgreener64 for additional reviews. My only concern is with numerical altlocs (that will be iterated first with this patch), but the pdb format only states that there should be no blank altlocs when we have disorder, so...

jgreener64:
LGTM.

JoaoRodrigues:
Merged. Thank you @jgreener64 !

rob-miller:
Sorry I missed the ping (and to wake the zombie PR) - your code looks fine but mine (internal coords) not so easily fixed
 to handle blank altlocs.  Think I will leave that as 'not supported' - as you noted:

["If an atom is provided in more than one position, then a non-blank alternate location indicator must be used for each of the atomic positions."](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM)

JoaoRodrigues:
If the atoms are there in the Structure object, what would be the issue with your IC code? This change basically only changes the order in which your iterate over the disordered children.

rob-miller:
No, your code is fine and doesn't affect mine.  I wanted to support 5j2u to
convert to internal coords and back to an identical PDB file, but looks
like testing for the existence of altloc atoms without altloc IDs in my
code adds too much overhead for the benefit.

PR #3251, opened 2020-09-03
Update pre-commit settings for doc8 (Windows line endings)

MarkusPiotrowski:
This PR is related to issue #2580. It adjusts the `pre-commit` settings for the recent update in `doc8` so that line endings will not be checked.

In short: This will allow developers on Windows to change text files and use our `pre-commit` setup without the need to convert line endings to Unix style. 

I also added a line to `.gitattibutes` which ensures that text files will have Unix line endings when being committed. This works repository-wide, in contrast to user `git` settings (e.g. `core.autocrlf true` for Windows users) which may be set wrongly. This should ensure that we have uniform line endings in the project.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,11 @@
 # On Windows, git can be setup to convert text files to use
 # DOS/Windows newlines (\r\n) rather than Unix style (\n).
 # This should be harmless for most of our unit tests...
-#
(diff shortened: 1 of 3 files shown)


peterjc (approved):
I've not tested the ``.gitattributes`` value, it seems fine for Python files, and should be OK for text test files in general.

PR #3262, opened 2020-09-07
Changed atom has valid element check from PDBIO

JoaoRodrigues:
Realized PR #3250 introduces a bug in PDBIO when writing structures with unknown elements because "X" is not a valid element (in Bio.Data.IUPACData.atom_weights).

I thought of simply checking if element is "X", but I'd argue that the writer module should enforce the format and not necessarily the content. If users want to overload that particular column with whatever value, they should be able to as long as they comply with the format (2 characters max length).

@peterjc, you introduced the element check back in 2009 (at least according to git log!). Do you remember if there was a particular reason in mind back then? I could find a request in the old mailing list/bug tracker to parse elements, but that's it.

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -161,7 +161,7 @@ class PDBIO(StructureIO):
 
         if atom.element:
             element = atom.element.strip().upper()
-            if element.capitalize() not in atom_weights:
+            if element.capitalize() not in atom_weights and element != "X":
                 raise ValueError("Unrecognised element %r" % atom.element)
             element = element.rjust(2)
         else:
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -9,9 +9,6 @@
 # To allow saving of chains, residues, etc..
 from Bio.PDB.StructureBuilder import StructureBuilder
 
-# Allowed Elements
-from Bio.Data.IUPACData import atom_weights
-
 
 _ATOM_FORMAT_STRING = (
     "%s%5i %-4s%c%3s %c%4i%c   %8.3f%8.3f%8.3f%s%6.2f      %4s%2s%2s\n"
@@ -161,8 +158,8 @@ class PDBIO(StructureIO):
 
         if atom.element:
             element = atom.element.strip().upper()
-            if element.capitalize() not in atom_weights:
-                raise ValueError("Unrecognised element %r" % atom.element)
+            if len(element) > 2:
+                raise ValueError("Element string is too long: %r" % atom.element)
             element = element.rjust(2)
         else:
             element = "  "
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -9,6 +9,9 @@
 # To allow saving of chains, residues, etc..
 from Bio.PDB.StructureBuilder import StructureBuilder
 
+# Allowed Elements
+from Bio.Data.IUPACData import atom_weights
+
 
 _ATOM_FORMAT_STRING = (
     "%s%5i %-4s%c%3s %c%4i%c   %8.3f%8.3f%8.3f%s%6.2f      %4s%2s%2s\n"
@@ -158,8 +161,8 @@ class PDBIO(StructureIO):
 
         if atom.element:
             element = atom.element.strip().upper()
-            if len(element) > 2:
-                raise ValueError("Element string is too long: %r" % atom.element)
+            if element.capitalize() not in atom_weights:
+                raise ValueError("Unrecognised element %r" % atom.element)
             element = element.rjust(2)
         else:
             element = "  "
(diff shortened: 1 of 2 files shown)


peterjc:
Trying to find the exact change in the history, this added a sanity test similar to your proposal:

https://github.com/biopython/biopython/commit/453118cbde77c05eb4e4913f8621e61155da0ccb

peterjc:
Found it, 730a2150612aa152c1d856bc5c75198f5ae54f62 - from you @JoaoRodrigues

peterjc:
Here's the text of the original BugZilla issue http://bugzilla.open-bio.org/show_bug.cgi?id=2495 (filed by Hongbo Zhu in April 2008) as recorded in my emails, with triple back-ticks added for legibility on GitHub:

> Hi,
> 
> the current Bio.PDB.PDBParser does not parse column 77-78 from ATOM records
> in PDB files, where element symbols are (usually) stored for ATOM. We
> suggest BioPython to parse this information in the next version. The reasons
> are given as follows:
> 
> 1. The current remediated PDB format requires these symbols to be always
> present ( http://www.wwpdb.org/documentation/format3.1-20080211.pdf ),
> though in old PDB files (v2.3), these symbols are sometimes missing.
> 
> 2. In some cases it is not straightforward, if not impossible, to recognize
> hydrogen atoms by their identifiers in the remediated PDB files. e.g. in
> 1AWW,
> 
> ATOM    378 HD11 LEU A  25      46.755  -3.858   0.453  1.00  0.00
> H
> ATOM    379 HD12 LEU A  25      47.178  -2.160   0.234  1.00  0.00
> H
> ATOM    380 HD13 LEU A  25      47.054  -3.226  -1.165  1.00  0.00
> H
> ATOM    381 HD21 LEU A  25      49.453  -1.483   0.307  1.00  0.00
> H
> ATOM    382 HD22 LEU A  25      50.714  -2.537  -0.327  1.00  0.00
> H
> ATOM    383 HD23 LEU A  25      49.413  -1.984  -1.381  1.00  0.00
> H
> 
> In this PDB entry, chemical symbols (H) are not right justified in column
> 13-14 for hydrogen identifiers like for other elements. A bit extra work is
> required to figure it out.
> 
> What's more, sometimes it's even impossible to distinguish hydrogen from
> mercury without columns 77-78. From the PDB entry format description version
> 2.1:
> "Hydrogen naming sometimes conflicts with IUPAC conventions. For example, a
> hydrogen named HG11 in columns 13 - 16 is differentiated from a mercury atom
> by the element symbol in columns 77 - 78. Columns 13 - 16 present a unique
> name for each atom."
> 
> Therefore we strongly suggest PDBParser to cover column 77-78 for
> ATOM/HETATM records. We have looked at relevant code and it seems three
> files (Atom.py, PDBParser.py, StructureBuilder.py) needed to be revised
> marginally for integrating this update:
> 
> 1). in Atom.py CVS Revision 1.18
> 
> line 17: add one parameter "element" to the function Atom::__init__(...)
> ```
>     def __init__(self, name, coord, bfactor, occupancy, altloc, fullname,
> serial_number, element):
> ```
> 
> line 61: add line
> ```
>     self.element = element
> ```
> 
> add a set method:
> ```
>     def set_element(self, element):
>         self.element = element
> ```
> 
> add a public method:
> ```
>     def get_element(self):
>         return self.element
> ```
> 
> 
> 2). in PDBParser.py CVS Revision 1.20
> 
> line 161: add one line to parse element symbol in function
> ```
> PDBParser::_parse_coordinates(self, coords_trailer)
>     element=line[76:78].strip()
> ```
> 
> line 182: add one more parameter to init_atom():
> ```
>     structure_builder.init_atom(name, coord, bfactor, occupancy, altloc,
> fullname, serial_number, element)
> ```
> 
> 
> 3). in StructureBuilder.py CVS Revision 1.16
> 
> line 158: add one parameter "element" to the function
> ```
>     StructureBuilder::init_atom(self, name, coord, b_factor, occupancy,
> altloc, fullname, serial_number=None, element='')
> ```
> 
> line 190: add "element" to the initialization of Atom instance.
> ```
>     atom=self.atom=myAtom(name, coord, b_factor, occupancy, altloc,
> fullname, serial_number, element)
> ```

JoaoRodrigues:
Ah! I felt this was somewhat familiar but my git-fu is decidely worse than yours :) Thanks for checking! I think we can revert back to what we had before. 10 years and a PhD later, your initial proposal made more sense!

peterjc:
Checking for a known element seems reasonable to me, so I'd also be happy with allowing a known element or `X` as an unknown. I will defer to you and anyone else on Biopython's PDB team who wants to comment.

JoaoRodrigues:
That was my first idea, but I thought it might look clunky. I'll go with that to keep the current behavior and just work around the 'X' elements. Don't want to leave the 'bug' open for much longer.

JoaoRodrigues:
Also added tests to cover this, as it wasn't covered before.

peterjc (approved):
Looks good - I'd suggest a squash-and-merge to keep the history simple here.

JoaoRodrigues:
Merged! Thanks for the patience :)

PR #3266, opened 2020-09-09
Added methods to remove children of disordered Entities

JoaoRodrigues:

Disordered entities like DisorderedAtom and DisorderedResidues allow adding children to them but not removing. Removing is useful when editing structures. This PR adds methods to allow that to happen and simplifies, for instance, removing altlocs from a structure:

```python
s = ...  # structure
for atom in s.get_atoms():
    if atom.is_disordered():
        altlocs = list(atom.child_dict)
        altlocs.remove(atom.selected_child)
        for a in altlocs:
            atom.disordered_remove(a)
```

The methods do not remove the disordered entity completely even if it is empty. I thought that would be better left to the user to do it explicitly from the parent. They do change the `__repr__` of the disordered entity to signal that they are empty.

### DIFF
--- a/Bio/PDB/Atom.py
+++ b/Bio/PDB/Atom.py
@@ -7,8 +7,9 @@
 
 """Atom class, used in Structure objects."""
 
-import warnings
 import copy
+import sys
+import warnings
 
 import numpy as np
 
@@ -481,7 +482,7 @@ class DisorderedAtom(DisorderedEntityWrapper):
 
         """
         # TODO - make this a private attribute?
-        self.last_occupancy = -999999
+        self.last_occupancy = -sys.maxsize
         DisorderedEntityWrapper.__init__(self, id)
 
     # Special methods
@@ -492,7 +493,10 @@ class DisorderedAtom(DisorderedEntityWrapper):
 
     def __repr__(self):
         """Return disordered atom identifier."""
-        return "<Disordered Atom %s>" % self.get_id()
+        if self.child_dict:
+            return "<DisorderedAtom %s>" % self.get_id()
+        else:
+            return "<Empty DisorderedAtom %s>" % self.get_id()
 
     # This is a separate method from Entity.center_of_mass since DisorderedAtoms
     # will be unpacked by Residue.get_unpacked_list(). Here we allow for a very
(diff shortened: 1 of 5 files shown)


peterjc:
Are there pre-existing similar methods on the other objects for removing children? How would you remove a normal atom from a chain for example?

JoaoRodrigues:
There are. Entity has `detach_child()` that works both directly and via `__delitem__`. So you can remove an atom from a residue by doing `del residue["CA"]` for instance.

Lucioric2000:
I think that in the above example, the calling of altlocs.remove(atom.selected_child) is unnecesary, because altlocs is only a list not stated to be used afterwards. The object tree edition is done in the for cycle.

JoaoRodrigues:
>I think that in the above example, the calling of altlocs.remove(atom.selected_child) is unnecesary, because altlocs is only a list not stated to be used afterwards. The object tree edition is done in the for cycle.

I had a tiny typo in the for. The idea was to show how to remove all _other_ altlocs but the primary one, thus the `remove`. Is the syntax clear for you?

peterjc:
Thanks for clarifying the thoughts behind the method naming. That makes sense. I think that's all I can usefully offer for a review - who else would you suggest for a more informed second opinion? Maybe @etal?

JoaoRodrigues:
Without wanting to bother him too much (I've been pestering both of you lately with all these PRs...), @jgreener64 would be a good reviewer if @etal is not available!

jgreener64:
I'm not so familiar with the ways to modify entities or the internals of the disorder code, so am not sure I'll be able to thoroughly review this. Seems okay from a quick look though.

On a separate note, possibly for another PR, it would be good to add some material to the docs on adding or removing entities to structure objects.

JoaoRodrigues:
Thanks @jgreener64 ! 

Indeed. I was thinking about that this weekend - at some point in the past we discussed moving the tutorials to Jupyter notebooks. I remember @tiagoantao was leading this and indeed, there's a [repo](https://github.com/tiagoantao/biopython-notebook). Given the pervasiveness of conda and jupyter notebooks nowadays, I would be tempted to revist this approach to illustrate a bunch of examples (basic and not so basic) of our codebase. @peterjc , there are two folders in `Docs`, `cookbook` and `examples`. Could we maybe consider merging these two in a single folder (either called examples or cookbook, I like the latter) and moving the examples to notebook format?

Related: the "Going 3D" section of the tutorial could use a little reorganization and update.

JoaoRodrigues:
Meanwhile, I will merge this. I tested this branch on a few thousand PDB files (~8000) and there seem to be no outstanding issues.

peterjc:
I think (without checking the history) that the cookbook folder was superceded by the cookbook category on the then wiki based website: https://biopython.org/wiki/Category:Cookbook - I think ``Doc/cookbook`` could be retired, note we've talked about deprecating the restriction code anyway. This would also solve the confusion with two things called cookbook.

JoaoRodrigues:
Thanks. What do you think about moving these (including the recipes on the wiki/website) to a notebook format and storing them in the main source tree?

peterjc:
Does it have to be the same repository?

Could we can run automated continuous integration testing on the notebooks? Static documentation without tests came become out of date surprisingly often (and things like doctests and testing examples in the tutorial often catches gaps in our main test coverage).

JoaoRodrigues:
Nope, it can be a separate one. We can run CI on the notebooks, I did exactly the same for another repo of mine. I'll start with Bio.PDB and then we can see how to proceed? Shall I make a repo myself and maybe later transfer it or create one directly under the biopython org?

peterjc:
If you want to make one under https://github.com/biopython/ that sounds sensible to me, but by all means experiment first on your own repository first if you'd rather.

peterjc:
Having notebooks in a separate repository would make it easy to test both the latest Biopython master branch and one or more recent releases - that would be a plus.

JoaoRodrigues:
👍 Thank you! I'll give it a shot then on my own account first not to pollute the main account :)

PR #3268, opened 2020-09-10
default phyml test to single threaded

emollier:
By default the phyml test runs a MPI variant of the program that makes
the assumption of having access to an X server.  Since it may not be
the case on build servers, this outputs the message "No protocol
specified" in standard error, triggering a failure of the test.

This patch sets the variable PHYMLCPUS, if it is not in use, in order
to run phyml tests with the single threaded variant.  It is still
possible to override by setting with the environment variable if one
wishes to test explicitly the parallel case in the future.

### DIFF
--- a/Tests/test_phyml_tool.py
+++ b/Tests/test_phyml_tool.py
@@ -43,7 +43,7 @@ class AppTests(unittest.TestCase):
 
     def test_phyml(self):
         """Run PhyML using the wrapper."""
-        # Configure phyml to run in single threaded mode by default
+        # Stabilize phyml tests by running in single threaded mode by default
         if not os.getenv("PHYMLCPUS"):
             os.putenv("PHYMLCPUS", "1")
         cmd = PhymlCommandline(phyml_exe, input=EX_PHYLIP, datatype="aa")
--- a/Tests/test_phyml_tool.py
+++ b/Tests/test_phyml_tool.py
@@ -43,6 +43,7 @@ class AppTests(unittest.TestCase):
 
     def test_phyml(self):
         """Run PhyML using the wrapper."""
+        # Configure phyml to run in single threaded mode by default
         if not os.getenv("PHYMLCPUS"):
             os.putenv("PHYMLCPUS", "1")
         cmd = PhymlCommandline(phyml_exe, input=EX_PHYLIP, datatype="aa")
--- a/Tests/test_phyml_tool.py
+++ b/Tests/test_phyml_tool.py
@@ -43,8 +43,8 @@ class AppTests(unittest.TestCase):
 
     def test_phyml(self):
         """Run PhyML using the wrapper."""
-        if not os.getenv('PHYMLCPUS'):
-            os.putenv('PHYMLCPUS', '1')
+        if not os.getenv("PHYMLCPUS"):
+            os.putenv("PHYMLCPUS", "1")
         cmd = PhymlCommandline(phyml_exe, input=EX_PHYLIP, datatype="aa")
         # Smoke test
         try:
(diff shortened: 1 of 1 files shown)


emollier:
Hum, given the effects on code coverage, I wonder if the appropriate approach wouldn't be to filter out the known flaky error message after all, or fix MPI...

peterjc:
What code coverage numbers are you looking at? AppVeyor isn't installing phyml, but TravisCI should be via an apt package.

emollier:
Hi Peter,

I haven't looked at a figure in particular, but the overall
result.  I haven't had much occasion to work with coverage tools
so was a bit worried of the big blobs of red, and the 2% drop;
I will trust your judgement.

Kind Regards,
Étienne.

peterjc:
Which big red blobs and 2% drop?

https://github.com/biopython/biopython/runs/1099147542 and https://github.com/biopython/biopython/runs/1099147541 (the two links I see from the green tick second commit) say nothing changed.

However, https://github.com/biopython/biopython/runs/1099055339 from the first commit's red cross failed run says coverage went down - but that commit didn't run the any of the tests on Linux because the style test failed.

Tests/test_phyml_tool.py
```diff
@@ -43,6 +43,8 @@ class AppTests(unittest.TestCase):
 
     def test_phyml(self):
         """Run PhyML using the wrapper."""
+        if not os.getenv("PHYMLCPUS"):
+            os.putenv("PHYMLCPUS", "1")
```
peterjc, line 47: Took me a while to see this said "PHYML CPUs = 1", meaning single threaded. Maybe add a one line comment here explaining why you want it single threaded if not configured otherwise?

emollier: Good point, rereading myself the day after, I find it unclear as well.  I push a descriptive comment now.  Thanks for your review.

emollier:
Hi Peter,

> However, https://github.com/biopython/biopython/runs/1099055339 from the first commit's red cross failed run says coverage went down - but that commit didn't run the any of the tests on Linux because the style test failed.

Ah, yes, this is the result I was seeing on my side (-2.60%).  The grid had much more red-ish squares yesterday, so I might have freaked out a bit.  If I understand correctly, with the properly written test, the coverage remains unchanged, so no big deal actually.

Sorry for the confusion and thanks for your explanation,
Kind Regards,
Étienne.

peterjc:
OK, so you were seeing a coverage drop from (Linux  + Windows tests) to just (Windows tests) when the Linux tests were skipped.

peterjc:
I should probably have asked earlier, but is ``PHYMLCPUS`` documented anywhere? Is it specific to the MPI variant because I don't see it used on https://github.com/stephaneguindon/phyml

emollier:
Ouch, a good thing you ask!  On my system, this environment variable is documented in _/usr/share/doc/phyml/README.Debian_ which, as the name states, is Debian specific.  I'm not seeing an easy way to select the non MPI variant of phyml at _run_ time, since having MPI support seems decided at _build_ time in phyml source code; the Debian package builds the two variants and allows one to select the variant depending on the cores count.

So a more appropriate approach would be to filter out the message somehow.

peterjc:
Even if it is Debian specific, a lot of people use Debian. We should probably tweak the comment line though.

emollier:
Agreed, this also applies to derivatives, which are numerous.  Still, I feel a bit uncomfortable should this issue happen outside Debian context.  I'm pushing a small note to indicate where this PHYMLCPUS environment variable is applicable, it should probably do for now.

peterjc (approved):

peterjc:
Merged, thank you.

Issue #3125, opened 2020-07-23
Parsing non-protein PDF and mmCIF files with SeqIO problematic


peterjc:
This came up unexpectedly during #3121, where we noticed that parsing PDB files via SeqIO assumed them to be protein.

However, PDB files can contain RNA, DNA, protein (or mixes). e.g. https://www.rcsb.org/structure/6BNA is DNA (only)

Ideally each sequence returned via SeqIO would be explicitly labelled with the appropriate molecule type (DNA, RNA, or protein).

Using Biopython 1.77 (to avoid the current code flux with alphabet objects) and the "pdb-seqres" format:

```
$ python -c "from Bio import SeqIO; print(list(SeqIO.parse('6bna.pdb', 'pdb-seqres')))" 
[SeqRecord(seq=Seq('C', ProteinAlphabet()), id='6BNA:A', name='6BNA:A', description='PDB:6BNA 6BNA', dbxrefs=['PDB:6BNA', 'PDB:6BNA']), SeqRecord(seq=Seq('C', ProteinAlphabet()), id='6BNA:B', name='6BNA:B', description='PDB:6BNA 6BNA', dbxrefs=['PDB:6BNA', 'PDB:6BNA'])]
```

```
$ python -c "from Bio import SeqIO; print(list(SeqIO.parse('6bna.cif', 'cif-seqres')))" 
[SeqRecord(seq=Seq('C', ProteinAlphabet()), id='6BNA:A', name='6BNA:A', description='PDB:6BNA 6BNA', dbxrefs=['PDB:6BNA', 'PDB:6BNA']), SeqRecord(seq=Seq('C', ProteinAlphabet()), id='6BNA:B', name='6BNA:B', description='PDB:6BNA 6BNA', dbxrefs=['PDB:6BNA', 'PDB:6BNA'])]
```

i.e. Two records with sequence "C" but both labelled as protein. Same here using "pdb-atom" format (but with discontinuous warnings):

```
$ python -c "from Bio import SeqIO; print(list(SeqIO.parse('6bna.pdb', 'pdb-atom')))" 
/Users/xxx/miniconda3/lib/python3.7/site-packages/Bio/PDB/StructureBuilder.py:92: PDBConstructionWarning: WARNING: Chain A is discontinuous at line 973.
  PDBConstructionWarning,
[...]

PR #3275, opened 2020-09-13
Trim whitespace of PDB residue names when building residue.id

### LINKED ISSUE #3272: PDBParser and MMCIFParser assign HETATM ids differently
This issue comes from the benchmark in #3230. 

The first item in the residue id of HETATMs is `"H_" + res.resname`. Because the PDB format is a column-delimited format, we don't strip the empty spaces around residue names with less than 3 characters (e.g. MG) and end up with something like `H_ MG` in the id. The MMCIFParser, on the other hand, does strip this whitespace so we end up with `H_MG` for the same atom. This discrepancy makes it impossible to compare two structures in mmCIF and PDB formats - we get a lot of "mismatches".

A simple fix would be to strip the whitespace of the residue name in PDBParser and, if necessary, adapt PDBIO to pad/align residue names accordingly when writing to file.

I'm leaving this open for a little bit with a tag for newcomers since it's not a pressing issue and might be a good first issue for people looking into contributing to open-source.

Small test sample:

model.cif
```
data_1VQL
# 
loop_
_atom_site.group_PDB 
_atom_site.id 
_atom_site.type_symbol 
_atom_site.label_atom_id 
_atom_site.label_alt_id 
_atom_site.label_comp_id 
_atom_site.label_asym_id 
_atom_site.label_entity_id 
_atom_site.label_seq_id 
_atom_site.pdbx_PDB_ins_code 
_atom_site.Cartn_x 
_atom_site.Cartn_y 
_atom_site.Cartn_z 
_atom_site.occupancy 
_atom_site.B_iso_or_equiv 
_atom_site.pdbx_formal_charge 
_atom_site.auth_seq_id 
_atom_site.auth_comp_id 
_atom_site.auth_asym_id 
_atom_site.auth_atom_id 
_atom_site.pdbx_PDB_model_num 
HETATM 91426 MG MG    . MG  XL 33 .    ? 27.783  125.521 83.164  1.00 41.37  ? 8109 MG  Y MG    1 
#
```

model.pdb
```
HETATM91458 MG    MG Y8109      27.783 125.521  83.164  1.00 41.37          MG  
```

script.py
```python
from Bio.PDB import PDBParser, MMCIFParser

p_pdb = PDBParser(QUIET=1)
p_cif = MMCIFParser(QUIET=1)

s_pdb = p_pdb.get_structure("", "model.pdb")
s_cif = p_cif.get_structure("", "model.cif")

atom_pdb = [a for a in s_pdb.get_atoms()][0]
atom_cif = [a for a in s_cif.get_atoms()][0]

assert atom_pdb.get_full_id() == atom_cif.get_full_id()  # this should be True.
```

sgalpha01:

Closes #3272

### DIFF
--- a/Bio/PDB/StructureBuilder.py
+++ b/Bio/PDB/StructureBuilder.py
@@ -118,7 +118,7 @@ class StructureBuilder:
         if field != " ":
             if field == "H":
                 # The hetero field consists of H_ + the residue name (e.g. H_FUC)
-                field = "H_" + resname.strip()
+                field = "H_" + resname
         res_id = (field, resseq, icode)
         if field == " ":
             if self.chain.has_id(res_id):
--- a/Bio/PDB/StructureBuilder.py
+++ b/Bio/PDB/StructureBuilder.py
@@ -118,7 +118,7 @@ class StructureBuilder:
         if field != " ":
             if field == "H":
                 # The hetero field consists of H_ + the residue name (e.g. H_FUC)
-                field = ("H_" + resname).replace(" ", "")
+                field = "H_" + resname.strip()
         res_id = (field, resseq, icode)
         if field == " ":
             if self.chain.has_id(res_id):
--- a/Bio/PDB/StructureBuilder.py
+++ b/Bio/PDB/StructureBuilder.py
@@ -118,7 +118,7 @@ class StructureBuilder:
         if field != " ":
             if field == "H":
                 # The hetero field consists of H_ + the residue name (e.g. H_FUC)
-                field = "H_" + resname
+                field = ("H_" + resname).replace(" ", "")
         res_id = (field, resseq, icode)
         if field == " ":
             if self.chain.has_id(res_id):
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -28,11 +28,10 @@ def _res2aacode(residue, undef_code="X"):
 
     Non-amino acid are returned as "X".
     """
-    return (
-        _aa3to1_dict.get(residue, undef_code)
-        if isinstance(residue, str)
-        else _aa3to1_dict.get(residue.resname, undef_code)
-    )
+    if isinstance(residue, str):
+        return _aa3to1_dict.get(residue, undef_code)
+
+    return _aa3to1_dict.get(residue.resname, undef_code)
 
 
 def AtomIterator(pdb_id, structure):
@@ -58,7 +57,7 @@ def AtomIterator(pdb_id, structure):
         residues = [
             res
             for res in chain.get_unpacked_list()
-            if _res2aacode(res.get_resname().upper(), undef_code="X") != "X"
+            if _res2aacode(res.get_resname().upper()) != "X"
         ]
         if not residues:
             continue
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -12,20 +12,24 @@ import warnings
 
 from Bio import BiopythonParserWarning
 from Bio.Data.SCOPData import protein_letters_3to1
+from Bio.Data.IUPACData import protein_letters_3to1_extended
 from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
 from .Interfaces import SequenceIterator
 
 
 def _res2aacode(residue, undef_code="X"):
-    """Return a residue's type as a one-letter code.
+    """Return the one-letter amino acid code from the residue name.
 
-    Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
+    Non-amino acid are returned as "X".
     """
     onecode = protein_letters_3to1
     onecode.update((k, v) for k, v in {"XAA": "X", "XLE": "J", "PYL": "O"}.items())
 
-    return onecode.get(residue, undef_code)
+    if isinstance(residue, str):
+        return onecode.get(residue, undef_code)
+
+    return onecode.get(residue.resname, undef_code)
 
 
 def AtomIterator(pdb_id, structure):
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -11,11 +11,16 @@ import collections
 import warnings
 
 from Bio import BiopythonParserWarning
-from Bio.Data.SCOPData import protein_letters_3to1
-from Bio.Data.IUPACData import protein_letters_3to1_extended
+from Bio.Data.SCOPData import protein_letters_3to1 as scop_3to1
+from Bio.Data.IUPACData import protein_letters_3to1_extended as iupac_3to1_ext
 from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
 from .Interfaces import SequenceIterator
+
+
+_aa3to1_dict = {}
+_aa3to1_dict.update(iupac_3to1_ext)
+_aa3to1_dict.update(scop_3to1)
 
 
 def _res2aacode(residue, undef_code="X"):
@@ -23,13 +28,11 @@ def _res2aacode(residue, undef_code="X"):
 
     Non-amino acid are returned as "X".
     """
-    onecode = protein_letters_3to1
-    onecode.update((k, v) for k, v in {"XAA": "X", "XLE": "J", "PYL": "O"}.items())
-
-    if isinstance(residue, str):
-        return onecode.get(residue, undef_code)
-
-    return onecode.get(residue.resname, undef_code)
+    return (
+        _aa3to1_dict.get(residue, undef_code)
+        if isinstance(residue, str)
+        else _aa3to1_dict.get(residue.resname, undef_code)
+    )
 
 
 def AtomIterator(pdb_id, structure):
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -49,7 +49,8 @@ def AtomIterator(pdb_id, structure):
         residues = [
             res
             for res in chain.get_unpacked_list()
-            if seq1(res.get_resname().upper(), custom_map=protein_letters_3to1) != "X"
+            if seq1(res.get_resname().upper(), custom_map=protein_letters_3to1)
+            not in ("X", "")
         ]
         if not residues:
             continue
--- a/Bio/SeqIO/PdbIO.py
+++ b/Bio/SeqIO/PdbIO.py
@@ -17,6 +17,17 @@ from Bio.SeqRecord import SeqRecord
 from .Interfaces import SequenceIterator
 
 
+def _res2aacode(residue, undef_code="X"):
+    """Return a residue's type as a one-letter code.
+
+    Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
+    """
+    onecode = protein_letters_3to1
+    onecode.update((k, v) for k, v in {"XAA": "X", "XLE": "J", "PYL": "O"}.items())
+
+    return onecode.get(residue, undef_code)
+
+
 def AtomIterator(pdb_id, structure):
     """Return SeqRecords from Structure objects.
 
@@ -34,23 +45,13 @@ def AtomIterator(pdb_id, structure):
     See Bio.SeqIO.PdbIO.PdbAtomIterator and Bio.SeqIO.PdbIO.CifAtomIterator for
     details.
     """
-    from Bio.SeqUtils import seq1
-
-    def restype(residue):
-        """Return a residue's type as a one-letter code.
-
-        Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
-        """
-        return seq1(residue.resname, custom_map=protein_letters_3to1)
-
     model = structure[0]
     for chn_id, chain in sorted(model.child_dict.items()):
         # HETATM mod. res. policy: remove mod if in sequence, else discard
         residues = [
             res
             for res in chain.get_unpacked_list()
-            if seq1(res.get_resname().upper(), custom_map=protein_letters_3to1)
-            not in ("X", "")
+            if _res2aacode(res.get_resname().upper(), undef_code="X") != "X"
         ]
         if not residues:
             continue
@@ -68,7 +69,7 @@ def AtomIterator(pdb_id, structure):
             for i, pregap, postgap in gaps:
                 if postgap > pregap:
                     gapsize = postgap - pregap - 1
-                    res_out.extend(restype(x) for x in residues[prev_idx:i])
+                    res_out.extend(_res2aacode(x) for x in residues[prev_idx:i])
                     prev_idx = i
                     res_out.append("X" * gapsize)
                 else:
@@ -78,14 +79,14 @@ def AtomIterator(pdb_id, structure):
                     )
                     # Keep the normal part, drop the out-of-order segment
                     # (presumably modified or hetatm residues, e.g. 3BEG)
-                    res_out.extend(restype(x) for x in residues[prev_idx:i])
+                    res_out.extend(_res2aacode(x) for x in residues[prev_idx:i])
                     break
             else:
                 # Last segment
-                res_out.extend(restype(x) for x in residues[prev_idx:])
+                res_out.extend(_res2aacode(x) for x in residues[prev_idx:])
         else:
             # No gaps
-            res_out = [restype(x) for x in residues]
+            res_out = [_res2aacode(x) for x in residues]
         record_id = "%s:%s" % (pdb_id, chn_id)
         # ENH - model number in SeqRecord id if multiple models?
         # id = "Chain%s" % str(chain.id)
@@ -153,10 +154,6 @@ class PdbSeqresIterator(SequenceIterator):
 
     def iterate(self, handle):
         """Iterate over the records in the PDB file."""
-        # Late-binding import to avoid circular dependency on SeqIO in Bio.SeqUtils
-        # Not sure if this is really needed; Python can handle circular dependencies.
-        from Bio.SeqUtils import seq1
-
         chains = collections.defaultdict(list)
         metadata = collections.defaultdict(list)
 
@@ -175,10 +172,7 @@ class PdbSeqresIterator(SequenceIterator):
                 chn_id = line[11]
                 # Number of residues in the chain (repeated on every record)
                 # num_res = int(line[13:17])
-                residues = [
-                    seq1(res, custom_map=protein_letters_3to1)
-                    for res in line[19:].split()
-                ]
+                residues = [_res2aacode(res) for res in line[19:].split()]
                 chains[chn_id].extend(residues)
             elif rec_name == "DBREF":
                 #  ID code of this entry (PDB ID)
@@ -375,9 +369,6 @@ def CifSeqresIterator(source):
     Note the chain is recorded in the annotations dictionary, and any mmCIF
     _struct_ref_seq entries are recorded in the database cross-references list.
     """
-    # Late-binding import to avoid circular dependency on SeqIO in Bio.SeqUtils
-    from Bio.SeqUtils import seq1
-
     # Only import PDB when needed, to avoid/delay NumPy dependency in SeqIO
     from Bio.PDB.MMCIF2Dict import MMCIF2Dict
 
@@ -399,7 +390,7 @@ def CifSeqresIterator(source):
         records["_pdbx_poly_seq_scheme.asym_id"],
         records["_pdbx_poly_seq_scheme.mon_id"],
     ):
-        mon_id_1l = seq1(mon_id, custom_map=protein_letters_3to1)
+        mon_id_1l = _res2aacode(mon_id)
         chains[asym_id].append(mon_id_1l)
 
     # Build a dict of _struct_ref records, indexed by the id field:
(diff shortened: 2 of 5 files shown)


Bio/PDB/StructureBuilder.py
```diff
@@ -118,7 +118,7 @@ def init_residue(self, resname, field, resseq, icode):
         if field != " ":
             if field == "H":
                 # The hetero field consists of H_ + the residue name (e.g. H_FUC)
-                field = "H_" + resname
+                field = ("H_" + resname).replace(" ", "")
```
JoaoRodrigues, line 121: I would rather use `str.strip` here than `replace`. More explicit. You can call it on the residue name only.

JoaoRodrigues (changes requested):

JoaoRodrigues:
Thanks! This change breaks some tests, so it'd de great if you could fix those too. It should involve only changing the expected results on some tests. The logs for appveyor and Travis should tell you what tests to look at.

I also edited your PR text to make use of the closing keywords (at the end), to link automatically to the issue.

Bio/PDB/StructureBuilder.py
sgalpha01: Can I use `str.lstrip()`? It will be even more explicit. Or, should I simply use `str.strip()`?

JoaoRodrigues: Do you know that spaces always come the left? :)

JoaoRodrigues (commented):
Looks good! 

Going over the fix and the code I realized that also regular `ATOM` lines suffer from this problem. For instance, nucleotides (e.g. `DG`, `DC`) suffer from this same problem. In a structure coming from PDBParser, their residue names are ` DG` vs `DG` when parsed by MMCIFParser. We might tackle both issues by changing resname in `StructureBuilder` to `resname.strip()` the first time it comes through in `init_residue`. Then there's no need for this specific fix downstream.

Also, we should create new tests to check for this behavior. We already have two structures showing this 'problem' in our `Tests/PDB/` folder so we could simply write a new test in `test_PDB_SMCRA` that parses them both and compares them (chain by chain, residue by residue, atom by atom) to ensure they are the same. Would this be something you would be willing to do?

sgalpha01:
@JoaoRodrigues , thanks for reviewing this. I will try my best to do the needed.

sgalpha01:
> Looks good!
> 
> Going over the fix and the code I realized that also regular `ATOM` lines suffer from this problem. For instance, nucleotides (e.g. `DG`, `DC`) suffer from this same problem. In a structure coming from PDBParser, their residue names are ` DG` vs `DG` when parsed by MMCIFParser. We might tackle both issues by changing resname in `StructureBuilder` to `resname.strip()` the first time it comes through in `init_residue`. Then there's no need for this specific fix downstream.
> 
> Also, we should create new tests to check for this behavior. We already have two structures showing this 'problem' in our `Tests/PDB/` folder: `1LCD.pdb` and `1LCD.cif`. We could simply write a new test in `test_PDB_SMCRA` that parses them both and compares them (chain by chain, residue by residue, atom by atom) to ensure they are the same. Would this be something you would be willing to do?

Hello @JoaoRodrigues , I didn't understand the issue. Can you please elaborate more?
Also, did you mean the following changes:
``` Python3
...
        if field != " ":
            if field == "H":
                # The hetero field consists of H_ + the residue name (e.g. H_FUC)
                field = "H_" + resname.strip()
        res_id = (field, resseq, icode)
        if field == " ":
            if self.chain.has_id(res_id):
...
```
to something like:
``` Python3
...
        resname = resname.strip()  # Strips the whitespaces around resname

        if field != " ":
            if field == "H":
                # The hetero field consists of H_ + the residue name (e.g. H_FUC)
                field = "H_" + resname
        res_id = (field, resseq, icode)
        if field == " ":
            if self.chain.has_id(res_id):
...
```
I did so, but it broke two tests:
```
$ python run_tests.py --offline test_SeqIO_PdbIO.py
Skipping any tests requiring internet access
Python version: 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]
Operating system: nt win32
test_SeqIO_PdbIO ... FAIL
FAIL: test_atom_noheader (test_SeqIO_PdbIO.TestPdbAtom)
Parse a PDB with no HEADER line.
  File "D:\Development\biopython\biopython\Tests\test_SeqIO_PdbIO.py", line 167, in test_atom_noheader
    self.assertEqual(len(chains), 1)
AssertionError: 3 != 1

Ran 1 test in 5.078 seconds

FAILED (failures = 1)
```
and
```
$ python run_tests.py --offline test_SeqIO.py
Skipping any tests requiring internet access
Python version: 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]
Operating system: nt win32
test_SeqIO ... FAIL
FAIL: test_pdb_atom3 (test_SeqIO.TestSeqIO)
  File "D:\Development\biopython\biopython\Tests\test_SeqIO.py", line 5463, in test_pdb_atom3
    self.perform_test(
  File "D:\Development\biopython\biopython\Tests\test_SeqIO.py", line 478, in perform_test
    self.assertEqual(
AssertionError: 3 != 1 : Found 3 records but expected 1

Ran 1 test in 30.407 seconds

FAILED (failures = 1)
```

JoaoRodrigues:
Yes, I was even going to suggest placing that call to `.strip()` in `Bio/PDB/PDBParser.py` because it's the extra spaces are a 'feature' of the PDB format.

In summary, the PDB format is column-based so the residue name for each ATOM/HETATM record is located at columns 18, 19, and 20 (see [here for details](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM)). When you have a residue with a short name, such as `DG`, our parser still extracts those three characters from the line, making it ` DG` instead. Other formats (namely mmCIF) don't have this column-based format, so there are never empty spaces in the residue names. As we have it now, we cannot compare two structures easily because of issues like these. That's what I was trying to address with this issue. Is this clearer?

As for the test failures, this change might have consequences I did not antecipate. Both failures are in the `SeqIO` module. There's a type of reader (`pdb-atom`) that parses a PDB file and extracts the amino acid sequence. It's a bit of a rabbit hole, but the issue is with [this line](https://github.com/biopython/biopython/blob/master/Bio/SeqUtils/__init__.py#L307) in `Bio.SeqUtils.seq1`, which expects the input string to have a length that is divisible by 3. If you give it `DG` (length 2), it returns an empty list, which then eventually causes the error you are seeing. Since this is a very specific case, we can fix the call in `SeqIO/PdbIO.py` to workaround this issue. If you are not comfortable with fixing these other modules, just push the change to `PDBParser` and I can help you out in the other modules.

sgalpha01:
Hey @JoaoRodrigues , now the issue is clearer to me. I've pushed the changes to `PDBParser.py`. And, thank you for being so cooperative.

Also, may I bother you again as I will definitely need help with the other modules?

JoaoRodrigues:
Hi @sgalpha01 of course, thanks for the fixes. Have a go at fixing the two failing tests and let us know if you need help. This isn't as easy as I thought it would be, sorry about that.

sgalpha01:
I've created a test file, taking ideas from `test_SeqIO_PdbIO.py`, to check the issue.

test.py
``` Python3
import warnings

from Bio import BiopythonParserWarning, SeqIO
from Bio.PDB.PDBExceptions import PDBConstructionWarning


with warnings.catch_warnings():
    warnings.simplefilter("ignore", PDBConstructionWarning)
    warnings.simplefilter("ignore", BiopythonParserWarning)
    chains = list(SeqIO.parse("1LCD.pdb", "pdb-atom"))


print(chains)
```

This generates the following output:
``` Python3
[SeqRecord(seq=Seq('MKPVTLYDVAEYAGVSYQTVSRVVNQASHVSAKTREKVEAAMAELNYIPNR'), id='????:A', name='<unknown name>', description='????:A', dbxrefs=[]), 
SeqRecord(seq=Seq(''), id='????:B', name='<unknown name>', description='????:B', dbxrefs=[]), 
SeqRecord(seq=Seq(''), id='????:C', name='<unknown name>', description='????:C', dbxrefs=[])]
```

The test files can easily be resolved by modifying [this line](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Tests/test_SeqIO_PdbIO.py#L165) to `chains = list(SeqIO.parse("1LCD.pdb", "pdb-atom"))[:1]`. But, this doesn't solve the purpose. We need the chains list to be originally of length 1, right? I got the fact that this issue is coming due to the line you mentioned above, but can you give a little bit of clue about which module do I need to modify?

JoaoRodrigues:
Hi @sgalpha01 

I would suggest you follow the code/imports. Here's what I did when I checked:

1. That test calls `SeqIO.parse` with a file (`1LCD.pdb`) and a second argument `"pdb-atom"`. 
2. If you go to [Bio/SeqIO/__init__.py](https://github.com/biopython/biopython/blob/master/Bio/SeqIO/__init__.py) (since the import in the test is simply `from Bio import SeqIO`) and search for 'pdb-format' you see that it points to a class in a submodule: `Bio/SeqIO/PdbIO.PdbAtomIterator` [link](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L249). 
2. That function is actually quite short and besides parsing the PDB file, calls another function called `AtomIterator` defined at the top of the file, [here](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L20). 
3. [On line 54](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L54), there's an if statement that should skip processing a chain of the PDB file if there are no residues detected. So let's check how we find residues.
4. There's a list comprehension defined right before that if, [lines 49 to 53](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L49-L53) that uses a function called `seq1` to convert amino acid three-letter codes to one-letter codes: `seq1`. If the result of calling that function is _not_ "X", then the residue is accepted. Combining this with the if on line 54, we understand that `seq1` is the "gatekeeper" function here. So let's check it out.
5. `seq1` [is defined in Bio/SeqUtils/__init__.py](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqUtils/__init__.py#L256) and is quite short. Basically it expects a string or a list that has N elements, where N is a multiple of 3. Why? Because on line 307 you have a list comprehension that iterates over the input in strides of 3. Try running that function on something that is only 2 characters long, like `DG`. What do you get?
6. Now propagate that result to the other functions in `Bio/SeqIO`. Will that residues list be empty?

peterjc:
Somewhat related, see also #3125.

sgalpha01:
Hi @JoaoRodrigues , I have pushed some changes. Not sure if it will break some other modules. Can you please check it?

JoaoRodrigues:
Hi @sgalpha01 

That's a good fix, but it's masking a bit the problem. 

What about we call `protein_letters_3to1` directly, instead of through `seq1`? Since in this loop we call `seq1` on one residue at a time, it's more efficient to call `protein_letters_3to1` (it's a dictionary) and use the `dict.get` function to return "X" if a match is not found. For consistency, we can make this change in the function `restype` that is defined on [line 39](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L39) and use `restype` in the loop on lines 49-53. 

We then need to check in the rest of `PdbIO.py` if there are other places where `seq1` gets called like this. If there are, it might be worth moving that `restype` function to the global scope (and maybe rename it to `_res2aacode` to be 1) private and 2) more explicit) so we can call it throughout the code.

What do you think?

sgalpha01:
> Hi @sgalpha01
> 
> That's a good fix, but it's masking a bit the problem.
> 
> What about we call `protein_letters_3to1` directly, instead of through `seq1`? Since in this loop we call `seq1` on one residue at a time, it's more efficient to call `protein_letters_3to1` (it's a dictionary) and use the `dict.get` function to return "X" if a match is not found. For consistency, we can make this change in the function `restype` that is defined on [line 39](https://github.com/biopython/biopython/blob/36b5aee5652022d406e4a7529c1d23c525bf5d02/Bio/SeqIO/PdbIO.py#L39) and use `restype` in the loop on lines 49-53.
> 
> We then need to check in the rest of `PdbIO.py` if there are other places where `seq1` gets called like this. If there are, it might be worth moving that `restype` function to the global scope (and maybe rename it to `_res2aacode` to be 1) private and 2) more explicit) so we can call it throughout the code.
> 
> What do you think?

Yes, it would be a lot better. It will also increase the performance, ig. Okay, I'm on it.

sgalpha01:
I've replaced `restype` function to `_res2aacode` with global scope. To make it almost similar to `seq1` function with single residue input, I've written a script to check difference between its default mapping and `protein_letters_3to1` as `custom_map`.

test.py
``` Python3

from Bio.Data.IUPACData import protein_letters_3to1_extended
from Bio.Data.SCOPData import protein_letters_3to1

onecode = {k.upper(): v for k, v in protein_letters_3to1_extended.items()}
for k, v in onecode.items():
    if k not in protein_letters_3to1:
        print(k, v)

```
It gives the following output:
```
XAA X
XLE J
PYL O
```
So, I've added these 3 mappings to the map in `_res2aacode` .

Bio/SeqIO/PdbIO.py
```diff
@@ -17,6 +17,17 @@
 from .Interfaces import SequenceIterator
 
 
+def _res2aacode(residue, undef_code="X"):
+    """Return a residue's type as a one-letter code.
+
+    Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
+    """
+    onecode = protein_letters_3to1
```
JoaoRodrigues, line 25: You can simply make this `onecode = protein_letters_3to1_extended`, saves you the trouble of the update and retains exactly the same result as you'd have with `seq1`

sgalpha01:
Okay, something is weird. I ran the tests on my local machine, and it passed them all.

Bio/SeqIO/PdbIO.py
```diff
@@ -17,6 +17,17 @@
 from .Interfaces import SequenceIterator
 
 
+def _res2aacode(residue, undef_code="X"):
+    """Return a residue's type as a one-letter code.
```
JoaoRodrigues, line 21: I would change the docstring to: "Return the one-letter amino acid code from the residue name.". Type here is not really appropriate.

Bio/SeqIO/PdbIO.py
```diff
@@ -17,6 +17,17 @@
...
+
+    Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
```
JoaoRodrigues, line 23: And I'd write: "Non-amino acid residues return X"

JoaoRodrigues (changes requested):
Very nice! I think we're almost there! Good job catching most of the calls to `seq1` and friends throughout the module!

JoaoRodrigues:
Let's wait for the tests - maybe we missed something.

Bio/SeqIO/PdbIO.py
sgalpha01: But, `protein_letters_3to1_extended` contains a lot lesser mappings than `protein_letters_3to1`. Will it be okay to do so?

JoaoRodrigues:
The difference is in the inputs for `seq1` and the `restype` functions. The first takes a string (essentially, `residue.resname`) and the second one takes a residue (and translates to `residue.resname` internally). There are two options: either we remember where to pass `.resname` and where not to, or we just add a small check to the `_res2aacode` function to handle the two inputs:

```python
def _res2aacode(residue, undef_code="X"):
    """Return a residue's type as a one-letter code.

    Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
    """
    onecode = protein_letters_3to1
    onecode.update((k, v) for k, v in {"XAA": "X", "XLE": "J", "PYL": "O"}.items())

    if isinstance(residue, str):
        return onecode.get(residue, undef_code)

    return onecode.get(residue.resname, undef_code)
```

Since this is a private function, I think it's fine to fix this, what do you think?

sgalpha01:
> The difference is in the inputs for `seq1` and the `restype` functions. The first takes a string (essentially, `residue.resname`) and the second one takes a residue (and translates to `residue.resname` internally). There are two options: either we remember where to pass `.resname` and where not to, or we just add a small check to the `_res2aacode` function to handle the two inputs:
> 
> ```python
> def _res2aacode(residue, undef_code="X"):
>     """Return a residue's type as a one-letter code.
> 
>     Non-standard residues (e.g. CSD, ANP) are returned as 'X'.
>     """
>     onecode = protein_letters_3to1
>     onecode.update((k, v) for k, v in {"XAA": "X", "XLE": "J", "PYL": "O"}.items())
> 
>     if isinstance(residue, str):
>         return onecode.get(residue, undef_code)
> 
>     return onecode.get(residue.resname, undef_code)
> ```
> 
> Since this is a private function, I think it's fine to fix this, what do you think?

I think the latter would be better. In this way, we don't have to remember where to pass `.resname`.

Bio/SeqIO/PdbIO.py
sgalpha01: I've tried using `onecode = protein_letters_3to1_extended`, but it is causing tests to fail.

JoaoRodrigues: That's interesting because it's what `seq1` uses internally. Can you double check that the errors come from this change and not from something else?

sgalpha01: Yes, I've checked. Initially, the `custom_map` argument was set to `protein_letters_3to1` in `restype` function while calling `seq1`. This combined both the dictionaries, and this combined dictionary was then used.

JoaoRodrigues: Ah... I know what's up. This is confusing. We actually have two mappings called `protein_letters_3to1`, one in `Bio.Data.IUPACData` and one in `Bio.Data.SCOPData`. And one uses capitalized residue names while the other one does not.. This is messy. Let's keep it like this then, makes total sense.

JoaoRodrigues: Actually, let's do this: we can import both and merge them at the top of the file:

```python
from Bio.Data.SCOPData import protein_letters_3to1 as scop_3to1
from Bio.Data.IUPACData import protein_letters_3to1_extended as iupac_3to1_ext

...

_aa3to1_dict = {}
_aa3to1_dict.update(iupac_3to1_ext)
_aa3to1_dict.update(scop_3to1)


def _res2aacode(...):  # uses _aa3to1_dict
    ...
```

JoaoRodrigues: This way we get both dictionaries in, just like `seq1` and we keep them in sync with those dicts in other modules (no manual updates here).

sgalpha01:
Everything looks fine now

Bio/SeqIO/PdbIO.py
```diff
@@ -11,12 +11,30 @@
...
+_aa3to1_dict.update(iupac_3to1_ext)
+_aa3to1_dict.update(scop_3to1)
+
+
+def _res2aacode(residue, undef_code="X"):
+    """Return the one-letter amino acid code from the residue name.
+
+    Non-amino acid are returned as "X".
+    """
+    return (
```
JoaoRodrigues, line 31: I would prefer this being more explicit, but it's OK like this as well!

JoaoRodrigues (approved):
I think this is it! Thank you so much for all the changes and edits, great job!

I'll defer to @peterjc on the changes to `SeqIO/PdbIO`. It seems OK to me, but I'm not super familiar with that module.

Bio/SeqIO/PdbIO.py
peterjc: I agree, this version is five lines - an if/else would be four lines, and clearer.

Bio/SeqIO/PdbIO.py
```diff
@@ -34,22 +52,13 @@ def AtomIterator(pdb_id, structure):
...
-        return seq1(residue.resname, custom_map=protein_letters_3to1)
-
     model = structure[0]
     for chn_id, chain in sorted(model.child_dict.items()):
         # HETATM mod. res. policy: remove mod if in sequence, else discard
         residues = [
             res
             for res in chain.get_unpacked_list()
-            if seq1(res.get_resname().upper(), custom_map=protein_letters_3to1) != "X"
+            if _res2aacode(res.get_resname().upper(), undef_code="X") != "X"
```
peterjc, line 61: Having the "X" explicit here and in the private function definition seems redundant and potentially confusing. Here you set the undefined code, but 20 lines later you do not.

I would suggest making the function take two required arguments, and then all the calls to it will be consistent. Or since you always use "X" just hard code it, no second argument?

peterjc (changes requested):
Some stylistic comments, but the changes in SeqIO look fine in principle.

peterjc (approved):

JoaoRodrigues (approved):

JoaoRodrigues:
Thank you @sgalpha01 (and @peterjc  for reviewing)!

Issue #2835, opened 2020-04-11
Use assertEqual etc unittest methods


peterjc:
See #2700 for background, cross reference https://github.com/biopython/biopython/pull/2833#discussion_r406737022 e.g.

```
$ grep assertTrue *.py | grep "=="
common_BioSQL.py:        self.assertTrue(self.db.load(seqiter) == 1)
test_ClustalOmega_tool.py:        self.assertTrue(error.strip() == "" or
...
```

and:

```
$ grep assertFalse *.py | grep "!="
test_Restriction.py:        self.assertFalse(Acc65I != Asp718I)
```

Those should use ``assertEqual``, while:

```
$ grep assertTrue *.py | grep "!="
test_SearchIO_model.py:        self.assertTrue(hit11.id != "alt1")
test_SearchIO_model.py:        self.assertTrue(hit11.id != "alt11")
test_SearchIO_model.py:        self.assertTrue(hit11.id != "alt1")
```

and:

```
$ grep assertFalse *.py | grep "=="
test_PDB.py:        self.assertFalse(struct[0] == struct[1])
test_PDB.py:        self.assertFalse(struct[0] == [])  # __eq__ diff. types
test_PDB.py:        self.assertFalse(struct == model)
test_PDB.py:        self.assertFalse(res1 == res2)
test_PDB_StructureAlignment.py:        self.assertFalse(al.map12 == al.map21)
test_Restriction.py:        self.assertFalse(FormattedSeq(Seq("GATC")) ==
test_Restriction.py:        self.assertFalse(FormattedSeq(Seq("TAGC")) == Seq("TAGC"))
test_Restriction.py:        self.assertFalse(Acc65I == Asp718I)
```

Those shoud use ``self.assertNotEqual``

Not this cleanup work is not specific to dropping Python 3:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertNotEqual

However, some of the older usage probably did require at least Python 2.7:

https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual

peterjc:
See also https://github.com/biopython/biopython/issues/2700#issuecomment-594032554 where the ``A500`` check from ``flake8-assertive`` should catch more than my quick grep searches.

trkohler:
Hi!
Can I work on this? I'd like to help! 
It might be my first open-source issue so if it requires any special knowledge, please let me know

[...]

PR #3276, opened 2020-09-14
changed assertion styling for test_Restriction.py

lacrosse91:

This pull request addresses issue #2835

### DIFF
--- a/Tests/test_Restriction.py
+++ b/Tests/test_Restriction.py
@@ -229,27 +229,27 @@ class EnzymeComparison(unittest.TestCase):
         # Comparison of iso- and neoschizomers
         self.assertEqual(Acc65I, Acc65I)
         self.assertNotEqual(Acc65I, KpnI)
-        self.assertFalse(Acc65I == Asp718I)
+        self.assertEqual(Acc65I == Asp718I, False)
         # self.assertNotEqual(Acc65I, Asp718I) it doesn't work as expected
-        self.assertFalse(Acc65I != Asp718I)
+        self.assertEqual(Acc65I != Asp718I, False)
         self.assertNotEqual(Acc65I, EcoRI)
         self.assertTrue(Acc65I >> KpnI)
         self.assertFalse(Acc65I >> Asp718I)
 
         # Compare length of recognition sites
         self.assertFalse(EcoRI >= EcoRV)
-        self.assertTrue(EcoRV >= EcoRI)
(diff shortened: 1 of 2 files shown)


Tests/test_Restriction.py
```diff
@@ -229,27 +229,27 @@ def test_comparisons(self):
         # Comparison of iso- and neoschizomers
         self.assertEqual(Acc65I, Acc65I)
         self.assertNotEqual(Acc65I, KpnI)
-        self.assertFalse(Acc65I == Asp718I)
+        self.assertEqual(Acc65I == Asp718I, False)
         # self.assertNotEqual(Acc65I, Asp718I) it doesn't work as expected
-        self.assertFalse(Acc65I != Asp718I)
```
peterjc, line 234: This and the line above look like the problem cases, and I think that neither ``self.assertEqual`` nor ``self.assertNotEqual`` will work here? I've not tried that, but I did look at the source code:

https://github.com/python/cpython/blob/master/Lib/unittest/case.py

However, I think the originals using ``self.assertFalse`` are actually clearer than ``self.assertEqual(..., False)``.

I'm going to ask for a second opinion, but I would be inclined to leave these two lines as they are. We could add the special comment ``# noqa: A500`` to tell flake8 to ignore the warning from flake8-assertive (if installed).

@MarkusPiotrowski could you comment?

lacrosse91: Thanks for review.
Should I wait for his response?

peterjc: Wait for now - the second opinion might be different.

MarkusPiotrowski: `Bio.Restriction` overwrites the `==` and `!=` comparisons for some special meaning regarding neo- and isoschizomers. So what is actually tested here are the comparison operators and thus I think that the original tests are indeed clearer.

peterjc: Thanks Markus.

@lacrosse91 could you revert those two lines (and optionally add the ``# noqa: A500`` comment - which should make flake8-assertive happy and document this use of ``assertTrue`` is deliberate).

peterjc (changes requested):
See comments about the ``Acc65I == Asp718I`` and ``Acc65I != Asp718I`` tests.

lacrosse91:
@peterjc and @MarkusPiotrowski 
Thank you for advice.
please review the changes.

peterjc (approved):

PR #3285, opened 2020-09-23
Add version switching to API documentation

### LINKED ISSUE #2906: Version switching on the Sphinx API docs
We now have:

- https://biopython.org/docs/1.74/api/
- https://biopython.org/docs/1.75/api/
- https://biopython.org/docs/1.76/api/
- https://biopython.org/docs/1.77/api/
- ...
- https://biopython.org/docs/dev/api/

The webpage source for these exists here:

https://github.com/biopython/docs

The Sphinx configuration to build them is here:

https://github.com/biopython/biopython/tree/master/Doc/api

Can we add some JavaScript or otherwise like on ReadTheDocs to make it easy to toggle between version from within the browse?

This likely means some changes to our Sphinx configuration (e.g. the common templates), and therefore would require backporting to regenerate the old pages - which would be OK.

This could also resolve #2904.

MarkusPiotrowski:
With this PR the design of our API docs will be slightly changed to include version switching. This has already been applied to the existing API documentations for Biopython 1.74 to 1.78 (e.g. https://biopython.org/docs/latest/api/). 

Discussion on #2906 and https://github.com/biopython/biopython.github.io/issues/175#issuecomment-692629993

Closes #2906


Doc/api/conf.py
```diff
@@ -92,7 +111,9 @@
 }
 
 # To avoid import errors.
-autodoc_mock_imports = ["MySQLdb", "Bio.Restriction.Restriction", "Bio.Alphabet"]
+autodoc_mock_imports = ["MySQLdb", "Bio.Restriction.Restriction"]
+if version > "1.77":
```
peterjc, line 115: That's string comparison - it will only work if we have two digit minor versions, "1.xx".

But like the 1.75 check earlier, it will work for now.

peterjc (approved):

PR #3287, opened 2020-09-23
FixIssue3286

### LINKED ISSUE #3286: BioPython return label of SnapGene Feature, not title
### Setup

system as follows:

```
[GCC 6.3.0 20170516]
CPython
Linux-4.19.0-10-amd64-x86_64-with-debian-10.5
1.78
```
### Expected behaviour

I would like to get the names of all features found in >4000 SnapGene plasmid maps. I would expect the name of the features returned by BioPython to be the same as those returned by SnapGene Server's reportFeatures function.

### Actual behaviour

For some features, the name returned by BioPython is different from that returned by SnapGene Server's reportFeatures function. 

The problematic features are those that have both a title and a label attribute, for instance:

image

In the example above, BioPython reports the label `lac operator` while SnapGene Server returns the 
title `lac repressor encoded by lacI lac operator`. This would not be so much of a problem if I could find `lac repressor encoded by lacI lac operator` anywhere in the relevant SeqFeature object, but I cannot, unless, of course, I missed it.

### Steps to reproduce

I am attaching a relevant example SnapGene file.

[example.zip](https://github.com/biopython/biopython/files/5267092/example.zip)

gouttegd:
This pull request addresses issue #3286.

It changes the way the SnapGene parser handles the _name_ attribute carried by
features in SnapGene files. Previously, the value of that attribute was stored in the
_label_ qualifier, _if_ the feature did not have any _label_ qualifier; if a _label_
qualifier was present, then the value of the _name_ attribute was lost.

With this pull request, the value of the _name_ attribute is either stored in the
_label_ qualifier as previously, or under a new ad-hoc _name_ qualifier if the
feature already has a label that contains a different value. This way, no
information is lost when parsing the file, and a non-standard qualifier is only
used when required.

Closes #3286


Bio/SeqIO/SnapGeneIO.py
```diff
@@ -170,6 +167,16 @@ def _parse_features_packet(length, data, record):
                     qvalues.append(int(value.attributes["int"].value))
             quals[qname] = qvalues
 
+        name = _get_attribute_value(feature, "name")
+        if name:
+            if "label" not in quals:
+                # No explicit label attribute, use the SnapGene name
+                quals["label"] = [name]
```
peterjc, line 174: I don't fully understand why you want to do this here. Is this the rare corner case?

gouttegd: No, that would be the most common case actually.

To summarize, there are three possible cases:
* _name_ attribute present, _label_ qualifier absent (the most common case, at least in the files I have seen): then we store the name as a _label_ qualifier;
* _name_ attribute and _label_ qualifier both present and containing the same value (the second most common case): then we store that value as a _label_ qualifier;
* _name_ attribute and _label_ qualifier both present and containing different values (the “corner case” that gives issue #3286, and [which seems to occur in ~10% of files](https://github.com/biopython/biopython/issues/3286#issuecomment-697589615)): then we store both values, the name being stored in a non-standard _name_ qualifier.

(Theoretically, there could be a fourth case when no _name_ attribute is present, but I have never seen that and I strongly suspect SnapGene would consider that as invalid; still, I’d rather not assume that a _name_ attribute is always present, hence the initial test above.)

peterjc: OK, and this preserves the current/historic behaviour which mapped the name attribute to the label qualifier.

peterjc (approved):

Bio/SeqIO/SnapGeneIO.py
```diff
@@ -170,6 +167,16 @@ def _parse_features_packet(length, data, record):
...
+            elif name not in quals["label"]:
+                # The SnapGene name is different from the label,
```
MarkusPiotrowski, line 176: This works in the example case (where the name is longer than the label) but the other way round (e.g. name = "laqI", label = "lacI repressor") it would discard the name and only keeps the label. Am I right?

MarkusPiotrowski: Thus, we want to check for equality?

gouttegd: We do want to check for equality, and that is what we are doing here. The ``not in`` test is not intended to search for a substring within the qualifier, but to search for a value within the list of values associated with the qualifier (remember that each qualifier contains not a single value but an *array* of values, even if most cases the array contains a single element).

MarkusPiotrowski: My bad, each qualifier is a _list_...

peterjc: Yes, this often catches people out, but it models the worse case of duplicated qualifier keys in GenBank/EMBL files.

MarkusPiotrowski:
@gouttegd Thank you!

PR #3291, opened 2020-10-02
Refactor GenBank tests

Andrey-Raspopov:

Good morning!

I've taken several attempts at refactoring GenBank library, but it is truly a complex task, so I've decided to start from the other end. This commit is focused on moving test data away from the test logic and of course is a subject of a discussion.


peterjc:
What do you think @mdehoon? We started with a print-and-compare test, where a string representation the expected parser output was help in a text file. Then in #2124 you refactored to use the unittest framework with the expected parser output as python datatypes. Here it moves the expected output externally again, but into a JSON file.

If I was doing this from scratch, I'd probably start with a single file where all the parsed values are verified, supplemented by specific tests of corner cases (e.g. strange line wrapping, value-less qualifiers, atypical line breaks in structured comments, expected warnings from bad locations, bad quotes, ...).

PR #3307, opened 2020-10-13
Removed partial codon warning when testing amino acid sequences

kaskales:
Only changed the testing amino acid sequences so the warning isn't triggered. Amino acid sequences don't need to be divisible by 3, only codon sequences.

Trimmed three testing protein_seqs by one amino acid to prevent irrelevant partial codon warning from appearing when testing TranslationError of amino acid Seq() object.

Closes #...

### DIFF
--- a/Tests/test_seq.py
+++ b/Tests/test_seq.py
@@ -978,16 +978,6 @@ class TestTranslating(unittest.TestCase):
         seq = "GTGGCCATTGTAATGGGCCGCTGAAAGGGTGCCCGATAG"
         self.assertEqual("VAIVMGRWKGAR", Seq.translate(seq, table=2, to_stop=True))
 
-    def test_translation_on_proteins(self):
-        """Check translation fails on a protein."""
-        for s in protein_seqs:
-            with self.assertRaises(TranslationError):
-                Seq.translate(s)
-
-            if isinstance(s, Seq.Seq):
-                with self.assertRaises(TranslationError):
-                    s.translate()
-
     def test_translation_of_invalid_codon(self):
         for codon in ["TA?", "N-N", "AC_", "Ac_"]:
             with self.assertRaises(TranslationError):
--- a/Tests/test_seq.py
+++ b/Tests/test_seq.py
@@ -43,11 +43,11 @@ protein_seqs = [
     Seq.Seq("ATCGPK"),
     Seq.Seq("T.CGPK"),
     Seq.Seq("T-CGPK"),
-    Seq.Seq("MEDG-KRXR*"),
+    Seq.Seq("MEDG-KRX*"),
     Seq.MutableSeq("ME-K-DRXR*XU"),
-    Seq.Seq("MEDG-KRXR@"),
+    Seq.Seq("MEDG-KRX@"),
     Seq.Seq("ME-KR@"),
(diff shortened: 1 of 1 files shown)


peterjc:
I don't follow from the commit or pull request why you are changing this. If anything, it makes me wonder if there is a mistake in the tests?

kaskales:
There is no mistake in the test that I see. I was just curious and running the test_seq tests when I noticed a warning,

    BiopythonWarning: Partial codon, len(sequence) not a multiple of three. Explicitly trim the sequence or add trailing N before translation. This may become an error in future.
    BiopythonWarning,

that was being triggered by the `test_translation_on_proteins` test while assertRaises(TranslationError) was being caught. 

I just thought it might make the test pass cleaner if the warning wasn't being triggered, since it seems to be aimed for codon sequences and not about the amino acids being tested. Shortening the test amino acid sequences doesn't change any of the other tests.

MarkusPiotrowski:
From a quick inspection it looks to me the reason that the `test_translation_on_proteins` passes in the first place is because a `TranslationError` is thrown for the wrong reason, i.e. that a 'codon' of three amino acids is not found in the table.
https://github.com/biopython/biopython/blob/63ec6b5b4396b2b44b45efdfe8e28d727895f227/Bio/Seq.py#L2407-L2410

Since we cannot distinguish between nucleic and protein sequences in the `Seq` object any longer, the test which presumably has tested before that a protein cannot be translated, has become obsolete?

peterjc (approved):
Thanks @kaskales, ``test_translation_on_proteins`` was the scope I wanted clarified. Your change makes sense to me.

@MarkusPiotrowski Since the Seq object no longer distinguishes nucleotide versus protein, a protein/peptide "AAA" can be translated just fine but  "MED" fails as it does not make sense as a codon. The test motivation still makes sense to me.

Perhaps the length not a multiple of three warning should happen later in the translation (allowing the exceptions to trigger first). Also perhaps if using ``to_stop=True`` the warning should be skipped too (if an in-frame stop is found)?

MarkusPiotrowski:
> @MarkusPiotrowski Since the Seq object no longer distinguishes nucleotide versus protein, a protein/peptide "AAA" can be translated just fine but "MED" fails as it does not make sense as a codon. The test motivation still makes sense to me.

Exactly. Because protein sequences may be translatable, this test, described as "Check translation fails on a protein." doesn't make sense any longer. If the sequence was "ATG" or "MAS", the test wouldn't fail. 
There is another test, `test_translation_of_invalid_codons`, which could cover the "MED" case.

peterjc:
OK, if ``test_translation_of_invalid_codons`` covers it we can drop ``test_translation_of_invalid_codons`` 👍

kaskales:
Should I delete the `test_translation_on_proteins`, and push it to this PR, or make a new one for this?

peterjc:
Whichever is easier for you.

kaskales:
Are there any additional changes I should make for this PR?

peterjc (approved):
Over to @MarkusPiotrowski for second opinion (and merge)?

peterjc:
This was resolved another way in #3446, so belated closing this now. Sorry @kaskales for not following up on this sooner.

PR #3425, opened 2020-12-03
Enhancement SeqIO write preserves dbsource annotation

### LINKED ISSUE #920: genbank dbsource record not exported with SeqIO.write
Hello,

When I import a genbank file using Biopython Entrez.efetch and writing it as a genbank file with SeqIO write the "DBSOURCE" record / field is not exported. When fetching the same file with the standalone Entrez Edirect tool the DBSOURCE record is present (also online; see link). The dbsource record is also present in the biopython fetched version (i.e. accessible via annotations["db_source"] or when reading the raw output). So it appears to be at the level of SeqIO write. Is this a bug?

cheers,

Almer

Link to genbank test record:

http://www.ncbi.nlm.nih.gov/protein/667652385

SeqIO.write formatted genbank file:

[test_biopython_esearch.txt](https://github.com/biopython/biopython/files/439948/test_biopython_esearch.txt)

 Test script:  

[search_ncbi_test.py.txt](https://github.com/biopython/biopython/files/439970/search_ncbi_test.py.txt)

pcapel:

Closes #920
This PR adds a test to verify that the `db_source` key in the `SeqRecord.annotations` field is retained through the `SeqIO.write` method.
This was achieved by adding an `if` block and using the `_write_single_line` method inside of `GenBankWriter` in `Bio/SeqIO/InsdcIO.py`.


pcapel:
Ah darn. Looks like I did end up breaking a different test.

Oddly, running the test that is failing as a standalone is not working to recreate the failure. I have tried `python Tests/run_tests.py common_BioSQL.py` and it simply passes.

pcapel:
I figured out that it was a shared test file, so I was able to tighten the iteration loop a bit by testing `python Tests/run_tests.py --verbose test_BioSQL_sqlite3.py`.

So far what it looks like is that the BioSQL is reading the DBSOURCE back in as a List with the string as the content. I'm not super familiar with the file format for genbank, so is this an expected behavior?

peterjc:
The GenBank/EMBL parser does generally record feature annotations (SeqFeature ``.qualifiers`` dict) as lists - because any key can appear more than once, so lists of a single string a common there. But for the SeqRecord ``.annotations`` dict, simple strings is normal:

https://github.com/biopython/biopython/blob/biopython-178/Bio/GenBank/__init__.py#L898

Now for storing the SeqRecord ``.annotations`` dict in BioSQL, it copes with lists or plain strings:

https://github.com/biopython/biopython/blob/biopython-178/BioSQL/Loader.py#L751

But we may loose the distinguishment on loading them back again? That needs a little more close an examination. Tweaking the test may be a last resort.

Tests/test_SeqIO.py
```diff
@@ -3172,6 +3172,18 @@ def test_genbank21(self):
             messages,
         )
 
+    def test_genbank22(self):
+        """Test that genbank format write doesn't destroy db_source in annotations"""
+        record = SeqIO.read('GenBank/CAA46787.gb', 'genbank')
```
peterjc, line 3177: Could you use one of the existing test files, e.g. ``GenBank/protein_refseq.gb`` or ``dbsource_wrap.gb``?

pcapel: Excellent point. Will do.

pcapel:
> But we may loose the distinguishment on loading them back again? That needs a little more close an examination. Tweaking the test may be a last resort.

I placed a guard inside of the `GenBankWriter` prior to calling the `_write_single_line` method to ensure that the type was the expected string. This allows the tests to all pass.  I'd also be happy to dig into the loader code as well. 

I'm wondering why the test coverage report has dropped by so much though? Is that normal?

peterjc:
Right at this moment the test coverage is from AppVeyor running Windows only. The TravisCI tests take longer, cover more versions of Python, and more optional dependencies. The coverage should increase again...

CONTRIB.rst
```diff
@@ -234,6 +234,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - Peter Kerpedjiev <https://github.com/pkerpedjiev>
 - Peter Slickers <piet at domain clondiag.com>
 - Philip Bergstrom <https://github.com/phber>
+- Philip Capel <insert.available.name at gmail>
```
peterjc, line 237: You'd rather this than <https://github.com/pcapel>?

Bio/SeqIO/InsdcIO.py
```diff
@@ -1085,6 +1085,13 @@ def write_record(self, record):
             taxonomy = "."
         self._write_multi_line("", taxonomy)
 
+        if "db_source" in record.annotations:
+            # Hack around the issue of BioSQL loading a list for the db_source
```
peterjc, line 1089: Pragmatic, and less risky than touching the BioSQL code.

CONTRIB.rst
pcapel: This would be my first time adding myself to a file like this, so I wasn't entirely sure which would be best. I suppose pointing back to my github profile is more practical.

Bio/SeqIO/InsdcIO.py
pcapel: I thought so. My assumption was that this is being run as an internal utility, whereas the Loader for BioSQL was likely a public (user-facing) API. Addressing the fact that it's loading this annotation as a list would be a potentially breaking change. This does rely on the idea that there are never more than 1 actual `db_source` annotations, but I wasn't able to find any sort of specification for the file format that included a reference to `db_source`.

pcapel:
>The coverage should increase again...

Indeed it has! I think TravisCI got a little gummed up earlier. Glad to see this has cleared the checks!

peterjc (approved):
Are you happy for this to be squash merged into a single commit?

pcapel:
That works for me!

peterjc:
Done, thank you.

Issue #3304, opened 2020-10-09
Add Type Hints to existing methods/functions


sgalpha01:
Due to Dynamic nature of Python, sometimes it becomes difficult to check type of objects used. To make object type more explicit, Python introduced Type Hinting. 

More information can be found here:
- [PEP 483](https://www.python.org/dev/peps/pep-0483/)
- [PEP 484](https://www.python.org/dev/peps/pep-0484/)
- [typing](https://docs.python.org/3/library/typing.html)

This is just a small suggestion.

MarkusPiotrowski:
Dear @sgalpha,

thank you, I think we are already considering type hints and type annotations, see #2236 and #3165.

sgalpha01:
Oh, okay. Sorry, I didn't see that.

MarkusPiotrowski:
No problem, thank you for your motivation to contribute.

PR #3454, opened 2020-12-27
Generator typehints for Structure, Model, Residues, Atoms classes

rtviii:

Added some generator type-hints for structural classes in Bio.PDB's `Structure, Model, Chain, Residue, Atom`, in particular the getters like "`get_atoms`" etc. 
(It's frequently annoying to not have method hints on the generator or just those for the abstract `Entity`)

To João: `typing` is already present(at least) 7 months ago in Chain.py, added by Rob Miller.

### DIFF
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -7,10 +7,13 @@
 
 """Chain class, used in Structure objects."""
 
+from Bio.PDB.Atom import Atom
+from Bio.KEGG import Gene
+from Bio.PDB.Residue import Residue
 from Bio.PDB.Entity import Entity
 from Bio.PDB.internal_coords import IC_Chain
 
-from typing import Optional
+from typing import Generator, List, Optional
 
 
 class Chain(Entity):
@@ -20,7 +23,7 @@ class Chain(Entity):
     access atoms from residues.
     """
 
-    def __init__(self, id):
+    def __init__(self, id) -> None:
         """Initialize the class."""
         self.level = "C"
         self.internal_coord = None
@@ -133,7 +136,7 @@ class Chain(Entity):
 
     # Public methods
 
-    def get_unpacked_list(self):
+    def get_unpacked_list(self) -> List[Residue]:
         """Return a list of undisordered residues.
 
         Some Residue objects hide several disordered residues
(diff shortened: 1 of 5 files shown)


Bio/PDB/Chain.py
```diff
@@ -7,10 +7,13 @@
 
 """Chain class, used in Structure objects."""
 
+from Bio.PDB.Atom import Atom
+from Bio.KEGG import Gene
```
JoaoRodrigues, line 11: Is this used anywhere in the module?

JoaoRodrigues:
Hi @rtviii ,

Thanks for the PR. We've had previous discussions on whether to add type hints or not to the codebase (see #2236, #3165,  #3304). Personally I don't have much of an opinion on these but I don't think it would hurt adding more hints to the PDB module.

Bio/PDB/Chain.py
```diff
@@ -7,10 +7,13 @@
...
+from Bio.PDB.Residue import Residue
 from Bio.PDB.Entity import Entity
 from Bio.PDB.internal_coords import IC_Chain
 
-from typing import Optional
+from typing import Generator, List, Optional
```
JoaoRodrigues, line 16: It seems that both `typing.Generator` and `typing.List` will be deprecated soon (Python 3.9). See [here](https://docs.python.org/3/library/typing.html)

Bio/PDB/Residue.py
```diff
@@ -7,6 +7,8 @@
 
 """Residue class, used by Structure objects."""
 
+from Bio.PDB.Atom import Atom
+from typing import Generator
```
JoaoRodrigues, line 11: While we don't strictly enforce import orders (I think!), it would be nice not to mix standard library imports with internals of Biopython.

Bio/PDB/Structure.py
```diff
@@ -7,37 +7,42 @@
 
 """The structure class, representing a macromolecular structure."""
 
+from typing import Generator
 from Bio.PDB.Entity import Entity
 from Bio.PDB.internal_coords import IC_Chain
+from Bio.PDB.Atom import Atom
```
JoaoRodrigues, line 13: One of the nicer things of having a `Structure` and a `StructureBuilder` is that there isn't a strict dependency between what children `Structure` can have. Adding types adds all these imports that don't hurt much but would otherwise be completely unnecessary.  Also, if in the future we add another Entity subtype, we'd have to add it here to keep type hints working.

Is there a better way of handling this?

JoaoRodrigues (changes requested):

PR #3358, opened 2020-11-14
Move CI infrastructure to Github Actions

JoaoRodrigues:
_Disclaimer: Still a work in progress, probably rebases will happen._

This PR drops TravisCI and Appveyor as CI providers and moves to Github Actions. GHA offer several advantages over the current setup, namely:
- All major OSes are available for testing (Win, Linux, and MacOS).
- Simpler configuration (less config files).
- Free & Faster.

Roadmap:
- [x] Style checking using pre-commit.
- [x] Bonus: style checking only on files modified by PR to save time.
- [x] Check RST files.
- [x] Building steps (dist and wheel).
- [x] Offline tests (base dependencies).
- [ ] Offline tests (all dependencies).
- [x] Testing on PyPy.
- [x] Coverage.
- [x] Building Documentation with Sphinx

Possible Wishlist:
- [ ] Automatic deployment to conda/PyPI of development / tagged versions / releases.

References #3340

Issue #3340, opened 2020-11-09
Changes to TravisCI / Move CI to Github Actions


JoaoRodrigues:
TravisCI sent out an email this week announcing changes to their plans ([link](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)). The post is quite dense and it's hard to understand if this affects us but there seem to be new limitations to open-source projects. Also, we use the travis-ci.org portal, which will be closed down at the end of this year ([link](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions)) so we might take this opportunity to think about updating our CI infrastructure a little.

Currently, we use TravisCI and Appveyor to test on Windows and Linux, and it seems we don't test on MacOS. Moving to Github Actions would let us test on all three major OSes and a bunch of Python versions including PyPy. It would also allow us to run up to 256 jobs in parallel, and in my experience, it seems those jobs run faster than on Travis/Appveyor.

To kickstart this discussion and test drive a possible transition to Github Actions, I started a branch (purposedly out of date with the current master by a few commits) implementing the style checking, packaging, and simple testing steps. You can see the branch [here](https://github.com/JoaoRodrigues/biopython/tree/gh_actions_tests) and the file I have been working on [here](https://github.com/JoaoRodrigues/biopython/blob/gh_actions_tests/.github/workflows/basic.yml).

Thoughts? Suggestions?

peterjc:
If this is the end of free TravisCI testing (which as Joao says is hard to understand from the post), then so be it. We can be grateful for the years of free support they have given us. We could explore using OBF funds to pay for TravisCI, but speaking as current OBF president I would rather Biopython explore other free alternatives first.

Personally I also have experience using CircleCI, which should be a fairly simple change - the configuration file is different, but the approach is quite similar apart from the lack of TravisCI stages. They use the same business model TravisCI did, free for open source, paid for closed projects. I could have a go at that?

In parallel, GitHub Actions sounds like a good long term alternative - and may assist with more interactive bots on our pull requests too in future? I'd be delighted for Joao to take the lead on that.

JoaoRodrigues:
>If this is the end of free TravisCI testing (which as Joao says is hard to understand from the post), then so be it. We can be grateful for the years of free support they have given us. We could explore using OBF funds to pay for TravisCI, but speaking as current OBF president I would rather Biopython explore other free alternatives first.

[...]

PR #3458, opened 2020-12-31
Minimal CircleCI Linux testing integration

peterjc:

Seeks to address #3340, see also #3358 (GitHub actions).

Implements a minimal CircleCI configuration covering a single version of Python, no command line dependencies (e.g. blast - using a conda image might be better for this?), running the pre-commit style checks (including black) and our test suite (with coverage tracking).

Does not cache the (optional) Biopython pip dependencies (numpy, scipy, etc), since we don't have a requirements file or similar to give an easy way to key the cache. Does cache the pre-commit dependencies.

Does not run Sphinx/apidocs and update the API documentation (currently done on TravisCI).

Will require adding this repository to CircleCI, and the ``$CODECOV_TOKEN`` environment variable. Proof of principle here: https://app.circleci.com/pipelines/github/peterjc/biopython

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,69 @@
+# This is a special configuration file to run tests on Circle-CI via
+# GitHub notifications when changes are committed.
+#
+# This file is not intended for end users of Biopython.
+#
+# This uses a CircleCI provided Python image, see
+# https://circleci.com/developer/images/image/cimg/python
+
+version: 2.1
+orbs:
+  codecov: codecov/codecov@1.0.2
+jobs:
+  build:
+    docker:
+      - image: cimg/python:3.7
+    steps:
+      - checkout
+      - run:
+          name: Combine precommit config and python versions for caching
+          command: |
+            cat .pre-commit-config.yaml > pre-commit-deps.txt
+            python -VV >> pre-commit-deps.txt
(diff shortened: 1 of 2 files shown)


JoaoRodrigues:
Hi @peterjc, thanks for looking into this. I apologize for not getting back to this issue of testing earlier, the end of the year was quite busy. I never had time to go back to the Github Actions idea, but I did write the style checking/testing part. If you can/want, have a look [here](https://github.com/JoaoRodrigues/biopython/tree/gh_actions_tests). It runs quite nicely on all 3 OSes (in parallel) and I added a trick to only run linting on modified files by the PR, to save time.

Having said all that, this looks good. Would CircleCI be a good solution going forward, that supports all 3 major OSes, or would we still need Appveyor?

peterjc:
Potentially CircleCI could replace AppVeyor too - I've only used it Linux-only thus far.

Having more than one system ought to give us some resilience and potentially faster run time (e.g. Windows on service A, Linux on sevice B)? So having GitHub Actions too is worthwhile.

peterjc:
Applied and README badge added in f40bf0731483281bba7bb399053d2d9616a9f533

I was thinking adding the API documentation next would be wise if TravisCI stops soon.

Issue #3523, opened 2021-04-06
Expand `SeqUtils` `ProtParam` module


Caiofcas:
I had to calculate various values based in protein sequences for a ML model recently, and BioPython was by far the most convenient way to do it, but only for the methods currently covered in Bio/SeqUtils/ProtParam.py. I feel that the `ProteinAnalysis Class (and ProtParamData.py ) could be expanded to cover all the scales present in the [R Peptides](https://rdrr.io/cran/Peptides/) package, and other features, such as Chou's Pseudo AAC and k-peptide (such as Dipeptide and Tripeptide) composition.

Even though most of the features I'm proposing could be calculated using the `ProteinAnalysis.protein_scale()` method or directly from the sequence, I feel there is value adding standard methods for these. I saw that a similar proposal was made in #3124 , but that PR seems abandonned and I'm not proposing adding another module.

Would it make sense to add this to BioPython? Or maybe I should write a standalone package?

peterjc:
An incremental addition seems much easier for us to review and merge than a whole new module 👍

Caiofcas:
@peterjc Great! Will start working on the PR then :D

PR #3524, opened 2021-04-06
Adds other Hidrophobicity scales to `ProtParam`

Caiofcas:

Starts the work proposed in #3523 by adding a number of alternative Hidrophobicity scales as standard options to the module. The list of scales is inspired by https://rdrr.io/cran/Peptides/ , but only scales reported as hidrophobicity values in the original papers were added.

Also adds the paramenter _scale_ to `ProteinAnalysis.gravy` (mantaining the Kyte and Doolitle scale as the default option for compatibility), allowing for easy expansion of this list.

### DIFF
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -189,9 +189,14 @@ class ProteinAnalysis:
 
         return scores
 
-    def gravy(self):
+    def gravy(self, scale="KyteDoolitle"):
         """Calculate the gravy according to Kyte and Doolittle."""
-        total_gravy = sum(ProtParamData.kd[aa] for aa in self.sequence)
+        selected_scale = ProtParamData.gravy_scales.get(scale, -1)
+
+        if selected_scale == -1:
+            raise ValueError("scale: {} not know".format(scale))
+
+        total_gravy = sum(selected_scale[aa] for aa in self.sequence)
 
         return total_gravy / self.length
 
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -192,11 +192,12 @@ class ProteinAnalysis:
     def gravy(self, scale="KyteDoolitle"):
         """Calculate the GRAVY (Grand Average of Hydropathy) according to Kyte and Doolitle, 1982.
 
-        Utilizes the given Hidrophobicity scale, by default uses the original proposed
-        by Kyte and Doolittle (KyteDoolitle). Other options are: Aboderin, AbrahamLeo,
-        Argos, BlackMould, BullBreese, Casari, Cid, Cowan3.4, Cowan7.5, Eisenberg,
-        Engelman, Fasman, Fauchere, GoldSack, Guy, Jones, Juretic, Kidera, Miyazawa,
-        Parker,Ponnuswamy, Rose, Roseman, Sweet, Tanford, Wilson and Zimmerman.
+        Utilizes the given Hydrophobicity scale, by default uses the original
+        proposed by Kyte and Doolittle (KyteDoolitle). Other options are:
+        Aboderin, AbrahamLeo, Argos, BlackMould, BullBreese, Casari, Cid,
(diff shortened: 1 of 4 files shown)


Caiofcas:
@peterjc Sorry for tagging you directly, I'm apparently not able to request a code review through github. Could you take a look?

Bio/SeqUtils/ProtParam.py
```diff
@@ -189,9 +189,23 @@ def flexibility(self):
...
+        Argos, BlackMould, BullBreese, Casari, Cid, Cowan3.4, Cowan7.5, Eisenberg,
+        Engelman, Fasman, Fauchere, GoldSack, Guy, Jones, Juretic, Kidera, Miyazawa,
+        Parker,Ponnuswamy, Rose, Roseman, Sweet, Tanford, Wilson and Zimmerman.
+
+        New scales can be added in ProtParamData.
+        """
+        selected_scale = ProtParamData.gravy_scales.get(scale, -1)
+
+        if selected_scale == -1:
+            raise ValueError("scale: {} not know".format(scale))
```
peterjc, line 206: Typo: know -> known

Style: Please use an f-string rather than the format method

Ideally there would be a simple test for the unknown scale, using ``.assertRaises``

Bio/SeqUtils/ProtParamData.py
```diff
@@ -8,13 +8,245 @@
 # Turn black code style off
 # fmt: off
 
-# Kyte & Doolittle index of hydrophobicity
+# Hidrophobicity
+
+# 1 Kyte & Doolittle index of hydrophobicity
```
peterjc, line 13: Are the numbers (1 to 27) meaningful? If you can avoid this is will make any future additions easier.

Also perhaps do these in key order?

peterjc (changes requested):
Can you add at least one test using a new scale, and a simple test using an invalid scale please?

Bio/SeqUtils/ProtParamData.py
Caiofcas: I was doing it based on the Hidrophilicity and Surface accessibility values below, but the numbers are not meaningful by themselves. Will remove them.

By key order you mean the string in the third comment line? I believe they already are in this order with the exception of the `KyteDoolitle` scale, which is first because it is the default option, but i can also change it.

Caiofcas:
Will add the tests, was thinking of adding the tests for all the `ProteinAnalyisis` methods. Do you believe it is better to do this in the existing `test_SeqUtils.py` file or  to create a new `test_SeqUtils_ProteinAnalysis.py` file ( in the same vein as the `test_SeqIO_XXXX.py` files )?

Bio/SeqUtils/ProtParamData.py
peterjc: The current order makes sense then 👍

peterjc:
Use existing file ``Tests/test_ProtParam.py``?

Caiofcas:
Didn't see it... my bad

Caiofcas:
Any other changes necessary @peterjc ? I am working in another PR with some overlap so it would be better to fix all problems with this one first.

peterjc (approved):

peterjc:
@biopython/team-biopython Does anyone want to give a second opinion on this? If not, I think we can merge it at the end of the week.

MarkusPiotrowski (changes requested):
1. Could you change all occurrences of 'Hidro...' to 'Hydro...'?
2. You have added many new scales but added test for only two of them. Could you please add tests also for the other?

MarkusPiotrowski (approved):

peterjc:
Are you OK with this being a squash-and-merge, giving a single commit on the master branch?

Caiofcas:
> Are you OK with this being a squash-and-merge, giving a single commit on the master branch?

Yes. I realize my commit history is a bit messy, sorry about that, will pay more attention to that in the future.

peterjc:
Ah - I think I figured out why CircleCI is being very very slow to trigger, and in fact never triggered. I have just enabled the "Build forked pull requests" option.

Issue #3471, opened 2021-01-25
Update Bio.Application examples in Tutorial to use subprocess directly


peterjc:
See discussion on #2877, closed via #3344 declaring Bio.Application obsolete, suggesting using subprocess directly.

We need to update the Tutorial to translate the command line wrapper examples to use subprocess directly. Working on Linux/macOS only would be a reasonable first pass but ideally tested as working on Windows too.

The tutorial is written in LaTeX, source here https://github.com/biopython/biopython/tree/master/Doc/Tutorial - on Linux/macOS using the ``make`` command in the ``Doc/`` directory will build this for you.

There is scope for multiple contributors to work on this, most simply by working on different sections of the Tutorial.

Caiofcas:
I was starting to work on removing the `Bio.Application` calls in `chapter_phylo.tex`, but I'm confused if `Bio.Phylo.PAML` was also deprecated, since it seems very different from the other wrappers. In it's wiki page (https://biopython.org/wiki/PAML) and in the Class source code there's also no warning, so I don't know whether it should be removed from the tutorial or not.

It seems that it does provide some value beyond just a CLI wrapper, since it apparently manages the control files as well, and as such might be more troublesome to replace (this is more of a guess as i have not gone deep into what the module does exactly).

TLDR: Should `Bio.Phylo.PAML` also be removed from the tutorial?

peterjc:
The PAML stuff goes beyond just helping to build and run a command line string - it was not intended to be deprecated. So that stays please.

PR #3568, opened 2021-05-31
rm Bio.Applications from Blast section in the tutorial

Caiofcas:

Small progress on #3471 .

### DIFF
--- a/Doc/Tutorial/chapter_blast.tex
+++ b/Doc/Tutorial/chapter_blast.tex
@@ -208,36 +208,23 @@ of $0.001$ and produce XML output to the specified file (which we can then parse
 On my computer this takes about six minutes - a good reason to save the output
 to a file so you can repeat any analysis as needed.
 
-From within Biopython we can use the NCBI BLASTX wrapper from the
-\verb|Bio.Blast.Applications| module to build the command line string,
-and run it:
+From within python we can use the \verb|subprocess| module to build the command line
+string, and run it:
 
 \begin{minted}{pycon}
->>> from Bio.Blast.Applications import NcbiblastxCommandline
->>> help(NcbiblastxCommandline)
-...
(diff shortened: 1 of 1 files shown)


Doc/Tutorial/chapter_blast.tex
```diff
@@ -208,37 +208,24 @@ \subsection{Standalone NCBI BLAST+}
...
->>> blastx_cline
-NcbiblastxCommandline(cmd='blastx', out='opuntia.xml', outfmt=5, query='opuntia.fasta',
-db='nr', evalue=0.001)
->>> print(blastx_cline)
-blastx -out opuntia.xml -outfmt 5 -query opuntia.fasta -db nr -evalue 0.001
->>> stdout, stderr = blastx_cline()
+>>> import subprocess
+>>> cmd = "blastx -query opuntia.fasta -db nr -out opuntia.xml"
+>>> cmd += " -evalue 0.001 -outfmt 5"
+>>> subprocess.run(cmd, shell=True)
```
peterjc, line 218: Which OS have you tried this form on? If it works on Windows/macOS/Linux, then great.

Caiofcas: I have only tested on Linux, and don't have access to other OSes, thinking about it the only problem I could see is the `shell=True` having some weird behavior in Windows, but don't know enough about it to say for sure. Since it's such a simple example I would guess it would be fine.

peterjc: From memory, we're using a string for ``cmd`` here, so it should run on Windows - the difference may be in how the shell manifests (as a visible terminal window or not), and may also depend on if Python is run from a terminal or not.

Still, this is probably worth merging even without directly testing under Windows, but will have to wait a day or two as the current release is in progress.

peterjc:
Thank you

PR #3571, opened 2021-06-01
Fix LaTeX syntax to allow documentation to build.

JoaoRodrigues:
Building the PDF documentation is failing at the moment with the error: `! FancyVerb Error:  Extraneous input `{pycon}\end{}' between \end{minted} and line end`. Comparing with the rest of the document, seems that that `{pycon}` tag is extra in a few places in `chapter_align.tex`. Removing it allows for the documentation to build correctly.

Not a LaTeX expert so definitely needs review!

### DIFF
--- a/Doc/Tutorial/chapter_align.tex
+++ b/Doc/Tutorial/chapter_align.tex
@@ -2292,7 +2292,7 @@ TGAACT
  ||-|
 TGAACT
 <BLANKLINE>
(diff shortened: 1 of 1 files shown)


peterjc:
Well debugged. Last time I tried adding the Tutorial PDF to the CI, the latex dependency was too heavy.

PR #3572, opened 2021-06-01
Replace accession by locus id in EMBL header line

oschwengers:
Hello,
If the `SeqRecord`/`SeqFeature` API is used to create an in-memory data structure of an, _e.g._ bacterial genome annotation for subsequent serialization via `SeqIO`, the `GenBankWriter` writes both the *locus id* and the *accession*, the latter only if provided via the sequence annotations.

However, the `EmblWriter` writes the *accession* in both fields ignoring the *locus id*. 

I do not exactly know if this is a bug or a feature, i.e. something that is strictly required by the EMBL format - I could not find any specs on the web.

This became an issue when we annotated local genome assemblies for which no public *accession* is available - but an unique *locus id* is of course required to identify the records within multi EMBL/Genbank files.

I therefore replaced the *accession* by the *locus id* in the EMBL header line to my best understandings and in line with the behaviors of the `GenBankWriter`.

Closes #...

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -1268,10 +1268,17 @@ class EmblWriter(_InsdcWriter):
         # Get the taxonomy division
         division = self._get_data_division(record)
 
+        # Get the locus name/id in line with the GenBank output (line 714)
+        locus = record.name
+        if not locus or locus == "<unknown name>":
+            locus = record.id
+        if not locus or locus == "<unknown id>":
+            locus = self._get_annotation_str(record, "accession", just_first=True)
+        
         # TODO - Full ID line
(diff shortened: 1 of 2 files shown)


peterjc:
I forget now what exactly the GenBank writer logic was doing, but have dealt with far more unofficial GenBank files (i.e. not from NCBI or DDBJ) than unofficial EMBL files - that likely explains why this has not been looked at before.

peterjc:
It looks like your changes have broken some of the tests - including what might be some round-trip tests?

It is also possible you'll need to update the tests to match your desired behaviour.

oschwengers:
Thanks for the quick reply. I'll go through the list of failed tests and try to adopt these. However, this might take a while.

PR #3578, opened 2021-06-03
feat(SeqFeature): store description, evidence and status of sequence features

### LINKED ISSUE #3576: Access to UniProt annotation information
Is there a way in which biopython supports working with UniProt annotations?

Intended workflow:

1. Download either uniport XML or TXT release.
2. Query the annotation in these files by accession.

Example Question: How can biopython help to find the start and end position of the signal peptide of [pantetheinase](https://www.uniprot.org/uniprot/O95497)?

fabianegli:

Closes #3576

### DIFF
--- a/Bio/SeqIO/UniprotIO.py
+++ b/Bio/SeqIO/UniprotIO.py
@@ -455,6 +455,10 @@ class Parser:
             feature.type = element.attrib.get("type", "")
             if "id" in element.attrib:
                 feature.id = element.attrib["id"]
+            if "description" in element.attrib:
+                feature.description = element.attrib["description"]
+            if "evidence" in element.attrib:
+                feature.evidence = element.attrib["evidence"]
             for feature_element in element:
                 if feature_element.tag == NS + "location":
                     position_elements = feature_element.findall(NS + "position")
--- a/Bio/SeqIO/UniprotIO.py
+++ b/Bio/SeqIO/UniprotIO.py
@@ -455,12 +455,6 @@ class Parser:
             feature.type = element.attrib.get("type", "")
             if "id" in element.attrib:
                 feature.id = element.attrib["id"]
-            if "description" in element.attrib:
-                feature.qualifiers["description"] = element.attrib["description"]
-            if "evidence" in element.attrib:
-                feature.qualifiers["evidence"] = element.attrib["evidence"]
-            if "status" in element.attrib:
-                feature.qualifiers["status"] = element.attrib["status"]
             for feature_element in element:
                 if feature_element.tag == NS + "location":
                     position_elements = feature_element.findall(NS + "position")
--- a/Bio/SeqIO/UniprotIO.py
+++ b/Bio/SeqIO/UniprotIO.py
@@ -456,11 +456,11 @@ class Parser:
             if "id" in element.attrib:
                 feature.id = element.attrib["id"]
             if "description" in element.attrib:
-                feature.description = element.attrib["description"]
+                feature.qualifiers["description"] = element.attrib["description"]
             if "evidence" in element.attrib:
-                feature.evidence = element.attrib["evidence"]
+                feature.qualifiers["evidence"] = element.attrib["evidence"]
             if "status" in element.attrib:
-                feature.status = element.attrib["status"]
+                feature.qualifiers["status"] = element.attrib["status"]
             for feature_element in element:
                 if feature_element.tag == NS + "location":
                     position_elements = feature_element.findall(NS + "position")
--- a/Bio/SeqIO/UniprotIO.py
+++ b/Bio/SeqIO/UniprotIO.py
@@ -459,6 +459,8 @@ class Parser:
                 feature.description = element.attrib["description"]
             if "evidence" in element.attrib:
                 feature.evidence = element.attrib["evidence"]
+            if "status" in element.attrib:
+                feature.status = element.attrib["status"]
             for feature_element in element:
                 if feature_element.tag == NS + "location":
                     position_elements = feature_element.findall(NS + "position")
--- a/Bio/Align/__init__.py
+++ b/Bio/Align/__init__.py
@@ -977,14 +977,27 @@ class PairwiseAlignment:
 
         self[:, :]
 
-        which returns a copy of the PairwiseAlignment object, and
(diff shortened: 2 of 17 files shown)


fabianegli:
It may pay off to check if there are other attributes of feature that could be added with this PR.

fabianegli:
At least one more remains: `status`:  `git grep "^\s*<feature\ .*\ status=" Tests/`

`git grep "^\s*<feature\ " comlete_uniprot.xml` might reveal more.

peterjc (changes requested):
Adding UniProt specific attributes to the objects is not attractive. Can you not use the existing SeqFeatures qualifier dictionary instead?

fabianegli:
I see - and agree - but I don't understand what you mean by "SeqFeatures qualifier dictionary"?

fabianegli:
Just found out what you meant: https://github.com/biopython/biopython/blob/45eae8cd48c73cfd44c7bd84de168f26ad651ec0/Bio/SeqFeature.py#L85

peterjc:
Yes, that's what I was referring too.

fabianegli:
@peterjc local tests seem to hang at "test_NCBI_BLAST_tools". Is that something that I can easily fix to get the tests running locally? 

It states 

```
test_NCBI_BLAST_tools ... skipping. Install the NCBI BLAST+ command line tools if you want to use the Bio.Blast.Applications wrapper.
```
 and then freezes. No error, and seemingly no cpu/disk/network usage.

peterjc:
Once upon a time ``test_NCBI_qblast.py`` was quick, nowadays it can take an hour or more. You can use ``run_tests.py --offline`` to skip the online tests.

Bio/SeqIO/UniprotIO.py
```diff
@@ -455,6 +455,12 @@ def _parse_feature(element):
             feature.type = element.attrib.get("type", "")
             if "id" in element.attrib:
                 feature.id = element.attrib["id"]
+            if "description" in element.attrib:
+                feature.qualifiers["description"] = element.attrib["description"]
+            if "evidence" in element.attrib:
+                feature.qualifiers["evidence"] = element.attrib["evidence"]
+            if "status" in element.attrib:
+                feature.qualifiers["status"] = element.attrib["status"]
```
peterjc, line 463: I'm looking at that and thinking a for loop would be more elegant - especially if there might be more extra fields later on?

```python
for field in ["description", "evidence", "status"]:
    if field in element.attrib:
        feature.qualifiers[field] = element.attrib[field]
```

Or, if those are almost always present perhaps this:

```python
for field in ["description", "evidence", "status"]:
    try:
        feature.qualifiers[field] = element.attrib[field]
    except KeyError:
        pass
```

That is longer though, so maybe not?

peterjc:
This looks much better now. Would you be OK with me doing a squash-and-merge turning this into a single commit (for a simpler history)?

Bio/SeqIO/UniprotIO.py
fabianegli: I wonder if we can make it even more general so new attributes don't require a code change at all. 

```python
for field in [k for k in element.attrib.keys() if k not in ["id", "type"]]:
    feature.qualifiers[field] = element.attrib[field]
```

fabianegli: even the if statment for the id can be removed:

```python
            for field in [k for k in element.attrib.keys() if k != "type"]:
                feature.qualifiers[field] = element.attrib[field]
```

I wonder if type is that special that it has to be set even if it is not in the attribute list?

fabianegli: OK, it was all allready there... just a few lines up

https://github.com/biopython/biopython/blob/46a5c82e126713d22894301ba356c71e4c03732d/Bio/SeqIO/UniprotIO.py#L453-L454

fabianegli:
So, it seems I only added some tests ...

peterjc:
Laughs. Well, adding tests for something is good.

The NEWS entry will probably get replaced with a generic bit about improved test coverage later.

fabianegli:
Quick question: Why not show the qualifiers when repr? That would have prevented me from running down this rabbit hole - probably.

peterjc:
Not sure now - I added the ``__repr__`` back in 4b609092bb2cd977d09afa024b4285803cbf6f63 - some years back. I was probably thinking adding it would make the representation overly long, but perhaps showing ``, qualifiers=...,`` when the dict is non empty would be a good compromise?

fabianegli:
Why not go with the Python docs, and show a complete "evaluatable" representation of the object? It won't be all that big anyway,  AFAICT.

From https://docs.python.org/3.8/library/functions.html#repr

> Return a string containing a printable representation of an object. For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval, otherwise the representation is a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object. A class can control what this function returns for its instances by defining a __repr__ method.

Either would be fine with me, but the incomplete attempt to repr completely threw me off tracks

peterjc:
The complete evaluatable representation of the qualifiers dict will be very long, which is why showing ``, qualifiers=...,`` made sense to me. See also the representation of our ``Seq`` object which also uses dot dot dot.

fabianegli:
My question is basically this:
If there is a str representation containing the complete set of information (in this case the __str__() function), is it really added value to show an evaluatable string representation for repr that doesn't faithfully replicate the whole object? I think when the object can't be replicated by the output of the repr it should be just the name and address of the object as the Python docs indicates. 

I know this is opinionated, but I have doubts about the usefulness of the current state of repr.  Do you know of any use cases beyond a quick look at the object content?

If the repr needs to continue to show the custom string representation, I would definitely see the addition of `, qualifiers=...` as an improvement because it makes SeqFeature more transparent.

peterjc:
For complex objects like a SeqFeature or SeqRecord, my only use case of repr is something short for a quick look or debug/logging/error message. I really don't want it to fill the screen at the terminal, but I want the "key" information (and the Python default of object at some memory address is useless). In the case of the SeqFeature, I would want at least location and type.

From that starting point, adding ``, qualifiers=... `` for a non-empty dict is a minimal change.

fabianegli:
I think this we should move that to a new issue/PR?

PR #3595, opened 2021-06-08
indicate the existence of qualifiers in repr

### LINKED ISSUE #3593: Information in the SeqFeature repr is hidden
The current behavior of repr hides its `qualifiers` attribute, even if it is non-empty. It would be helpful in some situations to surface the qualifiers attribute in the repr as discussed in https://github.com/biopython/biopython/issues/3576.

A proposal for such a behaviour has been made by @peterjc in https://github.com/biopython/biopython/pull/3578.

fabianegli:
Closes #3593

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -279,12 +279,12 @@ class SeqFeature:
             answer += ", location_operator=%r" % self.location_operator
         if self.id and self.id != "<unknown id>":
             answer += ", id=%r" % self.id
+        if self.qualifiers:
+            answer += ", qualifiers=..."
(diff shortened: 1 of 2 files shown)


Bio/SeqFeature.py
```diff
@@ -283,6 +283,8 @@ def __repr__(self):
             answer += ", ref=%r" % self.ref
         if self.ref_db:
             answer += ", ref_db=%r" % self.ref_db
+        if self.qualifiers:
+            answer += ", qualifiers=..."
```
peterjc, line 287: Move before ref to match __init__ order

Issue #3600, opened 2021-06-09
Do not use assert outside testing


fabianegli:
There are multiple uses of assert inside the production code. This is bad practice because assert is removed when python is run with optimization, e.g. `python -0 some_script_using_biopython.py`. The problem with assert is that it is not a reliable flow control and input check for code that is run outside the programmers control.

Use the following command to find assert throughout the biopython codebase:

```shell
git grep "^[^.>#]*assert" | egrep -v "^Tests/"
```

Some uses in doctests (remove the `>` from the above regex to reveal them) are probably fine but not all the others.

fabianegli:
There are different use cases in the code AFACT:

1. Control Program flow: this includes input checks
    * These cases need to be replaced with proper if statements.
2. Tests: Testing assumptions about the codebase
    * These cases should be moved from `Bio` production code to the `Tests`

peterjc:
There is probably a flake8 plugin for this, not https://pypi.org/project/flake8-assertive/ which does something different.

fabianegli:
I have no idea if there is a tool that helps with this - but that would be cool :-)

Help wanted is fine with me. I am not territorial on this issue. Good first might be true for the part of the issue where program flow is the intended effect, the testing might be a bit more involved in some cases. There might be multiple PRs to solve this issue. Anyone who's new could go through one, few or multiple of the affected files. So yes, both flags are fine with me.

JoaoRodrigues:
I'm a little late to the party with the asserts but I see them as very
valuable checks, particularly in a library like ours. Sometimes we *do*
want to enforce a specific type or size or something and asserts are very
valid for this.

Do we ever build with -O? Asserts should anyway not be used for important
checks but for developer usage they can be useful.

A quarta, 9/06/2021, 10:17, Fabian Egli ***@***.***> escreveu:

mdehoon:
I agree with @JoaoRodrigues  . Also, for bug reports it's very useful to catch errors early.

fabianegli:
I might have been to vague about my intention by putting the focus on the functionality. I do not argue against the checks but merely point out that they are lost if python is invoked with optimization. In fact I opened the issue in the hope the checks would be changed from assert to if statements to ensure they are not removed by the optimization and always run and show up in bug reports, logs and error messages.

fabianegli:
@JoaoRodrigues  It does not matter how biopython is "built", but how python is invoked.

```shell
$ cat assert-test.py 
[...]

PR #3629, opened 2021-06-18
Remove assert bgzf #3600

vincentdavis:

issue #3600


Bio/bgzf.py
```diff
@@ -458,11 +458,14 @@ def _load_bgzf_block(handle, text_mode=False):
         subfield_data = handle.read(subfield_len)
         x_len += subfield_len + 4
         if subfield_id == _bytes_BC:
-            assert subfield_len == 2, "Wrong BC payload length"
-            assert block_size is None, "Two BC subfields?"
+            if subfield_len != 2:
+                raise ValueError("Wrong BC payload length")
+            if block_size is not None:
+                raise AttributeError("Two BC subfields?")
```
peterjc, line 464: Why an AttributeError?

peterjc:
Looks good in principle, but why are you using AttributeError in multiple places? It does not seem appropriate to me:
https://docs.python.org/3/library/exceptions.html#AttributeError

vincentdavis:
@peterjc Honestly part of why I wanted to work on this was to learn more about how to choose the correct errors.
I changed a couple to ValueErrors.
I left a couple Attribute errors, I think those are correct.

peterjc:
I don't claim to be an expert on the exceptions, but the two remaining uses of ``AttributeError`` here strike me as better being a ``ValueError`` when validating the method arguments:

```
        if filename and fileobj:
            raise AttributeError("Both filename and fileobj are defined")
```

Happy to call for a 2nd opinion, perhaps @mdehoon or @MarkusPiotrowski ?

fabianegli:
In the case of conflicting arguments I see two possible options:

1. ValueError
2. RuntimeError
    From the python docs:
    > Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

The ValueError is normally raised if ONE value is wrong, but here we are talking about incompatible arguments that can both be correct when supplied to the function without the other. This is why I think ValueError is the wrong choice for the code discussed above and the RuntimeError should be chosen.

fabianegli:
Also `"Both filename and fileobj are defined"` states what's wrong, but not how to use it.

peterjc:
I still think that would be a ``ValueError``, but agree the message should be clearer - e.g. "Supply either filename or fileobj, not both"

fabianegli:
It seems that the stackoverflow consensus is in agreement with @peterjc. Namely to use ValueError for incompatible arguments. Reference: https://stackoverflow.com/a/256260/6018688

vincentdavis:
@peterjc , @fabianegli 
I'll update in the next couple days, thanks for the feedback.

peterjc:
I have resolved the conflict, used the "Supply either filename or fileobj, not both" message, and switched to ValueError.

Assuming all the tests pass, I think we can squash-and-merge this now?

fabianegli:
Random thought: Not sure if there is a biopython standard for the formatting of error messages. The Python standard library normally does not capitalize the first letter of the message and does not use a full stop at the end of the message. I think, however, that if the first letter is capitalized, the message should be ended with a full stop to make it a complete sentence. And vice versa that if there is a full stop at the end of the message it should start capitalized.

peterjc:
I don't think we have ever discussed a consistent standard for formatting our error messages. It seems a nice idea (and could be enforced with a flake8-plugin or something in principle), but technically any change could break people's error handling code so does not seem worth the risk.

peterjc:
Thank you Vincent

PR #3664, opened 2021-07-10
skip atoms with invalid coords in PERMISSIVE mode

andreasntr:

Sometimes it may be better to get an incomplete 3D-structure than not getting any. The final user can then choose wether to keep the incomplete structure once built by examining it or throw it if too much information is missing, e.g. by counting the numer of atoms through `model.get_atoms()`.

### DIFF
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -211,10 +211,9 @@ class PDBParser:
                     y = float(line[38:46])
                     z = float(line[46:54])
                 except Exception:
-                    # Should we allow parsing to continue in permissive mode?
-                    # If so, what coordinates should we default to?  Easier to abort!
+                    # Skip the whole line if permissive
                     if self.PERMISSIVE:
-                        continue # Skip the whole line if permissive
+                        continue
                     raise PDBConstructionException(
                         "Invalid or missing coordinate(s) at line %i."
                         % global_line_counter
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -213,6 +213,8 @@ class PDBParser:
                 except Exception:
                     # Should we allow parsing to continue in permissive mode?
                     # If so, what coordinates should we default to?  Easier to abort!
+                    if self.PERMISSIVE:
+                        continue # Skip the whole line if permissive
                     raise PDBConstructionException(
                         "Invalid or missing coordinate(s) at line %i."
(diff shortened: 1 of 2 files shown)


andreasntr:
the test for bad atoms coordinates needs to be changed as well, since it expects a `PDBConstructionException` to be raised

JoaoRodrigues:
I am not sure it would be wise to make this change since the `PERMISSIVE` flag has a very well-defined behavior and this would be a rather big change. Also, PDB (or PQR or mmCIF) files are essentially coordinate files, so if a coordinate is missing, or isn't formatted properly, it is a bit deal and I'd be inclined to keep it as an exception.

Can you give an example or use-case where this behavior (ignoring atoms) would be desirable?

andreasntr:
When you don't really care about the full structure but only about the distribution of the cloud of atoms (points). But I understand it's a very limited use-case

rob-miller:
I think the issue is whether / how much PERMISSIVE should accept abuse of
the PDB format, in this case where one might work with no-density or
similar missing atom information by having ATOM/HETATM records without
coordinates instead of e.g. using SEQRES records to work it out as PDB
might dictate.

The alternative case is where the file is garbled in the middle so maybe
you should just dump it and move on, but the spec does say with
PERMISSIVE=True we will parse but some stuff will be missing.  In this case
I think I would certainly want a warning with the offending line instead of
just spinning by as the supplied patch does?  For a few PDB files you'll
see the messages while processing, while for bulk processing as Andrea says
you need to be doing other tests on a structure before accepting it whether
the file is valid PDB or not.

I think the correct solution is to change PERMISSIVE to an INT for how
permissive (how much abuse of the PDB format) to allow, but probably no one
else wants to do that :-) .

JoaoRodrigues:
Having multiple "degrees" of permissiveness is a slippery slope :) This is code for a parser of a (mostly) well-defined format. At some point, we have to draw the line and say "well, that's not the format any longer". I think this is the case with coordinates. The cases where PERMISSIVE acts on were based on actual PDB entries, that sometimes bend the format specifications when they are ambiguous. Ignoring coordinates might lead to a gross misinterpretation of the data.

My take is that whatever software is writing PDB files that cannot be parsed by our parser because of coordinate errors is itself buggy and needs fixing upstream.

@andreasntr you can always subclass or write your own parser and keep using the rest of Biopython on the Structure objects it creates. I'd recommend that as the simplest solution (short of fixing the upstream code!).

Issue #1383, opened 2017-09-01
Bio.PDB.PDBIO.py - increment serial number of the TER record


ChristopheDeprez:
### Setup

I am reporting a problem with Biopython version 1.70

### Expected behaviour

When writing out in PDB format, the serial number of the TER record should be one number greater than the serial number of the ATOM/HETATM preceding the TER.

Reference: http://deposit.rcsb.org/adit/docs/pdb_atom_format.html#TER

### Actual behaviour

When writing out in PDB format, the serial number of the TER record is the same as the serial number of the ATOM/HETATM preceding the TER.

### Solution:

Write out the TER record with an incremented atom_number.

```
--- Bio/PDB/PDBIO.py    2017-09-01 14:42:30.000000000 -0400
+++ Bio/PDB/PDBIO.py    2017-09-01 14:53:45.000000000 -0400
@@ -222,7 +222,7 @@
                                 atom_number += 1
                 if chain_residues_written:
                     fp.write("TER   %5i      %3s %c%4i%c                                                      \n"
-                             % (atom_number, resname, chain_id, resseq, icode))
+                             % (atom_number + 1, resname, chain_id, resseq, icode))
 
             if model_flag and model_residues_written:
                 fp.write("ENDMDL\n")
```

### Licensing:

I agree to the current contribution being dual licensed under both the original "Biopython License Agreement" and the more widely used "3-Clause BSD License".

peterjc:
@lennax could you look at this please?

Cross reference 7e185d8be91e170d8dd33c19d36fc265b2e9198f

JoaoRodrigues:
This will actually cause the TER statement and the the ATOM record right after to have the same serial number. Also needs to account for when we want to preserve numbering (how do we preserve numbering then if we *add* statements is another very good question).

ChristopheDeprez:
You're right, I was only considering the "preserve_atom_numbering" case!
In the not "preserve_atom_numbering" case, the serial number should be incremented after writing each record (after writing the TER as well as after writing the ATOM/HETATM records). I guess the following patch would cover both cases, because in the "preserve_atom_numbering" case, the serial number of the ATOM/HETATM records gets overwritten anyways. This assumes correct numbering of the TER record in the original PDB. 

```
--- PDBIO.py       2017-09-01 14:33:42.000000000 -0400
+++ PDBIO.py    2017-09-05 09:12:25.000000000 -0400
@@ -218,11 +218,11 @@
[...]

Issue #2292, opened 2019-10-04
Changing string in .gbk qualifiers introduces space after saving back to gbk


FantacticMisterFox:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
import sys; print(sys.version)

3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]

import platform; print(platform.python_implementation()); print(platform.platform())

CPython
Linux-5.2.17-200.fc30.x86_64-x86_64-with-fedora-30-Thirty

import Bio; print(Bio.__version__)

1.72
```

### Expected behaviour

I would expect that if I change the string of a qualifier in a sequence record this string is the same after saving the file to disk as a gbk file.

### Actual behaviour

Saving the string introduces white space at 52th and every (n*60+52)th position. Corresponds with line break in gbk file.

### Steps to reproduce

```
continous_string = "continous_string-------------------------------------------------------------------------"
record_dic = SeqIO.to_dict(SeqIO.parse("./ls_orchid.gbk", "genbank"))
record_dic["Z78533.1"].features[1].qualifiers["note"][0] = continous_string

SeqIO.write(record_dic.values(), "./new.gbk", "genbank")
new_record_dic = SeqIO.to_dict(SeqIO.parse("./new.gbk", "genbank"))

record_dic["Z78533.1"].features[1].qualifiers["note"][0] == continous_string
```
I used the ls_orchid.gbk file of the biopython cook book.

peterjc:
This is unavoidable given the line wrapping design of the GenBank format. Most qualifiers (including "note") assumes free text with space separation, with default line breaking at a suitable space which is not explicitly recorded. This is usually only a problem with very long chemical names, for example when describing an enzyme. Note that some qualifiers assume no spaces, e.g. "translation".

What you are trying to do is probably not suitable for the GenBank/EMBL file format.

What are you trying to record?

FantacticMisterFox:
Just a long ID. I mean I can remove all white spaces but as some qualifiers like the one you mentioned to not introduce white space but I assume that biopython does this internally. Anyhow if this is due to gbk than I will close the issue.

peterjc:
Thanks. The only way to do this in GenBank format would be to break the line length limit, so there is no good workaround other than accepting line breaks will be inserted on output (and give you extra spaces on parsing).

PR #989, opened 2016-11-12
Write correct fields in PDB TER record

lennax:
Slight modification of original PDBIO.diff
Closes https://redmine.open-bio.org/issues/2292

> Bio.PDBIO is happy to write TER records as "TER\n", which is inconsistent with PDB format specification.
> 
> The PDB format requires that TER records have some fields similar to ATOM records:
> '''The TER record has the same residue name, chain identifier, sequence number and insertion code as the terminal residue. The serial number of the TER record is one number greater than the serial number of the ATOM/HETATM preceding the TER.'''
> 
> [See http://www.wwpdb.org/documentation/format23/sect9.html#TER]
> 
> It leads to problem with programs that require correct TER records (like multiple structural alignment program MUSTANG), and crash when they are not found.
> 
> PDBIO.diff  - Proposed patch to PDBIO.py (656 Bytes) Michal J., 05/13/2007 05:31 PM

PR #3676, opened 2021-07-30
Pdb residue string

rob-miller:

This patch exposes functionality in PDBIO to generate a string for a single residue in PDB or PQR format.  The primary new functionality over the 'selector' mechanism is that residues can be generated with arbitrary chain IDs.  Relevant test code is added.

Less significant features are accessing as a string instead of e.g. using io.StringIO, documenting the preserve_atom_numbering parameter, addressing #1383, and minor cleanup of the PDBIO tests.

This functionality enables generating non-standard 'PDB' files like the attached for investigating residue sidechain interactions.  (suggest 'color by chain' if you load this into a viewer; each 'chain' is a group of potentially interacting PHE residues, with one member of each group superimposed on the N-CA-CB plane.)
[phe-pair.pdb.gz](https://github.com/biopython/biopython/files/6909336/phe-pair.pdb.gz)

With respect to #1383, the discussion appears to have stopped with the question of 'what to do if preserve_atom_numbering=True'.  I submit that this is not the default case, so (1) we should do the correct TER numbering in the default case, (2) for the non-default user this solution is correct for exporting correctly numbered (accounting for TER) input data, and (3) I think this exposes enough functionality for the non-default user to generate their desired output in the remaining cases.

With respect to the PDBIO tests, I believe that the `write_auto_numbering` and `write_preserve_numbering` test labels are currently swapped, and this fixes that as well.


JoaoRodrigues:
This makes the code in `PDBIO` a little more convoluted than it needs to be, in my opinion. Could you change the chains of the residues themselves instead before writing? What exactly would be the use case here? Trying to see if the change is worth it.

FWIW, I just created #3679 and was toying with the idea of breaking down the code in more functions, for easier test coverage, so this would play along those lines.

rob-miller:
I ran into issues trying to output and color (in a PDB viewer) arbitrary
sets of residues from multiple PDB files, preferably in a simple way to put
in a tutorial.  It seems like the existing selector method requires a bunch
of code to do this and processes the chain for every residue, when what I
really want is to re-use the lines already inside PDBIO.

With the current system to write a single residue I think I need two lines,
one to change the chain ID followed by a PDBIO.save() with a trivial
selector to catch the target residue.  Inside PDBIO.save() every time
though we open/close a file and spin through the chain just to find the
one/few target residues.  For flexibility on the user side I use
io.StringIO to collect the file writes back into strings because as I'm
processing the chain I've not decided that I want to write out this
residue, or maybe I want it after another residue later in the file, or
really I just want to send the data to stdout so I can pipe it into
something else. Doing things correctly I probably want to save/restore the
chain ID back to original state so I can re-use this code later.  After all
this PDB viewer Chimera complains that all my atom serial numbers are
messed up (always start with 1 for every save(), or use
preserve_atom_numbering and that will mess up across structures collected
in same file), and my TER lines don't have any data fields (not that I care
too much).

So the idea was just to pull out the PDBIO residue lines to an accessible
subroutine, but that insidious TER line complicates things because you
can't know when it will be needed.  I tried just storing the data for the
TER string in self, but this was slower in my tests.

I do agree it makes PDBIO more convoluted, but on the user side this way I
Initialize PDBIO's self.atom_number once and then freely mix/match
set_structure()s and residue and/or TER strings as desired - and the serial
numbers are all correct at the end too.

[Just to reiterate from the sample file above, all of this is only useful
for working with structurally aligned pieces so we're not just spitting out
residues randomly all over 3D space.]

Happy to merge in with your error message improvements if you like, and
happy to see it got you apparently catching up on some PDBIO work :-).

JoaoRodrigues:
Catching up indeed :)

The TER statement is something we can try to improve. Some tools write "naked" TER statements (simply TER), while others output a little more info. Most tools I know of don't actually read the info on the TER statement.

I like the idea of refactoring PDBIO to make it a little more modular. That's what I had in mind too when I looked at it yesterday. Would make it much simpler to unit test as well, as opposed to the current style of more like "integration" tests. Having said that, the records are pretty much written atom by atom (line by line), so bundling them would be annoying. We'd have to reverse the parser process in order to be robust, when writing is a much simpler aspect.

Can you post a sample of what you are trying to do? I have the feeling there's a simpler solution here.

rob-miller:
On Sun, 1 Aug 2021 at 19:57, João Rodrigues ***@***.***>
wrote:

> Can you post a sample of what you are trying to do? I have the feeling
> there's a simpler solution here.
>
>
Just quickly on this part, have you looked at the pdb file attached to the
first post on this PR?  You might have to go on GitHub, don't think it came
with the emails.

rob-miller:
> The TER statement is something we can try to improve. Some tools write "naked" TER statements (simply TER), while others output a little more info. Most tools I know of don't actually read the info on the TER statement.

lennax fixed Biopython's TER records in 7e185d8be91e170d8dd33c19d36fc265b2e9198f, but as you say I'd never noticed the other fields. Nonetheless it does seem like we should do the right thing, and I can't seem to load the redmine.open-bio.org link from that previous commit to see if that explains why it was an issue.

> I like the idea of refactoring PDBIO to make it a little more modular. That's what I had in mind too when I looked at it yesterday. Would make it much simpler to unit test as well, as opposed to the current style of more like "integration" tests. Having said that, the records are pretty much written atom by atom (line by line), so bundling them would be annoying. We'd have to reverse the parser process in order to be robust, when writing is a much simpler aspect.

I do think it is useful to bundle the atoms together as residues, and initially just had this as a convenience function inside my current internal_coords.py -- but assumed this would be criticized for code duplication.  I would *like* to replace the similar _pdb_atom_string in the Biopython internal_coords.py if we changed PDBIO's version to be public.  

> Can you post a sample of what you are trying to do? I have the feeling there's a simpler solution here.

Sorry, just noticed I had re-sequenced parts of my code before generating the previous sample and broke the serial numbers again.  Here's three more views on the same data since I'm playing with them:

[phe-pair-groups.pdb.gz](https://github.com/biopython/biopython/files/6918011/phe-pair-groups.pdb.gz)
[phe-pair-environs.pdb.gz](https://github.com/biopython/biopython/files/6918012/phe-pair-environs.pdb.gz)
[phe-pair-rings.pdb.gz](https://github.com/biopython/biopython/files/6918013/phe-pair-rings.pdb.gz)

Again, they are all intended to be color-by-chain in your structure viewer of choice, and this time have the serial numbers correct throughout.

I don't think you can actually generate these with the current PDBIO short of creating a new (SM)CRA object, and with that you would lose the ability to annotate different residues in the file as these examples do (another advantage of exposing this part of PDBIO as a subroutine).

peterjc:
Cross reference PR #989, to fix redmine 2292 https://redmine.open-bio.org/issues/2292 - from my emails:

> Bug #2292: Bio.PDBIO writes TER records without any required fields
> Author: Michal J.
> Status: Resolved
> Bio.PDBIO is happy to write TER records as "TER\n", which is inconsistent with PDB format specification.
> 
> The PDB format requires that TER records have some fields similar to ATOM records:
> '''The TER record has the same residue name, chain identifier, sequence number and insertion code as the terminal residue. The serial number of the TER record is one number greater than the serial number of the ATOM/HETATM preceding the TER.'''
> 
> [See http://www.wwpdb.org/documentation/format23/sect9.html#TER]
> 
> It leads to problem with programs that require correct TER records (like multiple structural alignment program MUSTANG), and crash when they are not found

(with a patch which I think is what Lenna committed)

rob-miller:
updated to incorporate #3679

rob-miller:
updated to incorporate #3882

@JoaoRodrigues I am still keen to get this in :-)

I want/need the ability to write arbitrary PDB residues with control over chain ID and atom numbering in order to study pairwise residue interactions like:

Screenshot 2022-04-08 at 15 21 56

I support doing this in #3774 and describe it in #3898.  I've implemented this functionality (`pdb_residue_string()`) in the internal_coords module, but really it should be in PDBIO - both logically and to benefit from improvements like #3679 and #3882. 

The implementation suggested here for PDBIO is much better than what I've done in internal_coords, and I would very much like to remove this code from internal_coords in time for a 1.80 release.

Abusing PDB files is an old and venerable technique, and the kind of thing I feel Biopython really should support!

rob-miller:
Some explanation on the image above and why it needs this PR:

The image shows 3 pairs of PHE residues from PDB structure 3PBL (Human dopamine D3 receptor), selected as putatively interacting on the basis that their C-alphas are within 6 angstroms of each other (there are much better and more stringent tests in the literature).  The N-Ca-C atoms (hedra) of the first residue in each pair (197-201, 345-346, and 390-394) are aligned using Chi1 transformation matrices exposed by the internal_coords module.

The attached PDB format file presents these pairs for analysis in PDB viewer software, as used to make the screenshot shown.  The key characteristics that I don't think are possible (at least without a lot more work) without this PR are:

* Each pair has its own chain ID (A, B, C)
* Each individual residue has a correctly formatted TER line
* Atom serial numbers are all sequential
* Each residue has a REMARK line annotating description and distance

Clearly the generated PDB file is not a 'structure', and various viewers will handle the issues this addresses differently.  The goal is to generate data that can be interactively selected, inspected and analysed with the minimum of noise - whether that's warnings and complaints or extraneous bonds.

[phe_pairs_4a.pdb.gz](https://github.com/biopython/biopython/files/8620717/phe_pairs_4a.pdb.gz)

rob-miller:
My reasons for wanting this merged:

1. It modularizes PDBIO.save(), which is usually considered a good thing, and factors out the residue generation in the same way the ATOM generation is factored out.  There is complication in handling the atom number for the TER string, but it is reasonable to expect to manage this at the chain level and expose this for manually generating PDB files.  Otherwise it is all the same code, so I really thought this would be a trivial change providing additional capability.
2. The current approach of supplying a select() function (a) doesn't support on-the-fly filtering of a chain as easily or efficiently because you have to generate that function and re-scan the chain each time, (b) doesn't support more complicated things like arbitrary chain IDs for coloring in viewers, (c) writes the TER line after each set of residues which may not be desired, (d) writes to a file or stringio which has to be configured separately so less flexible, and (e) probably some other stuff that I've forgotten.
3. It really should not be in the internal coords code, but it is there at least to support the examples in the documentation.  The support for this could and should be removed from internal coords.
4. Everything that writes PDB output should be together in one place so that changes/improvements like PR #3857 get picked up automatically.

peterjc:
Could you take a fresh look at this @JoaoRodrigues? Or maybe @etal or @lennax?

JoaoRodrigues:
Will do today.

rob-miller:
Update for 1.84

PR #3679, opened 2021-08-01
Improvements to PDBIO

### LINKED ISSUE #3438: PDBIo error when saving chain with more than one character
### Setup

system as follows:

```python

3.8.6 | packaged by conda-forge | (default, Nov 27 2020, 19:31:52) 
[GCC 9.3.0]
CPython
Linux-5.8.0-31-generic-x86_64-with-glibc2.10
1.78
```

### Expected behaviour

Using `PDBio.save` to save any chain that has more than one character as identifier should work, regardless on the length of the chain id.

### Actual behaviour

When subclassing `Select` for chain extraction, and using `PDBio` for saving, an exception is raised as follows.
```python
  File "PDBIO.py", line 370, in save
    s = get_atom_line(
  File "PDBIO.py", line 229, in _get_atom_line
    return _ATOM_FORMAT_STRING % args
TypeError: %c requires int or char

```

### Steps to reproduce

Minimal code to reproduce:

```python
from Bio.PDB import PDBList, PDBIO, FastMMCIFParser, Select

class ChainSelect(Select):
    def __init__(self, chain):
        self.chain = chain

    def accept_chain(self, chain):
        if chain.get_id() == self.chain:
            return 1
        else:
            return 0

pdbl = PDBList()
io = PDBIO()
parser = FastMMCIFParser(QUIET=True)

pdbl.retrieve_pdb_file('6kmw', pdir='.', file_format='mmCif')
structure = parser.get_structure('6kmw', '6kmw.cif')
io.set_structure(structure)
io.save('6kmw_aA.pdb', ChainSelect('aA'))
```

JoaoRodrigues:
Make error messages clearer by specifying the offending atom(s). In
addition, improve performance by accessing class attributes directly
instead of using get_xxx() methods.

Closes #3438.

### DIFF
--- a/Bio/PDB/PDBExceptions.py
+++ b/Bio/PDB/PDBExceptions.py
@@ -29,3 +29,10 @@ class PDBConstructionWarning(BiopythonWarning):
     """Define class PDBConstructionWarning."""
 
     pass
+
+
+# The SMCRA structure could not be written to file
+class PDBIOException(Exception):
+    """Define class PDBIOException."""
+
+    pass
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -6,6 +6,12 @@
 """Output of PDB files."""
 
 
+import warnings
+
+# Exceptions and Warnings
+from Bio import BiopythonWarning
+from Bio.PDB.PDBExceptions import PDBIOException
+
 # To allow saving of chains, residues, etc..
 from Bio.PDB.StructureBuilder import StructureBuilder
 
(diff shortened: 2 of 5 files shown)


JoaoRodrigues:
@rob-miller , if you could have a look to make sure things seem alright (I changed one of your tests), that would be great!

rob-miller:
Yes looks fine - but curious why?  Memory limits on the test hosts?

JoaoRodrigues:
Issues with accessing the file on Windows.

PR #3681, opened 2021-08-01
Refactor GHA CI workflow for speed.

JoaoRodrigues:
A first (or n-th) attempt to use cache to make things a little faster. I also added the `concurrency` keyword which runs only one action/workflow per github branch or reference, and a `timeout` of 20 minutes to the tests to prevent hanging.


JoaoRodrigues:
@peterjc I quit :) I have no idea how to fix the PyPy3 issues. Getting some dependencies built takes forever (10+ min) and testing generally speaking is very slow. 

I would maybe move PyPy3 CI to a different yaml file/workflow and try to make it optional (or run only on push or periodically), otherwise our entire CI is slowed down considerably. What do you think?

I got the docs error to go away, but the RST linting is failing (never worked in GHA) because of this error below. Could you take a look?

```

  # Check sort order (bash call work around for pipe character)
  bash -c \' grep "^- " CONTRIB.rst | LC_ALL=C sort -u -c -f \'
  # Check copyright date
  bash -c \' grep "1999-`date +'%Y'`" LICENSE.rst \'
  # Check no __docformat__ lines
  bash -c "if grep --include '*.py' -rn '^__docformat__ ' Bio BioSQL Tests Scripts Doc ; then echo 'Remove __docformat__ line(s), we assume restructuredtext.'; false; fi"
  # Black style check:
  black --check --diff .
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.6/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.6/x64/lib
sort: open failed: "'": No such file or directory
grep: -c: line 0: unexpected EOF while looking for matching `''
grep: -c: line 1: syntax error: unexpected end of file
Error: Process completed with exit code 2.
```

peterjc:
Pragmatic answer: Test pypy with as few dependencies as possible, I would suggest just numpy and mysqlclient as per the old TravisCI setup where again, the pypy tests were the slowest.

peterjc:
And yes, #3682 aims to fix the RST / Sphinx API doc testing

JoaoRodrigues:
Will split pypy3 into its own job (seems the simplest) and keep deps to a minimum, like the macos and windows builds.

peterjc:
I don't immediately see why that grep based linting is failing, but you could try dropping the ``bash -c ...`` bit which was a hack on TravisCI?

PR #3685, opened 2021-08-04
General improvements to GHA workflows

JoaoRodrigues:
An attempt to fix the maladies of the current CI workflow on GHA: slow PyPy, long running times, etc. Details in the commit message of the first commit, copied here for convenience:

- Cache Python installations on Linux for faster execution.
- Separate PyPy runners from standard Linux Python.
- CI dependencies are stored on a file for versioning and local testing.
- Time limits (15-20 minutes) on each testing step to avoid hanging.
- Concurrency limits to avoid multiple commits to the same PR to queue.


JoaoRodrigues:
@peterjc , this is a cleaner version of #3681 with some extra documentation and less trial-and-error commits.

peterjc:
Looks like it needs a conflict resolved, guessing there was a minor edit to the old YAML file?

JoaoRodrigues:
It does? On my end it says "branch has no conflicts with the base branch".

JoaoRodrigues:
Also, since we test on Windows, is there a reason to keep the appveyor CI runner?

peterjc:
Well if GitHub lets you merge it, great. It tells me "This branch cannot be rebased due to conflicts" - strange.

In theory GA can replace all our existing CI setups (TravisCI, CircleCI and AppVeyor), although we're not at that point yet. I think this is all we still need?:

- [ ] Upload coverage data to CodeCov
- [ ] Build the API docs with Sphinx and push that to GitHub Pages (was on TravisCI, now on CircleCI as of #3678)
- [ ] Non-Python dependency testing (e.g. ``paml`` - had lots on TravisCI, will be less as we remove application wrappers - see #3344)

JoaoRodrigues:
Thanks! Will merge then, if there is a conflict it will probably break.

The coverage data is already generated by GHA, I just don't upload it because we'd need to setup secrets I think.

The docs, if you can have a look at the error I mentioned in the [other commit](https://github.com/biopython/biopython/pull/3681#issuecomment-890686139) that'd be great. I can't figure it out... As for Sphinx, I can have a look at what you did for CircleCI and try to convert it.

As for the dependencies, my guess is that we can copy the code from Travis. The ideal scenario would be to have a file with the dependencies or their versions so we can cache them. Even a bash script to setup dependencies would work.

I'll open a new issue just to keep track of these bullet points.

peterjc:
The permissions for pushing to codecov are not fresh in my mind, but according to https://app.codecov.io/gh/biopython/biopython/settings no need:

> *Token not required for public repositories uploading from Travis, CircleCI, AppVeyor, Azure Pipelines or GitHub Actions.*

So should just be adding something like this:

```
conda install -c conda-forge codecov
codecov --file Tests/coverage.xml -X pycov -X gcov
```

Adding the Sphinx API doc building and push ought to be easy - you will need to generate a new deploy key on the docs dependency with write permissions (I think you have permissions to do that?), read https://github.com/biopython/biopython/blob/6950507e0ac9cac4576d39e41051f10bf908cfac/.github/deploy_docs.sh

With TravisCI we used Ubuntu Xenial packages via apt. Elsewhere I have used conda. We have previously fetched binaries of some tools pre-compiled from the author's website etc. Caching should mean this isn't a big speed overhead.

PR #3708, opened 2021-08-30
More spelling fixes via codespell

peterjc:

Continues work started in #3703.

Automating this looks painful, especially for the tests. It may however make sense for the documentation folder where there are less false positives to exclude.


peterjc:
May want to add something like this for checking the docs, not too long an ignore list:

```
$ codespell Doc/ --skip "*.html,*.log,*.txt,*.gbk,Doc/_minted-Tutorial/*" \
--ignore-regex '(^|\W)([A-Z]{2,3})(\W|$)' -L 'ser,hsa,hist,fpr'
```

Would want to replace one "noe" with "NOE", so that it and other 2 or 3 letter upper case acronyms are ignored via the regex.

PR #3717, opened 2021-09-07
Use https://pre-commit.ci/ for style checks

peterjc:

This pull request drops calling the tool ``pre-commit`` from our existing CI setups.

Separately we would enable Biopython on https://pre-commit.ci/ instead.

(Potential followup to enable their bot for applying changes including potentially black to pull requests, currently via 9c2be2db7e317a8ea8c3438b9b9324eb86b1ece4 that is off)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,30 +16,6 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Combine precommit config and python versions for caching
-          command: |
-            cat .pre-commit-config.yaml > pre-commit-deps.txt
-            python -VV >> pre-commit-deps.txt
-      - restore_cache:
-          keys:
-          - v1-precommit-deps-{{ checksum "pre-commit-deps.txt" }}
-      - run:
-          name: Style install
-          command: |
-            python --version
-            pip install pre-commit
-            pre-commit install
-            pre-commit install-hooks
-      - save_cache:
-          paths:
-            - ~/.cache/pre-commit
-            - ./venv
-          key: v1-precommit-deps-{{ checksum "pre-commit-deps.txt" }}
-      - run:
-          name: Style checks
-          command: |
-            pre-commit run -a
-      - run:
           name: Dependencies
           command: |
             python --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
   # since it is used in most actions.
   # Adapted from: https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d
   linux_prep:
-    needs: [style]
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -69,7 +68,6 @@ jobs:
         python setup.py sdist --formats=gztar,zip
 
   build_wheels:
-    needs: [style]
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
@@ -113,13 +111,13 @@ jobs:
   # tests install only numpy (lightweight version).
   # NOTE: PyPy only runs on Linux with minimal dependencies because it's running slowly.
   #
-  # Tests trigger only of the style and build steps succeed.
+  # Tests trigger only if the build steps succeed.
   #
 
   # Linux is where the thorough tests run. Install all dependencies and run all (offline) tests.
   test_linux:
     timeout-minutes: 15
-    needs: [style, build, build_wheels, linux_prep]
+    needs: [build, build_wheels, linux_prep]
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -169,7 +167,7 @@ jobs:
 
   test_macos:
     timeout-minutes: 15
-    needs: [style, build, build_wheels]
+    needs: [build, build_wheels]
     runs-on: macos-latest
     strategy:
       matrix:
@@ -202,7 +200,7 @@ jobs:
 
   test_windows:
     timeout-minutes: 15
-    needs: [style, build, build_wheels]
+    needs: [build, build_wheels]
     runs-on: windows-latest
     strategy:
       matrix:
@@ -235,7 +233,7 @@ jobs:
 
   test_pypy:
     timeout-minutes: 20  # a little more time
-    needs: [style, build, build_wheels]
+    needs: [build, build_wheels]
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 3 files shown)


peterjc:
Ah, I broke GitHub actions here:

```
Invalid workflow file : .github/workflows/ci.yml#L22
The workflow is not valid. .github/workflows/ci.yml (Line: 22, Col: 3): The workflow must contain at least one job with no dependencies.
```

JoaoRodrigues (approved):

JoaoRodrigues:
I didn't know about `pre-commit.ci`. Nice!

peterjc:
OK, that is now live on pre-commit.ci, badge using markdown:

[pre-commit.ci status](https://results.pre-commit.ci/latest/github/biopython/biopython/master)

It will be unknown until the next commit happens - which will probably be in the next day or so. Waiting and watching...

Note that I have only enabled this for the main Biopython repository (although it should see it is not configured on the rest and ignore them).

peterjc:
And we have a timeout:

https://results.pre-commit.ci/run/github/151541/1631052086.FAn4UAWoQBuM7pmmamPuPQ

```
biopython / biopython · #3662 
queued at: 2021-09-07 22:01:26

total time to completion: 3m18.2s

Check that executables have shebangs.....................................Passed
Check JSON...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
black....................................................................Passed
flake8...................................................................

(timeout after 180 seconds)
```

And again on a second rebase of that PR:

https://results.pre-commit.ci/run/github/151541/1631052605.o3KAzLC-TIm0IUBf194DSg

That isn't a complete surprise, flake8 does take a while to run nowadays - enough that I suspect a performance regression somewhere (perhaps in a plugin).

peterjc:
Timings on my desktop, all best of at least three, clearly highlight flake8 as the main time sink:

```
$ pre-commit --version
pre-commit 2.13.0
$ time pre-commit run -a
Check that executables have shebangs.....................................Passed
Check JSON...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
black....................................................................Passed
flake8...................................................................Passed
blacken-docs.............................................................Passed
rstcheck.................................................................Passed
doc8.....................................................................Passed
codespell................................................................Passed
Enforce https://doi.org/ style...........................................Passed

real	2m19.514s
user	2m31.016s
sys	0m5.065s
```

Then pre-commit except flake8, very quick:

```
$ time SKIP=flake8 pre-commit run -a
Check that executables have shebangs.....................................Passed
Check JSON...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
black....................................................................Passed
flake8..................................................................Skipped
blacken-docs.............................................................Passed
rstcheck.................................................................Passed
doc8.....................................................................Passed
codespell................................................................Passed
Enforce https://doi.org/ style...........................................Passed

real	0m6.570s
user	0m22.566s
sys	0m3.313s
```

This suggests that the vast majority of the time is in flake8 (and its plugins), and yes:

```
$ flake8 --version
3.9.2 (flake8-blind-except: 0.2.0, flake8-bugbear: 21.4.3, flake8-comprehensions: 3.6.1,
flake8_implicit_str_concat: 0.2.0, mccabe: 0.6.1, pycodestyle: 2.7.0, pyflakes: 2.3.1, rst-docstrings:
0.2.3) CPython 3.8.5 on Darwin
$ time flake8 Bio/ BioSQL/ Scripts/ Doc/examples/ Tests/ setup.py

real	1m47.655s
user	1m43.310s
sys	0m1.249s
```

There does seem to be a slight overhead running flake8 under pre-commit, versus separately.

Note everything seems to be run on a single thread. Same version of flake8, plugins not pinned, so there might be some variation there.

I don't think flake8 does any caching - unlike black, but that has the advantage of being far less configurable (although pre-commit.ci may not retain the black cache between runs).

peterjc:
In my tests outside of Biopython on other repositories, using pre-commit.ci was faster (for an all files check) than CircleCI or GitHub actions, and far easier to setup - no manual caching of the dependencies required. It also (for now at least) has a shorter queue time.

The much more interesting bit is the (currently disabled 9c2be2db7e317a8ea8c3438b9b9324eb86b1ece4 here) the option to have the pre-commit.ci bot apply any changes from our pre-commit configuration to pull requests.

e.g. If we had black setup not just to do ``--check``, but the default of actually editing the files (which I have considered doing on other projects purely for the git pre-commit stage), we automated the fairly common situation of a pull request needing black applied. If the PR author makes further local changes, they would have to either force push (overwrite the bot's changes), or incorporate them (e.g. rebase on the updated PR branch).

Another example, but less compelling, is codespell with ``-w`` will make what it considers to be unambiguous spelling corrections. In my experience in applying this to our code base in the last week, there is a fair chance of false positives - but that ought to be obvious during pull request review.

peterjc:
I have cherry-picked the badge to master as 091cbce1b96a6a8d17d65efef52611680c78d7b3

peterjc:
For reference, black timings on my macOS desktop with no black cache - best of three:

```
$ rm -rf ~/Library/Caches/black && time black --check Bio/ BioSQL/ Scripts/ Doc/examples/ Tests/ setup.py
All done! ✨ 🍰 ✨
565 files would be left unchanged.

real	0m32.543s
user	3m18.294s
sys	0m2.046s
```

Hopefully pre-commit.ci is caching this, doing that on CircleCI/GitHub Actions seems complicated (when would we update the black cache? Weekly? Every release?).

peterjc:
Proposing to squash-and-merge the PR now, which has been updated to only remove calling pre-commit from CircleCI (saving about 4 mins of a typical 18 mins run).

That would leave use with the new pre-commit.ci check (about 2 mins doing pre-commit checks except flake8, but all files) and GitHub actions style checks (about 20s doing all the pre-commit checks, but only for changed files).

This seems a much better use of the continuous integration server time.

peterjc:
Confirmed on https://github.com/pre-commit-ci/issues/issues/86 that pre-commit.ci does not retain the black cache, so this is a closer imitation of what they run (leaving aside what exactly gets parallelised):

```
$ rm -rf ~/Library/Caches/black && time pre-commit run -a --verbose
Check that executables have shebangs.....................................Passed
- hook id: check-executables-have-shebangs
- duration: 0.06s
Check JSON...............................................................Passed
- hook id: check-json
- duration: 0.06s
Debug Statements (Python)................................................Passed
- hook id: debug-statements
- duration: 1.6s
Fix End of Files.........................................................Passed
- hook id: end-of-file-fixer
- duration: 0.09s
Mixed line ending........................................................Passed
- hook id: mixed-line-ending
- duration: 1.41s
black....................................................................Passed
- hook id: black
- duration: 31.79s

All done! ✨ 🍰 ✨
565 files would be left unchanged.

flake8...................................................................Passed
- hook id: flake8
- duration: 128.75s
blacken-docs.............................................................Passed
- hook id: blacken-docs
- duration: 1.07s
rstcheck.................................................................Passed
- hook id: rstcheck
- duration: 1.04s
doc8.....................................................................Passed
- hook id: doc8
- duration: 1.01s
codespell................................................................Passed
- hook id: codespell
- duration: 0.31s
Enforce https://doi.org/ style...........................................Passed
- hook id: doi-link-style
- duration: 0.17s

real	2m48.491s
user	5m48.807s
sys	0m6.607s
```

JoaoRodrigues:
We could have black act on every PR, either pre-merge (https://github.com/psf/black/actions/runs/17913292/workflow) or post-merge.

It doesn't really matter to me tbh, I just prefer to keep things simple and avoid multiple services unless there's a clear advantage (and I don't see one that big here).

👍 good job with all the detective work with flake8, maybe we can let the maintainers of those plugins know that there is this slow performance.

peterjc:
I shouldn't be surprised there at GitHub actions recipes for black on PR now too. That's one less unique selling point - although I do like the minimal configuration for pre-commit.ci

I'll open a separate issue on black / flake8 performance. It looks like a minority of our files are having a disproportionate impact.

PR #3761, opened 2021-10-10
Fix Exonerate text parser to correctly determine strand from ':[revcomp]'

### LINKED ISSUE #3757: Parsing Exonerate text output with SearchIO does not assign target strand correctly
### Setup

system as follows:

```
Python 3.8.3 (default, May 19 2020, 13:54:14)
CPython
macOS-10.15.7-x86_64-i386-64bit
BioPython 1.78
```

### Expected behaviour

When parsing the output of an Exonerate search performed using the model `protein2genome`, `SearchIO.parse(<exonerate_output_file>, 'exonerate-text')` should assign hsp strands to `-` (converted to `-1) when the nucleotide hit is on the negative strand. 

### Actual behaviour

All hsps get assigned to the positive strand `+` (converted to `1`), regardless of whether the nucleotide hit is on the positive or negative strand.

### Steps to reproduce

Tested with Exonerate v2.2 and v2.4.

Using the attached Exonerate output file  [exonerate_results.fasta.zip](https://github.com/biopython/biopython/files/7263583/exonerate_results.fasta.zip) the SearchIO command is:

`exonerate_hits_from_alignment = list(SearchIO.parse('exonerate_results.fasta', 'exonerate-text'))`

Nucleotide hits are on the negative strand (target description contains `[revcomp]`). However:

```
for qresult in exonerate_hits_from_alignment:
    for hsp in qresult.hsps:
        print(hsp.hit_strand_all)
```

Output:
```
[1]
[1]
[1]
[1]
[1]
[1]
[1, 1]
[1]
```

I think this might be due to the code [here](https://github.com/biopython/biopython/blob/1e4acd3cc6a1ef1c79da187b0aaea8a34586d6d9/Bio/SearchIO/ExonerateIO/_base.py#L391); the strand is set based on the presence of the string `:[revcomp]` in the query or target description line, whereas the Exonerate output in version 2.2 and 2.4 contains `[revcomp]` instead , i.e. no preceeding colon.

Cheers,

Chris

bow:
This PR resolves #3757, where the `exonerate-text` parser does not correctly determine when a hit or query sequence lies on the negative strand.

Thanks to  @chrisjackson-pellicle , the root cause was determined to be the `:[revcomp]`, which was appended by Exonerate to the query / hit description. In more recent versions of Exonerate, the string has been updated to `[revcomp]` but this has not been followed by the parser, resulting in the negative strand always being set incorrectly as the positive strand.

In addition to the fix, this PR also refactored the functionality into a separate function.

Closes #3757

### DIFF
--- a/Bio/SearchIO/ExonerateIO/_base.py
+++ b/Bio/SearchIO/ExonerateIO/_base.py
@@ -311,6 +311,35 @@ def _parse_hit_or_query_line(line):
     return id, desc
 
 
+def _get_strand_from_desc(desc, is_protein, modify_desc=True):
+    """Determine the strand from the description (PRIVATE).
+
+    Exonerate appends ``:[revcomp]`` (versions <= 2.2) or ``[revcomp]``
+    (versions > 2.2) to the query and/or hit description string. This function
+    outputs '-' if the description has such modifications or '+' if not. If the
+    query and/or hit is a protein sequence, a '.' is output instead.
+
+    Aside from the strand, the input description value is also returned. It is
+    returned unmodified if ``modify_desc`` is ``False``. Otherwise, the appended
+    ``:[revcomp]`` or ``[revcomp]`` is removed.
+
+    """
+    if is_protein:
+        return ".", desc
+
+    suffix = ""
+    if desc.endswith("[revcomp]"):
+        suffix = ":[revcomp]" if desc.endswith(":[revcomp]") else "[revcomp]"
+
+    if not suffix:
+        return "+", desc
+
+    if modify_desc:
+        return "-", desc[:-len(suffix)]
+
+    return "-", desc
+
+
 class _BaseExonerateParser(ABC):
     """Abstract base class iterator for exonerate format."""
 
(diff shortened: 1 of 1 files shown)


chrisjackson-pellicle:
Hi @peterjc. I just noticed that the `[revcomp]` vs `:[revcomp]` issue doesn't appear to be related to the Exonerate version, as I'd thought, but rather on whether the target sequence has a description in the fasta header. If a description is present, Exonerate appends `:[revcomp]` to the query/hit description, whereas if it's absent, Exonerate appends `[revcomp]`. 

@bow's PR still fixes the issue, of course, but the docstring for `_get_strand_from_desc()` will need updating.

I've attached both a new Exonerate example file ([exn_24_m_protein2genome_revcomp_no_seq_desc.exn.zip](https://github.com/biopython/biopython/files/7559291/exn_24_m_protein2genome_revcomp_no_seq_desc.exn.zip); this has an example of both `[revcomp]` and `:[revcomp]`) along with a new test to be added to `class ExonerateTextCases(unittest.TestCase)` in `test_SearchIO_exonerate.py` ([test_exn_24_m_protein2genome_revcomp_no_seq_desc.py.zip](https://github.com/biopython/biopython/files/7559298/test_exn_24_m_protein2genome_revcomp_no_seq_desc.py.zip)). 

If these are used, I guess the [readme][1] needs to be updated to include:


```
Exonerate 2.4
-------------
exn_24_m_protein2genome_revcomp_no_seq_desc.exn           protein2genome model alignment
```

Apologies for any manual handling - I wasn't sure how else to add this to an existing PR (if that's even possible when it's not my PR?).

Finally, I think I'll need to wait until this PR is accepted before submitting a new PR for [#3769][2], as the current issue effects this as well (and hence any tests for #3769 with a revcomp hit won't pass until #3757 is fixed).




[1]: https://github.com/biopython/biopython/blob/8546c9622bf7ca7d257a60477f4b9b4e39492b02/Tests/Exonerate/README.txt
[2]: https://github.com/biopython/biopython/issues/3769

peterjc:
Thanks Chris - having another set of eye test this code is great.

Any Biopython maintainer (assuming the contributor ticked the box) would be able to update the pull request, but otherwise only the author. You could however make a new branch from that branch, add some more commits, and submit a new PR.

But as you suggest, it would be simpler as you suggest to merge this first (which I will do now given your testing), and then submit your follow up PR for the docstring.

Peter

Issue #3433, opened 2020-12-05
test suite execution failure with dssp version 4


emollier:
Greetings,

During an archive rebuild of the Debian operating system on Aarch64 platform, Lucas Nussbaum caught an issue with Biopython test suite relative to the DSSP test, which I could reproduce in a more usual x86_64 configuration.  See Debian [Bug#976542](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976542) for more information.

The test suite chokes on _test_PDB_DSSP.py_.  I am running the Biopython 1.78 test suite on top of _mkdssp_ version 4.0.0, which is built against the libcif++1 1.0.0:

```
$ mkdssp --version
mkdssp version 4.0.0 2020-11-27
$ dpkg -l | grep libcifpp
ii  libcifpp1  1.0.0-3  amd64  Library files for libcifpp
```

It turns out that the command _mkdssp_ version 4 now outputs mmCIF file format by default, instead of the DSSP format in the previous version 3.  I am able to get back the test suite to a working condition by passing explicitly the new option _--output-format=dssp_ to make sure the output file in the format expected by the test.  I have not suggested my patch because I believe it would break compatibility with _mkdssp_ 3, but I thought you would want to be aware of the issue.

Kind Regards,
Étienne.

peterjc:
I can see two ways to approach this and preserve support for older versions of the tool. First, make the code here work with the old output, or mmCIF equivalent (which Biopython should be able to parse):

https://github.com/biopython/biopython/blob/biopython-178/Bio/PDB/DSSP.py#L254

Second, invoke dssp with the extra command line argument. That could be done automatically by checking the tool version, or down to the caller (the test suite does know the version already).

I've not looked at the output, but the first approach seems more elegant?

peterjc:
A third option: Document we require an older version, and skip the test if version 4 is found. Pragmatic but not very satisfactory.

emollier:
Hi Peter,

Peter Cock, on 2020-12-05 15:21:13 -0800:
> I can see two ways to approach this and preserve support for older versions of the tool. First, make the code here work with the old output, or mmCIF equivalent (which Biopython should be able to parse):
> 
> https://github.com/biopython/biopython/blob/biopython-178/Bio/PDB/DSSP.py#L254

I like that idea, but I'm a bit concerned by the bugs section of
mkdssp(1) manual, stating in its BUGS section:

```
The mmCIF format currently lacks a lot of information
that was available in the old format like information
[...]

PR #3775, opened 2021-10-29
Add Support for DSSP v4.0

### LINKED ISSUE #3763: test_PDB_DSSP failure with mmcif_chain_dict[chain_id] KeyError: 'E'
Greetings,

When trying to upgrade the _python-biopython_ package to version 1.79 in Debian Sid, I'm getting the following unit test error:

```
ERROR: test_dssp_with_mmcif_file_and_different_chain_ids (test_PDB_DSSP.DSSP_tool_test)
Test DSSP generation from MMCIF which has different label and author chain IDs.
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9/build/Tests/test_PDB_DSSP.py", line 125, in test_dssp_with_mmcif_file_and_different_chain_ids
    dssp = DSSP(model, pdbfile)
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9/build/Bio/PDB/DSSP.py", line 476, in __init__
    chain_id = mmcif_chain_dict[chain_id]
KeyError: 'E'
```

It turns out that the _KeyError_ occurs a couple of lines before the _try: except KeyError:_ statement which, I suppose, is the object of the test.  Note that this occurs with the patch mentionned in the [issue #3433](https://github.com/biopython/biopython/issues/3433), in order to work around the issue caused by the new default file format, so there is a good chance this is a self inflicted breakage.  If need be, you can refer to the [full log on Salsa CI](https://salsa.debian.org/med-team/python-biopython/-/jobs/2061317/raw), on which I added a couple of _print_ in the loop, in an attempt to understand what happens.

Core components versions are:

```
3.9.7 (default, Sep 24 2021, 09:43:00)
CPython
Linux-5.14.0-1-amd64-x86_64-with-glibc2.32
```

Current versions of _mkdssp_ and _libcifpp_ available in Sid are:

```
$ mkdssp --version
mkdssp version 4.0.0 2020-12-18
$ dpkg -l | grep libcifpp
ii  libcifpp1  1.0.1-3  amd64  Library files for libcifpp
```

Kind Regards,
Étienne.

cbalbin-bio:

Closes #3763 #3433

This is a patch to the current DSSP parser to support DSSP v4.0 . The DSSP parser currently only handles up to v3.x.x. I think supporting v4.0 is important as that is what is install by default when using apt on Ubuntu 21.04 or greater. Currently it is not mentioned in the documentation that v4.0 is not supported and this could be undesirable to some users.

DSSP v4.0 shifts to a mmCIF format by default. You can output a DSSP v3.0 style report using the flag --output-format=dssp. While I am a big fan of mmCIFs gain in popularity. DSSP 4.0 s mmCIF out is missing some critical items such as accessibility; [see comment](https://github.com/biopython/biopython/issues/3433#issuecomment-739473781).
Thus I think it is better to stick with the classic DSSP format output.

DSSP v4.0 also now uses auth_asym_id for the chain IDs instead of label_asym_id when ran on mmCIF files [see comment](https://github.com/biopython/biopython/issues/3763#issuecomment-954334393). For older versions of DSSP, a mapping had to be done from the label to auth chain ID f85e4d91b27f72849f1288040519d6a3dd6537ab. This is now disabled when a DSSP v4.0 or higher executable is detected.

In order to enable the expected experience for the most users, whether using PDB or mmCIF format files, or DSSP v3 vs v4 I added a version check to the initializer of the DSSP class. This check uses code similar to that found the the DSSP unit test. If it is found that the executable has a version greater than or equal to '4.0.0' the flag --output-format=dssp is added to the dssp command. Also if the user is found to be using v4.0.0 or greater with an mmCIF file. The label to auth chain ID mapping is disabled as the results already contain the auth chain ID instead.

Thanks for your time,

Christian

### DIFF
--- a/Bio/PDB/DSSP.py
+++ b/Bio/PDB/DSSP.py
@@ -435,7 +435,7 @@ class DSSP(AbstractResiduePropertyMap):
 
     """
 
-    def __init__(self, model, in_file, dssp="dssp", acc_array="Sander", file_type="", dssp_version="3.9.9"):
+    def __init__(self, model, in_file, dssp="dssp", acc_array="Sander", file_type=""):
         """Create a DSSP object.
 
         Parameters
@@ -453,9 +453,6 @@ class DSSP(AbstractResiduePropertyMap):
         file_type: string
             File type switch: either PDB, MMCIF or DSSP. Inferred from the
             file extension by default.
-        dssp_version: string
-            Version of DSSP used to generate in_file if in_file is of DSSP
-            format. Else version will be automatically detected.
 
         """
         self.residue_max_acc = residue_max_acc[acc_array]
--- a/Bio/PDB/DSSP.py
+++ b/Bio/PDB/DSSP.py
@@ -204,7 +204,9 @@ residue_max_acc = {
     },
 }
 
+
 def version(version_string):
+    """Parse semantic version scheme."""
     return tuple(map(int, (version_string.split("."))))
 
 
@@ -234,7 +236,7 @@ def dssp_dict_from_pdb_file(in_file, DSSP="dssp", dssp_version="3.9.9"):
 
     DSSP : string
         DSSP executable (argument to subprocess)
-    
+
     dssp_version : string
         Version of DSSP excutable
 
@@ -474,9 +476,13 @@ class DSSP(AbstractResiduePropertyMap):
             # calling 'dssp' will not work in some operating systems
             # (Debian distribution of DSSP includes a symlink for 'dssp' argument)
             try:
-                version_string= subprocess.check_output([dssp, "--version"], universal_newlines=True)
+                version_string = subprocess.check_output(
+                    [dssp, "--version"], universal_newlines=True
+                )
                 dssp_version = re.search(r"\s*([\d.]+)", version_string).group(1)
-                dssp_dict, dssp_keys = dssp_dict_from_pdb_file(in_file, dssp, dssp_version)
+                dssp_dict, dssp_keys = dssp_dict_from_pdb_file(
+                    in_file, dssp, dssp_version
+                )
             except FileNotFoundError:
                 if dssp == "dssp":
                     dssp = "mkdssp"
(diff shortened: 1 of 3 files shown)


rob-miller:
On my ubuntu 22.04 install, dssp version 4.0.4 is happily reading .ent.gz and .cif.gz files (these are what my automated pdb mirror downloads).  Please might you add these file types to the accepted list in DSSP.__init__()?

IMHO we should not be failing on file name extension.  Instead trust the user to supply a valid format regardless of what their pipeline calls it and print the command line if if fails so they can work it out if there's a problem.

rob-miller:
This is working well for me (once I read the docs on how to skip the file type test) and I support merging.

May I suggest changing make_dssp_dict() to use as_handle() instead of open():

    from Bio.File import as_handle
    ...
    def make_dssp_dict(filename):
        with as_handle(filename, "r") as handle:
            return _make_dssp_dict(handle)

This maintains current functionality and further enables the user to get the dssp output from their own subprocess if/when the one in this code fails again.

Also

        dssp : string
            The dssp executable (ie. the argument to subprocess)

might be better described as something like

        dssp : string
            The name of the dssp executable ('dssp' or 'mkdssp', ie. the argument to subprocess)

rob-miller:
Just to note, I've run this on 6,145 mmCIF chains from a current Dunbrack PISCES nonredundant list, and the only issue is an alignment error on PDB code 3sqz chain A due to Cys 111 being a non-standard residue.  The PDB format file for this chain is handled correctly.

JoaoRodrigues:
Going back to this, thanks for the ping @rob-miller. I downloaded DSSP versions 3.x.x and 4.x.x (from their github repos) and installed them locally. I am having an issue with reading mmCIF files with version 4.x.x. The tests will pass on version 3.x.x but version 4.x.x produces an empty dssp_dict file. I am looking into it, but it might be related with issues with `label_` and `auth_` fields in the cif file.

JoaoRodrigues:
Uh oh, sorry... my `git pull origin master` created this mess!

JoaoRodrigues (approved):

PR #3777, opened 2021-11-03
defaultdict default_factory receives no arguments

samuelstanton:

There's a small bug in the SASA implementation. `ATOMIC_RADII` is a `defaultdict`, which is meant to handle KeyErrors from unspecified atomic radii. The problem is that defaultdicts do not pass any arguments to their `default_factory` function, so the `SharkeRupley.compute` call fails if it ever actually encounters a KeyError. The fix is trivial.

See the Python docs for more info: https://docs.python.org/3/library/collections.html#collections.defaultdict.default_factory


JoaoRodrigues:
I never got a notification about this, I'm sorry @samuelstanton ! Thanks for catching the bug, the fix is indeed trivial.

JoaoRodrigues (approved):

Issue #3780, opened 2021-11-05
ladderize fails in Bio.Nexus.Trees.Tree.to_string()


anders-was-here:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```python
3.7.9 (default, Aug 19 2020, 17:05:11) 
[GCC 9.3.1 20200408 (Red Hat 9.3.1-2)]
CPython
Linux-5.8.18-100.fc31.x86_64-x86_64-with-fedora-31-Thirty_One
1.79
```

### Actual behaviour

```
Traceback (most recent call last):
File "to_string.py", line 11, in <module>
    print(t.to_string(ladderize="LEFT"))
File "/home/andla/.local/lib/python3.7/site-packages/Bio/Nexus/Trees.py", line 732, in to_string
    subtrees = [newickize(sn, ladderize=ladderize) for sn in succnodes]
File "/home/andla/.local/lib/python3.7/site-packages/Bio/Nexus/Trees.py", line 732, in <listcomp>
    subtrees = [newickize(sn, ladderize=ladderize) for sn in succnodes]
File "/home/andla/.local/lib/python3.7/site-packages/Bio/Nexus/Trees.py", line 714, in newickize
    succnodes = ladderize_nodes(self.node(node).succ, ladderize=ladderize)
File "/home/andla/.local/lib/python3.7/site-packages/Bio/Nexus/Trees.py", line 700, in ladderize_nodes
    succnodes = zip(*succnode_terminals)[1]
TypeError: 'zip' object is not subscriptable
```

zip(...)[1] is not valid in Python 3. 

### Steps to reproduce

```python
from Bio.Nexus import Trees

t = Trees.Tree(
    "(((B 9:0.385832, (C 8:0.445135, C 4:0.41401)C:0.024032)B:0.041436,"
    "A 6:0.392496)A:0.0291131, t2:0.497673, ((E 0:0.301171,"
    "E 7:0.482152)E:0.0268148, ((G 5:0.0984167,G 3:0.488578)G:0.0349662,"
    "F 1:0.130208)F:0.0318288)D:0.0273876);"
)
print(t.to_string(ladderize="LEFT"))
```

peterjc:
That's Python 2 to 3 oversight. Try changing that line to:

```python
succnodes = list(zip(*succnode_terminals))[1]
```

If that works, maybe you could submit it as a pull request along with turning your example into a test case?

anders-was-here:
[...]

PR #3781, opened 2021-11-06
Fix ladderize (#3780)

anders-was-here:

Closes #...

### DIFF
--- a/Bio/Nexus/Trees.py
+++ b/Bio/Nexus/Trees.py
@@ -697,7 +697,7 @@ class Tree(Nodes.Chain):
                 if ladderize == "right" or ladderize == "RIGHT":
                     succnode_terminals.reverse()
                 if succnode_terminals:
-                    succnodes = zip(*succnode_terminals)[1]
+                    succnodes = list(zip(*succnode_terminals))[1]
                 else:
                     succnodes = []
             else:
--- a/Tests/test_Nexus.py
+++ b/Tests/test_Nexus.py
@@ -1035,6 +1035,16 @@ Root:  16
             ],
         )
 
+    def test_to_string(self):
+        t = Trees.Tree(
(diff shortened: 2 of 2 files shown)


anders-was-here:
Not sure why black failed on Tests/test_Nexus.py - too long row maybe?

erik-whiting:
hey @anders-was-here I'm not a maintainer but the pre-commit black fail had me curious and I think I figured it out. It doesn't like line 1046. I'm not exactly sure why, but I played around with it locally and this got rid of the pre-commit failure:
```python
tree_a_val = "((A 6,(B 9,(C 8,C 4))),t2,((E 0,E 7),(F 1,(G 5,G 3))));"
self.assertEqual(t.to_string(ladderize="LEFT"), "tree a_tree = " + tree_a_val)
```
Also, I think you should give that test a doc string (a like after the `def` line that's like `"""This tests some thing"""`)

anders-was-here:
> hey @anders-was-here I'm not a maintainer but the pre-commit black fail had me curious and I think I figured it out. It doesn't like line 1046. I'm not exactly sure why, but I played around with it locally and this got rid of the pre-commit failure:
> 
> ```python
> tree_a_val = "((A 6,(B 9,(C 8,C 4))),t2,((E 0,E 7),(F 1,(G 5,G 3))));"
> self.assertEqual(t.to_string(ladderize="LEFT"), "tree a_tree = " + tree_a_val)
> ```
> 
> Also, I think you should give that test a doc string (a line after the `def` line that's like `"""This tests some thing"""`)

After reading up on pre-commit and black my conclusion is that this was a false positive from the configured black version (black==19.10b0).

The code seems to be OK with the latest black (21.10b0).

peterjc:
Currently pre-commit is configured to use black 21.07b, so that might also have an issue with this line.

anders-was-here:
#3781

peterjc:
Thank you

Issue #3752, opened 2021-09-23
Apparently random fluctuations in coverage via CodeCov?


peterjc:
I would need to go back over some (ideally merged) pull requests to check, but my strong impression compared to when we ran TravisCI/CircleCI/AppVeyor to now with GitHub Actions/CircleCI/AppVeyor, the unit test coverage reported via CodeCov is much more random.

So, it that really the case?

If so, how can we fix it? One hypothesis is that CodeCov is reporting before all the CI tests have finished? e.g. If it only waits for at least one GitHub Actions report, rather than all the GitHub Actions reports?

seanrjohnson:
I think this happened with my pull request https://github.com/biopython/biopython/pull/3733 also. There were two CI errors, but neither of them seemed to be at all related to the code I wrote. One of them was this CodeCov error.

erik-whiting:
@peterjc can I take a run at this one? If so, can you tell me the significance of the tests ran in CircleCI vs AppVeyor? 

I think I can figure this out on my personal fork but it might involve writing out instructions for a maintainer.

peterjc:
It might be sensible to start looking purely at the coverage data locally on your own computer (without CodeCov or CircleCI or AppVeyor or GitHub Actions). Does re-running the same set of tests show fluctuating coverage? If so, can you narrow it down to specific test files or even test methods? Anything using random data is a candidate.

For online testing, you should be able to setup your own CodeCov account for a forked repository, and try pushing data there. One hypothesis is that uploading the data is failing at random (not that I have checked the logs in any details to rule this out).

Are we simply seeing https://about.codecov.io/blog/codecov-uploader-deprecation-plan/ which has a solution?

erik-whiting:
@peterjc I played around with codecov on a personal project last night and I have a few hypotheses. My first thought is that there are two `coverage.xml` files being generated--one by appveyor, the other by circle--and only one is shows in the PR comment. This is just a guess at this point though and should not take much time to confirm/deny.

I'll take your advice on running `coverage` locally and seeing if the coverage fluctuates. I'll also read chapter 21 of the Biopython tutorial but if there's any high-level concepts about the tests in CircleCI vs tests in AppVeyor you can tell me, I'm sure it'd be helpful.

peterjc:
Some of the individual tests do use random data (graph plotting comes to mind - as that one very rarely fails as a result of 'bad' data), so I expect a little noise. Those examples in the tests could be replaced with fixed data.

[...]

PR #3794, opened 2021-11-16
Updating codecov 2nd attempt

erik-whiting:

Addresses #3752

In this PR, I upgrade the codecov orb, and use _**one**_ test run to generate the coverage report. This seems to have fixed the allegedly random variations in coverage. I tested this by creating several PRs that randomly removed a file in the `Tests` directory, and then seeing if identical PRs would result in the same coverage report; they do.

There is one concern. I cannot tell what the difference between CirlceCI, GHA, and AppVeyor tests are. They are all run with the `--offline` flag, so the same code paths should be used in each. There is nothing to indicate that the tests run in AppVeyor implement different code paths than those in GitHub Actions or CircleCI for example. Perhaps the AppVeyor tests RDBMSs that CircleCI doesn't, but this is not something codecov would be able to detect or report on.

Please let me know if you have any questions, notes, criticisms, requests, or anything else.

This is a 2nd PR because I think codecov runs only once per PR
EDIT: That didn't work because Circle didn't rerun since it's already run its workload for this commit.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,17 +36,12 @@ build_script:
   - SET PATH=C:\Py;C:\Py\Scripts;C:\Py\Library\bin;%PATH%
   - conda config --set always_yes yes
   - conda update conda
-  - conda install coverage setuptools numpy
+  - conda install setuptools numpy
   - python setup.py build
 
 test_script:
   - python -c "import sys; print(sys.version)"
   - cd Tests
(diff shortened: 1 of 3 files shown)


peterjc:
The intention was all our CI platforms ran coverage, which would largely overlap with differences due to OS or dependencies.

I don't think this PR was going to work as it removed the call to run the tests via coverage. Closing.

PR #3819, opened 2021-12-05
Adding ability to calculate the isoelectric point by 18 methods.

lukasz-kozlowski:

Closes #...  This is my first contribution to Biopython, thus any feedback/help with fixing errors is appreciated.


Tests/test_ProtParam.py
```diff
@@ -139,11 +139,11 @@ def test_flexibility(self):
     def test_isoelectric_point(self):
         """Calculate the isoelectric point."""
         # Old test used a number rounded to two digits, so use the same
-        self.assertAlmostEqual(self.analysis.isoelectric_point(), 7.72, 2)
+        self.assertAlmostEqual(self.analysis.isoelectric_point(), 6.96, 2)
```
MarkusPiotrowski, line 142: Does this mean that the default behavior changed? This would be bad for backward compatibility.
Was there an error in the calculation before or do you set a different method as default?

Bio/SeqUtils/IsoelectricPoint.py
```diff
@@ -49,6 +142,11 @@ class IsoelectricPoint:
                  occurrences as integers, e.g. ``{"A": 3, "C": 0, ...}``.
                  Default: ``None``. If ``None``, the dic will be calculated
                  from the given sequence.
+    :pKa_scale: Different pKa scales. Need to be one of: Bjellqvist, DTASelect,
+            Dawson, EMBOSS, Grimsley, IPC2_peptide, IPC2_protein,
+            IPC_peptide, IPC_protein, Lehninger, Nozaki, Patrickios, ProMoST,
+            Rodwell, Sillero, Solomon, Thurlkill, Toseland, Wikipedia.
+            Default scale is 'IPC2_protein' (we assume the protein as an input)
```
MarkusPiotrowski, line 149: Although I can understand that you want your own method to be the default, I guess this is problematic for backward compatibility.

Bio/SeqUtils/IsoelectricPoint.py
```diff
@@ -143,15 +259,13 @@ def pi(self, pH=7.775, min_=4.05, max_=12):
...
-           which is above the theoretical maximum, when the protein is composed
-           exclusively of arginine.
+         - min\_: the minimum of the interval. Initial value set to 1.0 which is
+           below the theoretical maximum.
+         - max\_: the maximum of the the interval. Initial value defaults to 14,
+           which is above the theoretical maximum.
         """
         charge = self.charge_at_pH(pH)
-        if max_ - min_ > 0.0001:
+        if max_ - min_ > 0.001:
```
MarkusPiotrowski, line 268: Why did you change this?

Bio/SeqUtils/IsoelectricPoint.py
lukasz-kozlowski: It is not a matter of using my method as the default (if you can find/build a better model, I do not mind having it as the default, it is a normal thing in bioinformatics that after a few years some methods became obsolete). Currently, IPC2_protein/IPC2_peptide happens to be the most accurate*. I worked on isoelectric point subject for many years and developed multiple models. Biopython uses for pI calculation Henderson-Hasselbach equation with bisection as the main algorithm. Moreover, you need to set some pKa values for charged amino acids (Biopython uses Bjellqvist pKa set). Bjellqvist set is just one of the possible pKa sets you can use. I guess people are using it as it was one of the first introduced and it is widely implemented e.g. emboss. Yet, it is not the most accurate pKa set for proteins. I benchmarked this very extensively over the years, and even build new pKa sets that work better for proteins and peptides (for instance see http://ipc2.mimuw.edu.pl/results.html). For more details see: http://dx.doi.org/10.1186/s13062-016-0159-9 and http://dx.doi.org/10.1093/nar/gkab295

With the proposed commit, I want to give the users of biopython the opportunity to:
1) choose from 18 different pKa sets (not just Bjellqvist)
2) use as the default the most accurate* (currently it is IPC2_protein for proteins, but obviously as soon as a better isoelectric point prediction model will be available we should switch the default) 
 
* As mentioned above biopython uses a very simple approach (Henderson-Hasselbach equation with bisection), but more complicated and more accurate (and at the same time slower) algorithms exist. In the first table at http://ipc2.mimuw.edu.pl/results.html you can see models named IPC.protein.svr.19 or IPC.peptide.Conv2D that are even better than IPC2_protein, but they depend on support vector regression (sklearn) and deep learning (keras& TensorFlow). Thus my understanding is that biopython's developers want to stick to a simpler solution (Henderson-Hasselbach equation with bisection as it is pure python without any 3rd party involved), even if this compromises the accuracy a bit, rather than having dependencies like TensorFlow.

Tests/test_ProtParam.py
lukasz-kozlowski: Yes, the default method changed (previously Bjellqvist pKa sets inside of the algorithm have been used, now it is IPC2_protein set). The reason for the change is that IPC2_protein gives better results (http://ipc2.mimuw.edu.pl/results.html) in comparison to Bjellqvist set introduced in 1994.

mmokrejs:
Hi @lukasz-kozlowski and @MarkusPiotrowski,

I hope you don't mind — I took the liberty of rebasing and extending this work in PR #5200, with the help of AI (Claude Opus 4.6). Your original implementation was solid and the pKa tables are exactly preserved; I mainly added the missing test coverage and addressed the backward compatibility concern that @MarkusPiotrowski raised.

**Key changes vs this PR:**
- Default pKa scale remains **Bjellqvist** (no breaking changes to existing tests)
- Invalid scale names now **raise `ValueError`** instead of silently falling back
- Bisection tolerance unchanged at `0.0001` (search range widened to [1.0, 14.0])
- **56 new tests** across 9 test classes (targeting >95% coverage)
- Cross-validated against the standalone IPC2 tool (v2.0.1) — max difference 0.007

@lukasz-kozlowski — please don't feel offended that I picked this up. Your PR had been waiting for 4+ years and the feature is genuinely useful. I wanted to help get it across the finish line. If you'd prefer any adjustments or have concerns, please comment on #5200.

*Note: This comment was drafted with Claude Opus 4.6 assistance.*

Bio/SeqUtils/IsoelectricPoint.py
lukasz-kozlowski: agree, this was done only to be compatible with the standalone version

PR #3858, opened 2022-02-09
Supportintg LINK and SSBOND lines from PDB files

markf94:

This MR adds support for the commonly occuring LINK and SSBOND lines in PDB files.
    
So far, Biopython did not support LINK and SSBOND lines from PDB files despite them being very common as well as very important. For example, every cyclic peptide (e.g. [5XOA](https://www.rcsb.org/structure/5XO4)) relies on SSBOND lines to be fully described. In order to address this issue, this MR adds support for parsing from and writing LINK and SSBOND to PDB files. As a result, the `Structure` class now has two new methods: `get_links()` and `get_disulfide_bonds()` which allow users to retrieve the LINK and SSBOND information.

I'm anticipating that the details of the implementation are subject to change since I don't know you feel about the use `NamedTuple` in this context. I'm happy to change things around but overall I think this functionality would be a very useful addition to the Biopython library.


Bio/PDB/PDBIO.py
```diff
@@ -294,6 +286,92 @@ def _get_atom_line(
 
             return _PQR_ATOM_FORMAT_STRING % args
 
+    @staticmethod
```
JoaoRodrigues, line 289: Any particular reason to make this a `@staticmethod`?

Bio/PDB/Structure.py
```diff
@@ -7,9 +7,40 @@
...
 """The structure class, representing a macromolecular structure."""
 
+from typing import NamedTuple
+
+from Bio.PDB.Atom import Atom
+from Bio.PDB.Chain import Chain
 from Bio.PDB.Entity import Entity
 
 
+class DisulfideBond(NamedTuple):
```
JoaoRodrigues, line 17: This is a funny way of creating a namedtuple. I would rather avoid importing `typing` only to use the `NamedTuple` as a subclass. If you prefer this syntax, why not use a frozen dataclass? Otherwise, I believe the "standard" way of declaring namedtuple is simpler.

Bio/PDB/Structure.py
```diff
@@ -63,3 +94,85 @@ def internal_to_atom_coordinates(self, verbose: bool = False) -> None:
...
+    def _find_residue_by_id(self, chain: Chain, id_: int):
+        """
+        Return the residue with `id_` from the `chain`.
+
+        This is a workaround required due to Biopython's residue indexing e.g.
+        ('H_GLC', 100, 'A') instead of 100. For more info please refer to the
+        section "What is a residue id?" at
+        https://biopython.org/wiki/The_Biopython_Structural_Bioinformatics_FAQ
+        """
+        for residue in chain.get_residues():
```
JoaoRodrigues, line 107: This is going to be quite expensive.. but I assume we don't have a better solution unless we somehow cache things.

Bio/PDB/parse_pdb_header.py
```diff
@@ -15,8 +15,57 @@
 
 
 import re
+from typing import NamedTuple
```
JoaoRodrigues, line 18: Same comment as above.

Bio/PDB/Structure.py
```diff
@@ -63,3 +94,85 @@ def internal_to_atom_coordinates(self, verbose: bool = False) -> None:
...
+        Return intramolecular links of the Structure.
+
+        Each link consists of two Atom objects and their associated distance.
+        """
+        raw_links = self.header["links"]
+        links = []
+        for link in raw_links:
+            for model in self.get_models():
+                chain1 = model[link.res1.chain]
+                residue1 = self._find_residue_by_id(chain1, link.res1.id)
```
JoaoRodrigues, line 122: You might want to have a guard here in case the structure is edited and the residue/chain/atom no longer exists.

Bio/PDB/parse_pdb_header.py
```diff
@@ -318,6 +369,91 @@ def _parse_pdb_header_list(header):
...
+                        name=resname2, id=resid2, atom=atom2, chain=chainid2
+                    ),
+                    insertion_code1=insertion_code1,
+                    insertion_code2=insertion_code2,
+                    symmetry_operator1=symmetry_operator1,
+                    symmetry_operator2=symmetry_operator2,
+                    distance=distance,
+                )
+            )
         else:
```
JoaoRodrigues, line 457: This is quite a superfluous else, I think we can remove it!

JoaoRodrigues (changes requested):

Bio/PDB/PDBIO.py
peterjc: There are potentially quite a few examples in older Biopython code where we could do this - https://docs.python.org/2/library/functions.html#staticmethod says ``staticmethod`` was added in Python 2.2 and as a decorator in Python 2.4

Issue #3840, opened 2022-01-13
Insecure Temporary File using tempfile.mktemp()


peterjc:
This is a security issue spotted by @srikanthprathi reported by @JamieSlome on #3839, https://huntr.dev/bounties/bf1fb12c-7ed5-417a-912f-934006bcfcef/

Having review the private report, I considered it low risk, and with their agreement am posting it publicly:

> **Insecure Temporary File in biopython/biopython**
> biopython package is using the deprecated function tempfile.mktemp() which is not secure. Because a different process may create a file with this name in the time between the call to mktemp() and the subsequent attempt to create the file by the first process.
>
> Impact
> Availability will get affected because of this vulnerability.
>
> Remediation
> Use mkstemp() instead of tempfile.mktemp()

We appear to have two uses in the test suite, both using the same code style:

https://github.com/biopython/biopython/blob/master/Tests/test_Phylo_CDAO.py
https://github.com/biopython/biopython/blob/master/Tests/test_PhyloXML.py

And one use in the main code base in the PDB module for invoking the MSMS command line tool:

https://github.com/biopython/biopython/blob/master/Bio/PDB/ResidueDepth.py

PR #3870, opened 2022-02-22
Add remove_tmp option to get_surface() to reduce storage requirements

amorehead:

This change to the `get_surface` function in `ResidueDepth.py` allows callers of the function to decide whether the temporary files created for surface generation should be deleted after the requested surface has been generated and retrieved as a NumPy array. By default, the value for `remove_tmp` is set to `False` to maintain the original functionality of this function (for backwards compatibility).


peterjc:
Perhaps just delete them on success? That will happen eventually on Linux/macOS, not so sure what Windows does.

Given the temp files are put on the system temp folder in a randomly named folder, the user is almost certainly never going to look at them - even if the call fails, it would be hard.

MarkusPiotrowski:
I'm just curious, didn't we have an issue about security of `mktemp()` ? Given that `mktemp` is deprecated since Python version 2.3 (https://docs.python.org/3/library/tempfile.html) this would be an opportunity to switch to `mkstemp()`.

peterjc:
Yes, #3840 - that occurred to me too during the day. I agree, nice to fix both issues together :)

amorehead:
@peterjc, for context, the reason I am looking to add an explicit function argument to delete these temporary files is because, on my local HPC cluster, the `/tmp/`directory BioPython is writing to with these `mktemp` files is running out of space when I generate 1000s of surfaces in sequence.

peterjc:
I understand why you'd want this - I am suggesting it shouldn't be an option, it should be a behaviour change:

* Currently it does not delete the temp files (left for OS to do at some point)
* My suggestion is delete the temp files on success (leave them on failure for potential diagnosis; perhaps report the path in our error message?)
* Alternative suggestion is to always delete the temp files

peterjc:
Just modify your pull request to remove the optional argument?

(This doesn't address ``mktemp`` but you'd be welcome to work on that too, as a separate follow up pull request if you prefer)

JoaoRodrigues:
👍 by me, I can see this being useful for high-throughput usage in HPCs with disk quotas. I agree with @peterjc that the default should be deleting the temp file unless there is an error.

Bio/PDB/ResidueDepth.py
```diff
@@ -528,19 +530,45 @@ def get_surface(model, MSMS="msms"):
...
+    msms_tmp = tempfile.mktemp()
+    MSMS = MSMS + " -probe_radius 1.5 -if %s -of %s > " + msms_tmp
     make_surface = MSMS % (xyz_tmp, surface_tmp)
     subprocess.call(make_surface, shell=True)
+    face_file = surface_tmp + ".face"
     surface_file = surface_tmp + ".vert"
     if not os.path.isfile(surface_file):
         raise RuntimeError(
-            f"Failed to generate surface file using command:\n{make_surface}"
+            "Failed to generate surface file using command:\n%s" % make_surface
```
JoaoRodrigues, line 543: Any particular reason to drop the f-string formatting here?

Bio/PDB/ResidueDepth.py
```diff
@@ -528,19 +530,45 @@ def get_surface(model, MSMS="msms"):
...
         )
 
-    # read surface vertices from vertex file
+    # Read surface vertices from vertex file
     surface = _read_vertex_array(surface_file)
+
+    # Remove temporary files
+    if remove_tmp:
```
JoaoRodrigues, line 550: I'd put these files in a list and iterate, to avoid the deep chain of try/except statements:

```
for fn in [xyz_tmp, surface_tmp, ...]:
    try:
        os.remove(fn)
    except OSError:
        pass

**Review by JoaoRodrigues — CHANGES_REQUESTED** [review]

File: Bio/PDB/ResidueDepth.py (hunk 1 above)
Comment (peterjc) [reply Bio/PDB/ResidueDepth.py:543]: Good catch, yes - please put that back.

File: Bio/PDB/ResidueDepth.py (hunk 2 above)
Comment (amorehead) [reply Bio/PDB/ResidueDepth.py]: @JoaoRodrigues @peterjc 
I believe my [latest commit](https://github.com/biopython/biopython/pull/3870/commits/53317680aaf6bdce5818f3793cf8a43e353e1dcb) addresses both of your concerns fully. If not, let me know, and I can make any remaining changes necessary.

File: Bio/PDB/ResidueDepth.py [hunk 3]
```diff
@@ -50,21 +50,18 @@
 
 """
 
-
+import numpy
 import os
+import subprocess
```
peterjc, line 55: Nice to sort these :)

Bio/PDB/ResidueDepth.py
```diff
@@ -297,15 +295,15 @@ def _get_atom_radius(atom, rtype="united"):
...
+            and at_name.endswith(("1", "2", "3", "4", "5", "6", "7"))
     ):
         return _atomic_radii[21][typekey]
     elif (
-        resname in {"FS3", "FS4"}
-        and at_name.startswith("S")
-        and at_name.endswith(("1", "2", "3", "4", "5", "6", "7"))
+            resname in {"FS3", "FS4"}
+            and at_name.startswith("S")
+            and at_name.endswith(("1", "2", "3", "4", "5", "6", "7"))
```
peterjc, line 306: I suspect these indentation change and the one above will break black formatting rules?

Bio/PDB/ResidueDepth.py
amorehead: This should now be fixed in [this commit](https://github.com/amorehead/biopython/commit/b22226e5225866cc8a87b18d21da5cea5987e3d8).

Bio/PDB/ResidueDepth.py
amorehead: This should now be fixed in [this commit](https://github.com/amorehead/biopython/commit/b22226e5225866cc8a87b18d21da5cea5987e3d8).

peterjc: That fixed the blank line change, but not this indentation.

peterjc:
Can you install and run the tool black? See https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

More generally, in a pull request it is best not to edit unrelated code. Here I wonder if it was actually your text editor doing something automatically? That has caught me out before.

peterjc (approved):
Looks good to me, but we should squash this into one commit when merging it.

amorehead:
@peterjc,

You are absolutely right. It looks like my IDE (PyCharm) automatically formatted this file when I opened it (without me knowing it). I have just [squashed](https://github.com/biopython/biopython/commit/72d18a522633be0c4211accbbe1ce4e46da409aa) my 8 commits into one and pushed the changes to my current `patch-3` branch.

JoaoRodrigues:
Thanks @amorehead !

amorehead:
@peterjc @JoaoRodrigues My pleasure!

PR #3877, opened 2022-03-12
Use mkstemp instead of mktemp.

### LINKED ISSUE #3840: Insecure Temporary File using tempfile.mktemp()

JoaoRodrigues:

Closes #3840

### DIFF
--- a/Bio/PDB/ResidueDepth.py
+++ b/Bio/PDB/ResidueDepth.py
@@ -521,7 +521,7 @@ def get_surface(model, MSMS="msms"):
     # Make x,y,z,radius file
     atom_list = Selection.unfold_entities(model, "A")
 
-    xyz_tmp = tempfile.mktemp()
+    xyz_tmp = tempfile.mkstemp()
     with open(xyz_tmp, "w") as pdb_to_xyzr:
         for atom in atom_list:
             x, y, z = atom.coord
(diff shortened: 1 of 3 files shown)


peterjc:
Wouldn't this be shorter:

```
DUMMY = tempfile.NamedTemporaryFile(delete=False).name
```

Then you don't need all the ``DUMMY`` to ``DUMMY.name`` changes.

JoaoRodrigues:
> Ah, you mean there is an implicit deletion of the file once the `NamedTemporaryFile` goes out of scope?

No. Sorry, probably I wasn't super clear. I used `(...).name` pattern in one of the files I changed because we are looping over file names (both temporary and others) at some point in the code. For the other cases, I preferred sticking with using `DUMMY.name` where we needed them. But I can change it all to use that pattern, no problem!

peterjc (approved):
I don't think I understood your original intent with ``DUMMY.name``, but as it stands is a minimal diff which looks tidy.

PR #3882, opened 2022-03-17
Raise exception when writing structure that exceeds PDB format limits.

### LINKED ISSUE #1085: Prevent PDBIO from writing large structures not supported in PDB format
The PDB format specification imposes several limits on the number of atoms, residues, and chains that can be written to file. This is one of the reasons behind the move to the mmCIF format as the default structural data storage format. 

Since we have the ability to read both PDB and mmCIF structures, but only to write PDB files, there should be a check to prevent writing very large structures that break the format specifications: > 99.999 atoms, >9999 residues, >1 character chains, etc.

JoaoRodrigues:

Closes #1085

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -169,10 +169,6 @@ class PDBIO(StructureIO):
         else:
             record_type = "ATOM  "
 
-        # Ensure chain id isn't longer than 1 character
-        if len(chain_id) > 1:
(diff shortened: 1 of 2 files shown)


peterjc (approved):
Looks good to me - I expect there is good reason to move the chain ID ValueError check?

JoaoRodrigues:
Performance: no need to do the check at every atom - once per chain will do.

Thanks!

PR #3889, opened 2022-03-25
Add support for ignoring file extensions in codespell

### LINKED ISSUE #3888: codespell blocking documentation commit on attribute name
I have an attribute `.aks` in code that I'm documenting, but codespell blocks the commit in the tutorial because it thinks I mean 'ask'.  
There doesn't seem to be any in-file codespell directive to ignore the lines, is there some other suggested workaround?

erik-whiting:

Closes #3888 . I added a second regex flag for ignoring file extensions in codespell. Also left a comment to guide future devs if they need to ignore more file extensions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,7 +56,8 @@ repos:
         files: \.(rst|md|tex)$
         args: [
             --ignore-regex,
-            '(^|\W)([A-Z]{2,3})(\W|$)',
+            '(^|\W)([A-Z]{2,3})(\W|$)|(.aks)',
+            # last grouping above is for filenames. Add more with a pipe, e.g., (.aks|.txt|.asm)
             -L,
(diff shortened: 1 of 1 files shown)


erik-whiting:
To any reviewers: I think the codecov failure is a false-positive as I didn't update any files that can be tested

peterjc:
Yes, see #3752 - if no one can work out why that happens we may want to turn off the CodeCov PR check (it is already not required).

peterjc:
The usage in #3888 was not about filenames ending ``.aks`` but Python code referring to an attribute which would be accessed with ``.aks`` instead.

The regex is already complicated and I am unwilling to make it more complicated. While you can ignore words with the codespell regex setting (``--ignore-regex``), I think using ``-L`` would be far simpler here.

Also note ``.`` has a special meaning in a regex.

PR #3897, opened 2022-04-06
Fix broken reference in selected_child when copying a DisorderedAtom

### LINKED ISSUE #3896: Broken reference in selected_child when copying a DisorderedAtom
### Setup

system as follows:

```log
CPython (Python 3.10.4)
Linux-5.10.0-13-amd64-x86_64-with-glibc2.31
Biopython Version 1.80.dev0
```

Biopython version >=1.78 affected, due to a new way of handling copying of DisorderedAtoms.

### Expected behaviour

When copying a DisorderedAtom, is to be expected that it's reference to it's "best" child is one of it's own children. Represented in code as `DisorderedAtom.selected_child`

### Actual behaviour

When copying a DisorderedAtom, `DisorderedAtom.selected_child` and `DisorderedAtom.last_occupancy` does not get reset, and thus keeps pointing to an Atom on the original entry. When re-adding the Atoms, it also does not overwrite the `selected_child`, because `last_occupancy` is already set to the highest occupancy.

### Steps to reproduce

Minimal testcase:
```python
import os
import Bio.PDB


def get_pdb():
    # Download PDB if not present
    if (not os.path.exists('./pdb6nb8.ent')):
        pdbRetriever = Bio.PDB.PDBList()
        pdbRetriever.retrieve_pdb_file('6nb8', file_format='pdb', pdir='./')

    return './pdb6nb8.ent'


if __name__ == "__main__":
    pdb_path = get_pdb()
    parser = Bio.PDB.PDBParser()
    structure = parser.get_structure("structureID", pdb_path)
    # Get the disordered atoms, in chain L
    light_chain = structure[0].child_dict['L']
    atoms = [atom for atom in light_chain.get_atoms() if atom.is_disordered()]

    if len(atoms) == 0:
        exit('No disordered atoms found!')
    # Get the first disordered atom, and check it's selected_child
    atom = atoms[0]
    atom_copy = atom.copy()
    best_atom_of_copy = max(atom_copy.disordered_get_list(), key=lambda pos_atom: pos_atom.occupancy)
    assert best_atom_of_copy is atom_copy.selected_child
```

### Fix

I have already implemented a fix and will open a pull request mentioning this issue in a few minutes.

Bartvelp:

Closes #3896

Due to the agnostic nature of `DisorderedEntityWrapper.copy()`, it does not reset the `selected_child` and `last_occupancy` on a DisorderedAtom. This results in a broken reference in DisorderedEntityWrapper in the `selected_child`, because it never get's overwritten with one of the copied children Atoms.
This fix implement a new `copy()` method, specifically for `DisorderedAtom`s, which _does_ reset the relevant properties.

### DIFF
--- a/Bio/PDB/Atom.py
+++ b/Bio/PDB/Atom.py
@@ -564,3 +564,21 @@ class DisorderedAtom(DisorderedEntityWrapper):
         """
         for child in self:
             child.coord = np.dot(child.coord, rot) + tran
+
+    def copy(self):
+        """Copy disorderd atom recursively.
+
+        Loses parent relationship, and sets selected_child to own of it's own
+        children.
+        """
+        shallow = copy(self)
+        shallow.child_dict = {}
+        shallow.detach_parent()
+
+        shallow.selected_child = None
+        shallow.last_occupancy = 0
+
+        for child in self.disordered_get_list():
+            shallow.disordered_add(child.copy())
+
+        return shallow
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -44,6 +44,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - Aziz Khan <https://github.com/asntech>
 - Barbara Mühlemann <https://github.com/bamueh>
 - Bart de Koning <bratdaking gmail>
+- Bart Grosman <https://github.com/Bartvelp>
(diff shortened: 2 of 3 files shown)


Bartvelp:
By the way, I also fixed a typo in `test_copy_disordered_atom(self):` because it got the "copied" atoms from the original residue, instead of the copied one.

Issue #3885, opened 2022-03-22
SnapGene files with line breaks in feature qualifiers result in corrupted genbank files


biostu24:
### Setup
python: 3.10.0 (default, Nov 10 2021, 11:24:47) [Clang 12.0.0 ]
implementation: CPython
platform: macOS-10.16-x86_64-i386-64bit
BioPython version: 1.78 (bug also appears in 1.79)

### Expected behaviour
In rare cases, line break characters can appear in SnapGene features. In the binary SnapGene format, this appears as "\&#10;". During parsing by BioPython (to SeqRecord), these characters result in line breaks in the strings. Saving the SeqRecord as genbank should gracefully deal with these line break characters.

### Actual behaviour
Reading a SnapGene file containing line breaks returns a SeqRecord. Saving this SeqRecord in Genbank format results in a corrupted genbank formatted file. For example:
```
FEATURES             Location/Qualifiers
     source          1..9692
                     /label="source"
     promoter        1..229
                     /label="RSV  promoter"
                     /note="/application_notes=Strong promoter; drives
                     transcription
of viral RNA in packaging cells."
```
Notice how the final line of text is not correctly indented. This incorrect indenting will cause all genbank parsers to fail to read the file. 

The source of this issue is due line break characters in the source SnapGene file ("\&#10;"). This results in qualifier value strings with line breaks. Biopython is not dealing with these line breaks correctly

After saving a SeqRecord with line breaks as a Genbank file, then reloading using Seq.IO.read, the following error is produced:
```
Traceback (most recent call last):
  File "/Users/my_user_name/Desktop/plasmid_bug/test_bug.py", line 30, in <module>
    _ = SeqIO.read('bugged_plasmid.gb', 'genbank')
  File "/Users/my_user_name/miniconda3/envs/utility/lib/python3.10/site-packages/Bio/SeqIO/__init__.py", line 656, in read
    record = next(iterator)
  File "/Users/my_user_name/miniconda3/envs/utility/lib/python3.10/site-packages/Bio/SeqIO/Interfaces.py", line 73, in __next__
    return next(self.records)
  File "/Users/my_user_name/miniconda3/envs/utility/lib/python3.10/site-packages/Bio/GenBank/Scanner.py", line 516, in parse_records
    record = self.parse(handle, do_features)
  File "/Users/my_user_name/miniconda3/envs/utility/lib/python3.10/site-packages/Bio/GenBank/Scanner.py", line 499, in parse
[...]

PR #3907, opened 2022-04-22
Handle stray line breaks in GenBank Scanner

erik-whiting:

Closes (part of) #3885

This branch handles the possibility of an errant line break in a .gdb file. For example, I have been using a file like this for testing:
```
LOCUS       .                          4 bp    DNA              UNK 01-JAN-1980
DEFINITION  .
ACCESSION   <unknown id>
VERSION     <unknown id>
KEYWORDS    .
SOURCE      .
  ORGANISM  .
            .
     misc_feature    2
                     /label="buggy
feature"
ORIGIN
        1 aaaa
//
```

The general approach is this as follows
In the `parse_features` method of `GenBank::Scanner`:
* create an array (named `lines` of the next two lines in a file. 
* If the second line is alphanumeric, it is probably because it's supposed to be part of the line above it. In this case, join the two lines with a space and warn the user
* If the second line is normal, the file pointer has already reached it, so `readline()` would be incorrect. In this case, we pop the "good" line one out of the `lines` array. In the next iteration, if the `lines` array has only one item in it, the parser knows to use that as `line` instead of calling `readline`.

I'm not sure if this is correct so please let me know if I missed something important about how the lines are laid out.

### DIFF
--- a/Bio/GenBank/Scanner.py
+++ b/Bio/GenBank/Scanner.py
@@ -163,6 +163,7 @@ class InsdcScanner:
             self.line = self.handle.readline()
 
         features = []
+        lines = []  # For errant line break handling later
         line = self.line
         while True:
             if not line:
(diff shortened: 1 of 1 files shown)


peterjc:
Given an example like the one you showed, I would think something here would be a simpler fix?

```python
            if line[2 : self.FEATURE_QUALIFIER_INDENT].strip() == "":
                # This is an empty feature line between qualifiers. Empty
                # feature lines within qualifiers are handled below (ignored).
                line = self.handle.readline()
                continue
            if len(line) < self.FEATURE_QUALIFIER_INDENT:
                warnings.warn(
                    f"line too short to contain a feature: {line!r}",
                    BiopythonParserWarning,
                )
                line = self.handle.readline()
                continue
```

However, see #3885 discussion.

PR #3912, opened 2022-04-25
Remove sys.path hack in run_tests to avoid issues with compiled code

### LINKED ISSUE #3910: CI fails because of import errors with C extensions

JoaoRodrigues:

If the CI passes (this is the only way to test), this patch closes #3910.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,8 +36,8 @@ build_script:
   - SET PATH=C:\Py;C:\Py\Scripts;C:\Py\Library\bin;%PATH%
   - conda config --set always_yes yes
   - conda update conda
-  - conda install coverage setuptools numpy
-  - python setup.py build
+  - python -m pip install --upgrade coverage setuptools numpy
+  - python -m pip install .
 
 test_script:
   - python -c "import sys; print(sys.version)"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,17 +72,20 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    
+
     - uses: actions/cache@v2
       id: cache-pip
       with:
         path: ${{ env.pythonLocation }}
         key: ${{runner.os}}-${{ env.pythonLocation }}-${{ matrix.python-version }}-v2-${{ hashFiles('setup.py') }}-${{ hashFiles('**/ci-dependencies.txt') }}
 
+    - name: Install Python packaging tools
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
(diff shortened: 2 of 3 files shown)


JoaoRodrigues:
I don't understand why appveyor would fail, since all windows tests pass on GHA.. However, is the purpose of the Appveyor workflow to 1) test BioSQL and 2) upload the coverage results to codecov? These could be moved to GHA as well.

peterjc:
That makes sense. So we can no longer do this on a clean machine:

```
python setup.py build
python setup.py test
python setup.py install
```

Being able to test before install is helpful, but not essential given how most people now install pre-compiled wheels or other packages.

PR #3931, opened 2022-05-14
Replaced asserts in CelFile.py

valentin994:

already or do not wish to be listed. (*This acknowledgement is optional.*)

The checks are quite repetitive and could be put in a function but I wasn't sure if it was okay for me to put another function in the file as the scope of the tasks seems to only change asserts to if statements. I'm fine with either approach. 

As for the tests I'm not sure if it's necessary to add some new tests for the added exceptions?

Related #3600

ps. there is a typo on dual ~~licence~~ license, and ~~acknowledgement~~ acknowledgment in the agreement text

### DIFF
--- a/Bio/Affy/CelFile.py
+++ b/Bio/Affy/CelFile.py
@@ -336,32 +336,45 @@ def _read_v3(handle):
                     record.DatHeader["filename"] = filename
                     index += 1
                     field = line[index : index + 9]
-                    assert field[:4] == "CLS="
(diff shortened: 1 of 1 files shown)


peterjc (approved):
The position numbers are somewhat arbitrary at this point (since offset by the index variable), so I'm not bothered about Python vs 1 based counting.

peterjc:
Thank you

PR #3937, opened 2022-05-20
Update docs with working testing instructions

npars:

- PR #3912 introduced changes that broke the existing testing instructions
- Update testing instructions to work with latest test scripts
- Test instructions now work with setup.py script as well as through the Intellij test runner

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,7 +161,7 @@ at compile time:
 Then either download and decompress our source code, or fetch it using git.
 Now change directory to the Biopython source code folder and run::
 
-    python setup.py build_ext --inplace
+    pip install -e .
     python setup.py test
     sudo python setup.py install
 
@@ -184,7 +184,7 @@ Biopython includes a suite of regression tests to check if everything is
 running correctly. To run the tests, go to the biopython source code
 directory and type::
 
-    python setup.py build_ext --inplace
+    pip install -e .
     python setup.py test
 
 If you want to skip the online tests (which is recommended when doing repeated
--- a/README.rst
+++ b/README.rst
@@ -161,7 +161,7 @@ at compile time:
 Then either download and decompress our source code, or fetch it using git.
 Now change directory to the Biopython source code folder and run::
 
-    python setup.py build
+    python setup.py build_ext --inplace
     python setup.py test
     sudo python setup.py install
 
@@ -184,7 +184,7 @@ Biopython includes a suite of regression tests to check if everything is
 running correctly. To run the tests, go to the biopython source code
 directory and type::
 
-    python setup.py build
+    python setup.py build_ext --inplace
     python setup.py test
 
 If you want to skip the online tests (which is recommended when doing repeated


peterjc:
Good point, this needs fixing. I recall some pip changes about in-place builds, which may or may not be relevant here...

npars:
@peterjc do you have a link available? I found [this thread](https://discuss.python.org/t/the-difference-between-python-setup-py-build-ext-i-and-pip-install-e/3716/11) where it's mentioned that `build_ext -i` isn't available through pip yet. Not sure if anything has changed since then or if there's another way to get local tests working without needing to install every time.

peterjc:
I was hoping you knew - I'm familiar with ``pip install -e .`` but apparently that maps to ``python setup.py develop``?

npars:
I can give that a shot and see if it allows the tests to resolve the compiled modules correctly

npars:
@peterjc Local tests are working with `pip install -e .` command. It creates the `*.so` libraries in place in the source tree. Updated the instructions accordingly.

peterjc:
Any second opinions about suggesting ``pip install -e .`` prior to running our tests? I've not sat down to review the latest Python packaging documentation on this area...

e.g. @mdehoon or @JoaoRodrigues

npars:
Any update on this @peterjc / @mdehoon  / @JoaoRodrigues ?

peterjc:
Would the sudo install version now be redundant?

JoaoRodrigues:
This look good to me! Completely forgot about the testing instructions, thanks for catching this @npars !

JoaoRodrigues:
>Would the sudo install version now be redundant?
Yes.

peterjc:
Revisiting following #4140.

@etal Doing ``pip install .`` would touch the default Python's files, and if that is the system Python that would need admin rights (i.e. sudo).

peterjc:
This is the current suggestion on the PR:

```
pip install -e .
python setup.py test
sudo python setup.py install
```

@idoerg's suggestion on #4140 is:

```
python setup.py build
sudo python setup.py install
cd Tests
python run_tests.py
```

Note with setups like conda being more common, it less likely nowadays that ``sudo`` would be needed anymore, as you wouldn't need to touch the system Python. Personally for example, I am using:

```
pip install -e .
python setup.py test
```

npars:
My opinion is to just merge this as is since it will get the instructions working. If someone has a better approach than the instructions can be updated again in a follow up PR.

idoerg:
Does pip install from the latest source?

The reason I wanted to install from source in the first place was because
1.79 cannot parse Unipprot-Swissprot anymore (new FT record format). I did
not open an issue on that because 1.80 fixed the problem. I think it's
important to be able to install from source due to lags like that between
releases.

--

peterjc:
Certainly if we had merged this months ago, we'd have saved Iddo some pain.

The current diversity of Python installations (and ``sudo`` or not) probably makes a one-size-fits-all set of install instructions harder than it was 20 years ago.

peterjc:
``pip install .`` (with an optional ``-e`` for developer mode) means install the source code in the current directory ``.``, which in this case is a copy via git of the latest code.

PR #3957, opened 2022-06-16
Fix bug in rotation matrix transpose when using QCP methods

### LINKED ISSUE #3950: CEAlign does not correctly modify atom position
### Setup

I am reporting a problem with Biopython version 1.80dev. 
Full reproduction here: https://colab.research.google.com/drive/1Ybk9qlO920KkFZLJtCLuYpE9sa6ZoCw2?usp=sharing

### Expected behaviour

Same behavior as in Pymol CEAlign. 
image

### Actual behaviour
See notebook. RMS is correct but structures are not correctly superposed. 
image

Their coordinates change so the unit test passes but the result is not correct. 

### Steps to reproduce

Run all steps in this notebook: https://colab.research.google.com/drive/1Ybk9qlO920KkFZLJtCLuYpE9sa6ZoCw2?usp=sharing

JoaoRodrigues:

Closes #3950. In addition, replaces the QCP method in use by CEAligner by the pure python version, as the C version will be deprecated sooner or later.

### DIFF
--- a/Bio/PDB/QCPSuperimposer/__init__.py
+++ b/Bio/PDB/QCPSuperimposer/__init__.py
@@ -103,7 +103,7 @@ class QCPSuperimposer:
             -1.0,
         )
         rot = array([r0, r1, r2, r3, r4, r5, r6, r7, r8]).reshape(3, 3)
-        return (rmsd, rot.T, [q1, q2, q3, q4])
+        return (rmsd, rot, [q1, q2, q3, q4])
 
     # Public methods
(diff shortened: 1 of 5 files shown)


JoaoRodrigues:
@peterjc, pre-commit is failing (but passed locally), do you think it's just a temp glitch or something that may be related with the code changes?

peterjc:
We've seen a lot of time outs in the pre-commit online check recently, ignore it for this PR.

JoaoRodrigues:
Thanks!

PR #3967, opened 2022-06-22
Drop pre-commit.ci badge & configuration

### LINKED ISSUE #3958: pre-commit.ci time outs (just over 3m)
Recently many pull requests and some of the checks on the main branch are failing at pre-commit.ci due to time outs, we appear to now slip over the 3m limit.

I think the first master branch failure was https://github.com/biopython/biopython/commit/e33e229f07bcc4ffdcbb3b1ff21cbc93acd88137 - see https://results.pre-commit.ci/run/github/151541/1655293312.ldyU2MwLRgmi2MlTrhAOsw which got as far as codespell (the penultimate check).

More recent runs seem to only get as far as blacken-docs.

There have been no changes to our configuration https://github.com/biopython/biopython/blob/master/.pre-commit-config.yaml for 2 months, so this is not down to a version change in one of the tools.

Looking at a sample of earlier successful runs, the pre-commit run itself (after setup) was apparently taking about 2m10s - but had started to creep up. Perhaps a code or documentation change has hit a performance bottleneck in one of these tools?

Note we are already skipping flake8 itself on this check.

Note pre-commit.ci runs on all the files, it does not attempt to narrow this down to recently changed files (this functionality has been suggested and rejected).

One possible solution may be paying for a longer time out (although this does not seem to be "on the menu" in the GitHub marketplace options).

peterjc:
Separately we will disable the https://pre-commit.ci/ integration with GitHub.

As per discussion on #3958, this was redundant with the pre-commit tool checks already being run via GitHub Actions (and that was faster thanks to only checking recently changed files).

Closes #3958

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,13 +67,3 @@ repos:
         entry: '(?i)(doi:|dx\.doi\.org|http://doi\.org)'
         language: pygrep
         files: \.(rst|tex)$
-ci:
-    # Settings for the https://pre-commit.ci/ continuous integration service
-    autofix_prs: false
-    # Default message is more verbose
-    autoupdate_commit_msg: '[pre-commit.ci] autoupdate'
-    # Default is weekly
-    autoupdate_schedule: quarterly
-    # Currently pre-commit.ci checks all files, and flake8 takes too long
-    # and causes a timeout - so as a short-term workaround, skip it:
-    skip: [flake8]
--- a/README.rst
+++ b/README.rst
@@ -4,9 +4,6 @@
 .. image:: https://img.shields.io/conda/vn/conda-forge/biopython.svg?logo=conda-forge
    :alt: Biopython on the Conda package conda-forge channel
    :target: https://anaconda.org/conda-forge/biopython
-.. image:: https://results.pre-commit.ci/badge/github/biopython/biopython/master.svg
-   :target: https://results.pre-commit.ci/latest/github/biopython/biopython/master
-   :alt: pre-commit.ci status
 .. image:: https://img.shields.io/circleci/build/github/biopython/biopython.svg?logo=circleci
    :alt: Linux testing with CircleCI
    :target: https://app.circleci.com/pipelines/github/biopython/biopython


peterjc:
Actually, this isn't completely redundant - we will loose the automatic updating of the pre-commit tools, e.g. most recently https://github.com/biopython/biopython/pull/3895

We will instead have to periodically manually run ``pre-commit autoupdate`` to do this.

JoaoRodrigues:
I think the automated updating of the CI tools is a good thing actually. More control over how we run things and easier to reproduce locally. Otherwise we always have to check if our local CI config matches theirs. 👍 for me, good to merge!

peterjc:
To be clear, if we merge this (and disable the pre-commit.ci integration), we will lose the automated PR updating the pre-commit-hook's dependencies (currently set to quarterly as the weekly default was annoyingly often).

peterjc:
As per feedback on https://github.com/pre-commit-ci/issues/issues/131 if we want to retain the automatic updating of the pre-commit hook's dependencies (which I do like), we should keep the pre-commit continuous integration live but skip all/most of the hooks.

If we skipped (almost) everything then there is little reason to keep the README badge.

If we skipped just flake8 and black, then the CI would be really fast (not very useful, but more than just skipping everything). A sensible compromise?

peterjc:
OK, let's cut down the pre-commit.ci configuration, with a view to adding investigating ``pre-commit autoupdate`` via GitHub Actions.

JoaoRodrigues:
Regarding running autoupdate in an action, what do you think it's best:

1) pull request so that we can review changes to pre-commit everytime
autoupdate finds new versions. Example:
https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/

Or, 2) simply push to master directly. Example:
https://github.com/haddocking/pdb-tools/blob/master/.github/workflows/bump-version-on-push.yml

(1) requires a specific GH action and a bit more "work" to accept the PR,
but it gives us a little more control, (2) requires one of us to setup a
few secrets in the repo so that the commits aren't anonymous, but it's
painless thereafter.

A domingo, 26/06/2022, 07:34, Peter Cock ***@***.***>
escreveu:

peterjc:
I would prefer a PR, since occasionally updating the linting/style checks fails (e.g. new linting rule we're not following).

peterjc:
Another minor point is ``pre-commit autoupdate`` does not update sub-dependencies like the version of black used by blacken-docs. The author said this was too complicated with multiple hook languages (even if Python/PYPI looks straightforward).

On the other hand, the automated idea using secrets to push to master could be made conditional on the pre-commit tests still passing after updating the pre-commit dependencies?

peterjc:
As per my comment https://github.com/biopython/biopython/issues/3958#issuecomment-1177462669 my current inclination is to turn on the pre-commit.ci feature to apply fixers to pull requests (e.g. run black if needed).

For that, pre-commit.ci would only need to run actual fixers (e.g. black) and not validators (covered by GA), and should be fine with the 3m timeout.

PR #3968, opened 2022-06-23
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,6 +319,8 @@ jobs:
 
   # Flag step
   check_tests:
+    permissions:
+      contents: none
     needs: [test_linux, test_macos, test_windows]
     runs-on: ubuntu-latest
     steps:
@@ -366,6 +368,8 @@ jobs:
       shell: bash
     
   cleanup_wheels:
+    permissions:
+      contents: none
     if: always()
     needs: [build_wheels]
     runs-on: ubuntu-latest


peterjc:
Disregard the CircleCI failure:

```
  File "/home/circleci/project/biopython-1.80.dev0/Tests/test_phenotype_fit.py", line 70, in test_WellRecord
    self.assertAlmostEqual(w.y0, 45.879770069807989, places=5)
AssertionError: 45.879776570148486 != 45.87977006980799 within 5 places (6.5003404969843359e-06 difference)

```

Might be noise, or could be the recent numpy release - either way unrelated to this PR.

peterjc:
The links provided didn't explain to me why you think this change is wise.

For example, the page https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ talks about risks with ``pull_request_target`` which we're not using, and does not seem to mention the permissions-contents-none setting.

naveensrinivasan:
> The links provided didn't explain to me why you think this change is wise.
> 
> For example, the page https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ talks about risks with `pull_request_target` which we're not using, and does not seem to mention the permissions-contents-none setting.

It is a good practice. If you don't think so. I understand. Thanks.

PR #3979, opened 2022-07-11
Take the other object's class into consideration when comparing objec…

### LINKED ISSUE #3978: SeqFeature raises an unexpected AttributeError when compared to anything else than another SeqFeature
### Setup

system as follows:

```python
```

```bash
3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0]
CPython
Linux-5.15.0-40-generic-x86_64-with-glibc2.35
(...)
1.80.dev0
>>> 
```

### Expected behaviour

Comparing a SeqFeature object to any other Python object should always return a boolean - `True` or `False`.

### Actual behaviour

When comparing a SeqFeature object to a different object (for example, to a NoneType object), the operation raises an AttributeError instead of returning a simple False. The `__eq__` method is not prepared to handle any other objects that don't meet the SeqFeature interface.

### Steps to reproduce

```python
>>> from Bio.SeqFeature import SeqFeature
>>> seqfeature_obj = SeqFeature()
>>> seqfeature_obj2 = SeqFeature()
>>> seqfeature_obj == seqfeature_obj2
True
>>> seqfeature_obj == None
  File "/home/battlesheep/PycharmProjects/biopython/Bio/SeqFeature.py", line 292, in __eq__
    self.id == other.id
AttributeError: 'NoneType' object has no attribute 'id'
>>> 
```

Battlesheepu:
…ts using the __eq__ method

- [ X ] I hereby agree to dual licence this and any previous contributions under both

- [ X ] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``

- [ X ] I have added my name to the alphabetical contributors listings in the files

Closes #3978

Fixes an AttributeError raised whenever SeqFeature objects are compared to objects of any other type.

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -288,7 +288,8 @@ class SeqFeature:
     def __eq__(self, other):
         """Check if two SeqFeature objects should be considered equal."""
         return (
-            self.id == other.id
+            isinstance(other, SeqFeature)
+            and self.id == other.id
             and self.type == other.type
             and self.location == other.location
             and self.qualifiers == other.qualifiers
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -259,6 +259,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - Richard Neher <https://github.com/rneher>
 - Rob Miller <https://github.com/rob-miller>
 - Robert Ernst <https://github.com/rernst>
+- Robert Sawicki <https://github.com/Battlesheepu>
 - Rodrigo Dorantes-Gilardi <https://github.com/rodogi>
 - Rona Costello <https://github.com/RonaCostello>
 - Ryan Stecher <https://github.com/rystecher>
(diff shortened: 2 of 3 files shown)


mdehoon (approved):

peterjc (approved):
This looks sensible for the SeqFeature. Can you check the Seq and SeqRecord objects behave nicely too?

Battlesheepu:
Thanks @peterjc .

As for your question:
- SeqRecord behaves very nicely, [with an explicit error message explaining what to do](https://github.com/biopython/biopython/blob/master/Bio/SeqRecord.py#L794). I wouldn't change it.
- I'd say Seq [also has a nice logic implemented in the `__eq__` method](https://github.com/biopython/biopython/blob/master/Bio/Seq.py#L418). In my opinion this won't crash in weird ways just because somebody decided to compare a Seq object against some weird instance.

peterjc:
Merged, thank you :)

PR #1336, opened 2017-07-21
Move self-test for Bio.PDB.PSEA

souravsingh:
Fixes a part of #820

PR #3980, opened 2022-07-11
Move self test psea

valentin994:

related to  #1336 #820

I'm not sure if the pipeline job has psea already installed. If you want some more testing on this let me know, I was focused on covering all the code. I used the PDB files that were present in Tests/PDB. The psea function creates a  few`.sea` file, should I create a clean up job to delete those? If there are any other changes you want me to do please let me know.

### DIFF
--- a/Tests/test_PDB_PSEA.py
+++ b/Tests/test_PDB_PSEA.py
@@ -1,4 +1,4 @@
-# Copyright 2017 by Valentin Vareskic (valentin.vareskic@gmail.com).
+# Copyright 2022 by Valentin Vareskic (valentin.vareskic@gmail.com).
 # All rights reserved. This code is part of the Biopython distribution
 # and governed by its license.  Please see the LICENSE file that should
 # have been included as part of this package.
@@ -7,24 +7,92 @@
 
 import unittest
 from ftplib import FTP
-from Bio.PDB.PSEA import run_psea
+from Bio.PDB.PSEA import run_psea, psea, psea2HEC, annotate, PSEA
+from Bio.PDB import PDBParser
 
 
-# ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
-def setup():
-    with FTP("ftp.lmcp.jussieu.fr") as ftp:
-        ftp.login()
-        ftp.dir()
+class TestPDBPSEA(unittest.TestCase):
+    def test_run_psea(self):
+        psae_run = run_psea("PDB/1A8O.pdb")
+        self.assertEqual(psae_run, "1A8O.sea")
+
+    def test_psea(self):
+        psae_run = psea("PDB/2BEG.pdb")
+        self.assertEqual(psae_run, "ccccbbbbbbbccccbbbbbbbbbbc")
+
+    def test_psea_2HEC(self):
+        seq = psea("PDB/2BEG.pdb")
+        psae_run = psea2HEC(seq)
+        self.assertEqual(
+            psae_run,
+            [
+                "C",
+                "C",
+                "C",
+                "C",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "C",
+                "C",
+                "C",
+                "C",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "C",
+            ],
+        )
 
 
-# TODO Create unittest for run_psea(fname)
-def test_run_psea():
-    pass
+class TestPSEA(unittest.TestCase):
+    def test_get_seq(self):
+        p = PDBParser()
+        s = p.get_structure("X", "PDB/2BEG.pdb")
+        psea_class = PSEA(s[0], "PDB/2BEG.pdb")
+        self.assertEqual(
+            psea_class.get_seq(),
+            [
+                "C",
+                "C",
+                "C",
+                "C",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "C",
+                "C",
+                "C",
+                "C",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "E",
+                "C",
+            ],
+        )
 
 
 if __name__ == "__main__":
-    setup()
-# TODO Create unittest psea(pname)
-# TODO Create unittest psea2HEC(pseq)
-# TODO Create unittest annotate(m, ss_seq)
-# TODO Create unittest class PSEA
+    unittest.main()
--- a/Tests/test_PDB_PSEA.py
+++ b/Tests/test_PDB_PSEA.py
@@ -4,12 +4,15 @@
 # have been included as part of this package.
 
 """Tests for PDB PSEA."""
-import unittest
-from Bio.PDB.PSEA import run_psea, psea, psea2HEC, PSEA
-from Bio.PDB import PDBParser
-from subprocess import getoutput
-from Bio import MissingExternalDependencyError
+
 import os
+import unittest
+from subprocess import getoutput
+
+from Bio import MissingExternalDependencyError
+from Bio.PDB import PDBParser
+from Bio.PDB.PSEA import run_psea, psea, psea2HEC, PSEA
+
 
 os.environ["LANG"] = "C"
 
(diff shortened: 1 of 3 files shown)


peterjc:
This looks almost done.

You'll need to add some logic to detect if the psea binary is on the path or not (see e.g. ``Tests/test_Muscle_tool.py`` for an example, or the code on #1336), raising ``MissingExternalDependencyError`` as a signal to ``run_tests.py`` that the tests are being skipped rather than failing.

Also, we can drop the ``if __name__ == '__main__':`` block from ``Bio/PDB/PSEA.py`` since you're adding a proper test.

peterjc:
You haven't removed the if ``__name__ == '__main__':`` block from ``Bio/PDB/PSEA.py`` yet (perhaps done but not checked in as its a different file?)

valentin994:
Oh, I actually removed it from the test. 😅

peterjc:
Probably worth putting that back so that it can be invoked directly with ``python test_PDB_PSEA.py``

peterjc:
The psea executable check is failing on Windows

valentin994:
Okay, the test failed, but it doesn't seem my changes caused it.

Tests/test_PDB_PSEA.py
```diff
@@ -0,0 +1,114 @@
...
+"""Tests for PDB PSEA."""
+
+import unittest
+from Bio.PDB.PSEA import run_psea, psea, psea2HEC, PSEA
+from Bio.PDB import PDBParser
+from subprocess import getoutput
+from Bio import MissingExternalDependencyError
+import os
+
+if "command not found" or "'psea' is not recognized" in getoutput("psea -h"):
```
peterjc, line 15: As in the examples referred to, please add this line above:

```
os.environ['LANG'] = 'C'
```

We want to ensure the error messages are in US English to match the if-statement.

peterjc (changes requested):

peterjc:
From @JoaoRodrigues' comments on #1336 I think he has psea installed...

Tests/test_PDB_PSEA.py
```diff
@@ -0,0 +1,116 @@
...
+import unittest
+from Bio.PDB.PSEA import run_psea, psea, psea2HEC, PSEA
+from Bio.PDB import PDBParser
+from subprocess import getoutput
+from Bio import MissingExternalDependencyError
+import os
+
+os.environ["LANG"] = "C"
+
+if "command not found" or "'psea' is not recognized" in getoutput("psea -h"):
```
JoaoRodrigues, line 17: I would separate this into two lines for better readability and invert the if statement to test for the expected output. If I have my terminal language in e.g. french, it won't say 'command not found'.

e.g.
```
cmd_output = getoutput("psea -h")
if not cmd_output.startswith(...):
    raise MissingExternalDependencyError(...)
```

Tests/test_PDB_PSEA.py
```diff
@@ -0,0 +1,116 @@
...
+    raise MissingExternalDependencyError(
+        "Download and install psea from ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/. Make sure that psea is on path"
```
JoaoRodrigues, line 19: This link is dead for me, but I can get `p-sea` (note the dash in the name) here: http://bioserv.cbs.cnrs.fr/PSEA/DOWNLOAD/. Are these different versions?

Tests/test_PDB_PSEA.py
JoaoRodrigues, line 17: Actually, this conditional will always evaluate to `True` because it is interpreted as `if ("command not found") or ("'psea' is not recognized" in getoutput("psea -h"))` and the first string (command not found) evaluates to True.

Tests/test_PDB_PSEA.py
```diff
@@ -0,0 +1,116 @@
...
+                "E",
+                "E",
+                "E",
+                "E",
+                "C",
+            ],
+        )
+
+
+def clean_up():
```
JoaoRodrigues, line 106: This doesn't seem to be called anywhere. I'd call this in a `tearDown` method for all Test* classes, so that after every test, you remove the files.

JoaoRodrigues (changes requested):

JoaoRodrigues:
FYI, the output of `p-sea -h` is:

```$ psea -h
o-----------------------------------------------o
|                                               |
| p-sea :   Version 0.99 LMCP       Sept   96   |
|                                               |
o-----------------------------------------------o
| synopsis : phi psi omega alpha tau chi(s)     |
| and CA-1 CA+1 distance computing              |
|                                               |
| usage: sea pdb_file [out.sea]                 |
|                                               |
| ' ' : assign secondary structure              |
|                                               |
o-----------------------------------------------o
| options:                                      |
| -p  : computes phi, psi and omega             |
| -a  : computes alpha,tau and CA_CA+1 distance |
| -x  : computes chi1,chi2 and chi3             |
| -d  : computes CA-1_CA+1 distance             |
| -h  : displays this help                      |
| -m  : assign secondary structure              |
o-----------------------------------------------o
| if for any reason (missing atoms, angle not   |
| defined,...) the angle cannot be computed,    |
| it is set to 500.0                            |
|                                               |
o-----------------------------------------------o
| output              : on standard output      |
| comments and errors : on standard error       |
|                                               |
o-----------------------------------------------o
|                                               |
| The ouput values are in order (on one line):  |
| Atom_name Residue_name Residue_number [Phi \ |
| Psi Omega] [Alpha Tau CA_CA+1_Distance]    \ |
| [Chi1 Chi2 Chi3] [CA-1_CA+1_Distance]         |
|                                               |
o-----------------------------------------------o
```

Tests/test_PDB_PSEA.py
peterjc: Good catch - I didn't read that closely enough.

valentin994: First I was like where is the problem here, then it came to me like oooh 😄

Tests/test_PDB_PSEA.py
```diff
@@ -66,6 +68,10 @@ def test_psea_2HEC(self):
 
 
 class TestPSEA(unittest.TestCase):
+    def tearDown(self):
+        os.remove("2BEG.sea")
```
JoaoRodrigues, line 72: I'd make this a generic function that looks for and deletes ".sea" files, for simplicity and because you'll get an error if the test fails for some reason and these specific files aren't there. Then you can call the function in the two test classes.

Tests/test_PDB_PSEA.py
```diff
@@ -14,7 +14,9 @@
 
 os.environ["LANG"] = "C"
 
-if "command not found" or "'psea' is not recognized" in getoutput("psea -h"):
+
+cmd_output = getoutput("psea -h")
+if not cmd_output.startswith("o-----------------------------------------------o"):
     raise MissingExternalDependencyError(
```
JoaoRodrigues, line 20: You can shorten the string, no need for the full one.

JoaoRodrigues (changes requested):

Tests/test_PDB_PSEA.py
JoaoRodrigues: This is actually funny .. yeah, I agree, code should be the same!

peterjc:
An unrelated failure:

```
  File "/home/circleci/project/biopython-1.80.dev0/Tests/test_phenotype_fit.py", line 70, in test_WellRecord
    self.assertAlmostEqual(w.y0, 45.879770069807989, places=5)
AssertionError: 45.879776570148486 != 45.87977006980799 within 5 places (6.5003404969843359e-06 difference)

```

That was fixed in https://github.com/biopython/biopython/commit/a67e899126f3f0b2a426a238482cbcc338ba7495 so you can ignore this (or rebase to the latest master branch if you prefer).

JoaoRodrigues:
GitHub is messing up my comments so I'll write here. Almost there! 

Your last change to the removal of .sea files will fail (you import pathlib but use os). You also don't need the list comprehension - no need for a list. I'd stick to a simple for loop, pack it in a function like you had before, and call it in tearDown for both test classes.

JoaoRodrigues:
Looking good!

One last comment, purely stylistic. Imports are usually (but not always) sorted alphabetically, and then in the following order: standard library imports, then third-party modules (e.g. numpy), and finally local modules (e.g. Biopython in this case). 'Full' imports also usually come after 'partial' imports, so `import ...` comes first and `from ... import ...` comes later. So, your imports could be rewritten as:

```
import os
import unittest
from subprocess import getoutput

from Bio import MissingExternalDependencyError
from Bio.PDB import PDBParser
from Bio.PDB.PSEA import run_psea, psea, psea2HEC, PSEA
```

Again, nitpicking the style, the changes look good to me. Tests pass locally with p-sea installed.

JoaoRodrigues (approved):

valentin994:
I would run isort on it, but as you have pre-commit on it I was just leaving them as they are

JoaoRodrigues:
Yeah, there have been discussions about isort (e.g. #2828) but we never agreed on a strategy! Thanks for all the patience with the PR!

valentin994:
Glad I could help 😁

JoaoRodrigues:
Merged, thank you @valentin994 !

PR #3984, opened 2022-07-15
Extracts information about the relative orientation of chains in oligomeric structures

sachdved:

- [x ] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``

- [x ] I have added my name to the alphabetical contributors listings in the files

Currently, no functionality in the PDB parser extracts information from REMARK 350 about the orientations of the chains in their oligomeric state. I have written an additional function to extract the information accordingly and save it into the dictionary belonging to the header attribute of a structure.

### DIFF
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -226,7 +226,7 @@ def _parse_pdb_header_list(header):
         "source": {"1": {"misc": ""}},
         "has_missing_residues": False,
         "missing_residues": [],
-        "chain_orientations":{},
+        "chain_orientations":None,
     }
 
     pdbh_dict["structure_reference"] = _get_references(header)
--- a/Bio/PDB/parse_pdb_header.py
+++ b/Bio/PDB/parse_pdb_header.py
@@ -19,6 +19,38 @@ import re
 from Bio import File
 
 
+def _get_chain_orientations(inl):
+    ##automated approach to extracting information about the relative orientations of the chains from pdb header. This information is typically located at REMARK 350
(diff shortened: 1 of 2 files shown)


JoaoRodrigues:
Hi @sachdved,

Thanks for the contribution! I have a few comments regarding the code and the nomenclature you used. The documentation for REMARK 350 is found [here](https://www.wwpdb.org/documentation/file-format-content/format33/remarks2.html#REMARK%20350). These are not orientations per se, but more transformations to reconstruct the biological  assembly (or assemblies), which can be defined by the authors or automatically.

Regarding the code itself, before starting a proper code review, could you add a couple of test cases that will help us maintain your contribution in the future? You can pick a few PDB headers and use those to parse out the remarks and make sure that the output is consistent with what you would expect. Have a look in the `tests` folder (look for `test_PDB_*py files) for example of how we write tests.

Let me know if you have any questions about my comments.

sachdved:
Hi Joao - thanks for your comments. I edited the style, changed the nomenclature, and am still working on the unit tests. Unfortunately, I'm running into an issue with the comparison of dictionaries in the unit test package. I'll resolve this hopefully relatively soon.

sachdved:
Oh sorry, I think this should work now! I had to change the unit testing method to avoid issues with ambiguity emerging from comparing dictionaries. I addressed this by using numpy's assert equal method and unpacking the dictionaries like that. Hopefully that's alright.

sachdved:
Hi @JoaoRodrigues - just wanted to follow up on this and see if this should be merged or discarded!

PR #3998, opened 2022-07-25
Convert more f-strings

Caiofcas:

Some progress in #2510


peterjc:
How are you doing these? If by hand, that makes be very nervous about accidental mistakes.

Also something is triggering ``RecursionError: maximum recursion depth exceeded while calling a Python object`` on the tests.

Caiofcas:
Was doing them by hand, yes, but did somtehing wrong when testing locally, will fix it.

peterjc:
Is there good reason to replace ``str(expression)`` with ``f"{expression}"`` here? From a C programming background this is like a cast, and very clear and explicit. It might be faster too?

Caiofcas:
I had the impression it would be faster since it would avoid a function call, but doing some more research this appears to only be true for simpler types. I think I will close this PR and open some smaller ones to facilitate the review process.

peterjc:
OK - maybe focus on one sub-directory at a time? And I would suggest starting with just running flynt and black (automated so less chance of error) and see how much manual review is needed after that.

Some slightly out of date numbers on https://github.com/biopython/biopython/issues/2510#issuecomment-916793707

Issue #3989, opened 2022-07-18
Update argument handling in PDBList.py, possibly by creating a new module.


JoaoRodrigues:
The command-line interface for the `PDBList.py` module is quite clunky, using a series of complicated ifs and sys.argv calls to parse input from the user. It would be super nicer to simply use `argparse`, although this will break compatibility with previous versions.

The solution here might pass through copying the code to a new module and along with other issues (#3988, #3987) upgrade `PDBList.py` and break backwards compability. We could then slap a deprecation warning on PDBList to warn users to move to the new code.

RUCHI21CODES:
Can you please assign me this issue

peterjc:
@RUCHI21CODES No, there are already people working on this with open pull requests. If you can spot improvements or otherwise usefully comment on those PRs, that could be helpful.

onhgomat:
# pdblist_args.py
import argparse

def parse_args():
    parser = argparse.ArgumentParser(description='PDB List arguments')
    parser.add_argument('-i', '--input_file', help='Input file path', required=True)
    parser.add_argument('-o', '--output_file', help='Output file path', required=True)
    parser.add_argument('-v', '--verbose', help='Verbose mode', action='store_true')
    return parser.parse_args()



# PDBList.py
import pdblist_args

def main():
    args = pdblist_args.parse_args()
    # Use the parsed arguments here
    print(f'Input file: {args.input_file}')
    print(f'Output file: {args.output_file}')
    if args.verbose:
        print('Verbose mode enabled')

if __name__ == '__main__':
    main()

MANICHELLURII:
Hi, I’d like to work on this. Can you assign it to me?

peterjc:
@MANICHELLURII Maybe you could review #4032?

MANICHELLURII:
@peterjc  can you assign me to another issue?

peterjc:
With that old PR closed (understandable as it had gone stale and this would have been frustrating), you could take a fresh run at this.

ernestprovo23:
hey, id like to take a fresh run at this. saw the old PR #4032 went stale and @peterjc mentioned it was open for another attempt.

my approach: replace the sys.argv block with argparse subcommands (update, all, obsol, assemb, fetch). gonna use proper --format, --flat, --overwrite flags instead of the old -pdb/-d/-o style. will include tests for the parser.

taimurtahir2008:
hi i would like to make a contribution can u assign me to one. thanks

[...]

Issue #3988, opened 2022-07-18
Extend PDBList obsolete handling to mmCIF and mmtf formats as well as assemblies


JoaoRodrigues:
The current code in `PDBList.py` to move obsolete files expects a `.ent` extension, which is reserved for PDB-formatted files. We should extend this to depend on the file_format of choice (e,g. mmCIF, mmtf) and also to assemblies (if the original entry is labelled obsolete). Relevant code is [here](https://github.com/JoaoRodrigues/biopython/blob/4d7042ed1185505691c7275c50bb539ef871211a/Bio/PDB/PDBList.py#L389).

Fixing this involves keeping track of which extensions are used for which file formats, probably by encoding a dictionary to map file_format to extension, and then retrieving the extension based on the file_format provided to `update_pdb`.

kirandharmar867-maker:
hii I'd like to work on this issue if its still available.
please assign it to me.
Thanks!.

peterjc:
Given #3986 was merged, it would be good to revisit #4012 and #4014. @JulianMaurin might want to, but happily they did the appropriate licensing so that their work could be reused as a base.

kirandharmar867-maker:
hii @JoaoRodrigues .
can You pls Review my PR.
see https://github.com/biopython/biopython/pull/5157#issuecomment-3847705189.

PR #4012, opened 2022-07-30
Improve PDB server interoperability

### LINKED ISSUE #3987: Fix PDBList access to servers other than RCSB

JulianMaurin:
Hello :wave: 

This RP is intended to resolve issue #3987.

I'm not a PDB server specialist but I found information [here](https://www.wwpdb.org/ftp/pdb-ftp-sites) and [here](https://www.rcsb.org/docs/programmatic-access/file-download-services).

Introducing a new module `PDBServer` containing the eponymous class to organize server-related data and processing.
I think this implementation will allow us to evolve the `PDBList` module in a good direction.

It is fully backwards compatible. and thanks to lru cache we will avoid duplicated network queries.

### Protocols

As the introduction of the `PDBServer` module has greatly facilitated the implementation of this feature, I went a little further than the need expressed in the issue by implementing the ability to download files using the HTTPS protocol.

This need seemed to me necessary in view of the recommendations of the RCSB ([source](https://www.rcsb.org/docs/programmatic-access/file-download-services)):

> The HTTPS protocol offers more flexibility and less overhead and is strongly recommended for usage in scripts and third party software. HTTPS links for this service are described below. The FTP protocol will be gradually phased out.

______

______

Closes #3987

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -61,9 +61,13 @@ from Bio.PDB.PDBExceptions import PDBException
 
 @dataclasses.dataclass
 class PDBServer:
-    """Protein Data Server class.
+    """Dataclass to store parameters for connecting to PDB server.
 
-    In charge of the interactions with the server.
+    param domain: server domain name.
+    type domain: string
+
+    :param pdb_directory: path to the PDB data directory.
+    :type domain: string
     """
 
     domain: str
@@ -532,11 +536,12 @@ class PDBList:
             ftp = ftplib.FTP(pdb_server.split("/", 1)[0])
         ftp.login()  # anonymous
         pdb_directory = pdb_server.split("/", 1)[1]
+        ftp.cwd(pdb_directory)
         if file_format.lower() == "mmcif":
-            ftp.cwd(f"{pdb_directory}/data/assemblies/mmCIF/all/")
+            ftp.cwd("data/assemblies/mmCIF/all/")
             re_name = re.compile(r"(\d[0-9a-z]{3})-assembly(\d+).cif.gz")
         elif file_format.lower() == "pdb":
-            ftp.cwd(f"{pdb_directory}/data/biounit/PDB/all/")
+            ftp.cwd("data/biounit/PDB/all/")
             re_name = re.compile(r"(\d[0-9a-z]{3}).pdb(\d+).gz")
         else:
             msg = "file_format for assemblies must be 'pdb' or 'mmCif'"
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -35,18 +35,25 @@
 
 """Access the PDB over the internet (e.g. to download structures)."""
 
+from __future__ import annotations
 
+import collections
 import contextlib
 import ftplib
+import functools
 import gzip
 import os
 import re
 import shutil
+import socket
 import sys
+import time
 
 from urllib.request import urlopen
 from urllib.request import urlretrieve
 from urllib.request import urlcleanup
+
+from Bio.PDB.PDBExceptions import PDBException
 
 
 class PDBList:
@@ -84,15 +91,14 @@ class PDBList:
     http://www.pdb.org/.
     """
 
-    def __init__(
-        self, server="ftp://ftp.wwpdb.org", pdb=None, obsolete_pdb=None, verbose=True
-    ):
+    def __init__(self, server=None, pdb=None, obsolete_pdb=None, verbose=True):
         """Initialize the class with the default server or a custom one.
 
         Argument pdb is the local path to use, defaulting to the current
         directory at the moment of initialisation.
         """
-        self.pdb_server = server  # remote pdb server
+        self.pdb_server = self._get_pdb_server_url(server)
+
         if pdb:
             self.local_pdb = pdb  # local pdb file tree
         else:
@@ -111,6 +117,25 @@ class PDBList:
 
         # variable for command-line option
         self.flat_tree = False
+
+    def _get_pdb_server_url(self, server: str | PDBServer | None) -> str:
+        """Get remote PDB server url.
+
+        Handle legacy server declaration (str).
+        If no server specified, chose the fastest one.
+        """
+        pdb_server = None
+        if isinstance(server, str):
+            return "{server}/pub/pdb"
+        elif isinstance(server, PDBServer):
+            pdb_server = server
+        elif server is None:
+            pdb_server = get_fastest_server()
+
+        if pdb_server:
+            return f"ftp://{pdb_server.domain}{pdb_server.pdb_directory}"
+
+        raise TypeError(f"Unexpected server (server: {server}).")
 
     @staticmethod
     def _print_default_format_warning(file_format):
@@ -157,12 +182,12 @@ class PDBList:
             -rw-r--r--   1 1002     sysadmin    1327 Mar 12  2001 README
 
         """
-        path = self.pdb_server + "/pub/pdb/data/status/latest/"
+        url = f"{self.pdb_server}/data/status/latest"
 
         # Retrieve the lists
-        added = self.get_status_list(path + "added.pdb")
-        modified = self.get_status_list(path + "modified.pdb")
-        obsolete = self.get_status_list(path + "obsolete.pdb")
+        added = self.get_status_list(f"{url}/added.pdb")
+        modified = self.get_status_list(f"{url}/modified.pdb")
+        obsolete = self.get_status_list(f"{url}/obsolete.pdb")
         return [added, modified, obsolete]
 
     def get_all_entries(self):
@@ -170,7 +195,7 @@ class PDBList:
 
         Returns a list of PDB codes in the index file.
         """
-        url = self.pdb_server + "/pub/pdb/derived_data/index/entries.idx"
+        url = f"{self.pdb_server}/derived_data/index/entries.idx"
         if self._verbose:
             print("Retrieving index file. Takes about 27 MB.")
         with contextlib.closing(urlopen(url)) as handle:
@@ -202,7 +227,7 @@ class PDBList:
             ...
 
         """
-        url = self.pdb_server + "/pub/pdb/data/status/obsolete.dat"
+        url = f"{self.pdb_server}/data/status/obsolete.dat"
         with contextlib.closing(urlopen(url)) as handle:
             # Extract pdb codes. Could use a list comprehension, but I want
             # to include an assert to check for mis-reading the data.
@@ -290,17 +315,14 @@ class PDBList:
                 if file_format == "mmCif"
                 else "XML"
             )
-            url = self.pdb_server + "/pub/pdb/data/structures/%s/%s/%s/%s" % (
-                pdb_dir,
-                file_type,
-                code[1:3],
-                archive_fn,
+            url = (
+                f"{self.pdb_server}/data/structures/"
+                f"{pdb_dir}/{file_type}/{code[1:3]}/{archive_fn}"
             )
         elif file_format == "bundle":
-            url = self.pdb_server + "/pub/pdb/compatible/pdb_bundle/%s/%s/%s" % (
-                code[1:3],
-                code,
-                archive_fn,
+            url = (
+                f"{self.pdb_server}/compatible/pdb_bundle/"
+                f"{code[1:3]}/{code}/{archive_fn}"
             )
         else:
             url = f"http://mmtf.rcsb.org/v1.0/full/{code}"
@@ -479,17 +501,19 @@ class PDBList:
 
         # FTPLib is much faster than urlopen
         idx = self.pdb_server.find("://")
+        pdb_server = self.pdb_server
         if idx >= 0:
-            ftp = ftplib.FTP(self.pdb_server[idx + 3 :])
+            pdb_server = self.pdb_server[idx + 3 :]
+            ftp = ftplib.FTP(pdb_server.split("/", 1)[0])
         else:
-            ftp = ftplib.FTP(self.pdb_server)
+            ftp = ftplib.FTP(pdb_server.split("/", 1)[0])
         ftp.login()  # anonymous
-
+        pdb_directory = pdb_server.split("/", 1)[1]
         if file_format.lower() == "mmcif":
-            ftp.cwd("/pub/pdb/data/assemblies/mmCIF/all/")
+            ftp.cwd(f"{pdb_directory}/data/assemblies/mmCIF/all/")
             re_name = re.compile(r"(\d[0-9a-z]{3})-assembly(\d+).cif.gz")
         elif file_format.lower() == "pdb":
-            ftp.cwd("/pub/pdb/data/biounit/PDB/all/")
+            ftp.cwd(f"{pdb_directory}/data/biounit/PDB/all/")
             re_name = re.compile(r"(\d[0-9a-z]{3}).pdb(\d+).gz")
         else:
             msg = "file_format for assemblies must be 'pdb' or 'mmCif'"
@@ -548,9 +572,9 @@ class PDBList:
         archive_fn = archive[file_format] % (pdb_code.lower(), int(assembly_num))
 
         if file_format == "mmcif":
-            url = self.pdb_server + "/pub/pdb/data/assemblies/mmCIF/all/%s" % archive_fn
+            url = self.pdb_server + "/data/assemblies/mmCIF/all/%s" % archive_fn
         elif file_format == "pdb":
-            url = self.pdb_server + "/pub/pdb/data/biounit/PDB/all/%s" % archive_fn
+            url = self.pdb_server + "/data/biounit/PDB/all/%s" % archive_fn
         else:  # better safe than sorry
             raise ValueError("file_format '%s' not supported: %s" % file_format)
 
@@ -665,8 +689,47 @@ class PDBList:
         """Retrieve and save a (big) file containing all the sequences of PDB entries."""
         if self._verbose:
             print("Retrieving sequence file (takes over 110 MB).")
-        url = self.pdb_server + "/pub/pdb/derived_data/pdb_seqres.txt"
+        url = f"{self.pdb_server}/derived_data/pdb_seqres.txt"
         urlretrieve(url, savefile)
+
+
+PDBServer = collections.namedtuple("PDBServer", ["domain", "pdb_directory"])
+
+SERVERS = [
+    PDBServer("ftp.rcsb.org", "/pub/pdb"),
+    PDBServer("ftp.ebi.ac.uk", "/pub/databases/pdb"),
+    PDBServer("ftp.pdbj.org", "/pub/pdb"),
+]
+
+
+@functools.lru_cache(maxsize=None)
+def get_fastest_server(servers: list[PDBServer] | None = None) -> PDBServer:
+    """Fastest server among the available ones."""
+    servers = servers or SERVERS
+    fastest_server = None
+    fastest_timing = None
+
+    for server in servers:
+        server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        server_socket.settimeout(1)
+
+        time_before_connect = time.time()
+        try:
+            server_socket.connect((server.domain, 21))
+        except Exception:
+            continue
+        finally:
+            server_socket.close()
+        timing = time.time() - time_before_connect
+
+        if timing and (not fastest_timing or timing < fastest_timing):
+            fastest_server = server
+            fastest_timing = timing
+
+    if not fastest_server:
+        raise PDBException("Unable to connect to any PDB servers.")
+
+    return fastest_server
 
 
 if __name__ == "__main__":
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -66,12 +66,12 @@ class PDBServer:
     param domain: server domain name.
     type domain: string
 
-    :param pdb_directory: path to the PDB data directory.
+    :param pdb_dir_path: path to the PDB data directory.
     :type domain: string
     """
 
     domain: str
-    pdb_directory: str
+    pdb_dir_path: str
 
 
 SERVERS = [
@@ -159,7 +159,7 @@ class PDBList:
 
         if pdb_server:
             return urlunsplit(
-                ("ftp", pdb_server.domain, pdb_server.pdb_directory, None, None)
+                ("ftp", pdb_server.domain, pdb_server.pdb_dir_path, None, None)
             )
 
         raise TypeError(f"Unexpected server (server: {server}).")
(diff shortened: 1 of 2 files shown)


Bio/PDB/PDBServer.py
```diff
@@ -0,0 +1,173 @@
...
+        except Exception:
+            return None
+        return time.time() - time_before_connect
+
+
+@contextlib.contextmanager
+def _build_socket():
+    """Context manager auto closing socket."""
+    server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server_socket.settimeout(1)
```
JulianMaurin, line 122: Could be a little bit short, but checking all the servers could be long if the timeout is too important.

it seems to be a good compromise to me, but maybe we can introduce an environment variable to make it customizable.

JulianMaurin:
Hello @JoaoRodrigues , 

I just pushed a commit with more tests, we now validate the download from all servers using the `HTTPS` and `FTP` protocols.
Let me know if you think I missed something 🚀

JulianMaurin:
Not that proud of the last commit, but I would like to limit the scope of the PR. I think the introduction of the `PDBServer` module is already a good step.

I would like to continue the refactoring around `PDBList` in the pursuit of https://github.com/biopython/biopython/issues/3988 (as explained in the comment). The refactoring will enable to management of servers/URLs more efficiently, based on the sever and the format of the file to download (assembly included).

JoaoRodrigues:
Will check this tomorrow @JulianMaurin, I am not sure the `PDBServer` module is needed (as a separate module), but I need to see what you wrote. 

I had thought we could refactor `PDBList` into something called `wwpdb` (that's the name of the consortium) and then we could break backwards compatibility and write cleaner code, where these new functions would live.

JulianMaurin:
> Will check this tomorrow @JulianMaurin, I am not sure the `PDBServer` module is needed (as a separate module), but I need to see what you wrote.

Thanks @JoaoRodrigues !
I look forward to your feedback.

> I had thought we could refactor `PDBList` into something called `wwpdb` (that's the name of the consortium) and then we could break backwards compatibility and write cleaner code, where these new functions would live.

If I may humbly challenge that. I feel like this new module helps to introduce new capabilities with the minimum impact on legacy code. This way we can add fully tested, reliable and isolated new features step by step. Wich will gradually emancipating ourselves from legacy code, in order to reach a larger refactoring. 

My short-term vision would correspond more to the following diagram (thanks to [mermaid](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/)):

```mermaid
graph TD
    subgraph Bio.PDB
        A[PDBServer]-->C[PDBList]
        B[PDBFile]-->C[PDBList]
    end
```

With:
- `PDBServer`: Data and logic for interaction with PDB servers (solving this issue).
- `PDBFile/PDBFileFormat` (name to be define):  Data and logic to build PDB file path (locally and remotely). (solving https://github.com/biopython/biopython/issues/3988)
- `PDBList`: performs action according to the CLI (to be refactored or replaced solving https://github.com/biopython/biopython/issues/3989)

And of course we can imagine going further in the future. I like the notion of "acquisition", which I have encountered many times while working on other scientific software (for satellite or weather data). But we can talk about it later, as I said, step by step 🚀

JoaoRodrigues:
Hey @JulianMaurin, all very good points!

>If I may humbly challenge that. I feel like this new module helps to introduce new capabilities with the minimum impact on legacy code. This way we can add fully tested, reliable and isolated new features step by step. Wich will gradually emancipating ourselves from legacy code, in order to reach a larger refactoring.

Indeed, but we don't have to restrain ourselves to legacy code. One of the ideas with these changes is that we can move to a new, clean, module without the constraints of maintaining backwards compability. The goals of the module are to provide a nice and simple interface to download/maintain a local copy of the PDB database. Additionally, we can offer some filtering/search capabilities, but those are yet to be implemented.

I understand your points with modularizing parts of the code, but for example, the code in `PDBServer.py` could be reduced to a couple of simpler functions that still offer a lot of flexibility and be integrated in the same module as `PDBList` or `wwpdb`. This helps a lot with maintainability on the long-term, and also with usability from a developer perspective, as there are less modules to load and therefore less cognitive effort when writing new code. For instance, we don't reuse the socket code and the timing code anywhere else, so there's also no obvious immediate advantage to modularizing that so much.

For more practical and general feedback on your implementation. There are only 3 servers in the world for this database, and it's very unlikely that this will change in the near future. It would be less elegant and less modular, but much simpler to store a mapping (can even be a dataclass like you have) with the server domain, port, and subdomain where the data lives, rather than abstracting all of those details. These classes would be consumed by the `get_fastest_server` function and return the best one. Then a user with a custom server could simply derive their own class from this and plug it in to the rest of the machinery. Adding 4 functions and 5 classes for this is fantastic, but a bit overengineered for this specific use case! 😄

My suggestion would be to collapse some of your code into a simpler function. Here's a 'mock implementation' based on your own:

```
PDBServer = namedtuple("PDBServer", ["name", "port", "base_url", "data_dir"])  # could be your dataclass

SERVERS = {
    "RCSB": PDBServer(name="RCSB", port="21", base_url="ftp.rcsb.org", base_dir="pub/pdb/"),
    ...
}

@functools.lru_cache(maxsize=None)
def get_fastest_server(server_list=None):
    """Return the fastest server among the available servers."""
    if server_list is None:
        server_list = SERVERS

    fastest_server = None
    fastest_timing = None

    for server in server_list.values():
        server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        server_socket.settimeout(1)

        st_time = time.time()
        try:
            server_socket.connect((server.build_full_domain(protocol), server.port))
        except Exception:
            continue
        finally:
            server_socket.close()
        
        timing = time.time() - st_time
        if timing and (not fastest_timing or timing < fastest_timing):
            fastest_server = server
            fastest_timing = timing

    if not fastest_server:
        raise PDBServersConnectionError(protocol)

    return fastest_server
```

This does the same job, and encapsulates all the logic of "finding the fastest server" in one small function. It is simpler to test, simpler for a user/developer to grasp and re-use, etc. What do you think?

JulianMaurin:
Hello @JoaoRodrigues !
Thanks a lot for this first feedback. Super interesting, a lot of good points, some others where I do not fully agree (😁) but I think we will succeed to find a compromise easily. 

Some questions to do so: 

- You are talking about 3 servers but I found 4 of them (`wwpdb.org`, `rcsb.org`, `ebi.ac.uk`, `pdbj.org`). Is there one that is not valid?
- I agree that I need to reduce the abstraction and the biggest part of it is due to the handling of both protocols. So we can reach something simpler by moving to a unique protocol. Are you okay with directly switching to `HTTPS`, according to the remark I did in the description?
 
Be right back with refactoring implementing your suggestions :v:

JoaoRodrigues:
HTTPS is fine!

From the website:
```wwPDB:[ ftp://ftp.wwpdb.org](ftp://ftp.wwpdb.org/), https://files.wwpdb.org/, rsync://rsync.wwpdb.org
RCSB PDB (US):[ ftp://ftp.rcsb.org](ftp://ftp.rcsb.org/), https://files.rcsb.org/, rsync://rsync.rcsb.org (see the [download protocol below](https://www.wwpdb.org/ftp/pdb-ftp-sites#downloadPor))
RCSB PDB (US/AWS): https://s3.rcsb.org/
PDBe (UK): [ftp.ebi.ac.uk/pub/databases/pdb/](https://ftp.ebi.ac.uk/pub/databases/pdb/)
PDBj (Japan): [ftp.pdbj.org](https://ftp.pdbj.org/)
```

I'd stick to the regional sites: RCSB, EBI, and PDBj. wwPDB is probably US-based anyway.

JulianMaurin:
@JoaoRodrigues ,
Here is a new version that is way more minimalistic, according to your suggestions.
Checking the impact of switching to HTTPS, I changed my mind. Let's keep this for another issue if you agree.

Bio/PDB/PDBList.py
```diff
@@ -479,17 +501,19 @@ def get_all_assemblies(self, file_format="mmCif"):
 
         # FTPLib is much faster than urlopen
```
JoaoRodrigues, line 502: This is a shame really. FTPLib is really much faster for some tasks.. shame that they'll drop the protocol.

Bio/PDB/PDBList.py
```diff
@@ -665,10 +689,49 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
         """Retrieve and save a (big) file containing all the sequences of PDB entries."""
         if self._verbose:
             print("Retrieving sequence file (takes over 110 MB).")
-        url = self.pdb_server + "/pub/pdb/derived_data/pdb_seqres.txt"
+        url = f"{self.pdb_server}/derived_data/pdb_seqres.txt"
         urlretrieve(url, savefile)
 
 
+PDBServer = collections.namedtuple("PDBServer", ["domain", "pdb_directory"])
```
JoaoRodrigues, line 696: Just from a logical perspective, I'd move this to the top of the file. `PDBServer` first, then `PDBList`.

JoaoRodrigues: And following my comment above about adding some methods to `PDBServer`, maybe indeed a dataclass like you had before would be a better approach, what do you think?

JoaoRodrigues (changes requested):
Thanks for the compromise! I did like your `PDBServer` dataclass approach, I only wrote the namedtuple as a quick-and-dirty alternative. So feel free to revert to that design if you think it's better - it does give us more flexibility.

JulianMaurin:
Thanks! I will apply the suggestion later today or this week.

Bio/PDB/PDBList.py
JulianMaurin: In fact, it's debatable, but do not worry about the speed. Implementing parallel download will speed up the process so much that it will no longer be an issue (but that is another step!).

JulianMaurin:
So ! Squashing all commits preceding the review to get it clean.  
I did not know `urllib.parse.urljoin`, that's quite magic, thanks !

I applied most of your suggestions, but there are still some open topics:

1. Delegate the construction of the URL to `PBSServer`
According to [this comment](https://github.com/biopython/biopython/pull/4012#issuecomment-1207399567), I think we can dedicate an entire class to the URL construction (taking into account the nature and the format of the file). It will also be in charge of the construction of the local path. 
I totally agree to transform `PDBServer` to class, but I think she should only be in charge of the actions related to the exchange with the server (perfoming the requests, whether it is HTTP or FTP). Ideally, it would look something like `PDBServer.retrieve(pdb_file: PDBFile)`.

2. HTTPS
I would prefer to dedicate a PR to it. This will be simpler and cleaner once all requests are centralized in PDBServer

What do you think about that?

JoaoRodrigues:
RE: FTP vs HTTPS. It's not about download speed of the structures, it's this single operation that takes forever to get, for some reason. In any case, have a look at https://gist.github.com/JoaoRodrigues/761b456a110a654794f87294b226e14d for _really_ speeding up downloads :) Another PR.

As for your questions:

>Delegate the construction of the URL to PBSServer

A `PDBServer.retrieve` method might make sense yes.

>HTTPS

Agreed. We are already accumulating a lot of changes here that would be, I think, much better served with a full refactoring of the module. @sean-workman had shown some interest in refactoring the argument parsing part (bottom of the module), so maybe you can find a way that to collaborate on the refactoring?

sean-workman:
I've actually spent a bit of time this afternoon getting going with argparse on PDBList, not quite there yet though. I would happily work together to do a large scale refactoring of things! I was hoping to have something to submit for #3989 sometime in the next day or two.

JulianMaurin:
> have a look at [gist.github.com/JoaoRodrigues/761b456a110a654794f87294b226e14d](https://gist.github.com/JoaoRodrigues/761b456a110a654794f87294b226e14d) for really speeding up downloads 

Neat :ok_hand: 

> Agreed. We are already accumulating a lot of changes here

According to the last commit, aligning the last details, can we consider we are good with this PR ?

>  I think, much better served with a full refactoring of the module. @sean-workman had shown some interest in refactoring the argument parsing part (bottom of the module), so maybe you can find a way that to collaborate on the refactoring?

> I would happily work together to do a large scale refactoring of things!

Yes of course. I would be glad to help and review any works around this `PDBList`.

Thanks to this PR, I think I have a better understanding of the context and the expectation.  I would be in favour of a suite of small PR, to ease the review.

Bio/PDB/PDBList.py
```diff
@@ -35,18 +35,46 @@
...
 from urllib.request import urlcleanup
+from urllib.parse import urljoin
+from urllib.parse import urlunsplit
+
+from Bio.PDB.PDBExceptions import PDBException
+
+
+@dataclasses.dataclass
+class PDBServer:
+    """Protein Data Server class.
```
JoaoRodrigues, line 64: Suggestion. Usually docstrings are a little more descriptive in the first line and have some description of the attributes/methods. For example, here I'd write 

```python
class PDBserver:
    """Dataclass to store parameters for connecting to PDB server mirrors.

        :param domain: base domain address of the server.
        :type domain: string

        :param pdb_directory: sub-domain or URL to the PDB data folder.
        :type domain: string
    """
```

Bio/PDB/PDBList.py
```diff
@@ -35,18 +35,46 @@
...
+
+    In charge of the interactions with the server.
+    """
+
+    domain: str
```
JoaoRodrigues, line 69: I'd give these more descriptive names. Not sure if these are better but `base_url` and `pdb_path` seem a little better to me.

Bio/PDB/PDBList.py
```diff
@@ -112,6 +139,27 @@ def __init__(
...
+        if server is None:
+            pdb_server = get_fastest_server()
+        elif isinstance(server, str):
+            return urljoin(server, "pub/pdb/")
+        elif isinstance(server, PDBServer):
+            pdb_server = server
+
+        if pdb_server:
+            return urlunsplit(
+                ("ftp", pdb_server.domain, pdb_server.pdb_directory, None, None)
```
JoaoRodrigues, line 158: Again, I'd be careful about using `ftp` protocols here, since private servers might _not_ use that protocol. I'd default to https unless the `PDBServer` class explicitly points to an FTP. You can add that as a `scheme` (to match the urlsplit/unsplit nomenclature) attribute to the dataclass.

Bio/PDB/PDBList.py
```diff
@@ -112,6 +139,27 @@ def __init__(
...
+            )
+
+        raise TypeError(f"Unexpected server (server: {server}).")
```
JoaoRodrigues, line 161: Suggestion: `TypeError(f"Unsupported server option: {server}")`

Bio/PDB/PDBList.py
```diff
@@ -112,6 +139,27 @@ def __init__(
         # variable for command-line option
         self.flat_tree = False
 
+    def _get_pdb_server_url(self, server: str | PDBServer | None) -> str:
+        """Get remote PDB server url.
+
+        Handle legacy server declaration (str).
+        If no server specified, chose the fastest one.
+        """
+        pdb_server = None
```
JoaoRodrigues, line 148: Why don't you simply overload server is server is None?

Bio/PDB/PDBList.py
```diff
@@ -479,17 +524,19 @@ def get_all_assemblies(self, file_format="mmCif"):
 
         # FTPLib is much faster than urlopen
         idx = self.pdb_server.find("://")
+        pdb_server = self.pdb_server
         if idx >= 0:
```
JoaoRodrigues, line 528: This condition was introduced because `ftplib` will throw an error if you access for example `https://ftp.rcsb.org` or `ftp://ftp.rcsb.org`. I'd keep the mix usage of http/ftp (see my comment above) and leave this code as is (or even turn it into a dataclass method that returns the FTP connection object and handle this logic in there with the knowledge of the domain name. You could use urlsplit to convert a server string into a series of components you can use to build a PDBServer dataclass.

Bio/PDB/PDBList.py
```diff
@@ -479,17 +524,19 @@ def get_all_assemblies(self, file_format="mmCif"):
...
+            ftp = ftplib.FTP(pdb_server.split("/", 1)[0])
         else:
-            ftp = ftplib.FTP(self.pdb_server)
+            ftp = ftplib.FTP(pdb_server.split("/", 1)[0])
         ftp.login()  # anonymous
-
+        pdb_directory = pdb_server.split("/", 1)[1]
         if file_format.lower() == "mmcif":
-            ftp.cwd("/pub/pdb/data/assemblies/mmCIF/all/")
+            ftp.cwd(f"{pdb_directory}/data/assemblies/mmCIF/all/")
```
JoaoRodrigues, line 536: You can directly do `ftp.cwd(pdb_directory)` and then there's no need to attach it to the paths below.

JoaoRodrigues (changes requested):
We're getting there!

I think the biggest change I'd make is to handle all connections through `PDBServer`. In `PDBList.__init__`, I'd add a method to read in the `server` variable and convert it, if possible, to a `PDBServer` instance. This would involve parsing the scheme, base_url, and path from the string (which should be trivial). If there is no scheme, default to https as that's more widespread. Then you can create a couple of methods to return full paths to urls from the class, e.g.:

```python
def build_url(self, url):
    return urljoin(self.scheme, self.base_url, self.path, self.url)
```

This would make the `PDBServer` class a little more useful and remove a lot building/appending of paths throughout the code making it a lot simpler.

Let me know if you need some clarification on any of the points I raised. Thanks a lot for your contribution, it's really cleaning up the module and making it much nicer!

Bio/PDB/PDBList.py
JulianMaurin: :+1: (minus vocabulary)

Bio/PDB/PDBList.py
JulianMaurin: I do not understand your concern here, the feature is ISO with the previous version.

JulianMaurin (commented):
> I think the biggest change I'd make is to handle all connections through PDBServer.

This PR has already introduced a bug fix and a new feature.
IMHO, it will be more interesting to add capabilities to PDBServer as we refactor rather than making a big bang here.

Bio/PDB/PDBList.py
JoaoRodrigues: `domain_url` and `pdb_dir_path` look like a good compromise!

Bio/PDB/PDBList.py
JoaoRodrigues: That's not entirely accurate. If someone creates their own `PDBServer` dataclass, they'll hit this line.

I'd really go for the scheme being part of the class and, as I mentioned in my general comment, slightly refactor so that we always create a PDBServer instance at this point. It's more self-consistent.

Bio/PDB/PDBList.py
JoaoRodrigues: Why the redudant `server argument server: {server}` and not just `server argument :{server}`?

Bio/PDB/PDBList.py
JoaoRodrigues: Yes, but if you make the `urlunsplit` part a method of PDBServer (e.g. `self.get_pdb_url()`) you can reduce this to:
```python
        if server is None:
            server = get_fastest_server()  # assuming this returns a PDBServer instance
            return server.get_pdb_url()
        elif isinstance(server, str):
            return urljoin(server, "pub/pdb/")
        elif isinstance(server, PDBServer):
            return server.get_pdb_url()

    raise TypeError(....)
```

Although the point is moot if we simply convert a `server: str` to a `server: PDBServer` from the get go. Much simpler code here then: simply `PDBServer.get_pdb_url()`.

Bio/PDB/PDBList.py
JoaoRodrigues: I got confused sorry. This line is redundant though, since you redefine it in L529. You can just use `self.pdb_server` throughout or use `pdb_server` (to avoid the self call).

Bio/PDB/PDBList.py
JulianMaurin: Sorry to insist, but the domain is a subpart of the URL. I do not feel comfortable to do this confusion on purpose (cf. https://en.wikipedia.org/wiki/URL)

And why not `pdb_dir_path`, if you think we gain in readability

Bio/PDB/PDBList.py
JulianMaurin: It is a good practice that I have taken over the years, formatting log and error message as `Message (vars).`. 
In this specific and isolated case, it's hard to see the benefits, but when we have a (lot) of long text with tones of vars inside, it makes a big difference. This way all messages are uniform and readable and the log search is much easier.

But I'm not going to fight for this one, if you insist I can implement your suggestion as is.

Bio/PDB/PDBList.py
JulianMaurin: If of course, when someone edit the source code the software is not working as expected.

I think I see where you want to go at the end, but I think we already brought a lot of improvement in this PR, and I strongly think we must iterate step by step. 

Regarding the current implementation, we did not

Bio/PDB/PDBList.py
JoaoRodrigues: OK, let's keep `domain` and `pdb_dir_path`, compromise :)

Bio/PDB/PDBList.py
JoaoRodrigues: Biopython is often used as a library, so someone could look at your nice `PDBServer` dataclass and say "awesome, I'll write one for my own custom server" and then they run into trouble. I agree that there are a few changes here already, but since you're introducing the `PDBServer` class here, we can make it more extensible from the get-go so that when we move on to a more in-depth refactoring/rethinking of this module, we have less work :)

I trust you! I just want to make sure we're on the same page here before committing changes. If you truly feel this is out of scope for this PR, then I accept your argument and we can move on.

JulianMaurin: I was feeling like the conversation was stuck, obviously, I was wrong. Thank you for your effort.

Yes, the objective is to move from `PDBList.pdb_server: str` to `PDBList.pdb_server: PDBServer` in order to enjoy the future method of the dataclass. But I think it deserves a dedicated PR, I took interested in this work (as you noticed) and I would like to continue in an iterative way.

Bio/PDB/PDBList.py
JulianMaurin: 6baa32e22caa0b4f38be7df76df7fea5111b0ad8 :rocket:

Bio/PDB/PDBList.py
JoaoRodrigues: Alright, let's iterate then!

Bio/PDB/PDBList.py
JoaoRodrigues: I'd insist on this one, it's adding verbosity without adding any more information: `"Unexpected value for server argument: {server}"` is informative and concise.

JulianMaurin:
```suggestion
        raise TypeError(f"Unsupported server option: {server}")
```

JulianMaurin:
Hello @JoaoRodrigues :wave:
The next is ready (see: https://github.com/JulianMaurin/biopython/pull/3), if you think we can move forward 🚀

JulianMaurin:
Hello @JoaoRodrigues 
I hope you are well and that the summer went well! 

Are you okay to get back to this topic? :)

PR #4017, opened 2022-08-02
Pre-commit hook to prevent addition of assert in the production code

JulianMaurin:
Hello :wave: 

I would like to help with the issue https://github.com/biopython/biopython/issues/3600.

From my experience, I think that the best way to get rid of a piece of code is to start by preventing the addition of any new occurrence of the last.

With this in mind, a local pre-commit hook using `pygrep` is perfectly doing the job.

The negative point is the length of our exclusion list, but we can now refer to it to keep the track of the work to do.
ps: If the change is accepted I will write a comment on the issue thread to explain it.

edit: 
Running bandit `B101`

```
vulnerability check......................................................Failed
- hook id: bandit
- exit code: 1

[main]	INFO	profile include tests: None
[main]	INFO	profile exclude tests: None
[main]	INFO	cli include tests: B101
[main]	INFO	cli exclude tests: None
[main]	INFO	running on Python 3.7.13
Run started:2022-08-06 20:51:12.828079

Test results:
>> Issue: [B101:assert_used] Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
   Severity: Low   Confidence: High
   CWE: CWE-703 (https://cwe.mitre.org/data/definitions/703.html)
   Location: Bio/Align/__init__.py:997:12
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b101_assert_used.html
996	        for line in lines:
997	            assert m == len(line)
998	        path = []

```

_____

already or do not wish to be listed. (*This acknowledgement is optional.*)


peterjc:
Brute force, but it would work - was there really not a  suitable flake8 plugin?

My feeling is I'd be much happier about adding this is the exclude list was considerably shorter. i.e. Let's fix a bunch of these first? e.g. SeqIO, SearchIO and AlignIO are three logically linked batches which would really cut down that list of exceptions.

JulianMaurin:
Hello @peterjc, 
Thanks for the review. 

> Brute force, but it would work - was there really not a suitable flake8 plugin?

There is a checking rule on [bandit](https://github.com/PyCQA/bandit), `B101`, which is doing the job.
The negative point is the longer creation time of the virtual pre-commit environment, but the implementation is more reliable than the previous `pygrep`. I could notice false positives after a deep look.

Hook updated: ee3174942ff8a9b9629f8033a3a59ad92f4a2b58

> My feeling is I'd be much happier about adding this is the exclude list was considerably shorter.

Ok yes I understand, it was "a bit much". Removing the false positives and limiting the exclude pattern to the top module significantly reduces the list, are you okay with this one?

List updated: 10fca8b8501eb402b43a0f0818c858fc158eef86

> Let's fix a bunch of these first? 

I would like to keep this PR "single purpose". Making these updates may lead to discussions on a specific technical scope, which I would like to avoid here.

ps: If the "bandit" option takes shape, it could be interesting to open a permanent issue to invite people to introduce more bandit checks on the codebase (on the model of https://github.com/python-poetry/poetry/issues/3155). Some of the checks may be worth it ([B108](https://bandit.readthedocs.io/en/latest/plugins/b108_hardcoded_tmp_directory.html), [B110](https://bandit.readthedocs.io/en/latest/plugins/b110_try_except_pass.html), ...)

peterjc:
If it were not for the long exclude list, invoking bandit via https://pypi.org/project/flake8-bandit/ might be elegant. I doubt it would be much slower, but this is a concern with #3958 (if we continue to use that service).

That would let us use the flake8 ``# noqaL B101`` comments to skip specific issues - like the try/except pass on https://bandit.readthedocs.io/en/latest/plugins/b110_try_except_pass.html which as you note could be useful.

But I see bandit has its own ``#nosec B101`` notation, https://bandit.readthedocs.io/en/latest/config.html#exclusions - so limited benefit.

(And yes, actually fixing the asserts would not be appropriate on this pull request)

peterjc:
And yes, much as we had a help-wanted first-issue friendly tracking issue for flake8 violations, we could do the same with some of the bandit ones - if people think using bandit is a good idea overall.

JulianMaurin:
@peterjc , thanks for this review. 

So to recap: 
The exclusion list is okay as it is?
We have to swap from `bandit` to `flake8-bandit`? I am not convinced by this one, it seems to me to be an additional layer that brings nothing but possible bugs.
You would like to start fixing assert on this PR? Not convinced either, but why not if you think it's the right thing to do. In this case, can you list the files that you think are good candidates among the list please?

peterjc:
I was noting the existence of ``flake8-bandit``, and would have advocated using it if we were still using flake8 as our sole git pre-commit hook - but having moved to the pre-commit.com tool, the flake8 plugin adds very little. The only advantage is the consistency in using ``# noqa ...`` rather than ``# noseq ...`` in the code.

(A little history, we used to used ``flake8-black`` in order to check black style via flake8, but again we now call black directly from the pre-commit.com tool instead.)

I am not advocating fixing asserts on this PR, but on separate PRs as suggested on #3600. e.g. ``Bio/Align/exonerate.py`` would be a good first choice.

peterjc:
Just to note that the ruff linter (which Biopython is now using) can enforce the no-asserts rule from bandit.

PR #4029, opened 2022-08-10
Added quiet/verbose option to Bio.PDB.PSEA and updated tests

### LINKED ISSUE #3999: Add option to silence/capture stdout/stderr from Bio.PDB.PSEA
Spun from #3980.

Running `p-sea` through `Bio.PDB.PSEA` will always print output to standard out. We should refactor the code to use `subprocess.run` and add a keyword to the top-level `run_psea` function to control whether to print or capture the tool's output.

A mock implementation that adds some better error handling would be:

```import subprocess

def run_psea(pname, verbose=True):
    cmd = [...]
    try:
        p = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
    except subprocess.CalledProcessError as err:
        print(f"Command {cmd} failed")
        raise

    if verbose:
        print(p.stdout)
    ...
```

sean-workman:

Closes #3999 

I've added a verbose option to the `psea_run()` function in `Bio.PDB.PSEA` and set the default to False since I don't think most people would actually want the stdout to be printed. If you think that this might interfere with the existing code of some users I can switch the default.

We aren't able to use the try/except pattern for error handling originally suggested by @JoaoRodrigues in the issue because p-sea is coded to give an exit-status of 1 even when the program runs and outputs the `.sea` file correctly. Instead the function now checks that the stderr is empty and that the correct `.sea` output exists. I'm not sure if the `raise RuntimeError` is correct here, definitely looking for feedback on that.

The tests have been updated to capture the stderr and check whether or not it is empty with verbose set to True or False. This gets the job done and the tests work as expected, but I'm wondering if there is a cleaner approach to capturing the stdout.

### DIFF
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -17,7 +17,7 @@ ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
 """
 
 import subprocess
-from os.path import exists
+import os
 
 from Bio.PDB.Polypeptide import is_aa
 
@@ -38,12 +38,12 @@ def run_psea(fname, verbose=False):
     base = last.split(".")[0]
     cmd = ["psea", fname]
 
-    p = subprocess.run(cmd, capture_output=True)
+    p = subprocess.run(cmd, capture_output=True, universal_newlines=True)
 
     if verbose:
-        print(p.stdout.decode("utf-8"))
+        print(p.stdout)
 
-    if not p.stderr.decode("utf-8") and exists(base + ".sea"):
+    if not p.stderr.strip() and os.path.exists(base + ".sea"):
         return base + ".sea"
     else:
         raise RuntimeError("stderr not empty")
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -46,7 +46,7 @@ def run_psea(fname, verbose=False):
     if not p.stderr.strip() and os.path.exists(base + ".sea"):
         return base + ".sea"
     else:
-        raise RuntimeError("stderr not empty")
+        raise RuntimeError(f"Error running p-sea: {p.stderr}")
 
 
 def psea(pname):
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -36,8 +36,8 @@ def run_psea(fname, verbose=False):
     """
     last = fname.split("/")[-1]
     base = last.split(".")[0]
-
     cmd = ["psea", fname]
+
     p = subprocess.run(cmd, capture_output=True)
 
     if verbose:
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -17,11 +17,12 @@ ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
 """
 
 import subprocess
+from os.path import exists
 
 from Bio.PDB.Polypeptide import is_aa
 
 
-def run_psea(fname):
+def run_psea(fname, verbose=False):
     """Run PSEA and return output filename.
 
     Note that this assumes the P-SEA binary is called "psea" and that it is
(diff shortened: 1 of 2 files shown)


Bio/PDB/PSEA.py
```diff
@@ -30,12 +31,22 @@ def run_psea(fname):
...
+    Note that P-SEA will not write output to the terminal while run unless
+     verbose is set to True.
     """
-    subprocess.call(["psea", fname])
     last = fname.split("/")[-1]
     base = last.split(".")[0]
-    return base + ".sea"
+    cmd = ["psea", fname]
+
+    p = subprocess.run(cmd, capture_output=True)
```
JoaoRodrigues, line 41: You can get around using these `decode` calls by using `universal_newlines=True` or `text=True` (the latter is an alias for the format introduced for clarity but is only avaiable in Py 3.7 and onwards). Then, `p.stdout` and `p.stderr` will not be byte strings anymore.

Bio/PDB/PSEA.py
```diff
@@ -17,11 +17,12 @@
 """
 
 import subprocess
+from os.path import exists
```
JoaoRodrigues, line 20: I'm personally not a big fan of imports like this because it's hard for people to know what `exists` refers to. Is it a custom function of ours, or standard library? Unless it's something like `from module.submodule.package import class` (where the call would be super long).

Bio/PDB/PSEA.py
```diff
@@ -30,12 +31,22 @@ def run_psea(fname):
...
     base = last.split(".")[0]
-    return base + ".sea"
+    cmd = ["psea", fname]
+
+    p = subprocess.run(cmd, capture_output=True)
+
+    if verbose:
+        print(p.stdout.decode("utf-8"))
+
+    if not p.stderr.decode("utf-8") and exists(base + ".sea"):
```
JoaoRodrigues, line 46: Just for (paranoid) safety, I'd do `if not p.stderr.strip()`, in case there is some straggler whitespace in there.

Tests/test_PDB_PSEA.py
```diff
@@ -36,9 +38,21 @@ class TestPDBPSEA(unittest.TestCase):
...
+    def test_run_psea_verbose(self):
+        captured_ouput = io.StringIO()
+        sys.stdout = captured_ouput
+        psae_run = run_psea("PDB/1A8O.pdb", verbose=True)
+        sys.stdout = sys.__stdout__
         self.assertEqual(psae_run, "1A8O.sea")
+        self.assertTrue(captured_ouput.getvalue())
+
+    def test_run_psea_quiet(self):
+        captured_ouput = io.StringIO()
```
JoaoRodrigues, line 50: Neat!

JoaoRodrigues (approved):

Bio/PDB/PSEA.py
sean-workman: Very nice, I should have looked at the docs more! I've used the `universal_newlines` on the off chance that someone is trying to use <3.7, though I know the docs say that it isn't supported.

Bio/PDB/PSEA.py
sean-workman: Good point, I've just imported os now and done the full call of `os.path.exist()`.

Bio/PDB/PSEA.py
```diff
@@ -30,12 +31,22 @@ def run_psea(fname):
...
+
+    p = subprocess.run(cmd, capture_output=True, universal_newlines=True)
+
+    if verbose:
+        print(p.stdout)
+
+    if not p.stderr.strip() and os.path.exists(base + ".sea"):
+        return base + ".sea"
+    else:
+        raise RuntimeError("stderr not empty")
```
JoaoRodrigues, line 49: Sorry, I missed this. We should return the message here e.g. `raise RuntimeError(f"Error running p-sea: {p.stderr}")`

JoaoRodrigues (approved):

Bio/PDB/PSEA.py
sean-workman: That makes sense. My initial message was pretty useless, I now realize.

PR #4032, opened 2022-08-11
First pass at updating argument handling in PDBList

sean-workman:

This is a first pass at updating the argument parsing in Bio.PDB.PDBList, which looks to address #3989.

I think there are probably more sophisticated ways to implement the parsing with argparse but I'm pretty happy with the behaviour that I've tested thus far. While this doesn't really cut down on the number of lines after `if __name__ == "__main__"`, it does make things tidier and rely less on splitting/checking values in `sys.argv`, as well as allowing users to see usage information with the -h flag rather than when the script runs.

I'm looking for feedback on how I might improve what I've put together here, or if I should go back to the drawing board and figure something out more streamlined.

Thanks! :)

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -36,6 +36,7 @@
 """Access the PDB over the internet (e.g. to download structures)."""
 
 
+import argparse
 import contextlib
 import ftplib
 import gzip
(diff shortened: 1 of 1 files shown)


JulianMaurin:
Hello @sean-workman 👋 

First of all, well done. I'll checkout locally to look at this in more detail, but it looks nice.
The main thing I noticed is about the single/multi options. I think we can easily refactor it to rely on the same code (in the end, single is just multiple with only one value).

I would like to try to challenge your implementation to achieve the best possible result.
To do this I have some questions for the maintainers to have a better understanding of the expected result.

- What is the biopython policy about dependencies? Could we imagine relying on external lib like [typer](https://typer.tiangolo.com/)? (regarding [this comment](https://github.com/biopython/biopython/blob/master/setup.py#L108-L112) I guess there is probably some work to do so, but that would not be a big deal).

- Do we want to be fully back-forward compatible / to have both versions living at the same time (with a deprecation warning) / or to have a fresh start?

JoaoRodrigues:
- As for dependencies, we usually avoid third-party dependencies as much as possible unless they are absolutely essential for functionality.
- We might want to move to a new module for this PR. This will likely be better made nicely from scratch!

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,166 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    elif args.assemb:
+        # Get all assembly structures
+        pl.download_all_assemblies(file_format=file_format)
+
+    elif args.single:
+        pdb_code = args.single
+        # Get single PDB entry
+        pl.retrieve_pdb_file(
+            pdb_code, pdir=pdb_path, file_format=file_format, overwrite=overwrite
+        )
```
JulianMaurin, line 835: What do you think about reworking this method to:
`retrieve_pdb_file(pdb_code: list[str], pdir: str, file_format: str, overwrite: bool, with_assemblies: bool)`

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,166 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    )
+
+    # Parse arguments
+    args = parser.parse_args()
+
+    # Set path
+    pdb_path = args.pdb_path
+
+    # Instantiate PDBList class
+    pl = PDBList(pdb=pdb_path)
```
JulianMaurin, line 791: What about a (or multiple) new optional arg to set the server?

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,166 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    mode.add_argument(
+        "-multi",
+        help="retrieve a set of structures - accepts a comma separated list of PDB codes",
+    )
+
+    # Add mututally exclusive arguments for format
+    format = parser.add_mutually_exclusive_group()
+    format.add_argument(
+        "-pdb", action="store_true", help="downloads structures in PDB format"
+    )
```
JulianMaurin, line 760: I think we can swap to:
```suggestion
    format.add_argument(
        "-pdb", action='store_const', const="pdb", dest="format", help="downloads structures in PDB format"
    )
```

to avoid the if/else L794

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,166 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter, description=desc
+    )
+
+    # Add positional argument for pdb_path
+    parser.add_argument(
+        "pdb_path",
+        nargs="?",
+        default=os.getcwd(),
+        help="local path to use, defaults to the current working directory",
```
JulianMaurin, line 725: Nitpick: we can follow the grammar, question of style, the important is to be consistent 
```suggestion
        help="Local path to use, defaults to the current working directory.",
```

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,166 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+        "-o", action="store_true", help="overwrite existing structure files"
+    )
+    others.add_argument(
+        "-with-assemblies",
+        action="store_true",
+        help="downloads assemblies along with regular entries",
+    )
+
+    # Parse arguments
+    args = parser.parse_args()
```
JulianMaurin, line 785: I like to have the help when I do not use a CLI correctly 
```suggestion
    try:
        args = parser.parse_args()
    except:
        args.print_help()
        sys.exit(0)
```
question of style one more time I guess 😄

JulianMaurin:
Some testing would be cool, if you want to get inspiration from https://jugmac00.github.io/blog/testing-argparse-applications-the-better-way 

(nitpick, but I think naming the func `cli` is better than `main` 😄)

sean-workman:
Hey @JulianMaurin, thanks for the suggestions! I had refactoring things so that we could have a single argument for intaking PDB codes pretty high on my list of things to do, as it felt a bit silly to be repeating code there at the end. I'll jump on that and work through your other suggestions sometime tomorrow probably!

Bio/PDB/PDBList.py
sean-workman: I've updated the handling of a  `-codes` argument at the end so that it iterates through a list rather than reworking the method itself. Maybe changing the method itself will come in the new module that @JoaoRodrigues has mentioned.

Bio/PDB/PDBList.py
sean-workman: I think this is probably linked to #4012 and maybe needs to wait until that is resolved? Could certainly add more optional arguments but I don't want to alter the class in this PR.

Bio/PDB/PDBList.py
sean-workman: This was an excellent idea, but I instead just included an optional `-format` argument with the choice pdb, xml, or mmtf with a default value of mmCif.

Bio/PDB/PDBList.py
sean-workman: I generally like the look of all lower case in the help messages that I get and this was consistent with the style of the original docstring that currently prints when PDBList.py is called directly from the CLI.

Bio/PDB/PDBList.py
sean-workman: I think including something like this is great, but as it is currently written it doesn't work because if `parser.parse_args()" fails there is no `args` object to tap into. I would also want to make sure we avoid the bare except if we can.

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,134 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+
+    elif args.assemb:
+        # Get all assembly structures
+        pl.download_all_assemblies(file_format=file_format)
+
+    elif args.codes:
+        if "," in args.codes:
+            pdb_ids = args.codes.split(",")
+        else:
+            pdb_ids = [args.codes]
```
JulianMaurin, line 819:
```suggestion
        pdb_ids = args.codes.split(",")
```

will do the same :v:

Bio/PDB/PDBList.py
JulianMaurin: yes, I mean `parser.print_help()`, my bad

Bio/PDB/PDBList.py
JulianMaurin: :+1:

Bio/PDB/PDBList.py
JulianMaurin: not necessarily, but as you feel :)

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,134 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    )
+    mode.add_argument(
+        "-codes",
+        metavar="",
+        help="retrieve structure(s) - accepts a single PDB code or a comma separated list of PDB codes",
+    )
+
+    # Add optional format argument that select choice
+    parser.add_argument(
+        "-format",
```
JulianMaurin, line 756: ah, so you are finally introducing a breaking change? I didn't really understand if it was desired, to check

Bio/PDB/PDBList.py
```diff
@@ -701,88 +702,134 @@ def get_seqres_file(self, savefile="pdb_seqres.txt"):
...
+    elif args.assemb:
+        # Get all assembly structures
+        pl.download_all_assemblies(file_format=file_format)
+
+    elif args.codes:
+        if "," in args.codes:
+            pdb_ids = args.codes.split(",")
+        else:
+            pdb_ids = [args.codes]
+        for pdb_id in pdb_ids:
```
JulianMaurin, line 820: I looked in the code and there is obviously the use of `pdb_code` but also `pdb_id`.
IMHO, we should stick to one name to represent a concept (see: https://martinfowler.com/bliki/UbiquitousLanguage.html).

Bio/PDB/PDBList.py
sean-workman: If this is moving to a new module I think it's fine to introduce a breaking change in how the function is called. If there is still a plan to include `argparse` in `PDBList` I can use the `store_const`/`dest` solution you suggested.

Bio/PDB/PDBList.py
sean-workman: So weird, I must have written something funny that was iterating over the string before. Introduced the check because it was splitting my single PDB code into individual characters. Whoops haha

Bio/PDB/PDBList.py
sean-workman: Agreed and I've changed that now. I think that was a carryover from the messy argument handling before.

Bio/PDB/PDBList.py
sean-workman: Still not entirely sure if this is necessary. The default behaviour when the `.parse_args()` fails is to print the usage message along with what error occurred. I think if the user wants to see the full help output they can call it themselves, it just seems a bit verbose to spit it all out in one go.

If there is consensus that this is behaviour that we want I will just need to stop `argparse` from printing the usage/error message because it's giving redundant information right now.

e.g.

```
❯ python PDBList.py -updaaa
usage: PDBList.py [-h] (-update | -all | -obsol | -assemb | -codes ) [-format {pdb,xml,mmtf}] [-d] [-o] [-with-assemblies] [pdb_path]
PDBList.py: error: one of the arguments -update -all -obsol -assemb -codes is required
usage: PDBList.py [-h] (-update | -all | -obsol | -assemb | -codes ) [-format {pdb,xml,mmtf}] [-d] [-o] [-with-assemblies] [pdb_path]

    (c) Kristian Rother 2003, Wiktoria Karwicka & Jacek Smietanski 2016
    Contributed to Biopython
	...
```

PR #4034, opened 2022-08-12
Updates Bio.PDB.PSEA to run third-party tool in temporary directory

### LINKED ISSUE #4031: Bio.PDB.PSEA should use a temporary directory to run the third-party tool
Spun-off from #3999. `psea` should run in a temporary directory to avoid leaving unnecessary output files. For example:

```
def run_psea(...):
    curdir = os.path.abspath(os.curdir)
    with tempfile.TemporaryDirectory() as tmpdir:
        os.chdir(tmpdir)
        .... # run psea
        os.chdir(curdir)  # go back to where we started
```

sean-workman:

Closes #4031 

I've made changes to `run_psea()` so that it calls the `psea` program in a temporary directory that leaves no `[....].sea` files behind and returns the string output of `psea` rather than the `[....].sea` file name instead. To account for this, I changed the `psea()` function itself so that it takes in and parses the string output that is now returned from `run_psea()`.

I also altered the error message that is raised when something goes wrong from my initial commits because I found that when there is an error, it actually prints to stdout for some reason... Some confirmation of this behaviour from someone else would be good.

Finally, I had to comment out the unittest assertions for `run_psea()` that we currently have. I am looking for suggestions as currently I think we would need to return an f-string with a `os.path.abspath()` in there for the tests to run everywhere and I don't know if that's the best solution.

Thanks!

### DIFF
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -15,9 +15,9 @@ Comput Appl Biosci 1997 , 13:291-295
 
 ftp://ftp.lmcp.jussieu.fr/pub/sincris/software/protein/p-sea/
 """
-
-import subprocess
 import os
+import subprocess
+import tempfile
 
 from Bio.PDB.Polypeptide import is_aa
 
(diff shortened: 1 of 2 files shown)


sean-workman:
The tests now just run `psea` directly and compare its output (minus the first directory line) to the output of `run_psea()` (also minus the first directory line). Looks a bit messy and could be worth storing the strings being compared in variables for readability in the `assertEqual()` call, but it gets the job done for now.

PR #4073, opened 2022-09-07
Support multi-line qualifier values

v-raja:
Currently if a qualifier value is a multi-line string, SeqIO.write writes a genbank file in an incorrect format since it uses double-quotes for a multi-line qualifier value which isn't supported by the genbank file format. There are two ways to rectify this:
1. Use single quotes for qualifier values when writing a value which is a multi-line string
2. Break multi-line strings at their new-line character and use double-quotes as normal

This PR implements the second approach.

Closes #...

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -405,15 +405,21 @@ class _InsdcWriter(SequenceWriter):
             if len(line) <= self.MAX_WIDTH:
                 self.handle.write(line + "\n")
                 return
-            # Insert line break...
-            for index in range(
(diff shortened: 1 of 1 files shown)


peterjc:
I don't follow what the problem is from your description alone. Also this needs a test, which would probably explain it too. Probably best added to ``Tests/test_GenBank.py``?

peterjc:
Oh, those are rather pathological. I'd have to reread the spec, but I don't think it comments about this possibility.

peterjc:
Not sure why you're seeing errors about ``DBDRIVER``, used in the BioSQL tests (which include parsing GenBank files into a database).

PR #4120, opened 2022-10-07
switch to instance class contructor where usefull

jan-glx:

- [X] I have read the ``CONTRIBUTING.rst`` file, ~~have run ``pre-commit`` locally,~~ 
 and understand that continuous integration checks will be used to

This takes off the burden of having to override methods such as `__add__` of child classes such as `pydna.Dseqrecord` (See https://stackoverflow.com/a/14209708/1870254 , example line that would be made obsolete:        https://github.com/BjornFJohansson/pydna/blob/88dd388109428920092279bfb3b237c688318911/src/pydna/seqrecord.py#L642)

If this is judged as favorable, it might be good to check if there are other places (e.g. in `Seq.py`) where such a change would be useful.

### DIFF
--- a/Bio/SeqRecord.py
+++ b/Bio/SeqRecord.py
@@ -897,7 +897,7 @@ class SeqRecord:
         if not isinstance(other, SeqRecord):
             # Assume it is a string or a Seq.
             # Note can't transfer any per-letter-annotations
-            return SeqRecord(
+            return type(self)(
                 self.seq + other,
                 id=self.id,
                 name=self.name,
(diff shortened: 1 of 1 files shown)


mdehoon (approved):

peterjc:
This change evidently doesn't break our ``SeqRecord`` subclass for BioSQL, but that's the only example in our codebase. It really is more of interest to 3rd party code and libraries already subclassing us.

e.g. @kblin do you subclass the Biopython ``SeqRecord`` class?

kblin:
From a quick look at our source code, we wrap SeqRecord objects in some of our own, but we don't subclass

peterjc:
Thanks Kai - that's useful to know here.

peterjc:
Rebased to trigger a fresh CI run

peterjc:
Belatedly merging, thank you Jan.

Please let us know if you change you mind about appearing in the contributors listing.

PR #4139, opened 2022-10-20
raise and error if an invalid Newick unquoted label name is found while parsing the tree

### LINKED ISSUE #4134: Phylo.parse(file,"newick") does not throw error if name invalidly contains whitespace
### Setup

system as follows:

```
3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:52) 
[Clang 13.0.1 ]
CPython
macOS-12.6-x86_64-i386-64bit
1.79
```

### Expected behaviour

`Phylo.parse(file,"newick")` will raise an Exception when a tree name invalidly contains whitespace in the middle of a node name as specified in the specifications:
image
https://evolution.genetics.washington.edu/phylip/newicktree.html

### Actual behaviour

No Exception is raised. Instead, anything that follows the first blank is simply ignored. This is a problem, because I'm handwriting a Newick tree and occasionally forget to put a comma. In that case Phylo should complain. But it doesn't.

### Steps to reproduce

```python
from io import StringIO
from Bio import Phylo
newick_string = """(A:0.1,invalid name with whitespace:0.2,(C:0.3,D:0.4)E:0.5)F;"""
Phylo.draw_ascii(Phylo.read(StringIO(newick_string), "newick"))
```

Result:
```
  ______ A
 |
_|_____________ whitespace
 |
 |                                     _____________________ C
 |____________________________________|
                                      |_____________________________ D
```

Wrong because it should error.

amatria:

The issue is that the current regex expression that matches an unquoted label name returns several tokens for an invalid unquoted label. Therefore, instead of raising an error, the function ends up setting the node's name as the last token in the regex group that matched the unquoted label name expression.

The solution I have come up with is to return the entire unquoted label string with whitespaces, new lines, etc. included, and do the checking manually when _parse_tree encounters an unquoted label token.

Closes #4134


Tests/test_Phylo.py
```diff
@@ -91,26 +96,26 @@ def test_newick_read_scinot(self):
...
-        tree = Phylo.read(StringIO("((:1, B:-2):-5, (C:3, D:4):-2)"), "newick")
+        tree = Phylo.read(StringIO("((:1,B:-2):-5,(C:3,D:4):-2)"), "newick")
         distances = {-4.0: 1, -7.0: 1, 1: 1, 2: 1}
         for x in tree.get_terminals():
             entry = int(tree.distance(x))
             distances[entry] -= distances[entry]
             self.assertEqual(distances[entry], 0)
 
-        tree = Phylo.read(StringIO("((:\n1\n,\n B:-2):-5, (C:3, D:4):-2);"), "newick")
+        tree = Phylo.read(StringIO("((:\n1\n,\nB:-2):-5,(C:3,D:4):-2);"), "newick")
```
peterjc, line 118: You're removing the space following the new-line in this and the above similar examples.

Do you think the space was invalid, and ought not to have been accepted?

amatria: I am no Newick's tree notation expert so please bear with me :p.

As I far as I could understand, unquoted label names cannot contain any UNSAFE_CHARS, where the set of UNSAFE_CHARS is: `[' ', '\n', '\r', '\t', '\f', '\v']`. So, I assumed that any unquoted label name containing any of those chars should be flagged as an invalid name. That is why I decided to remove the whitespace in any unquoted label name, even if it is a leading whitespace. Moreover, all examples on Newick's trees that I have come across avoid using whitespaces at all costs. So, I further assumed that those trees should contain no unnecessary whitespaces.

If you don't agree with me, I can revert the changes I made to the tests. However, in such case I should update the logic that finds invalid chars in an unquoted label name string as well, since it would flag those examples as invalid names (the resulting name would be `•B`).

peterjc: You could be right - perhaps the parser was wrongly tolerating leading whitespace on the continuation lines. If so, an explicit test of that would be best.

@etal are you able to review this please?

amatria:
BUMP (@peterjc @etal)

etal:
The change seems reasonable, but I don't have a corpus of real-world Newick or Nexus files to test it against, so I'm afraid I can't be much help with the core decision to be made here.

PR #4170, opened 2022-11-18
Correct the number of flexibility scores

Ruibin-Liu:
For a sequence of length `L` and a window size `W`, the number of flexibility scores we can calculate is `L - W + 1`, not `L - W` as in the current code. For example, a sequence of length 9 should have 1 score, but it is '0' in the current version.

Closes #...

### DIFF
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -166,7 +166,7 @@ class ProteinAnalysis:
         weights = [0.25, 0.4375, 0.625, 0.8125, 1]
         scores = []
 
-        for i in range(self.length - window_size):
+        for i in range(self.length - window_size + 1):
             subsequence = self.sequence[i : i + window_size]
             score = 0.0
 


peterjc:
Assuming you are right, the test needs updating to match:

```
FAIL: test_flexibility (test_ProtParam.ProtParamTest)
Calculate protein flexibility.
  File "/home/circleci/project/biopython-1.81.dev0/Tests/test_ProtParam.py", line 146, in test_flexibility
    self.assertEqual(
AssertionError: 144 != 143 : Output length differs

```

Ruibin-Liu:
Okay, the test case is updated. I don't have access to the paper: http://dx.doi.org/10.1002/prot.340190207, so some one needs to finally verify it.

Ruibin-Liu:
@peterjc Thanks for the paper! My understanding was partially correct that there should be 1 score for a sequence length of 9 but the current implementation has another problem. The paper was using three neighbor correlated normalized flexibility parameters B_norm0, B_norm1, and B_norm2 in the sliding window average calculation, instead of  using only the averaged B_norm_avr in the current implementation. To follow the paper, we need to find out the 'rigidity' of the two neighbors for each residue using the definition of B_norm_avr and choose whether to use B_norm0, B_norm1, or B_norm2 based on how many rigid residues are surrounding the central residue. 

My two cents: the accuracy is low (correlation coefficient is <0.34) even with the correct implementation and thus we might instead need to find a newer method that has the best accuracy based on only sequence information. Also, is there a similar function in the structure based modules in biopython?

peterjc:
So maybe we should deprecate and remove this bit of code then?
https://biopython.org/wiki/Deprecation_policy

I don't recall anything else like this under ``Bio/SeqUtils/`` so yes, it might be time to look for a newer method if you need this functionality.

Ruibin-Liu:
I will take a look at the literature. If the method in the 1994 paper is still the best sequence-based flexibility predictor, I think we should keep it but still implement it correctly. A side question, do you know who implemented the current method? I am curious how the test data was generated/obtained. It seems to me the test was not about correctness of the implementation as to the paper but about whether any modification can pass the test itself.

peterjc:
You can look at the ``git blame`` view (within GitHub too) to work that out - you'll be going back decades I think though ;)

peterjc:
Original version from 19 years ago:

```python
	# Calculate the flexibility according to Vihinen, 1994.
	# No argument to change window size because parameters are specific for a window=9.	
	# the parameters used are optimized for determining the flexibility.
	def flexibility(self):
		Flex = ProtParamData.Flex.copy()
		Window=9
		Weights=[0.25,0.4375,0.625,0.8125,1]
		List=[]
		for i in range(self.length - Window):
			SubSeq=self.sequence[i:i+Window]
			score = 0.0
			for j in range(Window/2):
				score += (Flex[SubSeq[j]]+Flex[SubSeq[Window-j-1]]) * Weights[j]
			score += Flex[SubSeq[Window/2+1]]
			List.append(score/5.25)
		return List
```

Proposed version:

```python
    def flexibility(self):
        """Calculate the flexibility according to Vihinen, 1994.
        No argument to change window size because parameters are specific for
        a window=9. The parameters used are optimized for determining the
        flexibility.
        """
        flexibilities = ProtParamData.Flex
        window_size = 9
        weights = [0.25, 0.4375, 0.625, 0.8125, 1]
        scores = []

        for i in range(self.length - window_size + 1):  # <-- CHANGED LINE
            subsequence = self.sequence[i : i + window_size]
            score = 0.0

            for j in range(window_size // 2):
                front = subsequence[j]
                back = subsequence[window_size - j - 1]
                score += (flexibilities[front] + flexibilities[back]) * weights[j]

            middle = subsequence[window_size // 2 + 1]
            score += flexibilities[middle]

            scores.append(score / 5.25)

        return scores
```

To my eye, yours is the first actual change to the algorithm - everything else was style.

Ruibin-Liu:
The Vihinen 1994 paper followed the same algorithm with Karplus 1985 paper which clearly used the three B_norm values (the Karplus paper had a figure showing the Fortran code). If we want this method still usable and not misleading, I think we should implement the correct one at least.

I can do the coding and add the commits to this PR if you agree.

Ruibin-Liu:
According to the 1985 Karplus paper implementation, the neighbor-correlated B_norm values were only calculated for non-terminal residues, so the minimum sequence length when the window size is 9 is 11. The code is updated but the test case is not yet.

Bio/SeqUtils/ProtParam.py
```diff
@@ -161,25 +161,32 @@ def flexibility(self):
...
-        for i in range(self.length - window_size):
-            subsequence = self.sequence[i : i + window_size]
-            score = 0.0
+        # Calculate neighbor-correlated B_norm values first
+        for i, res in enumerate(self.sequence[1 : self.length - 1]):
+            bnorm_nc[i] = bnorm_0[res]
+            if bnorm[self.sequence[i]] < 1 or bnorm[self.sequence[i + 2]] < 1:
+                bnorm_nc[i] = bnorm_1[res]
+            if bnorm[self.sequence[i]] < 1 and bnorm[self.sequence[i + 2]] < 1:
+                bnorm_nc[i] = bnorm_2[res]
```
peterjc, line 179: If I am reading this right, what you have here is:

```python
if A or B:
   bnorm_nc[i] = bnorm_1[res]
if A and B:
   bnorm_nc[i] = bnorm_2[res]
```

This seems inefficient when both A and B are true, since the first if-statement and code do not need to be run. How about:

```python
if A and B:
   bnorm_nc[i] = bnorm_2[res]
elif A or B:
   bnorm_nc[i] = bnorm_1[res]
```

Ruibin-Liu: Yeah, these lines are not the most efficient. I was just following the original implementation. I will update it.

peterjc:
Ideally we'd need a review from someone using the protein functions, maybe you could take a look @sbassi?

peterjc:
@jgreener64 @JoaoRodrigues are you able to comment on protein flexibility scores?

jgreener64:
I'm not too familiar with this part of the code I'm afraid.

peterjc:
Thanks Joe. Guessing no names come to mind able to comment on protein flexibility scores?

peterjc:
Rebased and squashed to single commit - this can probably go into Biopython 1.85?

PR #4202, opened 2022-12-15
Fix MMCIFParser warning and IUPAC extended one letter codes

Truman-Xu:

Closes #...


peterjc:
What issue are you trying to fix? Those dictionaries were separate for a reason.

Truman-Xu:
But what is the harm in including those in the extended 3to1 dict?

peterjc:
There are PDB extensions to the core 3-letter protein names which are not IUPAC extensions.

peterjc:
There are PDB extensions to the core 3-letter protein names which are not IUPAC extensions.

Truman-Xu:
Sorry, I have trouble understanding this. Isn't the `protein_letters_3to1_extended` supposed to be everything but the 20 canonical residue names defined by IUPAC (i.e. everything but exclude `protein_letters_3to1`)? I thought `protein_letters_3to1_extended` would be a superset that also include the IUPAC defined extended residue names, and the current one does not include the reverse mapping from `protein_letters_1to3_extended` from IUPACData.py

Truman-Xu:
For example,
```python
from Bio.PDB.Polypeptide import protein_letters_3to1_extended
protein_letters_3to1_extended['SEC']   #SELENOCYSTEINE (U)
```
This would raise a KeyError

Truman-Xu:
Moreover, the 20 canonical residue names are indeed in the `protein_letters_3to1_extended` as expected

```python
from Bio.Data.PDBData import protein_letters_3to1
from Bio.Data.PDBData import protein_letters_3to1_extended

canonical_resnames = set(protein_letters_3to1.keys())
extended_resnames = set(protein_letters_3to1_extended.keys())
canonical_resnames.issubset(extended_resnames)
```
This would return `True`

JoaoRodrigues:
As Peter mentioned, these dictionaries refer to different sources. The dict in IUPACData refers to IUPAC nomenclature, while PDBData is a collection of data from structures deposited in the wwPDB. For example, `Xle` is a IUPAC aminoacid, but it does not match any monomer entry in the PDB.

The ones you did find in structures should be added to `PDBData`. IIRC, I updated the dictionary by parsing the `components-pub.cif` file made available by wwPDB, but I might have missed some indeed.

Truman-Xu:
Got it! Thank you for the explanations!

Bio/Data/PDBData.py
```diff
@@ -15,13 +15,21 @@
 # The 'fmt:off' lines prevent black for formatting the dictionaries.
 
 from Bio.Data.IUPACData import protein_letters_3to1 as _protein_letters_3to1
+from Bio.Data.IUPACData import (
```
JoaoRodrigues, line 18: Was this formatted by black?

Bio/PDB/MMCIFParser.py
```diff
@@ -215,7 +215,6 @@ def _build_structure(self, structure_id):
             except ValueError:
                 serial = atom_serial_list[i]
                 warnings.warn(
-                    "PDBConstructionWarning: "
```
JoaoRodrigues, line 218: This seems like an unrelated change, is it only for formatting? Does it avoid any repetition?

Bio/Data/PDBData.py
```diff
@@ -15,13 +15,21 @@
...
 protein_letters_3to1 = {k.upper(): v for k, v in _protein_letters_3to1.items()}
 protein_letters_1to3 = {v: k for k, v in protein_letters_3to1.items()}
+# Add the extended amino acid one letter codes from IUPACData.py
+protein_letters_3to1_extended_one_letter = {
+    k.upper(): v for k, v in _protein_letters_3to1_extended.items()
+}
 
 # fmt: off
 protein_letters_3to1_extended = {
+    **protein_letters_3to1_extended_one_letter,
```
JoaoRodrigues, line 32: Refer to my last reply in the general comment thread as to why I don't think this is necessarily appropriate. Some of the entries in `IUPACData` will _not_ be present in PDB files, and as such should not be included here. I'd rather this dictionary be updated manually rather than a bulk update.

JoaoRodrigues (changes requested):

Bio/PDB/MMCIFParser.py
peterjc: It would avoid some repetition in the default traceback message.

peterjc:
Closing as stale - but thank you for opening this, it looks like it did trigger several improvements to the code base.

Issue #4223, opened 2023-02-01
Add a continuous fuzz testing suite


silvergasp:
I've written a couple of fuzz tests and found a few (non-critical) bugs in some of the parsers. Before continuing any further is there any interest in accepting a PR with fuzz tests?

If you aren't familiar with fuzz testing, here is a bit of a run down ([from Wikipedia](https://en.wikipedia.org/wiki/Fuzzing));

> In [programming](https://en.wikipedia.org/wiki/Computer_programming) and [software development](https://en.wikipedia.org/wiki/Software_development), fuzzing or fuzz testing is an automated [software testing](https://en.wikipedia.org/wiki/Software_testing) technique that involves providing invalid, unexpected, or [random data](https://en.wikipedia.org/wiki/Random_data) as inputs to a [computer program](https://en.wikipedia.org/wiki/Computer_program). The program is then monitored for exceptions such as [crashes](https://en.wikipedia.org/wiki/Crash_(computing)), failing built-in code [assertions](https://en.wikipedia.org/wiki/Assertion_(software_development)), or potential [memory leaks](https://en.wikipedia.org/wiki/Memory_leak).

If this sounds like it'd be useful, as a follow-up step I'd like to suggest adding biopython to google/oss-fuzz. OSS-fuzz is Google's free open-source continuous fuzzing service. If biopython is integrated into oss-fuzz, the fuzz tests under biopython will be built and then run once a day, to search for bugs and vulnerabilities in biopython. This service can be integrated with the CI for biopython, so that the fuzz tests are run for 10min or so for every pull request, preventing buggy code from being merged.

peterjc:
This seems worth trying, although I think it would be most valuable for parsers for the well defined file formats. On the more ambiguous text formats I fear the fuzz testing could flag an unmanageable number of theoretical concerns. My hunch could be wrong though.

Exploring fuzz testing outside of CI would seem a sensible first step - there are likely more low hanging fruit (on top of the issues you alluded to) which you'd need to fix before using this within CI would be helpful in preventing new issues.

silvergasp:
> This seems worth trying, although I think it would be most valuable for parsers for the well defined file formats. On the more ambiguous text formats I fear the fuzz testing could flag an unmanageable number of theoretical concerns. My hunch could be wrong though.

[...]

PR #4224, opened 2023-02-01
feat(fuzzing): Add a fuzz test for the PDB parser

silvergasp:

This adds a coverage-guided fuzz test that attempts to find bugs in the PDB parser. See https://github.com/biopython/biopython/issues/4223 for a more general description of fuzz testing.


peterjc:
I've not clicked the approve-and-run as other than style testing, there would be no point right now - this doesn't (yet) install the dependency nor execute the new script on CI.

Guessing one of us would have to try it on our own machines...

Tests/fuzz_PDB_PDBParser.py
```diff
@@ -0,0 +1,37 @@
...
+    lines = [
+        fdp.ConsumeString(fdp.ConsumeIntInRange(0, max_line_length))
+        for _ in range(1, number_of_lines)
+    ]
+    parser = PDBParser()
+    parser.header = None
+    parser.trailer = None
+    # Make a StructureBuilder instance (pass id of structure as parameter)
+    parser.structure_builder.init_structure(fdp.ConsumeString(16))
+    parser._parse(lines)
```
silvergasp, line 28: Calling the internal `_parse` with our list of lines that was generated from the fuzz data. We can't call the public parsing API as it requires a path to a file, which is then read line by line. This reading from a file, stage isn't really well support in fuzz testing, and we generally want to avoid file IO in a fuzz test.

Tests/fuzz_PDB_PDBParser.py
```diff
@@ -0,0 +1,37 @@
...
+    parser.structure_builder.set_header(parser.header)
+    # Return the Structure instance
+    structure = parser.structure_builder.get_structure()
+    atoms = structure.get_atoms()
+    residues = structure.get_residues()
```
silvergasp, line 33: Just exercise some of the code to see how it reacts to the random input.

silvergasp (commented):
> I have to admit that I'm completely ignorant about fuzzy testing. Could you explain a little what this code does?

Sure :) Here's a simplified explanation of what a fuzz test harness does.
```mermaid
graph TD
A[Generate random data] --> B[System under test, PDB Parser]
B --> C[Measure code coverage]
C-->|High code coverage| D[Mutate origional input]
D --> B
C --> |Low code coverage| E[Discard input and generate newly random input]
E --> B
```
Code coverage is [just collected on the libraries that are in biopython](https://github.com/biopython/biopython/pull/4224/files#diff-80e8d34b63dfce0cb37bbdfa20e6cd4d1def79806ee78707071e1194e5933f26R8-R11). This ensures that the fuzzer [google/atheris](https://github.com/google/atheris) "focuses" on fuzzing only biopython and doesn't waste time trying to delve deeper into the other system libraries and dependencies.


> To include this, we'd have to somehow add also the dependency (atheris), preferrably on the dev/CI side and not user side.

I think with this fuzz harness there will be limited value in running it in the CI. This is simply because of the large volume of bugs (inputs that make the parser crash). This would be more useful as a local testing tool. Once the parser is a bit more robust against invalid/random inputs, I'd strongly recommend adding it to the CI :)

silvergasp:
> I've not clicked the approve-and-run as other than style testing, there would be no point right now - this doesn't (yet) install the dependency nor execute the new script on CI.
> 
> Guessing one of us would have to try it on our own machines...

Yeah here are some usage steps;
```
pip3 install atheris
python3 Tests/fuzz_PDB_PDBParser.py
```

The fuzzer will run forever **OR** until the parser crash's.

PR #4240, opened 2023-02-19
Fix DNA_NN4 thermodynamic table

vladkuz-vk:

According to SantaLucia & Hicks (2004), k should be calculated as **Ct / 4 * 1e-9**, where **Ct** is a total molar strand concentration (dnac1 + dnac2), it’s a more accurate way to calculate oligo and template concentrations. Moreover, there is an incorrect value in the thermodynamic table.
Found by me and Ricardas Ralys


MarkusPiotrowski:
@vladkuz-vk Thanks for your suggestions. I'll look up the thermodynamic table, it wouldn't be the first typo. Regarding the calculation of k, I however, disagree. I think Ct/4 is an oversimplification and is mostly valid if you do hybridiziations where both strands are present in more or less equal amounts. This is not the case for PCR, where the template is usually present in much lower amounts than the primer. Have you read the docstring for dnac1 and dnac2 in Tm_NN? You should show me a reference, where both versions are discussed and where the simpler formula was shown to be better.

vladkuz-vk:
@MarkusPiotrowski Yeah, I have read the docstring for dnac1 and dnac2. It seems to me a bit strange to divide Ct / 2 and assign each value for dnac1 (oligo) and dnac2 (template, for PCR equal 0). But if we set actual values (dnac1 = 50, dnac2 = 0), the result will differ. I see that dividing is the way to mimic Primer3 behavior.
From the biological meaning of the arguments, we should set dnac1 as the oligo concentration, and dnac2 as the template.
Also, there is an assumption in the Primer3:
> The value of this parameter (PRIMER_DNA_CONC) is less than the actual concentration of oligos in the initial reaction mix because it is the concentration of annealing oligos, which in turn depends on the amount of template (including PCR product) in a given cycle.

I'm confused where is the correct solution.

MarkusPiotrowski:
@vladkuz-vk:
(dnac1 + dnac2) / 4 is a special case of dnac1 - (dnac2 / 2) when dnac1 = dnac2

In the papers, in which the thermodynamic values for nearest neighbors were determined, they measured the melting behavior of short oligos which were present in equal amounts. So the simplification is valid in their case. 

What does this value Ct mean? It's the concentration of the free primer at Tm. At Tm, half of the possible dimers have formed, this is limited by the lower concentration of the template. So half of the template molecules (= dnac2/2) have bound to the primers (dnac1), so dnac1 - dnac2/2 primers are left.

vladkuz-vk:
@MarkusPiotrowski thank you for the explanation! Now I see.
I revert k formula and push only correct values in the DNA_NN4 thermodynamic table.

MarkusPiotrowski:
@peterjc Is `B028` "No explicit stacklevel keyword argument found." new and do we enforce it?

peterjc:
That's a new one https://github.com/PyCQA/flake8-bugbear#23213

https://github.com/PyCQA/flake8-bugbear says:

> B028: No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

Ignore?

MarkusPiotrowski:
I had the feeling that our warning messages are quite clear, so I wonder if we really need a higher stack level?

peterjc:
Yes, I'm inclined to see this new warning from bug-bear as excessive. They may well downgrade it in a future release (that's happened before).

MarkusPiotrowski:
@vladkuz-vk Can you update your local master and do a rebase from your dev branch?

vladkuz-vk:
@MarkusPiotrowski Done. Rebased and looks perfectly clean

MarkusPiotrowski:
@vladkuz-vk Thank you! The rebase was necessary, because we changed something in our style checks. 
One last thing, sorry, I didn't see this earlier: Can you please sort the authors' list in `NEWS.rst` alphabetically by first name, as in `CONTRIB.rst`? I know, it was already mixed before...

MarkusPiotrowski:
@vladkuz-vk Thank you!

PR #4249, opened 2023-03-01
Msa position mapping

### LINKED ISSUE #4248: `MultipleSeqAlignment` in `Bio.Align` should have a method for translating positions between sequences and the alignment
### Setup

```python
3.10.9 (main, Jan 11 2023, 09:18:20) [Clang 14.0.6 ]
CPython
macOS-10.16-x86_64-i386-64bit
1.82.dev0
```

### Expected behaviour

It would be very useful to have an inbuilt way for a `MultipleSeqAlignment` to calculate the positions in the *original* sequences (i.e. the original letter indexes, without the gap `-` characters); given a position in the `alignment` (or another sequence).  

As an example use case, I am aligning protein sequences from different PDB files and wish to compare homologous residues in 3D space.  I need to know the original position of a particular residue that has aligned to another, in order to index into the PDB files separately.  

```python
>>> from Bio import AlignIO
>>> align = AlignIO.read("Clustalw/opuntia.aln", "clustal")
>>> print(align)
Alignment with 3 rows and 9 columns
KTLK-E-ME Alpha
KTLK---ME Beta
KT-----LE Gamma
>>> align.map_position(3, "Gamma") # if no second sequence provided, returns the alignment column
8 
>>> align.map_position(pos=4, from_="Alpha", to_="Beta") # 'E' in Alpha is aligned to '-' in Beta; returns the previous position
3
>>> align.map_letter(2, "Gamma", "Alpha") # returns tuple of letter and position in Alpha; 'L' -> 'M'
('M', 6)
```
 
I've started making a PR for this, would love to have some feedback.  

I also think it would be good if you could supply an index slice (for either `pos`, the sequence `index`, or both) so that you could get several results back as a list.  So i'll work on that next. 

Thanks in advance! Cheers

kamurani:
- [] I hereby agree to dual licence this and any previous contributions under both

Closes #4248

### DIFF
--- a/Bio/Align/__init__.py
+++ b/Bio/Align/__init__.py
@@ -267,6 +267,8 @@ class MultipleSeqAlignment:
         If the sequence has a gap at the specified alignment position, the previous 
         non-gap position is returned. # TODO: should this just be -1 instead?
         """
+        if isinstance(pos, str):
+            raise TypeError("Invalid pos type: %r" % type(pos))
         if isinstance(index, int):
             # Assume it's a row number
             if index < 0:
@@ -284,6 +286,7 @@ class MultipleSeqAlignment:
         else:
             raise TypeError(f"Invalid index type: {type(index)}")
         
+        # TODO: if pos is slice
         if pos < 0:
             raise ValueError("Negative indexing not supported") 
             # TODO: use the alignment slicing method to get positions in reverse order; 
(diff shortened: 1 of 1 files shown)


Bio/Align/__init__.py
```diff
@@ -904,13 +1093,13 @@ def substitutions(self):
 
         Any weights associated with the sequences are taken into account when
         calculating the substitution matrix.  For example, given the following
-        multiple sequence alignment::
+        multiple sequence alignment:
```
peterjc, line 1096: Please revert this change, the double colon is RST markup indicating the following indented code is a literal block. Same for the change below.

peterjc:
I wonder if it would be best to more closely match the new alignment object's mapping via ``.indices`` and ``.inverse_indices``?

PR #4261, opened 2023-03-12
fix misleading comment in CompoundLocation.end documentation.

### LINKED ISSUE #4260: Misleading statement in CompoundLocation.end docstring
I believe the text in parentheses `(minus one given how Python counts from zero)` is incorrect, and should be deleted.

https://github.com/biopython/biopython/blob/c0be3a0b477f9cce38f26d91af538b2b07ca4720/Bio/SeqFeature.py#L1849

When I load a circular record where the first features spans the origin.

```python
>>> int(rec.features[0].location.end) == len(rec)
True
```

This also makes sense with python slicing, because to slice an array/SeqRecord of size 1, you would do `a[0:1]` to get the entire array/SeqRecord, not `a[0:0]`.

seanrjohnson:

Deleted a sentence in documentation.
Closes #4260

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -1846,8 +1846,7 @@ class CompoundLocation(Location):
         position.
 
         For the special case of a CompoundLocation wrapping the origin of
-        a circular genome this will match the genome length (minus one
-        given how Python counts from zero).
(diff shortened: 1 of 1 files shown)


peterjc:
Thank you!

PR #4273, opened 2023-04-01
Update calculation of secondary structure fraction

arpansahoo:

I believe there is a mistake in secondary_structure_fraction() method in ProtParam. "V, I, Y, F, W, L" and "E, M, A, L" are listed as the amino acids that tend to be in helix and sheet respectively, but it is actually the opposite.

I have fixed this mistake, and I also updated the amino acid list for each secondary structure based on information found in the scientific articles below:

For beta turn: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2142776/pdf/7756980.pdf (Table 2)
- Added D (aspartic acid) to the list

For alpha helix: https://www.nature.com/articles/srep38341
- Added K (lysine)

For beta sheet: https://www.nature.com/articles/362267a0 (Table 1)
- Added T (threonine)

### DIFF
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -323,7 +323,7 @@ class ProteinAnalysis:
         Amino acids in Turn: N, P, G, S, D.
         Amino acids in sheet: V, I, Y, F, W, L, T.
 
-        Note that, prior to v1.8.2, this method wrongly returned
+        Note that, prior to v1.82, this method wrongly returned
         (Sheet, Turn, Helix) while claiming to return (Helix, Turn, Sheet).
 
         Returns a tuple of three floats (Helix, Turn, Sheet).
--- a/Bio/SeqUtils/ProtParam.py
+++ b/Bio/SeqUtils/ProtParam.py
@@ -315,17 +315,17 @@ class ProteinAnalysis:
         Returns a list of the fraction of amino acids which tend
         to be in Helix, Turn or Sheet.
 
-        Amino acids in helix: V, I, Y, F, W, L.
-        Amino acids in Turn: N, P, G, S.
-        Amino acids in sheet: E, M, A, L.
+        Amino acids in helix: E, M, A, L, K.
+        Amino acids in Turn: N, P, G, S, D.
+        Amino acids in sheet: V, I, Y, F, W, L, T.
 
         Returns a tuple of three floats (Helix, Turn, Sheet).
         """
         aa_percentages = self.get_amino_acids_percent()
 
-        helix = sum(aa_percentages[r] for r in "VIYFWL")
-        turn = sum(aa_percentages[r] for r in "NPGS")
-        sheet = sum(aa_percentages[r] for r in "EMAL")
+        helix = sum(aa_percentages[r] for r in "EMALK")
+        turn = sum(aa_percentages[r] for r in "NPGSD")
+        sheet = sum(aa_percentages[r] for r in "VIYFWLT")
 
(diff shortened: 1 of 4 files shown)


arpansahoo:
@peterjc Could you take a look?

peterjc:
You appear to be right, meaning that for almost 20 years (since Biopython 1.24) rather than returning an estimate of (Helix, Turn, Sheet) it has actually been returning (Sheet, Turn, Helix):

https://github.com/biopython/biopython/blob/biopython-124/Bio/SeqUtils/ProtParam.py#L220

That's possible, but strange enough that I'd like a second opinion on this - ideally from someone currently doing structural biology. Maybe @jgreener64 or @JoaoRodrigues?

It might be less disruptive to change the documentation to say it returns (Sheet, Turn, Helix) but historically wrongly claimed to do the opposite order?

You would also need to update the tests too.

Either way, I think this needs a line or two in the NEWS file.

jgreener64:
From a quick look the PR seems correct. I don't work with these sequence tools much though.

jgreener64:
> I didn't want to change the return to `sheet, turn, helix` in case any software out there does something like `helix, turn, sheet = X.secondary_structure_fraction()`

I would agree with this approach.

NEWS.rst
```diff
@@ -18,12 +18,17 @@ iterators of the alignments to reset as well, which is bug prone. Instead,
 calling ``iter`` on a ``PairwiseAlignments`` object will now return itself. The
 iterator can be reset by calling the ``rewind`` method.
 
+Calling `secondary_structure_fraction` on a ``ProtParam.ProteinAnalysis``
+object historically returned (sheet, turn, helix) while claiming to return
+(helix, turn, sheet). This was fixed to correctly return (helix, turn, sheet).
+
```
peterjc, line 24: Add something like "... and the amino acids considered was revised as per recent literature."?

arpansahoo: Additionally, the amino acids considered were revised as per recent literature.

Bio/SeqUtils/ProtParam.py
```diff
@@ -315,17 +319,20 @@ def secondary_structure_fraction(self):
         Returns a list of the fraction of amino acids which tend
         to be in Helix, Turn or Sheet.
```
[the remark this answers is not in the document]
arpansahoo: I think that's a good idea. Would it be best to just add "according to Haimov and Srebnik, 2016; Hutchinson and Thornton, 1994; and Kim and Berg, 1993, respectively" at the end of "Returns a list of the fraction of amino acids which tend to be in Helix, Turn or Sheet"?

Bio/SeqUtils/ProtParam.py
```diff
@@ -313,19 +317,23 @@ def secondary_structure_fraction(self):
         """Calculate fraction of helix, turn and sheet.
 
         Returns a list of the fraction of amino acids which tend
-        to be in Helix, Turn or Sheet.
+        to be in Helix, Turn or Sheet, according to Haimov and Srebnik, 2016;
+        Hutchinson and Thornton, 1994; and Kim and Berg, 1993, respectively.
```
arpansahoo, line 321: Does this look good? @peterjc

peterjc:
Nice. Are you happy for me to squash-and-merge this into a single commit (simpler history)?

arpansahoo:
> Nice. Are you happy for me to squash-and-merge this into a single commit (simpler history)?

Sounds good to me

peterjc:
Merged, thank you!

PR #4293, opened 2023-04-29
Added MultipleSeqAlignment del method for removing sequences by index or slice

### LINKED ISSUE #4137: How to delete sequences from an alignment (Bio.Align.MultipleSeqAlignment)
### Setup

Is there a way to delete specific sequences from a Bio.Align.MultipleSeqAlignment object? It could be done either by its rec.id (ideal) or its index.

If there is not a _method_ for this, is it possible to do it 'manually' by creating a new object, e.g.,

```python
recs = recs[0:2] + recs[3:]
```

where I'm trying to remove the third (index 2) record? I know THIS doesn't work — it gives an error, **ValueError: When adding two alignments they must have the same length (i.e. same number or rows)**. But, is something like it possible?

Thanks very much!
Chase

siddhartasr10:

Closes #4137

Added support for deleting sequences from MultipleSeqAlignment objects, using the format requested in the issue. Added test that basically extends a basic test of an alignment of 3 sequences.

### DIFF
--- a/Tests/test_align.py
+++ b/Tests/test_align.py
@@ -176,6 +176,8 @@ class TestBasics(unittest.TestCase):
         alignment.append(SeqRecord(Seq(letters), id="mixed"))
         alignment.append(SeqRecord(Seq(letters.lower()), id="lower"))
         alignment.append(SeqRecord(Seq(letters.upper()), id="upper"))
+        alignment.append(SeqRecord(Seq(letters), id="duplicate"))
+        del alignment[3]
         self.assertEqual(alignment.get_alignment_length(), 26)
         self.assertEqual(len(alignment), 3)
         self.assertEqual(alignment[0].seq, letters)
--- a/Bio/Align/__init__.py
+++ b/Bio/Align/__init__.py
@@ -791,6 +791,13 @@ class MultipleSeqAlignment:
                     new.column_annotations[k] = v[col_index]
             return new
 
+    def __delitem__(self, index):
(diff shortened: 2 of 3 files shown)


siddhartasr10:
Let me know if have to change something, Im more than willing to help!

Tests/test_align.py
```diff
@@ -176,6 +176,8 @@ def test_basic_alignment(self):
         alignment.append(SeqRecord(Seq(letters), id="mixed"))
         alignment.append(SeqRecord(Seq(letters.lower()), id="lower"))
         alignment.append(SeqRecord(Seq(letters.upper()), id="upper"))
+        alignment.append(SeqRecord(Seq(letters), id="duplicate"))
+        del alignment[3]
```
peterjc, line 180: Maybe confirm the number of rows in the alignment after this?

siddhartasr10: I thought of it but didn't do it because the test already checks the number of rows and columns after the delete:
`181 self.assertEqual(alignment.get_alignment_length(), 26)` column check
`182 self.assertEqual(len(alignment), 3)`                                      row check

It does this to check for the appends before too, if its better I could put the test of the **del** method after it does those tests and check for the row length again just in case.

peterjc: Ah, that's fine then.

peterjc (approved):

peterjc:
Do you want your name in the NEWS file too (for the next release)?

siddhartasr10:
Nah its good

peterjc:
Merged, thank you.

PR #4319, opened 2023-06-04
Change set_angle to update related angles, fix errors in default internal_coordinate values, make_extended()

### LINKED ISSUE #4318: Coordinates update after setting chi angle on aromatic residue doesn't work for the whole ring
### Setup

system as follows:

3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:27:40) [GCC 11.3.0]
CPython
Linux-5.19.0-43-generic-x86_64-with-glibc2.35
1.81

### Expected behaviour

If you try to set a new chi angle on an aromatic residue, i.e. Histidine, chi2 I would expect that all atoms coordinates involved in the ring are updated accordingly to the new chi angle. 

### Actual behaviour

Just the coordinates of the atoms involved in the chi angle are updated, causing the ring to collapse or losing its aromatic structure.

### Steps to reproduce

```
def apply_transformation(residue, chain, transformation):
            angle = "chi" + str(2)

           # Set the new angle
            residue.internal_coord.set_angle(angle, transformation)

            # Update coordinates
            chain.internal_to_atom_coordinates(verbose=True)
    return residue
```

https://github.com/biopython/biopython/assets/25704885/3fd14d62-6629-4ff8-8577-491d361cadd3

rob-miller:

- [X ] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``

Closes #4318.

This PR changes the default behaviour of `IC_Residue.set_angle()` to modify overlapping angles by default, as most users would probably expect.   For example, with this change, setting Psi (N-CA-C-N) now updates the overlapping angle N-CA-C-O instead of generating a structure with clashing atoms.  The original behaviour can be restored by setting an optional parameter `overlap=False`.  Probably this was not noticed before because the tests look specifically at the changed angles and not the resulting coordinates; none of the previous tests were affected by this change.

This functionality was previously available only with the poorly named `bond_set()`, which is retained for compatibility.

Thanks to @niccolo93 for detailing the issue and highlighting the `bond_set()` workaround.

Other changes:

- fixes an unidentified issue in `read_PIC()` whereby the wrong default angles were selected in many cases.
- adds `IC_Chain.make_extended()` to change all psi and phi backbone angles of a chain to be those of an extended beta strand.  The default angles stored in `ic_data.py` generate the more popular alpha helix conformation in most cases.
- makes `bond_set()` and related `bond_rotate()` more robust
- more and better comments
- adds testing to check resulting coordinates for a default structure chain after `make_extended()` and sequentially modifying all chi angles

### DIFF
--- a/Bio/PDB/PICIO.py
+++ b/Bio/PDB/PICIO.py
@@ -305,6 +305,9 @@ def read_PIC(
         accpt = IC_Residue.accept_atoms
         if not all(ek[i].akl[atmNdx] in accpt for i in range(4)):
             return
+        dangle = float(dangle)
+        dangle = dangle if (dangle <= 180.0) else dangle - 360.0
+        dangle = dangle if (dangle >= -180.0) else dangle + 360.0
         da[ek] = float(dangle)
         sbcic.dihedra[ek] = ric.dihedra[ek] = d = Dihedron(ek)
         d.cic = sbcic
@@ -392,8 +395,6 @@ def read_PIC(
 
             if paKey in da:
                 angl = da[paKey] + dihedra_secondary_defaults[rdclass][1]
-                angl = angl if (angl <= 180.0) else angl - 360.0
-                angl = angl if (angl >= -180.0) else angl + 360.0
                 process_dihedron(
                     str(ek[0]),
                     str(ek[1]),
@@ -431,8 +432,6 @@ def read_PIC(
 
                 if paKey in da:
                     angl = da[paKey] + offset
-                    angl = angl if (angl <= 180.0) else angl - 360.0
-                    angl = angl if (angl >= -180.0) else angl + 360.0
                     process_dihedron(
                         str(ek[0]),
                         str(ek[1]),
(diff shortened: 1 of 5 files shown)


rob-miller:
#4312 add to 1.82 if possible :-)

The CI/test_linux issues look like something well past my code, please tell me if I need to do more work.

peterjc:
Agree, test failure unrelated (happening on the master branch too). Something  to do with ReportLab ``rlPyCairo``, looks somehow related to #4305 which fixed something similar.

Bio/PDB/PICIO.py
```diff
@@ -391,12 +391,15 @@ def default_dihedron(ek: List, ric: IC_Residue) -> None:
                 )
 
             if paKey in da:
+                angl = da[paKey] + dihedra_secondary_defaults[rdclass][1]
+                angl = angl if (angl <= 180.0) else angl - 360.0
+                angl = angl if (angl >= -180.0) else angl + 360.0
```
JoaoRodrigues, line 396: If you're using this operation often (I see it below at least once more), I'd make it a function.

JoaoRodrigues (approved):

JoaoRodrigues:
Looks good to me! Thanks @rob-miller! Happy to merge!

PR #4328, opened 2023-06-11
Fix __version__ value load to allow installation through poetry

### LINKED ISSUE #4279: Biopython is not installable as a git repository with Poetry
### Setup

system as follows:

```shell
❯ python -c 'import sys; print(sys.version)'
3.10.9 (main, Dec  7 2022, 01:12:00) [GCC 9.4.0]
```

```shell
❯ python -c 'import platform; print(platform.python_implementation()); print(platform.platform())'
CPython
Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
```

(This is an install issue, so certainly no install)
```shell
❯ python -c 'import Bio; print(Bio.__version__)'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'Bio'
```

### Expected behaviour

With the following in `pyproject.toml` under `[tool.poetry.dependencies]`:

```toml
biopython = { git = "https://github.com/biopython/biopython", rev="ca64586b1c6f6c8439c88e576fed091c923ad8f1" }
```

running `poetry lock` should update with the relevant biopython package.

### Actual behaviour

```shell
❯ poetry lock
Updating dependencies
Resolving dependencies... (0.0s)

Invalid version 'Undefined' on package biopython
```

### Steps to reproduce

Follow the above steps.

### Suggested Remediation

- Option A: Add a `setup.cfg` with version information as [here](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/)
- Option B: Move to [PEP 518 pyproject.toml](https://peps.python.org/pep-0518/)

joelibaceta:

Closes #4279

### DIFF
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -173,6 +173,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - João Vitor F Cavalcante <https://github.com/jvfe>
 - Joe Cora <https://github.com/JoeCora>
 - Joe Greener <https://github.com/jgreener64>
+- Joel Ibaceta <https://github.com/joelibaceta>
 - Joerg Schaarschmidt <https://github.com/schaarj>
 - Johann Visagie <wjv at domain cityip.co.za>
 - John Bradley <https://github.com/johnbradley>
--- a/setup.py
+++ b/setup.py
@@ -105,21 +105,6 @@ def can_import(module_name):
     except ImportError:
         return None
 
-
-def extract_variable_value(file_path, variable_name):
-    """Extract the value of a variable from a Python file."""
-    with open(file_path, "r") as file:
-        source_code = file.read()
-    tree = ast.parse(source_code)
-    for node in ast.walk(tree):
-        if isinstance(node, ast.Assign) and len(node.targets) == 1:
-            target = node.targets[0]
-            if isinstance(target, ast.Name) and target.id == variable_name:
-                value_node = node.value
-                value = ast.literal_eval(value_node)
-                return value
-
-
 # Using requirements.txt is preferred for an application
 # (and likely will pin specific version numbers), using
 # setup.py's install_requires is preferred for a library
@@ -215,10 +200,12 @@ EXTENSIONS = [
     Extension("Bio.SeqIO._twoBitIO", ["Bio/SeqIO/_twoBitIO.c"]),
 ]
 
-# We now define the Biopython version number in Bio/__init__.py
-# Here we can't use "import Bio" then "Bio.__version__" as that would
-# tell us the version of Biopython already installed (if any).
-__version__ = extract_variable_value("Bio/__init__.py", "__version__") or "undefined"
+def get_version():
+    for line in open("Bio/__init__.py"):
+        if line.startswith("__version__ = "):
+            return ast.literal_eval(line.split("=")[1].strip())
+    return "Undefined"
+__version__ = get_version()
 
 # We now load in our reStructuredText README.rst file to pass explicitly in the
 # metadata, since at time of writing PyPI did not do this for us.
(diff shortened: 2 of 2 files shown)


peterjc:
Clever!

I think you should define a default (the string undefined as before) in case the AST parsing fails.

(Sorry, I tried to comment earlier by email but it looks like GitHub was overloaded as it failed)

peterjc (dismissed):
Looks good to me. I'd have put the default in the ``extract_variable_value`` function but that's more style than anything else.

Anyone want to give a second opinion? @MarkusPiotrowski , @JoaoRodrigues or @mdehoon perhaps?

peterjc:
Out of interest, *why* does this work (parsing the AST) versus the existing simpler code using an eval?

joelibaceta:
@peterjc I tried to write explicit code according to the expected behavior (read a Python script and extract the value assigned to an especific variable). Personally, I believe it is important to distinguish between reading a plain text file and a file that adheres to a specific structure and hierarchy like a Python Script to avoid future errors. Using plain text reading may lead to errors if, for example, __version__ is mentioned in a comment before the assignment or if the script is refactored. This code aims to understand the file syntactically and identify the assigned variable's value, even when the file is modified in the future.

peterjc:
Yes, there are advantages to parsing the AST including security - but does this actually help with Poetry (issue #4279), and if so do you understand how/why?

joelibaceta:
@peterjc Poetry does not run the previous code to setup.py or load this file, its only evaluates the expression passed as version argument in `setup` function call https://github.com/biopython/biopython/blob/master/setup.py#L222  to solve Resolving dependencies, which immediately references this line to solve the expression https://github.com/biopython/biopython/blob/master/setup.py#L206  So, we have an error since this assignment is "undefined",  because nothing force to run this lines https://github.com/biopython/biopython/blob/master/setup.py#L207-L209

peterjc:
So this works by forcing poetry to run a function which parses the file?

Does this work (closer to the original version but avoiding ``eval``)?

```python
def get_version():
    for line in open("Bio/__init__.py"):
        if line.startswith("__version__ = "):
            # This will cope with a trailing comment,
            # expect 3 entries: module, expr, constant
            return ast.literal_eval(line[14:].strip())
    return "Undefined"
__version__ = get_version()
```

joelibaceta:
Yup that should works too

peterjc (commented):
Withdrawing PR approval pending further discussion (not sure I understand how/why this works yet for poetry to get a useful version string).

peterjc (approved):
Tested as described at https://github.com/biopython/biopython/issues/4279#issuecomment-1598153627

I personally prefer the version discussed in the comments (loop over the lines to find the ``__version__ = ...`` line rather than parse the entire file as AST), but am not overly bothered either way.

peterjc:
I propose to leave this open for a week for comments/objections, and if none merge it (with the expectation of including this before the 1.82 release but that does not seem critical given the nature of this issue being working with non-released code).

joelibaceta:
@peterjc I made the simplification of the method according to your suggestion, please review it and tell me what you think

peterjc (approved):
I've tested this approach locally, and it solves the original issue, and makes the version detection more robust.

I would propose to squash-and-merge this for a single commit in the history on the master branch.

peterjc:
Oh, it will need this minor issue fixed (from the flake8 checks):

```
setup.py:205:1: D103 Missing docstring in public function
```

joelibaceta (commented):
oops i forgot to include the docstring, solved in https://github.com/biopython/biopython/pull/4328/commits/f8962a12fdadf98ba42b3ea94dfa677ac7e431c3 @peterjc

peterjc:
Merged, thank you!

PR #4377, opened 2023-07-23
mypy pre-commit hook

speleo3:

Related to https://github.com/biopython/biopython/issues/2236

Revives work on pull request https://github.com/biopython/biopython/pull/3165 - But I removed the `mypyc` stuff from `setup.py` because it didn't work well with the doctests.

- Adds a `mypy` pre-commit hook and thus runs `mypy` in the CI pipeline.
- Adds type annotation to `Bio.SeqIO.QualityIO`, `Bio.SeqRecord` and related modules.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,6 +31,12 @@ repos:
             'flake8-rst-docstrings>=0.2.3',
             'pydocstyle>=6.0.0',
         ]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.4.1
+    hooks:
+    -   id: mypy
+        # equivalent to "files" in .mypy.ini
+        files: '^(Bio|BioSQL)/'
 -   repo: https://github.com/asottile/blacken-docs
     rev: 1.14.0
     hooks:
--- a/Bio/AlignIO/NexusIO.py
+++ b/Bio/AlignIO/NexusIO.py
@@ -13,6 +13,9 @@ See also the Bio.Nexus module (which this code calls internally),
 as this offers more than just accessing the alignment or its
 sequences as SeqRecord objects.
 """
+
+from typing import Iterator, Optional
+
 from Bio.Align import MultipleSeqAlignment
 from Bio.AlignIO.Interfaces import AlignmentWriter
 from Bio.Nexus import Nexus
@@ -24,7 +27,9 @@ from Bio.SeqRecord import SeqRecord
 
 
 # This is a generator function!
-def NexusIterator(handle, seq_count=None):
+def NexusIterator(
+    handle, seq_count: Optional[int] = None
+) -> Iterator[MultipleSeqAlignment]:
     """Return SeqRecord objects from a Nexus file.
 
     Thus uses the Bio.Nexus module to do the hard work.
@@ -51,6 +56,7 @@ def NexusIterator(handle, seq_count=None):
         )
 
     # TODO - Can we extract any annotation too?
+    annotations: Optional[SeqRecord._AnnotationsDict]
     if n.datatype in ("dna", "nucleotide"):
         annotations = {"molecule_type": "DNA"}
     elif n.datatype == "rna":
--- a/Bio/SeqIO/FastaIO.py
+++ b/Bio/SeqIO/FastaIO.py
@@ -22,6 +22,7 @@ from .Interfaces import _clean
 from .Interfaces import _get_seq_string
 from .Interfaces import SequenceIterator
 from .Interfaces import SequenceWriter
+from .Interfaces import _TextIOSource
 
 
 def SimpleFastaParser(handle):
@@ -140,7 +141,12 @@ def FastaTwoLineParser(handle):
 class FastaIterator(SequenceIterator):
     """Parser for Fasta files."""
 
-    def __init__(self, source, alphabet=None, title2ids=None):
+    def __init__(
+        self,
+        source: _TextIOSource,
+        alphabet=None,
+        title2ids=None,
+    ) -> None:
         """Iterate over Fasta records as SeqRecord objects.
 
         Arguments:
--- a/Bio/SeqRecord.py
+++ b/Bio/SeqRecord.py
@@ -1,6 +1,7 @@
 # Copyright 2000-2002 Andrew Dalke.  All rights reserved.
 # Copyright 2002-2004 Brad Chapman.  All rights reserved.
 # Copyright 2006-2023 by Peter Cock.  All rights reserved.
+# Copyright 2020 by Michael R. Crusoe
 #
 # This file is part of the Biopython distribution and governed by your
 # choice of the "Biopython License Agreement" or the "BSD 3-Clause License".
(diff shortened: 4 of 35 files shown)


BioSQL/BioSeq.py
```diff
@@ -536,21 +538,21 @@ def __del_seq(self):
...
 
-    def __del_dbxrefs(self):
+    @dbxrefs.deleter
+    def dbxrefs(self) -> None:
         del self._dbxrefs
 
-    dbxrefs = property(
-        __get_dbxrefs, __set_dbxrefs, __del_dbxrefs, "Database cross references"
-    )
-
```
peterjc, line 553: Looks like a nice bit of property modernisation there 👍

Bio/SeqIO/FastaIO.py
```diff
@@ -140,7 +143,12 @@ def FastaTwoLineParser(handle):
 class FastaIterator(SequenceIterator):
     """Parser for Fasta files."""
 
-    def __init__(self, source, alphabet=None, title2ids=None):
+    def __init__(
+        self,
+        source: Union[str, TextIO],
+        alphabet=None,
+        title2ids: Optional[Title2IDsFuncType] = None,
```
peterjc, line 150: The ``title2ids`` argument is deprecated and due to be removed soon - can we not bother typing it to avoid introducing ``Title2IDsFunctType`` and then removing it?

peterjc:
This looks exciting, thank you! CC @mr-c and @sbliven

Can someone point us at some introductory reading about the significance of ``mypy`` testing (done here via the pre-commit configuration, which is quick) versus the ``mypyc`` testing (not currently enabled but originally attempted on #3165)?

BioSQL/BioSeq.py
speleo3: Credit goes to @mr-c :)

Bio/SeqIO/FastaIO.py
speleo3: Makes sense. I've removed the annotation.

peterjc: I didn't realise initially this was based on the branch from 2 years ago.

mr-c:
> This looks exciting, thank you! CC @mr-c and @sbliven
> 
> Can someone point us at some introductory reading about the significance of ``mypy`` testing (done here via the pre-commit configuration, which is quick) versus the ``mypyc`` testing (not currently enabled but originally attempted on #3165)?

Short answer: using `mypy` instead of `mypyc` is fine. Let's get this merged!

peterjc (approved):
There may well be a few type annotations which are not quite right (especially as most of this was done a while back), but this is solid first step which we can improve upon.

Note there are a handful of code changes here too including some property modernisations in BioSQL, and some flag values changed from ``None`` to ``0`` to keep the flag as a constant type.

(I would be willing to pull a few of these out as separate PR if that would help reviewing this)

mr-c (approved):

.pre-commit-config.yaml
```diff
@@ -31,6 +31,11 @@ repos:
             'flake8-rst-docstrings>=0.2.3',
             'pydocstyle>=6.0.0',
         ]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.4.1
+    hooks:
+    -   id: mypy
+        files: '^Bio.*'
```
peterjc, line 38: This looks odd... we have no files named ``Bio.*`` where dot is literal. We do have ``Bio/*`` though (and ``BioSQL/*``)

speleo3: This is a regular expression

peterjc: I presume you want to match (Python) files under ``Bio/`` and ``BioSQL/``

The caret to left-anchor the search makes sense.

Why use dot (any char) and star (zero or more chars)?

speleo3: > I presume you want to match (Python) files under `Bio/` and `BioSQL/`

Yes correct. This essentially excludes `Tests` and `Scripts` for now.

It should match the `files` selection in [`.mypy.ini`](https://github.com/speleo3/biopython/blob/mypy/.mypy.ini#L2) which is used when you run `mypy` without any path arguments (not via `pre-commit`, which will pass paths of the committed files).

> Why use dot (any char) and star (zero or more chars)?

The `.*` is indeed redundant, I can remove it. Now that you pointed it out, I would actually prefer a more explicit pattern like `'^Bio/|^BioSQL/'` or `'^(Bio|BioSQL)/'`.

peterjc: I think ``'^(Bio|BioSQL)/'`` is much clearer too :smile:

peterjc (approved):
I'd like to merge this early next week baring any objections or requests to delay to allow for another review.

mdehoon (approved):

Bio/AlignIO/NexusIO.py
```diff
@@ -65,7 +68,7 @@ def NexusIterator(handle, seq_count=None):
             id=new_name,
             name=old_name,
             description="",
-            annotations=annotations,
+            annotations=cast(Dict[str, Union[str, int]], annotations),
```
speleo3, line 71: This unsafe `cast` can be replaced with a type annotation of `annotations`

Bio/SeqRecord.py
```diff
@@ -1399,6 +1430,41 @@ def translate(
...
+    def annotations(self, annotations: Optional[Dict[str, Union[str, int]]]) -> None:
+        # unnecessary, except to work around mypy bug: https://github.com/python/mypy/issues/4125
+        if annotations:
+            self._annotations = annotations
+        else:
+            self._annotations = {}
+
+    @annotations.deleter
+    def annotations(self) -> None:
+        # unnecessary, except to work around mypy bug: https://github.com/python/mypy/issues/4125
```
speleo3, line 1465: This bug has been fixed in mypy v0.990, I suggest to revert the workaround.

Bio/SeqRecord.py
```diff
@@ -786,31 +829,14 @@ def __len__(self):
...
-        raise NotImplementedError(_NO_SEQRECORD_COMPARISON)
-
-    def __gt__(self, other):
-        """Define the greater-than operand (not implemented)."""
-        raise NotImplementedError(_NO_SEQRECORD_COMPARISON)
-
-    def __ge__(self, other):
-        """Define the greater-than-or-equal-to operand (not implemented)."""
-        raise NotImplementedError(_NO_SEQRECORD_COMPARISON)
+    __lt__ = _deleted_seqrecord_comparison_method
```
peterjc, line 832: This is more consise, but we lose the docstring.

I would be happier to keep the repetitive stub methods, and annotate them.

speleo3: OK, I've reverted this. I also rebased on current master.

peterjc:
@speleo3 @mr-c I've merged this now, thank you both. I didn't want to delay this any longer, but would like to add something to the NEWS entry - and I think both your names should be acknowledged for the next release. Something like this maybe:

> We now have basic type hint annotations in various modules including Seq,
> SeqRecord, and SeqIO. This should help anyone using an editor or IDE with
> type-aware autocomplete.

Or is this already enough to do even more?

speleo3:
Thanks @peterjc for merging! The NEWS entry looks good to me. Besides auto-complete in IDEs, the type hinting will point out potential bugs, e.g. passing a SeqRecord to a function which only accepts a Seq object will not pass the mypy pre-commit.

mr-c:
@peterjc if a `py.typed` marker file is added to the root package, then typing tools will know to use the embedded types as per https://peps.python.org/pep-0561/#packaging-type-information

peterjc:
Thanks @speleo3, so how about:

> We now have basic type hint annotations in various modules including Seq,
> SeqRecord, and SeqIO. This should help anyone using an editor or IDE with
> type-aware autocomplete, or in conjunction with mypy as a pre-commit check
> this can catch passing the wrong types to Biopython functions/methods.

@mr-c if we can do that with the current partial type-annotation in Biopython, that sounds good. Would that just be a pull request to our ``setup.py`` file (given how we currently have the build setup)?

speleo3:
thanks @peterjc, I think this is a good NEWS entry now 👍

PR #4409, opened 2023-08-11
Enable mypy disallow_incomplete_defs setting.

peterjc:
This raises the bar for adding future annotations...

@speleo3 and/or @mr-c does making this stricter seem wise, and do the changes here look sane?

[We have some inconsistency in how the SeqIO handles are annotated, something for later]


mr-c (approved):

speleo3:
I haven't used `disallow_incomplete_defs` yet. Feels pretty strict, but I guess we can give it a try.

speleo3:
Regarding SeqIO handles: I like to propose some cleanup of `_IOSource` argument handling. Some could use `File.as_handle` for example. I already started some of that while working on the `mypy` branch and I'd be happy to make a pull request.

peterjc:
Sure, a cleanup of the type annotation of the handle arguments sounds good.

If you mean making more things agnostic about handle vs filename as string vs filename as path object, well maybe. The SeqIO design was the internal format-specific parsers would all expect a plain handle only - while the high level functions would be flexible. Also, we're quite likely to want to rework this to allow bytes handles in more places (which I hope could make the FASTA and FASTQ parsers faster).

speleo3:
> If you mean making more things agnostic about handle vs filename as string vs filename as path object, well maybe.

I don't think that I made it more agnostic, just more consistent and with added type checks. But please give it a critical review :-) https://github.com/biopython/biopython/pull/4410

mdehoon (approved):

PR #4412, opened 2023-08-12
Remove most `ignore_errors = True` from .mypy.ini

speleo3:

### DIFF
--- a/Bio/Align/bigbed.py
+++ b/Bio/Align/bigbed.py
@@ -75,6 +75,8 @@ Field = namedtuple("Field", ("as_type", "name", "comment"))
 class AutoSQLTable(list):
     """AutoSQL table describing the columns of an (possibly extended) BED format."""
 
+    default: "AutoSQLTable"
+
     def __init__(self, name, comment, fields):
         """Create an AutoSQL table describing the columns of an (extended) BED format."""
         self.name = name
--- a/Bio/Align/interfaces.py
+++ b/Bio/Align/interfaces.py
@@ -30,7 +30,7 @@ class AlignmentIterator(AlignmentsAbstractBaseClass):
     """
 
     mode = "t"  # assume text files by default
-    fmt = None  # to be defined in the subclass
+    fmt: str  # to be defined in the subclass
 
     def __init__(self, source):
         """Create an AlignmentIterator object.
@@ -142,7 +142,7 @@ class AlignmentWriter(ABC):
     """
 
     mode = "w"  # assume text files by default
-    fmt = None  # to be defined in the subclass
+    fmt: str  # to be defined in the subclass
 
     def __init__(self, target):
         """Create the writer object.
--- a/Bio/Align/interfaces.py
+++ b/Bio/Align/interfaces.py
@@ -13,6 +13,7 @@ use this module.  It provides base classes to try and simplify things.
 """
 from abc import ABC
 from abc import abstractmethod
+from typing import Optional
 
 from Bio import StreamModeError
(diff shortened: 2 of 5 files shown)


Bio/Nexus/Nexus.py
```diff
@@ -2133,7 +2133,7 @@ def terminal_gap_to_missing(self, missing=None, skip_n=True):
 
 
 try:
-    import cnexus
+    import cnexus  # type: ignore
```
speleo3, line 2136: This import actually looks wrong. The module name is `Bio.Nexus.cnexus`. Broken since Python 3 migration?

speleo3: Fix: https://github.com/biopython/biopython/pull/4413

peterjc: I think you're probably right about that - good catch.

Bio/Align/bigbed.py
```diff
@@ -75,6 +75,8 @@
 class AutoSQLTable(list):
     """AutoSQL table describing the columns of an (possibly extended) BED format."""
 
+    default: "AutoSQLTable"
```
peterjc, line 78: What does this do? Is it interpreted as a type hint by mypy (and other type aware systems)?

Bio/Align/interfaces.py
```diff
@@ -142,7 +142,7 @@ class AlignmentWriter(ABC):
     """
 
     mode = "w"  # assume text files by default
-    fmt = None  # to be defined in the subclass
+    fmt: str  # to be defined in the subclass
```
peterjc, line 145: Here and above, previously the attribute was initialised as None. Now it is not initialised until the subclass does it (and it will always be a string)?

Bio/Align/bigbed.py
speleo3: Yes it annotates `AutoSQLTable.default` as type `AutoSQLTable`. I you remove it, mypy reports:
```
Bio/Align/bigbed.py:167:1: error: "type[AutoSQLTable]" has no attribute "default"  [attr-defined]
    AutoSQLTable.default = AutoSQLTable(
    ^~~~~~~~~~~~~~~~~~~~
```

Bio/Align/interfaces.py
speleo3: As far as I understand, the subclass must define this variable. Whatever value we assign here would not be used anyway, so I decided to leave it uninitialized. We could assign an (empty) string which would be acceptable by mypy.

Without the change we get 14 errors like this:
```
Bio/Align/stockholm.py: note: In class "AlignmentWriter":
Bio/Align/stockholm.py:452:11: error: Incompatible types in assignment (expression has type "str", base class "AlignmentWriter" defined the type as "None")  [assignment]
        fmt = "Stockholm"
```

Side note: Type annotations without initialization is quite common with [dataclass](https://docs.python.org/3/library/dataclasses.html).

speleo3: > the subclass **must** define this variable

I have to correct myself: The subclass _may_ define this variable. The class `Bio.Align.bed.AlignmentWriter` currently does not define it. Under those circumstances, how about `fmt = "<undefined>"`? Or we go with `fmt: Optional[str] = None`

speleo3: Reverted to default `None`, with `Optional[str]` type. It's not my intention to change any behavior with this PR.

peterjc: No code change is simplest, although it is perhaps worth exploring why that subclass doesn't defined this.

PR #4436, opened 2023-09-04
bgzf: add utf-8 support

### LINKED ISSUE #2512: Unicode encoding in BGZF

ikravets:
Add UTF-8 encoding support to BGZF, both reader and writer.
There were some attempts to do this in the past. Hopefully, this time will be more successful.
The main problem with UTF-8 is multibyte characters, which would be split between BGZF blocks
if the writer performs a naive `str.encode()`. In this PR the writer detects such multibyte character,
which would be split between two blocks, and carries this character (as a whole) to the next block.

This implementation should be fully backward compatible with the current user code.

*WIP*: If the maintainers agree to merge this PR, I'd like to update the docs and add unit tests.

---

Closes #2512 (again)

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -220,6 +220,7 @@ Text Mode
 Like the standard library gzip.open(...), the BGZF code defaults to opening
 files in binary mode.
 
+FIXME change docs
 You can request the file be opened in text mode, but beware that this is hard
 coded to the simple "latin1" (aka "iso-8859-1") encoding (which includes all
 the ASCII characters), which works well with most Western European languages.
@@ -246,7 +247,7 @@ Instead, you will get the CR ("\r") and LF ("\n") characters as is.
 If your data is in UTF-8 or any other incompatible encoding, you must use
 binary mode, and decode the appropriate fragments yourself.
 """
-
+import codecs
 import struct
 import sys
 import zlib
@@ -257,11 +258,13 @@ _bgzf_magic = b"\x1f\x8b\x08\x04"
 _bgzf_header = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00\x42\x43\x02\x00"
 _bgzf_eof = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00BC\x02\x00\x1b\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00"
 _bytes_BC = b"BC"
+_supported_encodings = ("utf-8", "iso8859-1")
 
 
-def open(filename, mode="rb"):
+def open(filename, mode="rb", encoding=None):
     r"""Open a BGZF file for reading, writing or appending.
 
+    FIXME docs
     If text mode is requested, in order to avoid multi-byte characters, this is
     hard coded to use the "latin1" encoding, and "\r" and "\n" are passed as is
     (without implementing universal new line mode).
@@ -270,9 +273,9 @@ def open(filename, mode="rb"):
     binary mode, and decode the appropriate fragments yourself.
     """
     if "r" in mode.lower():
-        return BgzfReader(filename, mode)
+        return BgzfReader(filename, mode, encoding=encoding)
     elif "w" in mode.lower() or "a" in mode.lower():
-        return BgzfWriter(filename, mode)
+        return BgzfWriter(filename, mode, encoding=encoding)
     else:
         raise ValueError(f"Bad mode {mode!r}")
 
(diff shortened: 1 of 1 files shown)


peterjc:
This might work :)

The BGZF specification (just a couple of pages within the SAM/BAM) specification does not mention encodings. I have not checked the reference implementation source code (``bgzf.c`` and ``bgzip.c`` etc):

https://github.com/samtools/htslib

Do you think we should propose this as a clarification to the definition:

https://github.com/samtools/hts-specs/blob/master/SAMv1.tex

I imagine how this gets worded would be open to debate, e.g. if using BGZF on UTF-8 encoded unicode files:

* the writer *must* ensure multi-byte characters are not split between blocks, and the reader may assume this
* the writer *should* ensure multi-byte characters are not split between blocks, and the reader *cannot* assume this

I think those are the two logical approaches...

peterjc:
Or:

* say explicitly unicode and encodings is as per GZIP itself

I should reread our old issues. It looks like I was worried about Python text mode detecting newline mode ... but we expose the BGZF virtual offset anyway so that difference probably does not matter (even in ASCII or Latin1, Python text mode offsets differ from bytes offsets with Windows vs Unix new lines).

Bio/bgzf.py
```diff
@@ -812,23 +827,27 @@ def __init__(self, filename=None, mode="w", fileobj=None, compresslevel=6):
...
         self._handle = handle
         self._buffer = b""
         self.compresslevel = compresslevel
 
     def _write_block(self, block):
         """Write provided data to file as a single BGZF compressed block (PRIVATE)."""
-        # print("Saving %i bytes" % len(block))
-        if len(block) > 65536:
-            raise ValueError(f"{len(block)} Block length > 65536")
+        assert len(block) <= self._BLOCK_SIZE
```
peterjc, line 842: Avoid assets, the exception is preferable

Bio/bgzf.py
```diff
@@ -852,33 +871,58 @@ def _write_block(self, block):
...
+            self._write_bytes(data)
 
     def flush(self):
-        """Flush data explicitally."""
-        while len(self._buffer) >= 65536:
-            self._write_block(self._buffer[:65535])
-            self._buffer = self._buffer[65535:]
+        """Flush data explicitly."""
+        assert len(self._buffer) < self._BLOCK_SIZE
+        # TODO is writing an empty buffer/block an intended behavior?
```
peterjc, line 925: From memory yes, that is intended.

And again, explicit exceptions preferred over asserts.

peterjc:
For new lines, perhaps all we need to do in the reader in text mode to emulate universal read lines mode is something like map "\r\n" to "\n" and any "\r" to "\n" (subject to some testing vs the standard library behaviour). This seems tangential to the encoding though...

ikravets:
> This might work :)
> 
> The BGZF specification (just a couple of pages within the SAM/BAM) specification does not mention encodings. I have not checked the reference implementation source code (`bgzf.c` and `bgzip.c` etc):
> 
> https://github.com/samtools/htslib
> 
> Do you think we should propose this as a clarification to the definition:
> 
> https://github.com/samtools/hts-specs/blob/master/SAMv1.tex
> 
> I imagine how this gets worded would be open to debate, e.g. if using BGZF on UTF-8 encoded unicode files:
> 
>     * the writer _must_ ensure multi-byte characters are not split between blocks, and the reader may assume this
> 
>     * the writer _should_ ensure multi-byte characters are not split between blocks, and the reader _cannot_ assume this
> 
> 
> I think those are the two logical approaches...

Thanks for the pointers. I see `bgzf.c` implementation uses `void *data` in API, leaving all the encoding/decoding to the caller. It also seems that the callers ignore the character splitting issue completely. My guess is that UTF-8 is only allowed by SAM specs in some fields of the header (e.g. `@CO`) and they do not intend to seek into the middle of such field.

Bio/bgzf.py
```diff
@@ -852,33 +871,58 @@ def _write_block(self, block):
...
+        assert len(self._buffer) < self._BLOCK_SIZE
+        while len(data_bytes):
+            safe_size = self._BLOCK_SIZE - len(self._buffer)
+            if safe_size >= len(data_bytes):
+                safe_size = len(data_bytes)
+            else:
+                while safe_size > 0 and data_bytes[safe_size] >> 6 == 2:
+                    # the byte after the block boundary is not the first octet
+                    # of UTF-8 character (RFC 3629)
+                    safe_size -= 1
```
peterjc, line 903: This is the magic (the key bit of logic to prevent multi-byte characters being split).

https://www.rfc-editor.org/rfc/rfc3629 says "Character boundaries are easily found from anywhere in an octet stream." but I don't immediately see how.

Presumably this is only for utf-8, but will it cope with all the weird and wonderful things in unicode? e.g. emoji and zero-width-joins?

peterjc:
The merge conflicts look fairly simple to fix (ideally with a ``git rebase``).

PR #4468, opened 2023-10-26
Make hmmer program regex flexible enough to work on full program path on windows

### LINKED ISSUE #4361: HMMER Parser doesn't work with Windows
The regex defined here:

https://github.com/krdav/ANARCI_KD_copy/blob/5c4a17f6f0bc530b3e5154667e285eea1195a485/lib/python/anarci/Bio/SearchIO/HmmerIO/hmmer3_text.py#L24

Is unable to find any matches in this output banner from hmmscan:

'# C:\\msys64\\scr\\jou\\builds\\22-2\\build\\thirdparty\\bin\\Windows-x64\\hmmer\\bin\\hmmscan :: search sequence(s) against a profile database\n'

and as a result

https://github.com/biopython/biopython/blob/9965ece1d76adc8ee65904f87ea68c0b1206050a/Bio/SearchIO/HmmerIO/hmmer3_text.py#L20

Sometimes leaves query_start, query_end, hit_start, and hit_end as None. It's worse than assigning without correction, and should really be an error, or handled gracefully with its current downstream warning.

byerlyj:

Closes #4361

On Windows, the program name in the hmmscan scan preamble includes the full path to the executable. The current regex is unable to match, e.g. `C:\some\path\hmmscan`

This change updates `hmmer3_text._RE_PROGRAM` to be a more flexible in allowing characters preceding the program name, and adds a corresponding test

### DIFF
--- a/Tests/Hmmer/text_30_hmmscan_011.out
+++ b/Tests/Hmmer/text_30_hmmscan_011.out
@@ -1,0 +1,1425 @@
+# C:\msys64\scr\byerly\builds\NB\20231025 _\thirdparty\bin\Windows-x64\hmmer\bin\hmmscan :: search sequence(s) against a profile database
+# HMMER 3.1b2 (February 2015); http://hmmer.org/
+# Copyright (C) 2015 Howard Hughes Medical Institute.
+# Freely distributed under the GNU General Public License (GPLv3).
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+# query sequence file:             C:\Users\byerly\AppData\Local\Temp\tmpt7psecrp.fasta
+# target HMM database:             C:\msys64\scr\byerly\builds\NB\20231025 _\internal\lib\site-packages\anarci\dat\HMMs\ALL.hmm
+# output directed to file:         C:\Users\byerly\AppData\Local\Temp\tmp2u1a0mw3.txt
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+Query:       6FR3  [L=202]
+Scores for complete sequence (score includes all domains):
+   --- full sequence ---   --- best 1 domain ---    -#dom-
+    E-value  score  bias    E-value  score  bias    exp  N  Model    Description
+    ------- ------ -----    ------- ------ -----   ---- --  -------- -----------
+     3e-040  128.1   0.2   5.8e-040  127.2   0.0    1.6  2  human_A   
+   3.4e-038  121.4   0.5   8.4e-038  120.2   0.0    1.8  2  mouse_A   
+   5.4e-029   91.7   0.0   1.1e-028   90.7   0.0    1.5  2  human_D   
+   1.6e-028   90.6   0.2   4.7e-028   89.1   0.0    1.8  2  mouse_D   
+   1.4e-022   71.2   0.2   3.1e-022   70.0   0.0    1.6  2  mouse_B   
+   9.4e-021   65.1   0.0   2.2e-020   64.0   0.0    1.6  1  human_B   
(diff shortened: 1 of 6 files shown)


Tests/Hmmer/text_30_hmmscan_011.out
```diff
@@ -0,0 +1,1425 @@
+# C:\msys64\scr\byerly\builds\NB\20231025 _\thirdparty\bin\Windows-x64\hmmer\bin\hmmscan :: search sequence(s) against a profile database
+# HMMER 3.1b2 (February 2015); http://hmmer.org/
+# Copyright (C) 2015 Howard Hughes Medical Institute.
+# Freely distributed under the GNU General Public License (GPLv3).
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+# query sequence file:             C:\Users\byerly\AppData\Local\Temp\tmpt7psecrp.fasta
+# target HMM database:             C:\msys64\scr\byerly\builds\NB\20231025 _\internal\lib\site-packages\anarci\dat\HMMs\ALL.hmm
```
bow, line 7: I can see here that we also have a Windows path with space. Although this is handled by a different regex, could we still add an assertion to the test to make sure the value is parsed correctly?

Tests/test_SearchIO_hmmer3_text.py
```diff
@@ -1654,6 +1654,13 @@ def test_30_hmmscan_010(self):
         self.assertEqual(0, hit.domain_obs_num)
         self.assertEqual(0, len(hit))
 
+    def test_30_hmmscan_011(self):
+        """Parsing hmmscan 3.0 (text_30_hmmscan_011)."""
+        hmmer_file = get_file("text_30_hmmscan_011.out")
+        # Test getting program name when preamble contains full path and whitespace
+        for qresult in parse(hmmer_file, FMT):
+            assert qresult.program == "hmmscan"
+
```
bow, line 1663: This is a good start :+1:. One thing I might suggest here is to either add a little more detail to the assertion, for when the test fails.

Something like:
```
...
    assert qresult.program == "hmmscan", f"Expected 'hmmscan' in item index {idx}, found {qresult.program!r}"
...
```


Furthermore, let's add a little bit more for some light sanity checks. Would you mind adding:

* A test to make sure we have the expected number of `qresult`?
* For each qresult, a test to make sure have parsed its contents correctly. I'm thinking just testing for the number of parsed hits is enough.
* Optionally, also a test to make sure all HSPs are parsed correctly. Nothing too involved, something like the `score` of the last HSP should be enough.

bow (changes requested):
Thanks for opening the PR @byerlyj. It's great to see the issue being handled :slightly_smiling_face:.

I have some suggestions regarding which tests we should do. Please let me know if you need help with anything.

Over to you.

Tests/Hmmer/text_30_hmmscan_011.out
byerlyj: absolutely

Tests/test_SearchIO_hmmer3_text.py
byerlyj: done! let me know what you think

bow:
Thanks for the changes @byerlyj ~ I think this looks good to merge. I'll approve this and rebase this onto `master`.

byerlyj:
> Thanks for the changes @byerlyj ~ I think this looks good to merge. I'll approve this and rebase this onto `master`.
> 
> EDIT: Ah, sorry @byerlyj ~ one more thing. Not on the code. Just to cross check: `CONTRIB.rst` is updated but there is not yet an entry in `NEWS.rst` (with your name and a short description of the fix). Was this intentional? If not, would you mind adding such an entry?

honestly, it didn't seem newsworthy to me, haha. I can add it though!

PR #4486, opened 2023-11-06
Uniprot error

### LINKED ISSUE #4467: SeqIO OOM parse issue
### Setup

system as follows:

```python
```

3.8.8 (default, Apr 13 2021, 15:08:03) [MSC v.1916 64 bit (AMD64)]
CPython
Windows-10-10.0.19041-SP0
1.81

### Expected behaviour

I am using SeqIO.parse to load in Uniref50 XML (240gb) record by record to build an annotation set. Even though it is supposed to load one record at a time it is giving me OOM errors and crashing with 1.8Tb ram.

### Actual behaviour

Works fine for 8 gb file (Sprot XML)

```python
file_path = './uniref50.xml'
total=62149502
pubmed_pattern = re.compile(r'\((Ref\.\d+, )?(PubMed:\d+)(, Ref\.\d+)?(, PubMed:\d+)*\)')
seqs, ids, texts = [], [], []

with open(file_path, 'r') as handle:
    print('This prints')
    for record in SeqIO.parse(handle, 'uniprot-xml'):
        print('This does not print before OOM with 1.8Tb ram')
        annotations = record.annotations
        intersection = set(annotations.keys()) & set_keys
        seq_len = annotations.get('sequence_length', 0)
    
        if len(intersection) >= 5 and 50 < seq_len < 512:
            text = build_text(intersection, annotations, int_keys, regex_keys, text_keys, pubmed_pattern)
            seqs.append(str(record.seq))
            ids.append(record.id)
            texts.append(text)
        else:
            continue
```

### Steps to reproduce

Run the following code with uniref50 XML. The necessary dictionaries and functions can be fillers, they are not called when using the large files.

lurkingviolet:

Closes #4467 by adding an explicit check on the input xml namespaces that would fail with an informative error instead of running out of memory on large files that are the wrong format.  Ran performance tests on a large valid uniprot file (~800Mb file from #4467) before and after this fix to verify no observable performance impact (~700 seconds before and after).

### DIFF
--- a/Tests/test_SeqIO.py
+++ b/Tests/test_SeqIO.py
@@ -2539,6 +2539,25 @@ class TestSeqIO(SeqIOTestBaseClass):
             messages,
         )
 
+    def test_uniprot_xml_namespace(self):
+        uniref_file_name = (
+            "SwissProt/UniRef90_P99999.xml"  # a related but not uniproot format file
+        )
+        with self.assertRaises(ValueError) as context:
+            records = []
+            for record in SeqIO.parse(uniref_file_name, format="uniprot-xml"):
+                records.append(record)
+        self.assertRegex(
+            str(context.exception),
+            "http://uniprot.org/uniprot",
+            "Correct namespace in error",
+        )
+        self.assertRegex(
+            str(context.exception),
+            "http://uniprot.org/uniref",
+            "Unexpected namespace in error",
+        )
+
     def test_swiss20(self):
         sequences = ["MTMAAAQGKLSPDAIDNEVISNGSAKDYLDPPPAPLVDAG...SNYDAAV"]
         ids = ["F2CXE6"]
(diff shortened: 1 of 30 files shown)


Tests/test_SeqIO.py
```diff
@@ -2539,6 +2539,25 @@ def test_uniprot_xml5(self):
             messages,
         )
 
+    def test_uniprot_xml_namespace(self):
+        uniref_file_name = (
+            "SwissProt/UniRef90_P99999.xml"  # a related but not uniproot format file
```
peterjc, line 2544: typo "uniproot"

lurkingviolet: Good catch, I will fix and push again.

peterjc:
Merged into a single commit - thank you!

lurkingviolet:
Awesome thanks for merging and reviewing.

PR #4516, opened 2023-11-28
Allow qblast to accept lists of organisms to include/exclude

netogallo:

Allow qblast to accept an explicit list of organisms to include/exclude. That this is already possible to a limited degree by using the entrez query. However, I have a use case which requires me to build a complex list of organisms to exclude/include, which also queries qblast multiple times (with similar but not identical lists of organisms). This feature makes my use case easier to implement and prevents blast from having to do the entrez query as I already provide the correct names and taxids of the organisms.

Let me know if you are open to include this in biopython, I am happy to write a test if so.


Bio/Blast/NCBIWWW.py
```diff
@@ -131,6 +137,12 @@ def qblast(
                       manually set parameters like word size and e value. Turns
                       off when sequence length is > 30 residues. Default: None.
      - service        plain, psi, phi, rpsblast, megablast (lower case)
+     - organisms      A dictionary that defines the organisms that will be
+                      included/excluded in the search. The key is the name
+                      of the organism, following the taxonomy convention
+                      ie. "Bacteria (taxid:2)" and the value is a boolean
```
peterjc, line 143: ie -> eg

I assume (please confirm) this works with any taxonomy node like "Bacteria" and not just leaf nodes (like a specific species)?

Bio/Blast/NCBIWWW.py
```diff
@@ -206,8 +218,34 @@ def qblast(
...
     }
 
+    organisms = {} if organisms is None else organisms
+    for i, (organism, exclude) in enumerate(organisms.items()):
+        if not isinstance(organism, str):
+            raise TypeError("The keys of the 'organisms' parameter must be strings.")
+
+        if ORGANISM_REGEX.match(organism) is None:
+            raise ValueError(
+                "Organisms must be specified following the taxonomy convention. ie. 'Bacteria (taxid:2)'"
```
peterjc, line 231: ie -> eg

peterjc:
Is this explained in the latest qblast documentation from the NCBI?

Bio/Blast/NCBIWWW.py
netogallo: On my use cases, it is indeed the case. And it does set difference between domains, eg.

`
{
  "Bacteria (taxid:2)": False,
  "E. coli (taxid:562)": True
}
`

Will give you results that are bacteria but not "E. coli". This particular behavior is what I am relying on for my use case.

netogallo:
Unfortunately, I could not find any official documentation for this functionality. I worked it out by inspecting the web requests performed by the "NCBI Blast" web page and reverse engineered the  functionality. For this reason, I can understand you might not want to include the changes.

PR #4526, opened 2023-12-05
Parse origin-spanning locations containing the entire sequence

### LINKED ISSUE #4524: Bug when parsing origin-spanning features that span an entire circular sequence
### Setup

system as follows:

```python
3.11.6 (main, Oct  6 2023, 11:55:02) [Clang 13.0.0 (clang-1300.0.29.30)]
CPython
macOS-12.7-x86_64-i386-64bit
1.81
```

### Expected behaviour

In a circular sequence of length 4, the location `3..2`, which spans the entire sequence spanning the origin and starting at position 3, should give a location `join{[2:4](+), [0:2](+)}`.

### Actual behaviour

That feature is parsed as `[2:2](+)`, which is an empty slice that when extracted gives an empty sequence.

### Steps to reproduce

cc @peterjc 

Happy to suggest a fix and include a test if this is considered a bug. See the comments for observed vs. expected behaviour.

```python
from Bio.SeqIO import read
from io import StringIO

example = """
LOCUS       pj5_00001                  4 bp ds-DNA     circular     05-DEC-2023
DEFINITION  .
COMMENT     Imported using the Genbank importer. File name: pj5_00001 (1).gb
            teselagen_unique_id: 5adf735aa1811801e17d8aac
     misc_feature    3..2
                     /label="forward"
     misc_feature    complement(3..2)
                     /label="reverse"
ORIGIN
        1 acgt
//
"""

seq = read(StringIO(example), 'genbank')

print(seq.features[0].location) # [2:2](+) / Should be join{[2:4](+), [0:2](+)}
print(seq.features[1].location) # [2:2](-) / Should be join{[0:2](-), [2:4](-)}

print(seq.features[0].extract(seq).seq) # Gives empty sequence
print(seq.features[1].extract(seq).seq) # Gives empty sequence

example = """
LOCUS       pj5_00001                  4 bp ds-DNA     circular     05-DEC-2023
DEFINITION  .
COMMENT     Imported using the Genbank importer. File name: pj5_00001 (1).gb
            teselagen_unique_id: 5adf735aa1811801e17d8aac
     misc_feature    4..2
                     /label="forward"
     misc_feature    complement(4..2)
                     /label="reverse"
ORIGIN
        1 acgt
//
"""

seq = read(StringIO(example), 'genbank')

print(seq.features[0].location) # join{[3:4](+), [0:2](+)}
print(seq.features[1].location) # join{[0:2](-), [3:4](-)}

print(seq.features[0].extract(seq).seq) # Gives TAC
print(seq.features[1].extract(seq).seq) # Gives GTA

```

manulera:

Closes #4524

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -909,7 +909,7 @@ class SimpleLocation(Location):
         except ValueError:
             pass
         else:
-            if 0 <= s <= e:
+            if 0 <= s < e:
                 return SimpleLocation(s, e, strand)
         # Try general case
         try:
@@ -941,7 +941,7 @@ class SimpleLocation(Location):
             # Attempt to fix features that span the origin
             s_pos = Position.fromstring(s, -1)
             e_pos = Position.fromstring(e)
-            if s_pos > e_pos:
+            if s_pos >= e_pos:
                 # There is likely a problem with origin wrapping.
                 # Create a CompoundLocation of the wrapped feature,
                 # consisting of two SimpleLocation objects to extend to
--- a/Tests/test_SeqFeature.py
+++ b/Tests/test_SeqFeature.py
@@ -337,6 +337,26 @@ class TestExtract(unittest.TestCase):
         self.assertEqual(type(sequence), Seq.Seq)
         self.assertEqual(sequence, "ccaatgg")
 
+    def test_origin_spanning_location(self):
+        """Test location spanning origin."""
+        # Regular origin-spanning sequence
+        self.assertEqual(
+            str(SimpleLocation.fromstring("4..2", 4, True)), "join{[3:4], [0:2]}"
+        )
+        self.assertEqual(
+            str(SimpleLocation.fromstring("complement(4..2)", 4, True)),
+            "join{[0:2](-), [3:4](-)}",
+        )
+
+        # Origin-spanning location containing the entire sequence
+        self.assertEqual(
+            str(SimpleLocation.fromstring("3..2", 4, True)), "join{[2:4], [0:2]}"
+        )
+        self.assertEqual(
+            str(SimpleLocation.fromstring("complement(3..2)", 4, True)),
+            "join{[0:2](-), [2:4](-)}",
+        )
+
 
 if __name__ == "__main__":
     runner = unittest.TextTestRunner(verbosity=2)
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -113,6 +113,7 @@ possible, especially the following contributors:
 - Jacob Byerly (first contribution)
 - João Rodrigues
 - Joe Greener
+- Manuel Lera Ramirez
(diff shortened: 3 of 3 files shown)


Bio/SeqFeature.py
```diff
@@ -909,7 +909,7 @@ def fromstring(text, length=None, circular=False):
         except ValueError:
             pass
         else:
-            if 0 <= s <= e:
```
manulera, line 912: @peterjc this also had to be changed. I don't think it breaks anything, but flagging in case you can think of something

Bio/SeqFeature.py
```diff
@@ -941,7 +941,7 @@ def fromstring(text, length=None, circular=False):
             # Attempt to fix features that span the origin
             s_pos = Position.fromstring(s, -1)
             e_pos = Position.fromstring(e)
-            if s_pos > e_pos:
+            if s_pos >= e_pos:
```
peterjc, line 944: Will you add the else-clause with ``raise LocationParserError(f"invalid feature location '{text}'")`` as well?

Bio/SeqFeature.py
peterjc: I thought initially it might break single base locations, but this is after the start - 1 has been applied, so I think it is OK.

Tests/test_SeqFeature.py
```diff
@@ -337,6 +337,26 @@ def test_reference_in_compound_location_sequence(self):
...
+
+        # Origin-spanning location containing the entire sequence
+        self.assertEqual(
+            str(SimpleLocation.fromstring("3..2", 4, True)), "join{[2:4], [0:2]}"
+        )
+        self.assertEqual(
+            str(SimpleLocation.fromstring("complement(3..2)", 4, True)),
+            "join{[0:2](-), [2:4](-)}",
+        )
+
```
peterjc, line 359: Yay -  that looks like a simple but comprensive test based on your motivating example.

Bio/SeqFeature.py
manulera: Added with https://github.com/biopython/biopython/pull/4526/commits/e3a5ae0f6926e467b667454e0d6e0716c3736f6b

peterjc: Oh - I suppose now that we've fixed the equality edge case, that no longer applies.

manulera: It seems this messes up some things for alignments, for locations like this `<437..706` that I did not know existed. Not sure how this particular case should be handled

peterjc: Welcome to the joys and murkier corners of GenBank locations. That's a before position, not used much nowadays.

It looks like we should revert e3a5ae0f6926e467b667454e0d6e0716c3736f6b - I was wrong to suggest adding that. This particular bit of code is only trying to find nice and simple origin wrapping locations.

PR #4592, opened 2024-01-19
Parallelize PDBList download methods

Will-Tyler:
### Acknowledgements

### Description

In this pull request, I parallelize the PDB download methods, which significantly reduces the time spent downloading if multiple threads are used. In each of the download methods, I add a parameter, `max_num_threads`, that allows the user to specify the maximum number of threads to use.

I also updated the docstrings and tried to make them more consistent. I decided to remove the type information in the docstrings because the API docs show the type information in the parameter list.

### Testing

The PDBList unit tests do not cover the download methods. I manually tested the four download methods (`download_pdb_files`, `download_all_assemblies`, `download_entire_pdb`, and `download_obsolete_entries`) to confirm that the multi-threading works as expected.

For the docstring updates, I built the documentation and inspected the API docs to check my changes.

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -36,14 +36,15 @@
 """Access the PDB over the internet (e.g. to download structures)."""
 
 import contextlib
+import functools
 import gzip
 import json
 import os
 import re
 import shutil
 import sys
-
-from typing import List, Tuple
+from concurrent.futures import ThreadPoolExecutor
+from typing import List, Optional, Tuple
 from urllib.request import Request
 from urllib.request import urlcleanup
 from urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


Bio/PDB/PDBList.py
```diff
@@ -412,7 +415,13 @@ def update_pdb(self, file_format=None, with_assemblies=False):
...
 
     def download_pdb_files(
-        self, pdb_codes, obsolete=False, pdir=None, file_format=None, overwrite=False
+        self,
+        pdb_codes,
+        obsolete=False,
+        pdir=None,
+        file_format=None,
+        overwrite=False,
+        max_num_threads: Union[int, None] = None,
```
peterjc, line 424: I believe the convention is to use ``Optional[int]`` instead.

Or more simply, default to zero and use that as no limit?

peterjc: Might also not bother with adding type annotation, or do the whole function if you fancy?

Will-Tyler: I decided to add type annotations to all the parameters, and I use Optional[int] now as you suggested.

Bio/PDB/PDBList.py
```diff
@@ -412,52 +413,51 @@ def update_pdb(self, file_format=None, with_assemblies=False):
                     print(f"Obsolete file {old_file} is missing")
 
     def download_pdb_files(
-        self, pdb_codes, obsolete=False, pdir=None, file_format=None, overwrite=False
+        self,
+        pdb_codes: List[str],
```
peterjc, line 417: I think you can do just ``list[str]`` lower case without an import?

[Note we still support Python 3.8 in Biopython which might prevent that?]

JoaoRodrigues:
@Will-Tyler, very nice idea! 

I think there's a simpler (and more efficient) way of speeding up the downloads for regular entries. See [this gist](https://gist.github.com/JoaoRodrigues/761b456a110a654794f87294b226e14d) for an implementation. The overall idea is to request batches of IDs from the PDB itself (I'm not sure this works for PDBe and PDBj, need to check) and then unzip them locally. So, to download 1000 structures, instead of making 1000 requests, you make 10, each grabbing 100 structures. In my hands this has been a _lot_ faster. Do you think this would be interesting to merge with these changes here?

Will-Tyler:
I agree that batch downloading is a good idea, and I added it to my personal list of ideas for Biopython tasks. I am also happy to create an issue for looking into batch downloading (and checking if it can be done from the wwPDB server). However, I'd like to do it as part of a follow-up pull request.

JoaoRodrigues:
Fair enough, let me test this locally then this weekend 👍

Bio/PDB/PDBList.py
Will-Tyler: If I remember correctly, I encountered "type is not subscriptable" errors in the GitHub checks, which led me to use the types from the `typing` module.

JoaoRodrigues:
Tested locally, works great, thanks @Will-Tyler !

PR #4593, opened 2024-01-19
Implement deep equality in PDB model classes (breaking change)

Will-Tyler:
### Acknowledgements

### Description

In this pull request, I implement deep equality checking in the PDB model classes (`Entity`, `Structure`, `Model`, `Chain`, `Residue`, `Atom`, `DisorderedEntityWrapper`, and `DisorderedAtom`). Deep equality checking means that if the Python interpreter is comparing two entities in a tree, then the interpreter recursively compares the children of the entities as well. Although this is technically a breaking change, I think this new definition of equality makes much more sense. In the previous definition of equality, two structures could be equal even if they have different atoms, residues, chains, or models.

### Testing

I add a unit test that would fail with the previous definition of equality.


JoaoRodrigues:
I like this idea, but I think we should support both equality tests. For instance, if I want to compare two conformers, I might want to check they have the same chain/residue/atom composition but the coordinates are irrelevant. I might also want to check only if two chains have the same residues, and an equality based on residue name here is enough.

Why don't we spin this out as a new class, `StructureComparison`, that allows for this type of strict comparison?

Will-Tyler:
What should be the implementation for `==`? I am new to bioinformatics, but I was expecting a full deep equality check similar to how equality is checked between two Python dictionaries. I tried to implement the definition of equality that I would expect in the context of PDB structures. For example, I would find it strange to say that two PDB structures are equal if they have different atoms or even if they have the same atoms, but those atoms have different coordinates—likewise for other types of entities with children.

I think I do understand the first comparison that you mentioned though. What about introducing a method to check if two structures have the same composition without checking the coordinates? (Let me know if there is a better term.

This also got me thinking that it might be useful to have a method to check for mathematical isomorphism.

For the second comparison, I am having some trouble understanding. When would you be interested in checking that the residue names are equal without regard for the atoms that make up the residues? Is it just that it is more efficient since the atoms should be the same if the residue name is the same? What about introducing a separate method for this type of comparison as well?

peterjc:
I've only dabbled with PDF (etc) files, but it certainly can make sense to just check the list of residues/atoms are them same even if their locations are different. A clear example here would be alternative arrangements of the same structure in space (e.g. different models of a protein from NMR).

Which comparison should be the default of object equality and why is another discussion, as is should we change the historical behaviour -- which is a very high bar given how old this code is (and how many users is has).

JoaoRodrigues:
Equivalence here is not strictly defined because it depends on the application.

You can have two structures of the same molecule, where only the coordinates change. The molecules are _chemically_ the same, but structurally different (conformers). Another example is isomerism - again, same molecular composition (at the atom level), but different coordinates. This is one example.

Another example is for example to check chain/residue composition of a structure. Taking the examples above, if you want to compare two conformers of the same molecule you first need to know they are indeed the same molecule. So first you'd check they have the same atoms and then if that's true, compare the coordinates. A more simple application is simply comparing two chains by their residue identities, e.g. chain A is Ala-Glu-Lys-Ala, is chain B the same? That's valuable on its own.

There is indeed isomorphism in structures as well. A typical example is hydrogen atoms being treated as equivalent in many calculations. Take an Alanine residue for example (structure [here](https://en.wikipedia.org/wiki/Alanine)). The positions of the 3 hydrogen atoms in the side-chain (that CH3 group in the image) are different, but for most biological/chemical purposes, those three atoms are equivalent. You don't really care if two structures of Ala have a coordinate correspondence of say H1-H2, H2-H3, H3-H1; you just deem them equivalent as long as everybody else is in the same place.

Long story short - I think there is value in having a comparison class/function that takes two structures and does this sort of recursive deep comparison. I think it should also go and compare coordinates in _some_ situations (maybe optionally) where that is useful. But it should definitely not be the default comparison between two structures.

JoaoRodrigues:
Adding to my answer above.

>What should be the implementation for ==? I am new to bioinformatics, but I was expecting a full deep equality check similar to how equality is checked between two Python dictionaries. I tried to implement the definition of equality that I would expect in the context of PDB structures. For example, I would find it strange to say that two PDB structures are equal if they have different atoms or even if they have the same atoms, but those atoms have different coordinates—likewise for other types of entities with children.

For coordinate comparison (and to say if two structures are the same) we usually look at RMSDs of equivalent atomic positions (https://en.wikipedia.org/wiki/Root-mean-square_deviation_of_atomic_positions). We have a couple of modules to do that. And there is where that isomorphism I mentioned above comes into play, but we need to know the bond topology of the molecule for that. I have some solutions for that if you want to go that route.

>For the second comparison, I am having some trouble understanding. When would you be interested in checking that the residue names are equal without regard for the atoms that make up the residues? Is it just that it is more efficient since the atoms should be the same if the residue name is the same? 

Experimentally-determined structures, like those you get from the PDB, don't always have all the atoms. This is usually linked to the resolution of the structure (higher, more detail) but not always. So you can have one structure of protein A determined at 1Å resolution and another at 4Å resolution, and they are the same protein, but whereas the first has all the atoms the second only has alpha carbons because that's all they could confidently model in.

Also, keep in mind that Biopython is used by different crowds. Risking angering people with this broad generalization, biologists look at proteins like a string of letters while chemists look at them like a collection of atoms. So for a biologist, two proteins are the same if they have the same sequence of residues, whereas a chemist will gasp and ask about the coordinates.

Will-Tyler:
Thank you for the clear explanation. I think I understand better now, and I agree that the deep strict equality should not be default comparison.

For the sake of incrementally delivering small changes, I will focus on just adding a method to check for deep strict equality in this pull request. But I am interested in adding more comparisons in future pull requests if you think they would be useful. I also want to look at the coordinate comparison modules that you mentioned. Taking a quick glance now, it looks like there is code to rotate/translate atoms to superimpose structures on each other and minimize the RMSD, which seems useful for determining geometric isomorphism.

Anyway, I will work on making deep strict equality checking a separate method for now and consider adding more comparison methods / changing the `==` implementation in a future pull request.

Bio/PDB/Atom.py
```diff
@@ -262,6 +263,40 @@ def __sub__(self, other):
...
+        :type other: Atom
+        :return: Whether the atoms are strictly equal
+        :rtype: bool
+        """
+        if not isinstance(other, type(self)):
+            return False
+
+        are_equal = (
+            self.name == other.name
+            and np.allclose(self.coord, other.coord)
```
JoaoRodrigues, line 286: I'd push this one to the end, as it's the most expensive call.

Bio/PDB/Atom.py
```diff
@@ -262,6 +263,40 @@ def __sub__(self, other):
...
+            and self.bfactor == other.bfactor
+            and self.occupancy == other.occupancy
+            and self.altloc == other.altloc
+            and self.fullname == other.fullname
+        )
+
+        if are_equal:
```
JoaoRodrigues, line 293: Why not add these to the `are_equal` above with:

```python
getattr(self, "element", None) == getattr(self, "element", None)
...
```

We'd have to update this function in any case if the attributes change so doesn't seem a lot more work to have the checks explicit there.

Bio/PDB/Entity.py
```diff
@@ -183,6 +183,33 @@ def id(self, value):
...
+            return False
+
+        if self.id != other.id:
+            return False
+
+        if len(self.child_list) != len(other.child_list):
+            return False
+
+        for left_child, right_child in zip(self.child_list, other.child_list):
+            assert hasattr(left_child, "strictly_equals")
```
JoaoRodrigues, line 207: Isn't this covered by the `isinstance` above?

JoaoRodrigues (changes requested):
LGTM, some small suggestions.

JoaoRodrigues:
There's one scenario that I can think of where two molecules are exactly the same but these tests can fail. Take this example: `struct2` is a copy of `struct1` but it was translated by [1, 0, 0]. All atoms are in the exact same relative orientation (to each other), but the test will fail. Thus, I am not sure adding an equivalence check for coordinates might be such a good idea after all.

Bio/PDB/Entity.py
```diff
@@ -435,6 +462,34 @@ def get_id(self):
...
+        :return: Whether the two entities are strictly equal
+        :rtype: bool
+        """
+        if not isinstance(other, type(self)):
+            return False
+
+        if self.id != other.id:
+            return False
+
+        # Check that the children dictionaries have the same keys
```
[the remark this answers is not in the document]
Will-Tyler: I don't think this catches the case where `other.child_dict` has a child that `self.child_dict` does not have.

Bio/PDB/Entity.py
Will-Tyler: I think that in standard usage of the Entity class, you would never expect this assertion to fail. All children should be other entities or atoms, all of which have `strictly_equals`. But I don't think that the assertion is logically or technically guaranteed to be true, so I add the assertion statement to indicate that there is an issue in case it fails.

Will-Tyler:
I think we should add a separate method to test for structural isomorphism. That is, there exists a rigid motion (a rotation and a translation) such that the structures are strictly equal when the rigid motion is applied to one of the structures. Perhaps we can check if the internal coordinates are the same to test for this. This would be useful for the translation case that you described. I can do this in a follow-up pull request.

Bio/PDB/Entity.py
```diff
@@ -435,6 +462,37 @@ def get_id(self):
...
+        :rtype: bool
+        """
+        if not isinstance(other, type(self)):
+            return False
+
+        if self.id != other.id:
+            return False
+
+        if not self.selected_child.get_id() == other.selected_child.get_id():
+            return False
```
Will-Tyler, line 483: I added this in my latest push. Disordered entities have to have the same selected child to be considered strictly equal.

Bio/PDB/Entity.py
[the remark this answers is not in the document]
JoaoRodrigues: Oh yeah, sorry, true, nevermind!

Bio/PDB/Entity.py
JoaoRodrigues: Alright!

JoaoRodrigues:
>I think we should add a separate method to test for structural isomorphism. That is, there exists a rigid motion (a rotation and a translation) such that the structures are strictly equal when the rigid motion is applied to one of the structures. Perhaps we can check if the internal coordinates are the same to test for this. This would be useful for the translation case that you described. I can do this in a follow-up pull request.

Being very honest, in ~20 years in the field, I have yet to find a use for a comparison tool like that. I think what is most useful (practically so) is doing a recursive comparison of 'labels' on chains, residues, atoms, etc, and skip coordinates. Comparing coordinates (as is, without any rotation/translation optimization) is too brittle, whereas doing a proper structural comparison is tricky because you might have structure A be a subset of structure B, and that can count as a match for certain people and applications.

I'd be more comfortable if we just dropped the coordinate comparison here, but I also want to be mindful of your time and effort, so if you feel strongly about it, what about adding an optional argument to the comparison method to use coordinates or not?

Will-Tyler:
I appreciate the honesty—as someone who is trying to enter the field, I do not want to work on something that is not useful.

> What about adding an optional argument to the comparison method to use coordinates or not?

If I am understanding correctly, it seems that there are several notions of equality or sameness that can be useful in different contexts. For me, the strict notion of equality (which includes comparing the coordinates) was useful because I implemented a parser that I needed to test. And you are saying the most useful comparison is to compare the labels. In my opinion, these different notions should be different methods.

Rather than having one method named `equals` (or something similar) with parameters to change its behavior like
```python
structure.equals(other_structure, strict=True)  # Compare most things, including coordinates
structure.equals(other_structure)  # Just recursively compare the labels
```
I think it would be clearer to have different methods, like
```python
structure.strictly_equals(other_structure)  # Compare most things, including coordinates
structure.nominally_equals(other_structure)  # Just recursively compare the labels
```

Let me know your thoughts, and I will defer to your opinion.

JoaoRodrigues:
Sorry, busy weeks recently.. 

I vote for your last option, I think it's the clearest:

```python
structure.strictly_equals(other_structure, compare_coordinates=True)  # Compare most attributes, including the coordinates
structure.strictly_equals(other_structure)  # Compare the same attributes, except for the coordinates
```

JoaoRodrigues:
I had missed the notification about this. Sorry! Merging now!

JoaoRodrigues (approved):

JoaoRodrigues:
Thanks @Will-Tyler !

Will-Tyler:
Thanks for taking the time to review and explain the concepts!

PR #4598, opened 2024-01-22
Add SeqIO support for GFA files.

### LINKED ISSUE #4597: SeqIO.parse() support type GFA
Can SeqIO support reading GFA files into SeqRecord objects?

michaelfm1211:

Closes #4597

The title is pretty self-explanatory. This PR adds support for GFA files as documented at https://gfa-spec.github.io/GFA-spec/GFA1.html and https://gfa-spec.github.io/GFA-spec/GFA2.html.


Bio/SeqIO/GfaIO.py
```diff
@@ -0,0 +1,89 @@
...
+
+def GfaIterator(source):
+    """Parser for GFA 1.x files.
+
+    Documentation: https://gfa-spec.github.io/GFA-spec/GFA1.html
+    """
+    with as_handle(source) as handle:
+        for line in handle:
+            if line == "\n":
+                warnings.warn("GFA data has a blank line.", BiopythonWarning)
```
peterjc, line 40: Maybe add a ``continue`` line here? Otherwise won't it fail at ``fields[0]`` a few lines down?

Bio/SeqIO/GfaIO.py
```diff
@@ -0,0 +1,89 @@
...
+            fields = line.strip("\n").split("\t")
+            if fields[0] != "S":
+                continue
+            if len(fields) < 3:
+                raise ValueError(
+                    f"Segment line must have name and sequence fields: {line}."
+                )
+
+            if fields[2] == "*":
+                # no sequence data
```
peterjc, line 51: We might still want to yield a record with unknown sequence? Are these common in real GFA output?

Bio/SeqIO/GfaIO.py
```diff
@@ -0,0 +1,89 @@
+"""Bio.SeqIO support for the Graphical Fragment Assembly format."""
```
peterjc, line 1: Best put a bit more context here, including links for the file format.

Bio/SeqIO/__init__.py
```diff
@@ -420,6 +425,8 @@
     "gck": GckIO.GckIterator,
     "genbank": InsdcIO.GenBankIterator,
     "genbank-cds": InsdcIO.GenBankCdsFeatureIterator,
+    "gfa": GfaIO.GfaIterator,
```
peterjc, line 428: Assuming we can't easily detect the version from the file contents (which would be ideal, then we just need "gfa" as a single format), maybe "gfa1" and "gfa2" as format names?

Adamtaranto:
Would it make sense to load the segment tags into the annotations dict for the SeqRecord?

peterjc:
Yes, except for the length and checksum perhaps?

```
Tag | Type | Description
-- | -- | --
LN | i | Segment length
RC | i | Read count
FC | i | Fragment count
KC | i | k-mer count
SH | H | SHA-256 checksum of the sequence
UR | Z | URI or local file-system path of the sequence. If it does not  start with a standard protocol (e.g. ftp), it is assumed to be a local  path.
```

Bio/SeqIO/GfaIO.py
michaelfm1211: Oops, yep you're right. I'll add a commit to fix this later today.

michaelfm1211:
@peterjc I just pushed a commit that should fix these things. I ended up putting the LN and SH tags in the annotations dict anyway for sake of simplicity. Is this okay?

Bio/SeqIO/GfaIO.py
```diff
@@ -29,7 +42,68 @@ def _check_tags(seq, tags):
...
-def GfaIterator(source):
+def _tags_to_annotations(tags):
+    """Build an annotations dictionary from a list of tags (PRIVATE)."""
+    annotations = {}
+    for tag in tags:
+        parts = tag.split(":")
+        if len(parts) < 3:
+            raise ValueError(f"Segment line has invalid tag: {tag}")
+        annotations[parts[0]] = parts[2]
+
```
Adamtaranto, line 53: Add check that `parts[0]` does not contain any whitespace. Can happen if users add their own tags and delimit with spaces instead of tabs.
Would check whole tag for whitespace but I think JSON type tags can contain spaces.

Adamtaranto:
Looks good!

What do you think about storing the annotations as a (TYPE:VALUE) tuple? i.e. `annotations = {'LN': ('i', 100)}`

That would allow you to reconstruct a GFA segment line if you want to support write to GFA later on.

Bio/SeqIO/GfaIO.py
michaelfm1211: Good idea.
One thing in the spec that I just now noticed is that `parts[0]` should match the RegEx `[A-Za-z][A-Za-z0-9]`, which would cover the whitespace issue too.

Adamtaranto (approved):
Looks good to me. Thanks for working on this!

michaelfm1211:
@peterjc Any thoughts on these changes?

Bio/SeqIO/GfaIO.py
```diff
@@ -0,0 +1,177 @@
+"""Bio.SeqIO support for the Graphical Fragment Assembly format.
+
+This format is output by many assemblers and includes linkage information for
+how the different sequences fit together, however, we just care about the
+segment (sequence) information.
+
+Documentation:
+    Version 1.x: https://gfa-spec.github.io/GFA-spec/GFA1.html
+    Version 2.0: https://gfa-spec.github.io/GFA-spec/GFA2.html
```
peterjc, line 9: This does not render nicely as RST, maybe make it into a bullet point list explicit?

Bio/SeqIO/GfaIO.py
```diff
@@ -0,0 +1,177 @@
...
+            if fields[2] == "*":
+                seq = Seq(None, length=0)
+            else:
+                seq = Seq(fields[2])
+
+            tags = fields[3:]
+            _check_tags(seq, tags)
+            annotations = _tags_to_annotations(tags)
+
+            yield SeqRecord(seq, name=fields[1], annotations=annotations)
```
peterjc, line 140: In general the ``.id`` is the primary identifier used in ``SeqIO``.

Leave ``.name`` as the default, or use the same value if you prefer.

peterjc:
Can you add one of the good GFA files to the ``test_SeqIO.py`` file? That ought to do some basic conversions like GFA to FASTA which might catch some interoperability issues (e.g. was not setting ``.id`` which I noticed by eye).

michaelfm1211:
@peterjc I just pushed a new commit that should fix those issues and add a test for GFA to test_SeqIO. Thanks for catching those things.

peterjc:
I think this deserves a mention in the NEWS file - and you'd be welcome to add yourself to the contributors for the next release and in the CONTRIB file. You can use the partly anonymised name on your GitHub profile if you want, or even just a github username (there are a couple of precendents).

peterjc (approved):

michaelfm1211:
@peterjc Great, thanks! I just pushed a new NEWS entry. Is it descriptive enough? If so, then everything should be ready.

peterjc:
Can we add "(first contribution)" to your NEWS.rst entry, and add the name and GitHub profile to CONTRIB.rst too?

Bio/SeqIO/GfaIO.py
```diff
@@ -29,7 +42,68 @@ def _check_tags(seq, tags):
...
+    for tag in tags:
+        parts = tag.split(":")
+        if len(parts) < 3:
+            raise ValueError(f"Segment line has invalid tag: {tag}")
+        annotations[parts[0]] = parts[2]
+
+        # Check type of the tag and raise warning on a mismatch. These RegExs
+        # are part of the 1.0 standard.
+        if parts[1] not in "AifZJHB":
+            warnings.warn(f"Tag has invalid type: {parts[2]}", BiopythonWarning)
```
Adamtaranto, line 63: I think the warning should actually print the type {parts[1]} here not the value field {parts[2]}.

Adamtaranto:
Added a comment. I think this bit should print {parts[1]}.

https://github.com/biopython/biopython/pull/4598/commits/60217f2ae7cd987a25c8d299fab7b4edf85ac111#r1467640306

Bio/SeqIO/GfaIO.py
michaelfm1211: Yep, thanks! I must have missed that while copying from the other tag type warnings.

michaelfm1211:
@Adamtaranto I just pushed a new commit that should fix that, thanks! It also fixes the NEWS entry.

peterjc:
Would you be happy with a squash-and-merge into a single commit?

michaelfm1211:
@peterjc Yes, a squash and merge would be good.

Adamtaranto:
@michaelfm1211 Noticed a bug in _check_tags(). 

Comment on the commit isn't showing up here so linking: https://github.com/biopython/biopython/commit/ea1fbfedcadd60855969355ec2e9463ceede8b0e

michaelfm1211:
@Adamtaranto Added a response to your comment on the commit, but I'll summarize it here too just for record-keeping:

In my opinion this potential bug isn't really a bug. Either way of looking at it, the potential bug doesn't affect the logic of the parser.

This PR (and the associated issue) is closed and unlikely to receive much attention, so if you'd like to get more opinions on this I would open a new issue.

PR #4613, opened 2024-01-31
add biomoltrans to pdb header parsing and mmtf writing

jamesmkrieger:

Added biomolecular assembly transformations to pdb header parser and mmtfio _save_structure. This means that this information can be read from pdb and included when saving to mmtf

This goes with using these functionalities in ProDy (see https://github.com/prody/ProDy/pull/1816)


peterjc:
Black would reformat Tests/test_PDB_parse_pdb_header.py - it might be that 3-line for statement I commented on, or something else?

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -104,6 +104,8 @@ def _save_structure(self, filepath, select):
             header_dict["structure_method"] = []
         else:
             header_dict["structure_method"] = [header_dict["structure_method"]]
+        if header_dict["biomoltrans"] == "":
```
JoaoRodrigues, line 107: This code would be much nicer if we used `dict.get` instead of `setdefault`, or simply 'if key not in dict: set key`

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -117,19 +119,20 @@ def _save_structure(self, filepath, select):
...
-            ]:  # 3 lines for 3x3 rotation and last column translations
-                matrix_items.extend([float(item) for item in line.split()])
-            matrix_items.extend([0.0, 0.0, 0.0, 1.0])
-
-            encoder.set_bio_assembly_trans(
-                bio_assembly_index=key,
-                input_chain_indices=[chain_ids.index(c) for c in value[0]],
-                input_transform=matrix_items,
-            )
+        if biomoltrans is not None:
```
JoaoRodrigues, line 122: This for me is very weird, because of the comment I made above. `header_dict["biomoltrans"]` should simply throw a KeyError if it's not defined, instead of being defined as None. At best, this should be replaced by something like:

```
biomoltrans = header_dict.get("biomoltrans")
if biolmoltrans is not None:
    # do stuff
```

JoaoRodrigues (changes requested):
Thanks for the contribution @jamesmkrieger. Would changing the code a little to address my comments be reasonable for you?

jamesmkrieger:
> Thanks for the contribution @jamesmkrieger. Would changing the code a little to address my comments be reasonable for you?

You're welcome. Of course!

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Would you like me to change the entries above too?

jamesmkrieger: I think the following should be good:
```
        if header_dict.get("biomoltrans") in ["", None]:
            header_dict["biomoltrans"] = None
```

jamesmkrieger: I tested this with an empty dict and it worked

jamesmkrieger: However, it looks like having a defaultdict initially as str gives the empty string when given any missing key like the current code:
```
In [1]: from collections import defaultdict

In [2]: header_dict = defaultdict(str, {})

In [3]: header_dict["biomoltrans"]
Out[3]: ''
```

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Because of using defaultdict, it was originally giving empty string and not a KeyError and we then set it as None. 

So, now asking if it's not None makes sense. I guess this is more secure against future changes though

Bio/PDB/mmtf/mmtfio.py
peterjc: I looked at this and wondered about the style, but it was consistent with the existing code so didn't flag it.

I wouldn't use a defaultdict here - that might have side effects downstream for people using the dictionary.

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -98,12 +98,14 @@ def _save_structure(self, filepath, select):
 
         # The header information is missing for some structure objects
         header_dict = defaultdict(str, self.structure.header)
```
JoaoRodrigues, line 100: I would simply keep this as a dict. The defaultdict will always return a value even if the key is not in the dictionary. That way you could simplify the code below like this:

```python
header_dict = dict(self.structure.header)  # copy
if "resolution" not in header_dict:
    header_dict["resolution"] = None

if "structure_method" not in header_dict:
    header_dict["structure_method"] = []

# Later around ln 123-ish
if "biomoltrans" in header_dict:
    # do the whole matrix stuff

File: Bio/PDB/parse_pdb_header.py [hunk 4]
```diff
@@ -15,10 +16,50 @@
...
+            elif line[13:18] == "BIOMT":
+                biomt = biomolecule[currentBiomolecule]
+                if line[13:19] == "BIOMT1":
+                    if applyToChains == []:
+                        applyToChains = biomt[0]
+                    biomt.append(applyToChains)
+                elif line[13:19]:
+                    applyToChains = []
+                biomt.append(line[23:])
+    return dict(biomolecule)
```
JoaoRodrigues, line 60: Nice touch! I'm picky about this!

Bio/PDB/parse_pdb_header.py
```diff
@@ -15,10 +16,50 @@
...
+    # REMARK 350
+    # REMARK 350 BIOMOLECULE: 1
+    # REMARK 350 AUTHOR DETERMINED BIOLOGICAL UNIT: MONOMERIC
+    # REMARK 350 APPLY THE FOLLOWING TO CHAINS: A
+    # REMARK 350   BIOMT1   1  1.000000  0.000000  0.000000        0.00000
+    # REMARK 350   BIOMT2   1  0.000000  1.000000  0.000000        0.00000
+    # REMARK 350   BIOMT3   1  0.000000  0.000000  1.000000        0.00000
+    biomolecule = defaultdict(list)
+    for line in inl:
+        if re.search(r"\AREMARK 350", line):
```
JoaoRodrigues, line 40: Would it simpler to just do `if line.startswith(...)` instead of calling `re.search`?

Tests/test_PDB_parse_pdb_header.py
```diff
@@ -170,6 +172,24 @@ def test_parse_pdb_with_remark_99(self):
...
+                "1": [
+                    ["A", "B"],
+                    "  1.000000  0.000000  0.000000        0.00000            \n",
+                    "  0.000000  1.000000  0.000000        0.00000            \n",
+                    "  0.000000  0.000000  1.000000        0.00000            \n",
+                ]
+            },
+        )
+
+        header = parse_pdb_header("PDB/1LCD.pdb")
```
JoaoRodrigues, line 190: I would separate these two tests, one to check for 2XHE and another to check that 1LCD does not have the remark.

JoaoRodrigues (changes requested):
Thanks!

Bio/PDB/parse_pdb_header.py
jamesmkrieger: Yes

Tests/test_PDB_parse_pdb_header.py
jamesmkrieger: good point

jamesmkrieger: It may well be that the second test doesn't pass anymore. I need to check what the behaviour is now following these changes.

jamesmkrieger: > It may well be that the second test doesn't pass anymore. I need to check what the behaviour is now following these changes.

ok, this is still like that

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Ok, changing this broke the tests so I need to change them accordingly 

@jamesmkrieger note to self

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -115,6 +116,22 @@ def _save_structure(self, filepath, select):
...
+            for key, value in biomoltrans.items():
+                matrix_items = []  # list of 16 items of 4x4 matrix
+                for line in value[1:]:
+                    # 3 lines for 3x3 rotation and last column translations
+                    matrix_items.extend([float(item) for item in line.split()])
+                matrix_items.extend([0.0, 0.0, 0.0, 1.0])
+
+                encoder.set_bio_assembly_trans(
+                    bio_assembly_index=key,
+                    input_chain_indices=[chain_ids.index(c) for c in value[0]],
```
JoaoRodrigues, line 131: Small nitpick: for small structures this won't matter but for larger ones (e.g. viral capsids) with hundreds of chains, it might be faster to compile the chain_ids into a dict with enumerate and use it here. `list.index` is O(n).

```python
chain_id_to_idx = {v: k for k, v in enumerate(c.id for c in self.structure.get_chains())}
```

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -174,9 +191,11 @@ def _save_structure(self, filepath, select):
                         )
                         encoder.set_chain_info(
                             chain_id=next(chain_id_iterator),
-                            chain_name="\x00"
-                            if len(chain.get_id().strip()) == 0
-                            else chain.get_id(),
+                            chain_name=(
+                                "\x00"
+                                if len(chain.get_id().strip()) == 0
```
JoaoRodrigues, line 196: I know this is not part of the PR but since you're here... this could be simplified to `"\x00" if not chain.id.strip() else chain.id.strip()`

JoaoRodrigues (approved):
If the tests are passing, I'd be OK to include this. See my comments for a couple of minor stylistic changes that you can consider optional.

jamesmkrieger:
> If the tests are passing, I'd be OK to include this. See my comments for a couple of minor stylistic changes that you can consider optional.

The tests are not all passing now

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Ok

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Makes sense. Thanks

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: Based on the tests, it's already being used in a way that needs it to be like that and people might get more unexpected effects if we change it. I think it's fine to have an empty field rather than a key error. 

There are also downstream programs like BioTite that require there to be some way of accessing every field even if it's empty so I think the original implementation makes sense here.

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: I have restored it to a defaultdict instead

> Ok, changing this broke the tests so I need to change them accordingly
> 
> @jamesmkrieger note to self

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -100,6 +100,7 @@ def _save_structure(self, filepath, select):
         header_dict = defaultdict(str, self.structure.header)
         if header_dict["resolution"] == "":
```
JoaoRodrigues, line 101: With the `defaultdict`, this key access will automatically create the key if it does not exist. I'd simply check with `dict.get` instead, like this:

```python
if header_dict.get("resolution") is None:  # does not create key
    header_dict["resolution"] = None
...
```

JoaoRodrigues (approved):

Bio/PDB/mmtf/mmtfio.py
peterjc: Yes - that surprising behaviour of the defaultdict caused me a major memory leak once!

Bio/PDB/mmtf/mmtfio.py
jamesmkrieger: I just spotted that this calculation had already been done outside the loop and put into a variable so I'm changing to using that

Bio/PDB/mmtf/mmtfio.py
```diff
@@ -128,7 +128,7 @@ def _save_structure(self, filepath, select):
 
                 chain_id_to_idx = {
                     v: k
-                    for k, v in enumerate(c.id for c in self.structure.get_chains())
+                    for k, v in enumerate(chain_ids)
```
peterjc, line 131: I think black would put this all on one line now

peterjc:
@JoaoRodrigues OK to squash-and-merge?

JoaoRodrigues:
Done, sorry, really busy week... I usually catch up on weekends!

PR #4627, opened 2024-02-13
Make SeqIO's import of MultipleSeqAlignment lazy.

peterjc:
If the file system is being slow (e.g. our cluster under heavy load), then the overhead of importing MultipleSeqAlignment from Bio.Align is noticeable if using Bio.SeqIO - but it is rarely needed.


mdehoon (commented):
I have not objections, but I am surprised that the import is slow. Once Bio.Align is compiled to pyc files, importing it should be fast.
I tried a bit, and it seems that importing Bio.File from Bio.Align.substitution_matrices contributes most to the import time.

peterjc:
Today our cluster has been stuggling to the point there is a lag of seconds for filename tab completion, and I'm using ``nano`` over ``emacs`` simply down to it opening less config files. There is some randomness though, so reliable timings are not easy - but I think change helped.

PR #4638, opened 2024-02-27
Fix inconsistent restriction digest cutting behaviour

### LINKED ISSUE #4604: Inconsistent enzyme.search behaviour when cutting sequence falls outside sequence
### Setup

system as follows:

```python
```

```
3.11.6 (main, Oct  6 2023, 11:55:02) [Clang 13.0.0 (clang-1300.0.29.30)]
CPython
macOS-12.7-x86_64-i386-64bit
1.83
```

### Steps to reproduce

cc: @BjornFJohansson

Let's start with the digestion of the following sequence `TAAAAAAAAAAAAGCCGGCAAAAAAATAAAAA` with the restriction enzyme that makes two cuts `NmeDI`. I am using this example to show what happens at both ends, but the same applies if you use one-cutters only. After digestion we would get 3 fragments:

```
T      AAAAAAAAAAAAGCCGGCAAAAAAA       TAAAAA
ATTTTT      TTTTTTTCGGCCGTTTTTTTATTTT       T
```

If we use the enzyme search, we get the positions (one-based) where these cuts will occur:

```python
from Bio.Seq import Seq
from Bio.Restriction import NmeDI

seq = Seq('TAAAAAAAAAAAAGCCGGCAAAAAAATAAAAA')
print(NmeDI.search(seq))

# Prints `[2, 27]`
```

If we start trimming the molecule from the sides, the sequence where the enzyme would make the cut and leave and overhang will disappear, and the behaviour is different when we trim the left and right side:

```python

for i in range(1,7):
    print(i, NmeDI.search(seq[i:]), NmeDI.search(seq[:-i]))

# Prints:
# 1 [26] [2, 27] < As soon as the initial T is removed, the left site is not returned anymore
# 2 [25] [2, 27]
# 3 [24] [2, 27]
# 4 [23] [2, 27]
# 5 [22] [2, 27]
# 6 [21] [2]     < Until the final T is removed, the right site is still returned
```

I don't know what the right thing to do is. My feeling is that probably the behaviour of the first trimming on either side should work, and the rest not? This would mean accepting a cut like this, where the cutting sites lie exactly at the end, and single-strand DNA is generated (see below). I am not sure if biologically a cut would happen with further trimming.

```
      AAAAAAAAAAAAGCCGGCAAAAAAA       TAAAA
TTTTT      TTTTTTTCGGCCGTTTTTTTATTTT
```

manulera:

Closes #4604

@peterjc I have also refactored the code, since there was some code repeated up to three times in the different `_drop` functions. Let me know if that makes sense.


manulera:
Hi @peterjc I am not sure test that does not pass is actually wrong. The failed test can be reproduced with `python run_tests.py "Bio.Restriction"`.

I am not sure, but If I understand the error correctly, it is checking that the output of `a.print_that()` in the docstring of the file `Bio/restricntion/__init__.py` matches the actual content of the docstring. The output would of course be much longer than that. Maybe a flag to not run this test in that docstring should be included?

peterjc:
I would find this easier to read if the fix and the refactoring were done as two separate commits. Would that be easy to do in principle, and is your git knowledge up to it?

(I've not looked at the change, nor quite how the refactoring works - but agree there was duplicated code)

manulera:
Hi @peterjc I splitted the change as you requested

Bio/Restriction/Restriction.py
```diff
@@ -548,11 +547,13 @@ def _drop(cls):
...
-        take = itertools.takewhile
         if cls.dna.is_linear():
-            cls.results = list(drop(lambda x: x <= 1, cls.results))
-            cls.results = list(take(lambda x: x <= length, cls.results))
+
+            def filtering_function(cut_on_watson):
+                cut_on_crick = cut_on_watson - cls.ovhg
+                return all(1 < x <= length for x in (cut_on_watson, cut_on_crick))
+
+            cls.results = list(filter(filtering_function, cls.results))
```
peterjc, line 556: A style query, as I didn't immediately remember how filter worked (not indeed the drop and take methods in the older code). See e.g. https://realpython.com/python-filter-function/#replacing-filter-with-a-list-comprehension

```python
# Generating a list with filter()
list(filter(function, iterable))

# Generating a list with a list comprehension
[item for item in iterable if function(item)]
```

How about something like:

```python
def filtering_function(cut_on_watson):
    cut_on_crick = cut_on_watson - cls.ovhg
    return (1 < cut_on_watson <= length) and (1 < cut_on_crick <= length)

cls.results = [cut for cut in cls.results if filtering_function(cut)]
```

Although I feel this can also be improved on...

peterjc:
The refactoring as a separate commit is much clearer to me - thank you!

Bio/Restriction/Restriction.py
manulera: I have made that change

peterjc:
Over to you @MarkusPiotrowski - this looks good to me from a style/technical point of view.

MarkusPiotrowski:
Yes, I'll have a look at it.

manulera:
Hi @MarkusPiotrowski just following up on the review request.

manulera:
Hi @peterjc @MarkusPiotrowski this has been pending since February, not sure if you can find someone else to review the PR. The changes are fairly small:
* Refactor to remove code duplication: https://github.com/biopython/biopython/pull/4638/commits/81c0bdb1938579020eb58ee4c60f16f6a69431c1
* Feature implementation (-5 +7 lines) + tests: https://github.com/biopython/biopython/pull/4638/commits/bb13298ac46e75738644d293f20116b1cb9578f6
* Refactor of feature based on @peterjc review: https://github.com/biopython/biopython/pull/4638/commits/a2d2e4880ce34ed5796e783c379a9196b582d9f2

manulera:
Hi @peterjc pinging again. I have implemented this on my side, but I would like to use a version of biopython with this working. Can someone else review the PR?

peterjc (approved):
Waiting on @MarkusPiotrowski for the final say.

manulera:
Hi @MarkusPiotrowski any chance you will be able to review this PR anytime soon?

MarkusPiotrowski:
@manulera : Can you add another test to show that e.g. `BsaI.search(Seq("GGTCTCATAAAA")` now gives the same result as `BsaI.search(Seq("GGTCTCATAAAA")`  ( = [] )?

Otherwise, this looks good to me, working as I would expect it to work.

manulera:
Hi @MarkusPiotrowski I am not sure what you mean, both code examples you gave contain the same statement `BsaI.search(Seq("GGTCTCATAAAA")`.

There is a test for that (see [Tests/test_Restriction.py](https://github.com/biopython/biopython/pull/4638/files#diff-8418489ad1873bf96b480e8e1643f3b27a6c6376c74ca5bda42a282eccff045f)), but from what we discussed in #4604 the expected result is [8], not an empty list. An empty list should be if there was one fewer `A` (line below), since there would be no base on the right side of the cut.

```python
        # Examples from https://github.com/biopython/biopython/issues/4604
        self.assertEqual(BsaI.search(Seq("GGTCTCATAAAA")), [8])
        self.assertEqual(BsaI.search(Seq("GGTCTCATAAA")), [])
```

Screenshot 2024-01-25 at 11 15 17

MarkusPiotrowski:
Sorry, copy/pasting from the console didn't work as expected. What I really meant is:
`BsaI.search(Seq('GGTCTCGT'))`
and
`BsaI.search(Seq('GGTCTCGT').reverse_complement())`

manulera:
Ok! I included that test. Thanks for reviewing. Should be ready to merge when tests pass

MarkusPiotrowski (approved):

manulera:
I think the tests that fail are unrelated to my change

MarkusPiotrowski:
They are possibly related to #4751 #4752 

We need to wait until this is sorted out.

peterjc:
We should retest this now that the master branch is working again (several regressions from numpy v2 and other library updates).

MarkusPiotrowski:
@manulera Can you do a rebase on the current master? See https://biopython.org/wiki/GitUsage

manulera:
There it is!

peterjc:
Merged, thank you!

PR #4641, opened 2024-02-28
Static typing for Bio.PDB

speleo3:
`mypy` type checking was disabled for `Bio.PDB` and the module started to accumulate broken type annotations.

- Enable type checking for `Bio.PDB`
  - With some exceptions. Notably the newly added `PDBMLParser` -- maybe @Will-Tyler wants to fix the typing there?
- Fix type annotation errors
- Add more type annotations

Related to https://github.com/biopython/biopython/issues/2236

----

### DIFF
--- a/Bio/PDB/Entity.py
+++ b/Bio/PDB/Entity.py
@@ -23,20 +23,20 @@ from Bio.PDB.PDBExceptions import PDBConstructionException
 if TYPE_CHECKING:
     from Bio.PDB.Atom import Atom
 
-Child = TypeVar("Child", bound=Union["Entity", "Atom"])
-Parent = TypeVar("Parent", bound=Optional["Entity"])
+_Child = TypeVar("_Child", bound=Union["Entity", "Atom"])
+_Parent = TypeVar("_Parent", bound=Optional["Entity"])
 
 
-class Entity(Generic[Parent, Child]):
+class Entity(Generic[_Parent, _Child]):
     """Basic container object for PDB hierarchy.
 
     Structure, Model, Chain and Residue are subclasses of Entity.
     It deals with storage and lookup.
     """
 
-    parent: Optional[Parent]
-    child_list: List[Child]
-    child_dict: Dict[Any, Child]
+    parent: Optional[_Parent]
+    child_list: List[_Child]
+    child_dict: Dict[Any, _Child]
     level: str
 
     def __init__(self, id):
@@ -212,7 +212,7 @@ class Entity(Generic[Parent, Child]):
         """
         return self.level
 
-    def set_parent(self, entity: Parent):
+    def set_parent(self, entity: _Parent):
         """Set the parent Entity object."""
         self.parent = entity
         self._reset_full_id()
@@ -228,7 +228,7 @@ class Entity(Generic[Parent, Child]):
         del self.child_dict[id]
         self.child_list.remove(child)
 
-    def add(self, entity: Child):
+    def add(self, entity: _Child):
         """Add a child to the Entity."""
         entity_id = entity.get_id()
         if self.has_id(entity_id):
@@ -237,7 +237,7 @@ class Entity(Generic[Parent, Child]):
         self.child_list.append(entity)
         self.child_dict[entity_id] = entity
 
-    def insert(self, pos: int, entity: Child):
+    def insert(self, pos: int, entity: _Child):
         """Add a child to the Entity at a specified position."""
         entity_id = entity.get_id()
         if self.has_id(entity_id):
--- a/Bio/PDB/Entity.py
+++ b/Bio/PDB/Entity.py
@@ -13,6 +13,7 @@ import warnings
 
 from collections import deque
 from copy import copy
+from typing import Any, Dict, Generic, List, Optional, Protocol, TypeVar
 
 import numpy as np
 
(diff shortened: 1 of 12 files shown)


Bio/PDB/Entity.py
```diff
@@ -13,20 +13,37 @@
 
 from collections import deque
 from copy import copy
+from typing import Any, Dict, Generic, List, Optional, Protocol, TypeVar
```
peterjc, line 16: I'm not familiar with Protocol and TypeVar here, but these Entity changes aside it all looks fine to me.

Bio/PDB/StructureBuilder.py
```diff
@@ -185,15 +186,15 @@ def init_residue(self, resname: str, field: str, resseq: int, icode: str):
     def init_atom(
         self,
         name: str,
-        coord: np.array,
+        coord: np.ndarray,
```
peterjc, line 189: Why?

speleo3: [np.array](https://numpy.org/doc/stable/reference/generated/numpy.array.html) is not a type but a function which returns an [np.ndarray](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html).

mypy reports:
```
error: Function "numpy.core.multiarray.array" is not valid as a type  [valid-type]
```

Bio/PDB/Entity.py
speleo3: I found a slightly more elegant solution which doesn't need `Protocol`. Can bind directly to `Union["Entity", "Atom"]` and to `Optional["Entity"]` instead.

Bio/PDB/StructureBuilder.py
peterjc: Thanks - I guess I'll get used to it, but I haven't been type annotating with numpy yet.

I presume that is long standing and doesn't need a particularly recent numpy dependency?

Bio/PDB/Entity.py
```diff
@@ -13,20 +13,32 @@
...
 
 from Bio import BiopythonWarning
 from Bio.PDB.PDBExceptions import PDBConstructionException
 
+if TYPE_CHECKING:
+    from Bio.PDB.Atom import Atom
 
-class Entity:
+Child = TypeVar("Child", bound=Union["Entity", "Atom"])
+Parent = TypeVar("Parent", bound=Optional["Entity"])
```
peterjc, line 27: Maybe make these private for now, rather than expanding the public API? i.e. Add leading underscore.

Bio/PDB/PDBList.py
```diff
@@ -484,14 +484,14 @@ def get_all_assemblies(self, file_format: str = "") -> List[Tuple[str, str]]:
...
-            "Content-Length": len(data),
+            "Content-Length": str(len(data)),
         }
         request = Request("https://search.rcsb.org/rcsbsearch/v2/query", data, headers)
         with urlopen(request) as response:
             assemblies = json.loads(response.read().decode("utf-8"))["result_set"]
 
         # We transform the assemblies to match the format that they have historically been returned in.
-        def transform(assembly: dict) -> tuple[str, str]:
+        def transform(assembly: dict) -> Tuple[str, str]:
```
peterjc, line 494: I believe (and am happy to be informed otherwise), that since Python 3.9 (and PEP-0585) it is preferred to use tuple over Tuple (via an import from typing).

Bio/PDB/Entity.py
```diff
@@ -13,20 +13,32 @@
 
 from collections import deque
 from copy import copy
+from typing import TYPE_CHECKING, Any, Dict, Generic, List, Optional, TypeVar, Union
```
peterjc, line 16: Repeating myself from a comment about Tuple, but likewise rather than importing Dict and List from typing, shouldn't we just used dict and list (lower case)?

peterjc:
Unclear why the CI failed, perhaps something in the macOS environment - it may sort itself out next time, I have not bothered with a re-run:

```
 gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/usr/local/opt/sqlite/include -I/usr/local/opt/sqlite/include -I/Users/runner/hostedtoolcache/Python/3.10.13/x64/include/python3.10 -c Bio/Align/_codonaligner.c -o build/temp.macosx-11.7-x86_64-cpython-310/Bio/Align/_codonaligner.o
gcc: error: unable to locate xcodebuild, please make sure the path to the Xcode folder is set correctly!
gcc: error: You can set the path to the Xcode folder using /usr/bin/xcode-select -switch
error: command '/usr/bin/gcc' failed with exit code 71
```

Bio/PDB/PDBList.py
speleo3: Shouldn't we drop Python 3.8 support first? I still see it in `README.rst`, `setup.py`, and `.github/workflows/ci.yml`.

peterjc: Ah - good point. We've warned that our Python 3.8 is coming to an end, probably we should move to deprecating it formally in our next release. That should be in line with the end of Python 3.8 security updates in Oct 2024 https://devguide.python.org/versions/

JoaoRodrigues:
LGTM, I have no comments on this - I personally don't see the point of typing in Python tbh 😄

peterjc:
@speleo3 Squash-and-merge OK with you?

speleo3:
> Squash-and-merge OK with you?

yes perfect, thanks!

Bio/PDB/Entity.py
speleo3: will do a big search-and-replace once we've dropped Python 3.8 :-)

Bio/PDB/StructureBuilder.py
speleo3: > I presume that is long standing and doesn't need a particularly recent numpy dependency?

that's correct.

PR #4650, opened 2024-03-05
SSBond parsing addition - 2nd attempt

suchanek:

- [ x] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``

I've added SSBOND parsing to parse_pdb_header.py. This adds an SSBOND key to the PDB header structure, allowing parsing of disulfide bonds. I'm using this with https://github.com/suchanek/proteusPy . please remove the prior pull request.

Closes #...


Bio/PDB/parse_pdb_header.py
```diff
@@ -21,6 +20,23 @@
 from Bio import File
 
 
+# Added to parse SSBOND records from the PDB header and add them back to the header dict
+# Eric G. Suchanek, PhD 11/25/22
+def _get_ssbond(inl):
```
peterjc, line 25: This is a private function (leading underscore) so we don't insist on a docstring, but the comments here might as well be transformed into a docstring. Note there are style rules (one line summary, blank line, more text etc), so do run the pre-commit checks as per https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

Bio/PDB/parse_pdb_header.py
```diff
@@ -333,6 +354,13 @@ def _parse_pdb_header_list(header):
                 pdbh_dict["author"] += auth
             else:
                 pdbh_dict["author"] = auth
+        # parse SSBOND records and add them to the Header dict
+        elif key == "SSBOND":
+            ssbond_numb += 1
+            ssb = _get_ssbond(tail)
+            ssbdict = {ssbond_numb: ssb}
+            pdbh_dict["ssbond"].update(ssbdict)
+        # print(f'SSBOND: {tail}')
```
peterjc, line 363: You can remove the commented out print statement.

peterjc (changes requested):
Currently this records a dictionary of 4-entry tuples, but only from reading the source code can you guess what they are and even then the names are cryptic: ``(prox, dist, chn, chn2)``.

For a start, please expand those into full words for clarity - and then perhaps use a named tuple or a dictionary instead of a tuple?

What ever you do, I think the SSBOND parsing needs to be documented in a public docstring.

Also, this really needs a test. I suggest extending the unit test already parsing ``Tests/PDB/1A8O.pdb`` (one SSBOND line) or better ``Tests/PDB/7DDO.pdb`` which has seven.

suchanek:
Inadvertently removed due to my inexperience. Will restore and incorporate your comments. Thank you.

Cheers,
Eric

—

Eric G. Suchanek, PhD

Do good, be good, make the world a better place. It only takes ONE person to change the world.

***@***.***

Bio/PDB/parse_pdb_header.py
```diff
@@ -166,7 +182,8 @@ def parse_pdb_header(infile):
             record_type = line[0:6]
             if record_type in ("ATOM  ", "HETATM", "MODEL "):
                 break
-            header.append(line)
+            else:
```
JoaoRodrigues, line 185: No need for the else here. The for loop will run until it his one of the record_types above and then will break (stop the for-loop).

Bio/PDB/parse_pdb_header.py
```diff
@@ -219,6 +236,7 @@ def _parse_remark_465(line):
 
 def _parse_pdb_header_list(header):
     # database fields
+    # added ssbond 11/20/22 Eric G. Suchanek
```
JoaoRodrigues, line 239: I'd remove these comments. Git will track these changes and keep your authorship intact if anyone wishes to know who wrote this code.

Bio/PDB/parse_pdb_header.py
```diff
@@ -333,6 +354,13 @@ def _parse_pdb_header_list(header):
                 pdbh_dict["author"] += auth
             else:
                 pdbh_dict["author"] = auth
+        # parse SSBOND records and add them to the Header dict
+        elif key == "SSBOND":
```
JoaoRodrigues, line 358: Why not do something like in lines `259-261` above and use the `_get_ssbond` method to directly grab the ss bond info from the header? That would make the function much easier to test (a single call) and you can do all the manipulation you need inside of it.

In other words:

```python
pdbh_dict["ssbond"] = _get_ssbond(header)
```

Bio/PDB/parse_pdb_header.py
```diff
@@ -21,6 +20,23 @@
 from Bio import File
 
 
+# Added to parse SSBOND records from the PDB header and add them back to the header dict
+# Eric G. Suchanek, PhD 11/25/22
+def _get_ssbond(inl):
+    # SSBOND   1 CYS A   26    CYS A   84                          1555   1555  2.04
+    # CYS A   26    CYS A   84                          1555   1555  2.0
+    # print(inl)
+    tok = inl.split()
```
JoaoRodrigues, line 29: Please do not use split when dealing with PDB files. The format is character-delimited, so you might run into cases where e.g. the residue number is large enough to leave no space between itself and the next field and split will fail.

See [here](https://www.wwpdb.org/documentation/file-format-content/format33/sect6.html#SSBOND) for the description of the SSBOND records.

JoaoRodrigues (changes requested):

PR #4651, opened 2024-03-05
Remove type annotations from strictly_equals methods in Bio.PDB

JoaoRodrigues:
For some reason, all checks passed on #4593 but after merging mypy started complaining. The issue seems to be overloading `strictly_equals` in Entity subclasses (and in Atom.py) violates some typing rule. I cannot figure out a way to handle this, so I simply removed the type annotation from these methods..

### DIFF
--- a/Bio/PDB/Atom.py
+++ b/Bio/PDB/Atom.py
@@ -267,7 +267,7 @@ class Atom:
         diff = self.coord - other.coord
         return np.sqrt(np.dot(diff, diff))
 
-    def strictly_equals(self, other: "Atom", compare_coordinates: bool = False) -> bool:
+    def strictly_equals(self, other, compare_coordinates: bool = False) -> bool:
         """Compare this atom to the other atom using a strict definition of equality.
 
         Indicates whether the atoms have the same name, B factor, occupancy,
--- a/Bio/PDB/Entity.py
+++ b/Bio/PDB/Entity.py
@@ -201,16 +201,14 @@ class Entity(Generic[_Parent, _Child]):
         self._id = value
         self._reset_full_id()
 
-    def strictly_equals(
-        self, other: "Entity", compare_coordinates: bool = False
-    ) -> bool:
+    def strictly_equals(self, other, compare_coordinates: bool = False) -> bool:
         """Compare this entity to the other entity for equality.
 
         Recursively compare the children of this entity to the other entity's children.
         Compare most properties including names and IDs.
 
         :param other: The entity to compare this entity with
-        :type other: Entity
+        :type other: Entity or Atom
         :param compare_coordinates: Whether to compare atomic coordinates
         :type compare_coordinates: bool
         :return: Whether the two entities are strictly equal
--- a/Bio/PDB/Residue.py
+++ b/Bio/PDB/Residue.py
@@ -43,9 +43,7 @@ class Residue(Entity["Chain", "Atom"]):
         full_id = (resname, hetflag, resseq, icode)
         return "<Residue %s het=%s resseq=%s icode=%s>" % full_id
 
-    def strictly_equals(
-        self, other: "Residue", compare_coordinates: bool = False
-    ) -> bool:
+    def strictly_equals(self, other, compare_coordinates: bool = False) -> bool:
         """Compare this residue to the other residue using a strict definition of equality.
 
         The residues are equal if they have the same name, identifier,


peterjc (approved):
Pragmatic - I've only started doing type annotations in my own code fairly recently, debugging them is still a big of mystery sometimes.

bzoracler:
This happens because types of function/method arguments are contravariant (the Python typing specification, especially the [Variance of generic types section on `● Callable`](https://typing.readthedocs.io/en/latest/source/generics.html#variance-of-generic-types), is highly relevant).

The easiest fix is to use [`typing.Self`](https://docs.python.org/3/library/typing.html#typing.Self) like this (demo: [mypy Playground](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&gist=46930a3c21f78b494a003ec954a98343))

```python
class Entity(Generic[_Parent, _Child]):
    def strictly_equals(self, other: Self, compare_coordinates: bool = False) -> bool: ...

class Residue(Entity["Chain", "Atom"]):
    def strictly_equals(self, other: Self, compare_coordinates: bool = False) -> bool: ...

class Atom:
    def strictly_equals(self, other: Self, compare_coordinates: bool = False) -> bool: ...
```

However, since this was only introduced in Python 3.11, and Biopython probably (?) doesn't want `typing_extensions` to be a runtime dependency, you can either do this

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # This is fine regardless whether it is actually installed, because all type checkers rely on their own `typing_extensions` `.pyi` stubs, not any runtime installation.
    # ---
    # mypy: https://github.com/python/mypy/blob/master/mypy/typeshed/stdlib/typing_extensions.pyi
    # pyright: https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/typeshed-fallback/stdlib/typing_extensions.pyi
    # pytype: https://github.com/google/pytype (`[typeshed @ <commit hash>]`)
    # pyre: https://github.com/facebook/pyre-check/blob/main/stubs/typeshed/typeshed/stdlib/typing_extensions.pyi
    from typing_extensions import Self

class Entity(Generic[_Parent, _Child]):
    def strictly_equals(self, other: "Self", compare_coordinates: bool = False) -> bool: ...
```

Or this ([mypy Playground](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&gist=bcab7f7c42bdc888058bfa4fceb66b0b)):

```python
_EntityT = TypeVar("_EntityT", bound="Entity")
_ResidueT = TypeVar("_ResidueT", bound="Residue")
_AtomT = TypeVar("_AtomT", bound="Atom")

class Entity(Generic[_Parent, _Child]):
    def strictly_equals(self: _EntityT, other: _EntityT, compare_coordinates: bool = False) -> bool: ...

class Residue(Entity["Chain", "Atom"]):
    def strictly_equals(self: _ResidueT, other: _ResidueT, compare_coordinates: bool = False) -> bool: ...

class Atom:
    def strictly_equals(self: _AtomT, other: _AtomT, compare_coordinates: bool = False) -> bool: ...
```

bzoracler:
@peterjc PR made at https://github.com/biopython/biopython/pull/4653

A side-question - would Biopython accept PRs made purely for improving type annotations? Perhaps sparsely, on a per-module basis.

peterjc:
Personally I am happy with adding type annotations, although not all module owners may be keen (e.g. some modules are under fairly active revision where this might hamper a developer unfamiliar with them). It might be best to start a tracking issue and confirm this on a module by module basis?

peterjc:
We'll probably close this in favour of #4653...

bzoracler:
> Personally I am happy with adding type annotations, although not all module owners may be keen (e.g. some modules are under fairly active revision where this might hamper a developer unfamiliar with them). It might be best to start a tracking issue and confirm this on a module by module basis?

That's certainly a lot of modules in Biopython to track!

I understand the concern about familiarity with typing and active development. There's a way around this by using [`.pyi` type stubs](https://typing.readthedocs.io/en/latest/source/stubs.html#stubs) which live *beside* the runtime module. However, if we do this, the symbols in the runtime module (`.py` file) and stub module (`.pyi` file) need to be in-sync, otherwise symbols added or deleted in the runtime module won't be picked up by type-checkers or other linters. `.pyi`s also come with their own linting rules and formatting guides, so that may complicate the CI process (but hopefully not by much). The [root numpy directory](https://github.com/numpy/numpy/tree/main/numpy) is an example of what that may look like.

The tracking issue to track module-by-module sounds great, although I think it should come from a core/frequent developer who is more familiar with Biopython and are more familiar with its contribution history and module owners. Given the number of modules and module owners in this project, perhaps it'd be better for the owners themselves, or a core developer, to add to a *gradual* list of modules or sub-packages in Biopython which welcome type-annotation contributions (and mention whether they'd like them inline or as a `.pyi` stub). That way we don't get angry owners or frustrated contributors :)

peterjc:
For now I'd suggest Bio.PDB (already started), Bio.bgzf (self contained), and then maybe Bio.SeqIO (which may in turn mean doing some of Bio/Seq.py, Bio/SeqRecord.py, Bio/SeqFeature.py)

PR #4654, opened 2024-03-07
Feat/add bgzf typing

bzoracler:

---

As discussed in https://github.com/biopython/biopython/pull/4651#issuecomment-1980850966, this is a conversion of `Bio.bgzf` to [`--strict`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict) static typing. Setting this as a draft PR for feedback.

Part of this required making some runtime changes, which are done in the following commits:
 - https://github.com/biopython/biopython/commit/45fd6093d8c46bf186a04f9ddb7570caa1e5ea22
 - https://github.com/biopython/biopython/commit/a83311c47301f28b27b6884e602642f76f285d5a
 - https://github.com/biopython/biopython/commit/6fba8bb98302921e498f2e525c24d8ac1adf89a0
 - https://github.com/biopython/biopython/commit/79c981b8cb864bdfaeb0278dfd6f665a87b70c8b

I *believe* these taken together are fully compatible with the usage of the module as given by the specifications, but this needs thorough review.

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -516,13 +516,13 @@ def _load_bgzf_block(handle, text_mode=False):
     if expected_size != len(data):
         raise RuntimeError("Decompressed to %i, not %i" % (len(data), expected_size))
     # Should cope with a mix of Python platforms...
-    crc = zlib.crc32(data)
-    if crc < 0:
-        crc = struct.pack("<i", crc)
+    checksum = zlib.crc32(data)
+    if checksum < 0:
+        crc = struct.pack("<i", checksum)
     else:
-        crc = struct.pack("<I", crc)
+        crc = struct.pack("<I", checksum)
     if expected_crc != crc:
-        raise RuntimeError(f"CRC is {crc}, not {expected_crc}")
+        raise RuntimeError(f"CRC is {crc!r}, not {expected_crc!r}")
     if text_mode:
         # Note ISO-8859-1 aka Latin-1 preserves first 256 chars
         # (i.e. ASCII), but critically is a single byte encoding
@@ -638,11 +638,12 @@ class BgzfReader(tp.Generic[_BufferT]):
                 "Must use a read mode like 'r' (default), 'rt', or 'rb' for binary"
             )
         # If an open file was passed, make sure it was opened in binary mode.
-        if fileobj:
+        if fileobj is not None:
             if fileobj.read(0) != b"":
                 raise ValueError("fileobj not opened in binary mode")
             handle = fileobj
         else:
+            assert filename is not None
             handle = _open(filename, "rb")
         self._text = "b" not in mode.lower()
         self._empty_buffer: _BufferT
@@ -842,11 +843,12 @@ class BgzfWriter:
         if filename and fileobj:
             raise ValueError("Supply either filename or fileobj, not both")
         # If an open file was passed, make sure it was opened in binary mode.
-        if fileobj:
+        if fileobj is not None:
             if fileobj.read(0) != b"":
                 raise ValueError("fileobj not opened in binary mode")
             handle = fileobj
         else:
+            assert filename is not None
             if not _is_writing_appending_mode(mode.lower()):
                 raise ValueError(f"Must use write or append mode, not {mode!r}")
             if "a" in mode.lower():
@@ -874,12 +876,6 @@ class BgzfWriter:
             raise RuntimeError(
                 "TODO - Didn't compress enough, try less data in this block"
             )
-        crc = zlib.crc32(block)
-        # Should cope with a mix of Python platforms...
-        if crc < 0:
-            crc = struct.pack("<i", crc)
-        else:
-            crc = struct.pack("<I", crc)
         bsize = struct.pack("<H", len(compressed) + 25)  # includes -1
         crc = struct.pack("<I", zlib.crc32(block) & 0xFFFFFFFF)
         uncompressed_length = struct.pack("<I", len(block))
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -258,11 +258,13 @@ import zlib
 from builtins import open as _open
 
 if tp.TYPE_CHECKING:
+    import _typeshed
     import io
     import collections.abc as cx_abc
     from typing_extensions import TypeAlias, TypeGuard
 
 _BufferT = tp.TypeVar("_BufferT", str, bytes)
+_Self = tp.TypeVar("_Self")
 # fmt: off
 # Writing and updating modes
 _ReadingBinaryMode: TypeAlias = tp.Literal["rb", "br"]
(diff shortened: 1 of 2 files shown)


bzoracler:
Some of these style failures conflict with black's defaults. I'll just disable the codes for this file for now.

Bio/bgzf.py
```diff
@@ -476,13 +540,13 @@ def _load_bgzf_block(handle, text_mode=False):
...
-    crc = zlib.crc32(data)
-    if crc < 0:
-        crc = struct.pack("<i", crc)
+    checksum = zlib.crc32(data)
+    if checksum < 0:
+        crc = struct.pack("<i", checksum)
     else:
-        crc = struct.pack("<I", crc)
+        crc = struct.pack("<I", checksum)
     if expected_crc != crc:
```
peterjc, line 548: You still have some ``crc`` usage despite changing the varible name above...

Bio/bgzf.py
```diff
@@ -593,30 +675,37 @@ def __init__(self, filename=None, mode="r", fileobj=None, max_cache=100):
...
+            self._empty_buffer = b""  # type: ignore[assignment]
+            self._newline = b"\n"  # type: ignore[assignment]
         self._handle = handle
         self.max_cache = max_cache
-        self._buffers = {}
-        self._block_start_offset = None
-        self._block_raw_length = None
+        self._buffers: dict[int, tuple[_BufferT, int]] = {}
+        # Is there a safe default for a non-`None` integer?
+        self._block_start_offset: int = None  # type: ignore[assignment]
```
peterjc, line 703: Optional[int] perhaps?

We can't used zero here.

Bio/bgzf.py
```diff
@@ -247,19 +247,56 @@
...
+    import _typeshed
+    import io
+    import collections.abc as cx_abc
+    from typing_extensions import TypeAlias, TypeGuard
+
+_BufferT = tp.TypeVar("_BufferT", str, bytes)
+_Self = tp.TypeVar("_Self")
+# fmt: off
+# Writing and updating modes
+_ReadingBinaryMode: TypeAlias = tp.Literal["rb", "br"]
```
peterjc, line 270: Wow, esentially an enum - does this work nicely in a compatible code editor for suggesting values? Until now I'd have just annotated this as ``str``.

Bio/bgzf.py
```diff
@@ -556,7 +620,25 @@ class BgzfReader:
...
+    @tp.overload
+    def __init__(self: BgzfReader[str], *, mode: _ReadingTextMode = "r", fileobj: tp.BinaryIO, max_cache: int = 100) -> None: ...
+    @tp.overload
+    def __init__(self: BgzfReader[bytes], filename: _typeshed.StrPath, mode: _ReadingBinaryMode, *, max_cache: int = 100) -> None: ...
+    @tp.overload
+    def __init__(self: BgzfReader[bytes], *, mode: _ReadingBinaryMode, fileobj: tp.BinaryIO, max_cache: int = 100) -> None: ...
+    # fmt: on
+    def __init__(
+        self,
+        filename: _typeshed.StrPath | None = None,
```
peterjc, line 637: At what point can we use ``XXX | None`` rather than ``Optional[XXX]``?

Bio/bgzf.py
```diff
@@ -247,19 +247,56 @@
...
+@tp.overload
+def open(filename: _typeshed.StrPath, mode: _WritingAppendingBinaryMode) -> BgzfWriter[bytes]: ...
+@tp.overload
+def open(filename: _typeshed.StrPath, mode: _ReadingTextMode) -> BgzfReader[str]: ...
+@tp.overload
+def open(filename: _typeshed.StrPath, mode: _WritingAppendingTextMode) -> BgzfWriter[str]: ...
+# fmt: on
+def open(
+    filename: _typeshed.StrPath, mode: str = "rb"
+) -> BgzfReader[str] | BgzfWriter[str] | BgzfReader[bytes] | BgzfWriter[bytes]:
```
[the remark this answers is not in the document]
bzoracler: The addition of `from __future__ import annotations` ([PEP 563](https://peps.python.org/pep-0563/)) at the top of this module implicitly converts all annotations in the module to strings. Essentially, this means that typing features from newer versions of Python can be used in older versions of Python, as long as those features aren't grammar changes (causing syntax errors).

The `|` syntax is introduced in Python 3.10 in [PEP 604](https://peps.python.org/pep-0604/). However, because of the stringification of the annotations, this will work in Python 3.8.

Bio/bgzf.py
bzoracler: Some language server protocols might work well with `typing.Literal` - I use PyCharm and unfortunately it doesn't seem to do so :(

Bio/bgzf.py
bzoracler: This is deliberate. Static type checkers won't like `crc` being of two different types:
 - `crc` is first inferred to be an `int` from the line `crc = zlib.crc32(data)`
 - Type checkers will subsequently complain on the line `crc = struct.pack(...)`, because `struct.pack(...)` returns a `bytes` object.

peterjc: Ah, I've hit that kind of issue myself - using a new variable name does seem to be clearest. I may have not read this change carefully enough.

Bio/bgzf.py
peterjc: Yes, I have had to introduce occasional ``assert isintance(...)`` lines in my own project. As you say, largely visual noise :(

Bio/bgzf.py
```diff
@@ -247,19 +247,56 @@
...
 _bgzf_magic = b"\x1f\x8b\x08\x04"
 _bgzf_header = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00\x42\x43\x02\x00"
 _bgzf_eof = b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00BC\x02\x00\x1b\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00"
 _bytes_BC = b"BC"
 
 
-def open(filename, mode="rb"):
+# fmt: off
+@tp.overload
+def open(filename: _typeshed.StrPath, mode: _ReadingBinaryMode = "rb") -> BgzfReader[bytes]: ...
```
peterjc, line 289: Is it these lines triggering ``D103 missing docstring in public functions``? That feels like a false positive - indeed https://github.com/PyCQA/pydocstyle/issues/419 agreed.

However, pydocstyle is now deprecated. I will look at droping this from our flake8 configuration and enabling it via ruff as suggested.

XM-Helico: Yep - `@typing.overload`ed functions are usually considered fully-fledged functions, so docstring style-checkers may complain if you don't add something to them. I don't use flake8 in my own projects, but I had a similar experience with https://github.com/HunterMcGushion/docstr_coverage.

I don't know if I'd consider it strictly a false positive, just because if you have more than 3 overloads that have wildly different signatures, it's probably not conducive to cram all of their explanations into one function's docstring.

PR #4655, opened 2024-03-07
Pre-commit tools and black update

peterjc:

Some minor housekeeping with the style checking.


peterjc:
Cross reference https://github.com/PyCQA/pyflakes/issues/804 for E704 false positives

JoaoRodrigues:
Not necessarily related to this PR, but should we have a look at all the linters/rules we use and maybe standardize them? Right now in our CI workflows I see:
- Our CI runs pre-commit
- We have a separate pre-commit.ci service

In the CI workflow, we have:
- flake8
- black
- ruff
- etc.

Could we maybe consider dropping black and flake for ruff alone? Will that make a big difference in style, whereas it might simplify our CI quite a bit?

peterjc:
I've been moving some my own stuff to ruff in favour of flake8 and black, but it doesn't cover everything we're using. Dropping pydocstyle on this PR is a step in that direction, happy to take that further 👍

JoaoRodrigues:
👍 👍 Are there any rules you're following when dropping the linters? Like, run with ruff only, then run the old linters to see if it missed anything?

peterjc:
You have to enable non-default ruff checks, which ought to be fine if we're already using the original it was based on (e.g. flake8-bugbear), but occasionally there are differences.

I'm hoping once ruff adds array formatting we can use it in place off black.

peterjc:
@JoaoRodrigues was that an implicit PR approval?

PR #4662, opened 2024-03-14
Apply isort (forcing single lines, not sorting by type) via ruff

peterjc:
Done using ruff v0.3.2 with this command:

```
    $ ruff check --fix --select=I \
      --config=lint.isort.force-single-line=true \
      --config=lint.isort.order-by-type=false \
      BioSQL/ Bio/ Tests/ Scripts/ Doc/ setup.py
```

This does also do some blank line changing before/after imports, and I removed a couple of redundant comments about standard library vs local imports.

[*initially*] This focused on modules I look after, but ulitmately I'd like to apply something like this globally. [*Update: Done later*]

Cross reference #2828


JoaoRodrigues:
Looks good to me!

JoaoRodrigues:
Why not apply globally, not to have sort of two standards in the code base?

>

peterjc:
I think this initial set of files suffices to show the feel of the changes, dropping brackets for multiple entry imports etc.

If there are no objections, I would be happy to expand the pull request to apply this to the whole code base - and add this import sorting to the automated checks.

peterjc:
Expanded to apply import sorting to the entire codebase as suggested, now a monster pull request!

JoaoRodrigues:
👹 👍 👍

JoaoRodrigues:
That's weird. The mmtf tests are failing only on MacOS and Windows because the dependency is missing, but going fine on Linux. We should maybe add a try/except to that `import mmtf` line? I wonder if changing the order of the imports triggered some sort of import caching in unittest, since it used to first import the Bio package that tests for mmtf and issues a warning.

peterjc:
Very odd, it worked locally but I do have ``mmtf`` installed - and on the other CI platforms. But as you say, macOS:

```
...
test_kNN ... ok
test_mmtf ... loading tests failed:
Failed to import test module: test_mmtf
  File "/Users/runner/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
  File "/Users/runner/work/biopython/biopython/Tests/test_mmtf.py", line 16, in <module>
    import mmtf
ModuleNotFoundError: No module named 'mmtf'

test_mmtf_online ... skipping. Error: Install mmtf to use Bio.PDB.mmtf (e.g. pip install mmtf-python)

test_motifs ... ok
...
```

Note https://github.com/biopython/biopython/blob/master/Tests/test_mmtf.py just has ``import mmtf`` while https://github.com/biopython/biopython/blob/master/Tests/test_mmtf_online.py has ``from Bio.PDB.mmtf import MMTFParser`` which in turn calls ``https://github.com/biopython/biopython/blob/master/Bio/PDB/mmtf/__init__.py`` which has the try/except:

```python
try:
    from mmtf import fetch, parse
except ImportError:

    raise MissingPythonDependencyError(
        "Install mmtf to use Bio.PDB.mmtf (e.g. pip install mmtf-python)"
    ) from None
```

So yes, for consistency we can add the try/except to ``test_mmtf.py``

peterjc:
I understand the mmtf failure now. After sorting it does:

```python
import mmtf  # may fail
...
from Bio.PDB.mmtf import MMTFParser, MMTFIO
```

Presorting we did this:

```python
from Bio.PDB.mmtf import MMTFParser, MMTFIO  # catches failure 
...
import mmtf
```

JoaoRodrigues:
That's what it is... thanks for figuring that out!

JoaoRodrigues (approved):

peterjc:
I don't expect all the automatically assigned reviewers to review this - but would like more eyes on this before merging.

Note this PR would close #2828 by adopting the ruff implementation of an isort style.

bzoracler:
As these are large formatting commits affecting hundreds of files, it might be helpful to skip these commits (along with those in https://github.com/biopython/biopython/pull/4655) for ease of use in git blame?

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

peterjc:
@bzoracler good idea. For similar reasons I used something like black@example.org for the bulk style changes a while back.

peterjc:
Rebased to deal with the merge conflict in the configuration, and compressed the isort commits into one with a ruff@example.org dummy author.

Assuming we merge this, I think a followup PR adding a ``.git-blame-ignore-revs`` file for this and previous bulk edits like black formatting and some of style changes would make sense.

peterjc:
The Python 3.10 failure was #4640 (since fixed). I'll rebase again for the merge conflicts...

peterjc:
Rebased for trivial version updates in the pre-commit config; and some whitespace cleanup spotted while comparing branches.

peterjc:
Redone against master, and this time splitting the automatic step from the manual comment cleanup.

Issue #4661, opened 2024-03-14
PDB.PDBIO.sav add end_ter = True


pippo1990:
as noticing from https://stackoverflow.com/questions/78141949/modifying-values-in-pdb-file-and-save-as-new/78149130#78149130  could we have end_ter = TRUE in  def save(self, file, select=_select, write_end=True, preserve_atom_numbering=False): (https://github.com/biopython/biopython/blob/d1f1aa83a417386c075471f8c22624ebb9aa3cd0/Bio/PDB/PDBIO.py#L298 )

so that settibg end_ter = False it will skip :

if chain_residues_written:
                    fhandle.write(
                        _TER_FORMAT_STRING
                        % (atom_number, resname, chain_id, resseq, icode)
                    )

peterjc:
It sounds like you have a change in mind, perhaps a pull request would make it easier for us to understand your intent?

JoaoRodrigues:
From the SO question, I don't see a reason to make this change (tbh neither
for the optional END statement). Both statements are features of the file
format and any parser will parse them correctly (or ignore them).

What use case would you have where you need to write a pdb file without ter
statements?

pippo1990:
guess I'll have to install git and git it a try

peterjc:
If you only want to edit a single file, then the GitHub editor in the browser works pretty well.

pippo1990:
> From the SO question, I don't see a reason to make this change (tbh neither for the optional END statement). Both statements are features of the file format and any parser will parse them correctly (or ignore them). What use case would you have where you need to write a pdb file without ter statements?



> From the SO question, I don't see a reason to make this change (tbh neither for the optional END statement). Both statements are features of the file format and any parser will parse them correctly (or ignore them). What use case would you have where you need to write a pdb file without ter statements?

I have no idea we should ask https://stackoverflow.com/users/18253668/irene-ferfoglia , since the write_end is there , a write_ter could be plausible ? Don't know I am not expert, just tring github , for first time in my life ... I am thrilled, Thanks for your patience

peterjc:
Well, you are learning about GitHub, pull requests, and automated style checking - but I agree with @JoaoRodrigues this really needs a justification, like a motivating example.

rob-miller:
[...]

PR #4664, opened 2024-03-14
Update PDBIO.py PDBIO.save with parameter write_ter

pippo1990:
added write_ter = True as parameter of PDBIO.save to have a way to decide to get or not TER at the end of pdb 

corrected order of parameter in def save to have backward compatibility  

and 

As per the automated checks, style wise just use if chain_residues_written and write_ter:

Closes #...

### DIFF
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -294,7 +294,7 @@ class PDBIO(StructureIO):
             return _PQR_ATOM_FORMAT_STRING % args
 
     # Public methods
-    def save(self, file, select=_select, write_end=True, preserve_atom_numbering=False):
+    def save(self, file, select=_select, write_end=True, preserve_atom_numbering=False, write_ter = True):
         """Save structure to a file.
 
         :param file: output file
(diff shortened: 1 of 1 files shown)


Bio/PDB/PDBIO.py
```diff
@@ -294,7 +294,7 @@ def _get_atom_line(
             return _PQR_ATOM_FORMAT_STRING % args
 
     # Public methods
-    def save(self, file, select=_select, write_end=True, preserve_atom_numbering=False):
+    def save(self, file, select=_select, write_end=True, preserve_atom_numbering=False, write_ter = True):
```
JoaoRodrigues, line 297: This should be `write_ter=True`, per style guidelines, but the style check should catch this.

JoaoRodrigues (commented):
_Technically_ this is OK, but I wonder about the utility of the change. What is the motivation behind this change? What is the use-case for _not_ writing the TER statements?

peterjc:
This replaced pull request #4663 (better to just update the original PR next time, by updating the branch), and is to address issue #4661

PR #4666, opened 2024-03-15
Enable ruff check --fix, and pre-commit.ci autofix PRs

peterjc:
This means when run locally, ruff will apply fixes where possible. The order is changed so that black is now run afterwards for any style fix needed on top. I find this helpful when working at the command line.

Next, this will also happen on the pre-commit.ci service, and if changes are made it will update the associated pull request.

i.e. If anyone submits a minor edit on GitHub via their browser which doesn't follow the style checks, then their PR is automatically updated for them - very easy.

If you are working at the command line (or in an editor) with the pre-commit hook setup as per CONTRIBUTING.rst, again very little changes except you don't have to run black or ruff check by hand.

The only potential for surprise is if you bypass the pre-commit hooks: your online branch will have the fixes, leaving your local branch out of date. You would then have to pull the changes, or make equivalent fixes locally and force push your version online overriding the automatic fixes.

This will become more valuable if/when we include isort as per #4662 

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
     # Run the Ruff linter.
     -   id: ruff
         args: [
+            '--fix',
             '--extend-select=B,C4,D,ISC,UP',
             '--extend-ignore=E501,F401,F841,D105,D2,D4',
             '--extend-ignore=B007,B009,B010,B018,B028,B904,UP031',
@@ -90,7 +91,7 @@ repos:
         files: \.(rst|tex)$
 ci:
     # Settings for the https://pre-commit.ci/ continuous integration service
-    autofix_prs: false
+    autofix_prs: true
     # Default message is more verbose
     autoupdate_commit_msg: '[pre-commit.ci] autoupdate'
     # Default is weekly
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,11 +13,6 @@ repos:
     -   id: end-of-file-fixer
         files: \.py$
     -   id: mixed-line-ending
--   repo: https://github.com/psf/black
-    rev: 24.2.0
-    hooks:
-    -   id: black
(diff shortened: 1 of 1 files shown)


peterjc:
I originally intended to do this in #4322.

I personally have been using the pre-commit.ci service to apply fixers to pull requests automatically - example at https://github.com/peterjc/thapbi-pict/pull/473 doing import sorting.

JoaoRodrigues:
Sounds good!

One question - if I remember correctly, we got pre-commit.ci in place because GHA was timing out with the style checks + tests. With the move to ruff, which is substantially faster, will still we need to split the style check between GHA and pre-commit.ci?

JoaoRodrigues (approved):

peterjc:
For linting along, using pre-commit.ci and GHA is redundant, although not entirely a bad thing as occasionally one might be unavailable.

However, I rather like the pre-commit.ci autofix PR functionality (which this PR would enable). I've not explored how easy it would be to get something similar from GHA. I'm also a little nervous about an overly complicated GHA setup for maintenance and vendor lock in.

JoaoRodrigues:
Sounds good!

Auto fix as in, make changes directly? We'd need an auto commit action, but that's rather simple to setup. See an example here for how I automatically change the version with each merge: https://github.com/haddocking/pdb-tools/blob/master/.github/workflows/bump-version-on-push.yml

>I'm also a little nervous about an overly complicated GHA setup for maintenance and vendor lock in.
Ah! I'm worried about the opposite 😄 - fragmentation over multiple services that makes maintenance harder! What we can do is split our single CI workflow in multiple smaller runs - that should help with understanding it a little better. I can look into that sometime.

peterjc:
Updated for the (minor) merge conflict, and also broke up the commit changes into smaller logical units.

peterjc:
If there are no objections, I'll merge this in a day or two - and then look at rebasing the other related changes like isort on #4662

Issue #4437, opened 2023-09-05
 Coping with missing restriction enzymes


marcus-r-kelly:
### Setup

I am reporting a problem with Biopython 1.81, Python 3.8.17, and operating
system as follows:

```python
import sys; print(sys.version)
import platform; print(platform.python_implementation()); print(platform.platform())
import Bio; print(Bio.__version__)
```
```
|--> python3 -c "import sys; print(sys.version)
|-dquote> import platform; print(platform.python_implementation()); print(platform.platform())
`-dquote> import Bio; print(Bio.__version__)"
3.8.17 | packaged by conda-forge | (default, Jun 16 2023, 07:06:00)
[GCC 11.4.0]
CPython
Linux-4.18.0-477.21.1.el8_8.x86_64-x86_64-with-glibc2.10
1.81
```

A number of restriction enzymes, hosted by rebase, are not in the `Restriction` module as class instances or in `Bio.Restriction_Dictionary.rest_dict`. An example-- the one that prompted this issue, is [I-SceI](http://rebase.neb.com/rebase/enz/I-SceI.html). It makes sense that the hyphen precludes the class instance, but this should not be an issue for dictionary keys.

Ideally, one would then create one's own restriction enzymes, but it is not documented how to do this.

I'm unsure whether this is a "coding" issue or a "documentation" issue-- either fixing omissions from rebase or documenting how to instantiate a new enzyme would fix my issue.

peterjc:
Is it in the latest version of REBASE? If so, you/we need to rerun the conversion script ``Scripts/Restriction/rebase_update.py`` to update ``Bio/Restriction/Restriction_Dictionary.py`` - see e.g. this old pull request #3823

Usually that is quite straightforward, do you fancy giving this a go?

marcus-r-kelly:
It is.. not straightforward. The `Scripts` folder in which these scripts are supposed to be has not been installed by conda. Having created the `Scripts` folder manually inside the `Restriction` folder, I then downloaded `ranacompiler.py` and `rebase_update.py` following [this website's directions](http://home.cc.umanitoba.ca/~psgendb/local/biopython-1.55.old/Doc/cookbook/Restriction/Restriction.html). However, this version of ranacompiler uses type hinting and that is breaking backward compatibility with IPython:
```
python3 ranacompiler.py
Traceback (most recent call last):
  File "ranacompiler.py", line 110, in <module>
    def load_enzyme_ids(file) -> dict[str, int]:
[...]

PR #4685, opened 2024-03-30
Updating REBASE parser on Restriction Enzyme Dictionary creation script  

Rolando-at-Arzeda:

Address issue: #4437

This Pull Request will be updating the parser on the Restriction Enzyme Dictionary creation script (/Scripts/Rebase/ranacompiler.py) to support all EMBOSS files generated by REBASE or self created using the EMBOSS  "rebaseextract" command. Minor changes on the output text were also done to make the messages more understandable.

REBASE will add the missing restriction Type I enzymes to embossa_[ers].XXX starting from the next release, as we requested. I would recommend generating the BioPython Restriction Enzyme Dictionary from "embossa" instead, as it will add some 3000 enzymes (+ the Type I) to the current dict. This is particularly important to be able to check any recombinant DNA construct against relevant restriction enzymes natural to the host, or used on the DNA construction process. 

I can add the changes to the relevant code and the newly generated dictionary later on another Pull Request.

### DIFF
--- a/Scripts/Restriction/ranacompiler.py
+++ b/Scripts/Restriction/ranacompiler.py
@@ -94,6 +94,8 @@ def read_enzyme_record(handle):
     for line in handle:
         key, value = line[:2], line[5:].rstrip()
         if key == "ID":
+            # Sync the replacement of "- and ." by "_" or enzyme will not be found in the bairoch list
+            value = value.replace("-", "_").replace(".", "_")
             record = {"ID": value}
         elif key == "AC":
             record["AC"] = value
@@ -452,7 +454,8 @@ class DictionaryBuilder:
             try:
                 enzyme_id = enzyme_id_dict[name]
             except KeyError:
-                print(f"Could not find REBASE enzyme ID for {name}: omitting")
+                # be more specific, this is a warning it just omits the REBASE URL
+                print(f"WARNING : Could not find REBASE enzyme ID for {name}: omitting url")
                 enzyme_id = None
             cls = newenzyme(name, enzyme_id)
             #
@@ -516,7 +519,7 @@ class DictionaryBuilder:
 
         update = os.getcwd()
         with open(os.path.join(update, "Restriction_Dictionary.py"), "w") as results:
-            print("Writing the dictionary containing the new Restriction classes...")
+            print("Writing the dictionary containing the new Restriction classes ... ", end='')
             results.write(start)
             results.write("rest_dict = {}\n")
             results.write("\n")
@@ -528,8 +531,8 @@ class DictionaryBuilder:
                         % (double_quote_repr(key), double_quote_repr(value))
                     )
                 results.write("}\n\n")
-            print("OK.\n")
-            print("Writing the dictionary containing the suppliers data...")
+            print("OK.")
+            print("Writing the dictionary containing the suppliers data ... ", end='')
             results.write("\n")
             results.write("# Turn black code style off\n# fmt: off\n")
             results.write("\n")
@@ -540,8 +543,8 @@ class DictionaryBuilder:
                 for value in suppliersdict[name]:
                     results.write("    %s,\n" % double_quote_repr(value))
                 results.write(")\n\n")
-            print("OK.\n")
-            print("Writing the dictionary containing the Restriction types...")
+            print("OK.")
+            print("Writing the dictionary containing the Restriction types ... ", end='')
             results.write("\n")
             results.write("typedict = {}\n")
             results.write("\n")
@@ -551,7 +554,7 @@ class DictionaryBuilder:
                     results.write("    %s,\n" % double_quote_repr(value))
                 results.write(")\n\n")
             results.write("# Turn black code style on\n# fmt: on\n")
-            print("OK.\n")
+            print("OK.")
 
     def install_dict(self):
         """Install the newly created dictionary in the site-packages folder.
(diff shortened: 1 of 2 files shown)


Scripts/Restriction/ranacompiler.py
```diff
@@ -94,6 +94,8 @@ def read_enzyme_record(handle):
     for line in handle:
         key, value = line[:2], line[5:].rstrip()
         if key == "ID":
+            # Sync the replacement of "- and ." by "_" or enzyme will not be found in the bairoch list
+            value = value.replace("-", "_").replace(".", "_")
```
peterjc, line 98: I'm guessing you changed this as per @MarkusPiotrowski's comment https://github.com/biopython/biopython/issues/4437#issuecomment-1711617443 

However I'm concerned that in addition to fixing looking up the enzyme ID from the Bairoch file, it prevents access to the unmodified ID (e.g. in end user scripts).

This needs a test please.

Scripts/Restriction/ranacompiler.py
```diff
@@ -452,7 +454,10 @@ def build_dict(self):
             try:
                 enzyme_id = enzyme_id_dict[name]
             except KeyError:
-                print(f"Could not find REBASE enzyme ID for {name}: omitting")
+                # be more specific, this is a warning it just omits the REBASE URL
+                print(
+                    f"WARNING : Could not find REBASE enzyme ID for {name}: omitting url"
+                )
```
peterjc, line 460: This seems like a useful logging clarification

peterjc:
I wouldn't have bothered with the whitespace changes to the logging (changing new lines and tabs), it just makes it hard to review your functional code changes here.

peterjc:
Do you have an example file generated with ``rebaseextract`` which the old script couldn't understand? That might be helpful for us in reviewing this, and might even be useful as a test case (although as I recall, we don't currently test the compiler script in the CI tests).

Scripts/Restriction/ranacompiler.py
peterjc: Never mind, in refreshing the REBASE files on https://github.com/biopython/biopython/pull/4687 I see we don't currently expose the original ID string anyway (just the numerical ID).

e.g. ``CviKI-1`` is only visible as ``CviKI_1``.

Rolando-at-Arzeda: Hi Peter,  Yes good catch from @MarkusPiotrowski back in September! 

It does not change (much) the output of the script, the enzyme is always added to the BioPython dictionary but if the enzyme exists in the Bairoch file, it can't look up the correct REBASE ID to add the proper URL in the Restriction_Dictionary.py. As in my comment below, I also added the proper change on the logging message.
BTW there are enzymes that might still get caught on this because they might not be in the bairoch file .. 

Agreed on doing more tests, specially that just today the new REBASE release is out, with several new changes.

By  "it prevents access to the unmodified ID (e.g. in end user scripts)." do you mean access from Restriction_Dictionary.py ? Because if the enzyme had a dash or a colon on the name, it was already replaced by the old script, this change only adds the proper ID (REBASE number) and allows creating the URL.

Scripts/Restriction/ranacompiler.py
Rolando-at-Arzeda: Yes as per my comment before, this was needed on the output because the enzyme is added but just the URL is omitted, before it looked like "the enzyme" was omitted.

Rolando-at-Arzeda:
You are right, we change the output format to make it more consistent on the number of blank lines .. but it was just output .. On the tests, @peterjc, I was going to ask you to hold the PR until we had the chance to test this with the new REBASE release (of today) 

@MarkusPiotrowski I see you as a reviewer .. I am testing the changes with the new REBASE release of today, that Dana was making some changes and additions and noticed that something else breaks .. I will fix that, and do the tests again.

Peter, I will add to this comments the file created by EMBOSS "rebaseextract" from the new REBASE release.

peterjc:
Curious - I didn't notice any problems with the April 2024 release on #4687 

(My plan was to do the update with the old script, and then repeat it with your revised script. This would separate REBASE changes from your changes).

Rolando-at-Arzeda:
> Curious - I didn't notice any problems with the April 2024 release on #4687
> 
> (My plan was to do the update with the old script, and then repeat it with your revised script. This would separate REBASE changes from your changes).

The error deals with the (not) opening the bairoch files, to load the enzyme IDs, if you had the old files in the directory where you were running the script, and/or you said "no" to the update .. you needed to reload the enzyme ids (or return the bairoch filename in some case) and it was apparently not happening .. 

BTW The mini-release on "emboss" files only added 2 enzymes .. those are just cherry-picked enzymes, the real changes and additions happens in "embossa". I really think we should move to the "new" emboss.

Rolando-at-Arzeda:
@peterjc, @MarkusPiotrowski  I finished with the code revision and the new tests, I am changing the PR back to draft to be able to add the next commit. I cleaned up some errors related to the bairoch file not being loaded properly, and when you decline to do the update and still use the old version or don't have the bairoch file.  

I tested :

1. Building the dict with the (version 404) emboss, embossa(renamed) and the self generated using the EMBOSS rebaseextract
2. Building the dict with the old version (403) saying "no" to the update (that was failing before)
3. Trying building the dictionary with any of the emboss or bairoch file missing (it now aborts gracefully, it was failing when the bairoch file was missing)

PR #4687, opened 2024-04-01
Update Bio.Restriction to REBASE version 404 (April 2024)

peterjc:

### DIFF
--- a/Bio/Restriction/Restriction_Dictionary.py
+++ b/Bio/Restriction/Restriction_Dictionary.py
@@ -12,7 +12,7 @@
 
 """Restriction Analysis Libraries.
 
-Used REBASE emboss files version 309 (2023).
+Used REBASE emboss files version 404 (2024).
 """
 
 rest_dict = {}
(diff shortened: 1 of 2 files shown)


peterjc:
@Rolando-at-Arzeda and/or @MarkusPiotrowski - is this good to merge as is, a minor update with just 2 new enzymes in the short list which makes it into the default EMBOSS set?

Rolando-at-Arzeda:
Well it is what we have these day @peterjc .. if we want to update the restriction enzyme dictionary that will be it ..

20 years ago when the bulk of this work was done there was only a single emboss file, but now we have two, the short list with a selection made by the REBASE maintainers and the full set of enzymes. People will really appreciate if BioPython move to use the full list. 

I finished the revision of the ranacompiler script now and it will work with all EMBOSS formats, If we decide we could change to "embossa" and do a release with the big database, I can take care of updating the scripts and do all kind of tests again, adding all to my pending PR ..

I can also keep taking care of the REBASE dictionary Script code, updates and other requests in the future ..

peterjc:
I'll merge this now. I am a little nervous about expanding this to larger REBASE set given the way the module creates a class for each, but let's keep that discussion on #4437.

PR #4703, opened 2024-04-14
Warn on writing INSDC-noncompliant feature and qualifier keys

michaelfm1211:

From the [DDBJ/ENA/GenBank Feature Table Definition](https://www.insdc.org/submitting-standards/feature-table/#3.1):
> 3.1 Naming conventions
> Feature table components, including feature keys, qualifiers, accession 
> numbers, database name abbreviations, and location operators, are all named 
> following the same conventions. Component names may be no more than 20 
> characters long  (Feature keys 15, Feature qualifiers 20) and must 
> contain at least one letter. The following characters are permitted to 
> occur in feature table component names: 
> 
> * Uppercase letters (A-Z) 
> * Lowercase letters (a-z) Numbers (0-9) 
> * Underscore (_) 
> * Hyphen (-) 
> * Single quotation mark or apostrophe (') 
> * Asterisk (*) 

Currently, Biopython allows you to read or write INSDC files with feature and qualifier keys that don't follow these rules without warning. This PR warns `BiopythonWarning` if you try to write a noncompliant feature or qualifier key.

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -380,7 +380,7 @@ class _InsdcWriter(SequenceWriter):
     )
 
     def _write_feature_qualifier(self, key, value=None, quote=None):
-        if not all(ch in _allowed_table_component_name_chars for ch in key):
+        if not _allowed_table_component_name_chars.issuperset(key):
             warnings.warn(
                 f"Feature qualifier key '{key}' contains characters not"
                 " allowed by standard.",
@@ -458,7 +458,7 @@ class _InsdcWriter(SequenceWriter):
         assert feature.type, feature
 
         f_type = feature.type.replace(" ", "_")
-        if not all(ch in _allowed_table_component_name_chars for ch in f_type):
+        if not _allowed_table_component_name_chars.issuperset(f_type):
             warnings.warn(
                 f"Feature key '{f_type}' contains characters not allowed by"
                 " standard.",
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -33,6 +33,7 @@ http://www.ebi.ac.uk/imgt/hla/docs/manual.html
 import warnings
 
 from datetime import datetime
+from string import ascii_letters, digits
 
 from Bio import BiopythonWarning
 from Bio import SeqFeature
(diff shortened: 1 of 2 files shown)


peterjc:
I think this is a good idea in principle.

I would define ``allowed_chars = ascii_letters + digits + "_-'*"`` once at module level as a private variable, e.g. ``_allowed_qualifier_key_chars = ascii_letters + digits + "_-'*"``

Also I suspect but have not checked it would be quicker if this was a set, and we used set intersection to check for invalid characters.

I would find it helpful if the warnings included the problematic key name (but probably not the location string), which would avoid too many similar warnings with the default settings (each unique warning is shown once).

Bio/SeqIO/InsdcIO.py
```diff
@@ -376,6 +380,19 @@ class _InsdcWriter(SequenceWriter):
     )
 
     def _write_feature_qualifier(self, key, value=None, quote=None):
+        if not all(ch in _allowed_table_component_name_chars for ch in key):
```
peterjc, line 383: Replace that with a set operation: ``_allowed_table_component_name_chars.issuperset(key)``

michaelfm1211:
@peterjc  Good ideas, thanks.

I just added a new commit that renames `allowed_chars` to  `_allowed_table_component_name_chars` (to match the language in the [spec](https://www.insdc.org/submitting-standards/feature-table/#3.1)) in the module level and mentions the offending feature or qualifier key name.

I did a quick test to compare set intersection and `all`, and on my machine `all` consistently runs a bit faster. This is the code I used to test:
```python
from timeit import timeit

print("With set intersection:")
print(timeit("""
from random import choice
from string import ascii_letters, digits, ascii_lowercase

_allowed_table_component_chars = set(ascii_letters + digits + "_-'*")

for i in range(100000):
    txt = ''.join(choice(ascii_lowercase) for i in range(20))
    set(txt) == _allowed_table_component_chars ^ set(txt)
""", number=1))

print("With all():")
print(timeit("""
from random import choice
from string import ascii_letters, digits, ascii_lowercase

_allowed_table_component_chars = set(ascii_letters + digits + "_-'*")

for i in range(100000):
    txt = ''.join(choice(ascii_lowercase) for i in range(20))
    all(ch in _allowed_table_component_chars for ch in txt)
""", number=1))
```

peterjc (changes requested):

Bio/SeqIO/InsdcIO.py
michaelfm1211: Cool, I hadn't thought of `. issuperset`. I just check it and it is faster than `all()`. I'll add another commit in a second.

michaelfm1211:
@peterjc I pushed a new commit that uses set intersection instead.

peterjc (approved):
Looks good, and you're in the CONTRIB and NEWS files already.

Is a squash-and-merge OK with you?

michaelfm1211:
@peterjc Yes, that would be perfect.

peterjc:
Merged, thank you.

PR #4715, opened 2024-05-01
fix typo

kamurani:
Simple typo fix in docstring


peterjc (approved):
Looks good, thanks.

This is one of the things codespell would fix
https://github.com/codespell-project/codespell/blob/v2.2.6/codespell_lib/data/dictionary.txt#L25060

However it looks like in #3709 I didn't enable this for the Python code. I wonder why...

PR #4724, opened 2024-05-09
Fix byte mode detection on `BgzfWriter`

DavidCain:
In commit c4a47ffff7f3e7de32ab3d8846983d3531ea63b4, there was an
attempt to detect handles which are not in binary mode. However, there's
a core problem here in that most handles open in write mode *cannot* be
read from. For example:

```python
>>> handle = open('/tmp/demo.txt', 'wb')
>>> handle.read(0)
io.UnsupportedOperation: read
```

This means that one cannot pass a plain writeable file object to
`BgzfWriter`:

```python
>>> Bio.__version__
'1.83'
>>> Bio.bgzf
AttributeError: module 'Bio' has no attribute 'bgzf'
>>> from Bio import bgzf
>>> with open('/tmp/demo.txt', 'wb') as handle:
...     bgzf.BgzfWriter(fileobj=handle)
...
    if fileobj.read(0) != b"":
       ^^^^^^^^^^^^^^^
io.UnsupportedOperation: read
```

If eager detection of the handle's mode is desired, one can check
`handle.mode` as was done before -- I opt here to just hardcode handling
of `io.StringIO`. The simplest option in my opinion is to just rely on
ducktyping -- the `_write_block()` method will try to write the data --
either the handle accepts `bytes` and all is well, or it fails.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -91,6 +91,9 @@ the representation of origin-spanning locations. See issue #4611.
 
 Updated ``Bio.Restriction`` to the April 2024 release of REBASE.
 
+A bug in ``bgzf`` was resolved, restoring the ability to pass a file handle
+directly to ``BgzfWriter``.
+
 As in recent releases, more of our code is now explicitly available under
 either our original "Biopython License Agreement", or the very similar but
 more commonly used "3-Clause BSD License".  See the ``LICENSE.rst`` file for
@@ -100,6 +103,7 @@ Many thanks to the Biopython developers and community for making this release
 possible, especially the following contributors:
 
 - Anil Tuncel (first contribution)
+- David Cain
 - Fabio Zanini (first contribution)
 - Joao Rodrigues
 - Judith Bernett (first contribution)
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -247,6 +247,7 @@ If your data is in UTF-8 or any other incompatible encoding, you must use
 binary mode, and decode the appropriate fragments yourself.
 """
 
+import io
 import struct
 import sys
 import zlib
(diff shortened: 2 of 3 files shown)


peterjc:
Oh my, we missed that on #4270 - and therefore we're missing something on the test suite too.

peterjc:
Do you agree to dual licence this and any previous contributions under both the Biopython License Agreement AND the BSD 3-Clause License.

Also do you want to be named in the NEWS file etc?

DavidCain:
> Do you agree to dual licence this and any previous contributions under both the Biopython License Agreement AND the BSD 3-Clause License.
> 
> Also do you want to be named in the NEWS file etc?

Yes, I agree to dual-license this work, and sure -- feel free to mention me in NEWS and such (I should already be in CONTRIB.rst!). Thanks, @peterjc

peterjc (approved):
I'll merge it in the next few days (leaving you a chance to update the NEWS file to add your name if you want).  Thanks!

DavidCain:
Thanks, @peterjc -- just added!

NEWS.rst
```diff
@@ -91,6 +91,9 @@ the representation of origin-spanning locations. See issue #4611.
 
 Updated ``Bio.Restriction`` to the April 2024 release of REBASE.
 
+A bug in ``bgzf`` was resolved, restoring the ability to pass a file handle
+directly to ``BgzfWriter``.
```
peterjc, line 95: Based on historical precedent, I wouldn't normally have added a news entry for a relatively minor bug fix, but it can stay unless the release notes get too long.

(Plenty of other projects go to the other extreme of reporting everything which would be excessive here)

DavidCain: Thanks, Peter! By all means please feel free to purge this entry later -- my feelings won't be hurt if this super trivial fix doesn't make the release notes! 😄

Issue #4747, opened 2024-06-18
mmCIF writer should always write initial data_ token


JoaoRodrigues:
As shown in #4744, we should always write a `data_` token at the beginning of a mmCIF file, otherwise we mess up the parsing. The key issue is with `MMCIF2Dict`, where we expect the first line to be a `data_` token which is always treated differently (see [here](https://github.com/biopython/biopython/blob/cca7983d2e054db56123295034f4eefc919cb7fa/Bio/PDB/MMCIF2Dict.py#L29)). I can't find a definite format spec saying that `data_` is mandatory, but since `data_` indicates the start of a data block, I think that's implicit.

We should do two things to mitigate this issue:
1. Ensure we always write a `data_` block header in mmcifio.py
2. Check that a mmcif file starts with `data_` otherwise fail early.

thepug432:
Could I take this?

thepug432:
The above should hopefully fix the issue.

JoaoRodrigues:
Sure :) Thanks for picking it up so quickly, I'll have a look tomorrow at
your PR.

A terça, 18/06/2024, 19:29, Kiril Bourakov ***@***.***>
escreveu:

thepug432:
This should be closed.

JoaoRodrigues:
Indeed!

PR #4748, opened 2024-06-18
 mmCIF writer should always write initial data_ token #4747 

thepug432:
These changes should do the two things described in #4747 . Please tell me if there are any problems, and I will do my best to fix them. 

Closes #...

### DIFF
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -33,7 +33,7 @@ class MMCIF2Dict(dict):
             self[token[0:5]] = token[5:]
             if not token[0:5].startswith("data_"):
                 raise ValueError(
-                    f"The mmCIF file {filename} does not start with 'data_'"
+                    "The input mmCIF file must begin with a 'data_' directive."
                 )
             i = 0
             n = 0
--- a/Bio/PDB/MMCIF2Dict.py
+++ b/Bio/PDB/MMCIF2Dict.py
@@ -31,6 +31,10 @@ class MMCIF2Dict(dict):
             except StopIteration:
                 return  # for Python 3.7 and PEP 479
             self[token[0:5]] = token[5:]
+            if not token[0:5].startswith("data_"):
+                raise ValueError(
+                    f"The mmCIF file {filename} does not start with 'data_'"
+                )
             i = 0
             n = 0
             for token in tokens:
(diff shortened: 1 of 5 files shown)


Bio/PDB/MMCIF2Dict.py
```diff
@@ -31,6 +31,10 @@ def __init__(self, filename):
             except StopIteration:
                 return  # for Python 3.7 and PEP 479
             self[token[0:5]] = token[5:]
+            if not token[0:5].startswith("data_"):
+                raise ValueError(
+                    f"The mmCIF file {filename} does not start with 'data_'"
```
JoaoRodrigues, line 36: I'd write: "The input mmCIF file must begin with a 'data_' directive."

JoaoRodrigues (approved):
Besides that comment on the error message, this looks good to me!

Would you mind adding a couple of tests to make sure this works as intended? One test could be trying to parse a file without the `data_` directive and asserting the exception is raised. Another would be parsing a file without an id, e.g. `parser.get_structure("", "1ctf.pdb")`, which results in an empty `data_val` in mmcifio and making sure you can write the structure to mmCIF and read it again.

Bio/PDB/MMCIF2Dict.py
thepug432: Absolutely. I'll do that hopefully either today or tomorrow.

thepug432: A point on the error message, however, it may be better to have something like "The input mmCIF file {filename} must begin with a 'data_' directive.", as the filename may make debugging a little easier.

thepug432:
Changes committed, please tell me if any changes are needed.

JoaoRodrigues:
Thank you! Seems like a lot of our own test files do not have `data_` entries! Would you mind fixing them? See for example: https://ci.appveyor.com/project/biopython/biopython/builds/50053066/job/lwu5su4iefi1byc7

thepug432:
Done. If I understand the code correctly, only .cif files are required to start with data_, which makes a_structure.cif the only file I could find that does not do so.

JoaoRodrigues:
I'm expecting failures because of other issues at the moment, but both CircleCI and Appveyor passed the PDB tests so this should be OK (given that all other builds pass too).

Thanks @thepug432 !

JoaoRodrigues:
Annoying but we have to wait for the other stuff to get fixed to merge this. @peterjc, unless you can push this through?

peterjc:
There seems to be multiple unexpected failures there, not just the alignment 64 bit integer stuff (about to be fixed on #4752). Let's wait on that first to be sure...

thepug432:
Done

peterjc:
Thank you both, rebased and merged.

PR #4749, opened 2024-06-20
Remove error message if wheel is not installed

hoodmane:
Since https://github.com/pypa/setuptools/pull/4369 was merged, wheel is vendored into setuptools and no
longer needs to be installed for `bdist_wheel` to work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,15 +37,6 @@ except ImportError:
         "Try running: python -m ensurepip"
     )
 
-if "bdist_wheel" in sys.argv:
-    try:
-        import wheel  # noqa: F401
-    except ImportError:
-        sys.exit(
-            "We need both setuptools AND wheel packages installed "
-            "for bdist_wheel to work. Try running: pip install wheel"
-        )
-
 
 # Make sure we have the right Python version.
 MIN_PY_VER = (3, 9)
--- a/setup.py
+++ b/setup.py
@@ -31,11 +31,28 @@ try:
     from setuptools import setup
     from setuptools import Command
     from setuptools import Extension
+    from setuptools import __version__ as setuptools_version
 except ImportError:
     sys.exit(
         "We need the Python library setuptools to be installed. "
         "Try running: python -m ensurepip"
     )
+
+
+setuptools_version_tuple = tuple(int(x) for x in setuptools_version.split("."))
+if setuptools_version_tuple < (70, 1) and "bdist_wheel" in sys.argv:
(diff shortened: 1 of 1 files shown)


agronholm:
To answer my own question, that comes from Biopython's own `setup.py`. Why then do you say that setuptools contains a "regression"?

agronholm:
Biopython's setup makes the unfounded assumption that `wheel` is required to be installed for the `bdist_wheel` command to work. This is the cause of the failure.

hoodmane:
Okay not a regression then. Sorry @agronholm I should have looked into it further before blaming you =)

hoodmane:
Okay I reworked the PR to remove that error message instead of capping setuptools. Thanks!

peterjc:
So if we have setuptools 70.1.0 or later, we don't need to check for the wheel package.

But on older setup tools, the check and warning are still valuable.

Making the check conditional would be more elegant, but not essential.

peterjc:
The Circle CI failure is odd, possible stochastic:

```
  File "/home/circleci/project/biopython-1.84.dev0/Tests/test_phenotype_fit.py", line 68, in test_WellRecord
    self.assertAlmostEqual(w.slope, 48.190618284831132, places=4)
AssertionError: np.float64(48.19069017500249) != 48.19061828483113 within 4 places (np.float64(7.189017135544873e-05) difference)

```

Anyway, like the AppVeyor failure #4751 it is unrelated to your changes.

hoodmane:
> But on older setup tools, the check and warning are still valuable.
>
> Making the check conditional would be more elegant.

Setuptools implemented [PEP 517's get_requires_for_build_wheel](https://peps.python.org/pep-0517/#get-requires-for-build-wheel) in [this commit](https://github.com/pypa/setuptools/commit/b59076e9f12c295aca42927f2a544e46fd31f386) first released in setuptools v38.2.5 made on Dec 24, 2017. I think most build frontends have been respecting PEP 517 for about as long, including the pip build frontend where people are most likely to see this message. So I think it's been unneeded for over 6 years.

But I updated it to do the check for setuptools <70.1 as you suggest. I think it might make more sense just to remove it since it is 6 years outdated. Whichever you prefer.

peterjc:
Rebased now that the issues on the recent regressions on the default branch have been fixed.

covracer (approved):
This also fixes
```
~/code/biopython$ pip install --use-pep517 --no-cache-dir --force-reinstall .
```

PR #4782, opened 2024-07-29
PDB structure query engine prototype

Will-Tyler:
### Acknowledgements

### Overview

This draft pull request ia a prototype of a query language and engine for PDB structures. The query language is based on [PyMOL's selection language](https://pymolwiki.org/index.php/Selection_Algebra).

I do not intend for this pull request to be merged. Instead, the goal of this pull request is to get high-level feedback and make sure I am on the right path before investing more effort (testing and documentation) into a specific approach.

### Query language

The prototype language supports the following boolean operators: `not`, `and` and `or`.

The language supports control over the order of operations with parentheses.

The language supports selection by the following types of identifiers: `model`, `chain`, `resn`, and `name`.

See the descriptions of these operators in the [PyMOL selection algebra](https://pymolwiki.org/index.php/Selection_Algebra).

### Query engine

The query engine includes two components: a parser and an evaluator. The parser processes the query string and produces a syntax tree that represents the query. The evaluator processes the syntax tree and produces a list of boolean match indicators indicating whether each atom in the structure is selected by the query. With the match indicators, the engine builds a new PDB structures containing all and only the atoms selected by the query.

I implement the parser using the [pyparsing module](https://pyparsing-docs.readthedocs.io/en/latest/index.html).

The evaluator works by recursively evaluating the syntax tree produced by the parser. If a node in the syntax tree represents an operation, then any child trees of the operation node are evaluated to give the operands. The operation is then executed with the operands. The evaluator uses boolean NumPy array to indicate which atoms are included in the selection as a result of each operation.

Finally, the engine builds a new PDB by iterating over the atoms of the original structure and including the ones that are selected by the query.

### References

- [PyMOL Selection Algebra](https://pymolwiki.org/index.php/Selection_Algebra)
- [PyParsing homepage](https://pyparsing-docs.readthedocs.io/en/latest/index.html)
- [NumPy documentation](https://numpy.org/doc/stable/)

### DIFF
--- a/Bio/PDB/Selection.py
+++ b/Bio/PDB/Selection.py
@@ -8,10 +8,26 @@
 """Selection of atoms, residues, etc."""
 
 import itertools
+from copy import copy
+
+import numpy as np
 
 from Bio.PDB.Atom import Atom
+from Bio.PDB.Chain import Chain
 from Bio.PDB.Entity import Entity
+from Bio.PDB.Model import Model
 from Bio.PDB.PDBExceptions import PDBException
+from Bio.PDB.Residue import Residue
+from Bio.PDB.Structure import Structure
+
+try:
+    import pyparsing as pp
+except ImportError:
+    from Bio import MissingPythonDependencyError
+
+    raise MissingPythonDependencyError(
+        "Install pyparsing to use Bio.PDB.Selection (e.g. pip install pyparsing)"
+    ) from None
 
 entity_levels = ["A", "R", "C", "M", "S"]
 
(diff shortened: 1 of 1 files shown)


Will-Tyler:
Accidentally clicked create pull request... I am still working on the description.

Will-Tyler:
Not sure why the tests are failing because of the missing pyparsing dependency, whereas the test program skips other tests that are missing a dependency. Perhaps because I am updating a test file rather than adding a new one?

peterjc:
We use ``MissingPythonDependencyError`` to skip an entire file. For a single test method, you can probably just use ``return`` if the dependency is missing?

peterjc:
Or detect the dependency at the top, and make defining the test method conditional.

PR #4784, opened 2024-08-03
FOGSAA PairwiseAligner implementation

### LINKED ISSUE #4776: [feature] implement FOGSAA for global sequence alignment
I would like to propose adding the option to use the "Fast Optimal Global Sequence Alignment Algorithm" (FOGSAA) [1] both for pairwise and multiple sequence alignments. Given that FOGSAA provides alignments of identical quality to those produced by the Needleman-Wunsch (NW) algorithm, it may even be beneficial to set FOGSAA as the default option for global alignments.

The FOGSAA has an average time complexity of O(m+n) and outperforms NW especially on sequence alignments with high sequence identities. This enhancement could offer substantial performance improvements in sequence alignment tasks. 

[1] _Chakraborty A, Bandyopadhyay S. FOGSAA: Fast Optimal Global Sequence Alignment Algorithm. Scientific reports. 2013;3(1). doi:https://doi.org/10.1038/srep01746_

Edit 1: typo 
‌

michaelfm1211:

Closes #4776

This PR adds an implementation of the [Fast Optimal Global Alignment Algorithm (FOGSAA)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3638164/) to `Bio.Align.PairwiseAligner`. Right now this PR is still a work in progress and not ready to be merged. The implementation is still in its infancy and there are likely a number of bugs. I wanted to make a PR to give time for feedback on the code and discussion on whether or not adding FOGSAA to `PairwiseAligner` is even a good idea.

The majority of this PR is taken and adapted from [Angana Chakraborty's C++ code](https://www.isical.ac.in/~bioinfo_miu/FOGSAA.htm). In the linked issue @CallMeMisterOwl has indicated the author has OK'd the code for use in BioPython, but we'll probably want to check in again if this PR will eventually be merged.

Right now only the DNA scoring algorithm has been added. Matrix scoring for proteins and alignment are still TODO. To make writing the code easier for me, I haven't put everything in macros like the other algorithms, but this will be done before the final version. There are some flaws in the DNA scoring implementation. Chakraborty's code only uses integers for scoring, but the rest of `PairwiseAligner` uses `double`s. Before the final version this will be addressed (it will require replacing/changing the queue data structure used in the original). The code also creates a full 2D alignment matrix, but some sources (ex, [this README](https://github.com/kjenova/fogsaa/blob/master/README)) say the algorithm can be done without it. There are almost certainly more optimizations that can be done too.

`PairwiseAligner` selects the best algorithm by the parameters given. In the linked issue, it was suggested that FOGSAA be made the default algorithm, but IMO that won't be a good idea until after the final version has been battle-tested in production. For now, I've added a setter to the `.algorithm` property of `PairwiseAligner`. I think this setter is a good idea on its own right, but if others disagree it can be removed in the final version.

I'll update this description as progress continues.

**Update August 9 (commit https://github.com/biopython/biopython/pull/4784/commits/793ff19af506c4efa52ff0aa2aad7ea9cec3adca)**: Support for scoring with a scoring matrix should now be implemented. The code has now been moved into macros similar to the other algorithm implementations.

**Update August 14 (commit https://github.com/biopython/biopython/pull/4784/commits/3f398f57e5aa0c68c04448ecb6a9adc784514e98)**: Per mdehoon's suggestions, the `.algorithm` setter has been removed. The FOGSAA algorithm is now selected by setting the `mode` attribute of `PairwiseAligner` to `"fogsaa"`.

**Update August 22**: Commits https://github.com/biopython/biopython/pull/4784/commits/b0c365f85a0ae9bc0c78ac3f2511cfe4105d7a0e and https://github.com/biopython/biopython/pull/4784/commits/98eb4f5355937c92f240322b2cefef5692a1afd5 should now add basic support for aligning two sequences with FOGSAA.

**Update August 27 (commit [`463aabc`](https://github.com/biopython/biopython/pull/4784/commits/463aabc11a708cc158ae27dcc6d3c86a062ffa3f))**: Support for non-integer scores should now be implemented. This was done by replacing the priority queue implementation used in Chakraborty's code with a max heap.

**Update September 28**: End gap scores and affine gaps should now work, and warnings will be raised if potentially problematic gap scores are used for FOGSAA. Numerous bugs fixes and some speedups have also been added.


Bio/Align/_pairwisealigner.c
```diff
@@ -3796,13 +3802,81 @@ Aligner_get_algorithm(Aligner* self, void* closure)
...
         default:
             break;
     }
     return PyUnicode_FromString(s);
 }
 
+static int
+Aligner_set_algorithm(Aligner* self, PyObject* value, void* closure)
+{
+    if (value == Py_None) {
```
mdehoon, line 3825: I haven't tried the code, but it seems that with this method the user is able to choose an algorithm that is inconsistent with the gap scores.

michaelfm1211: Good catch. I pushed another commit today that should fix this. Can you take a look at it?

mdehoon: I think that the algorithm should be read-only. The appropriate algorithm is chosen automatically based on the gap scores and mode. Other than for FOGSAA, there is no point in choosing a different algorithm, because then the algorithm is either inconsistent with the gap scores, or it is slower than the automatically selected one (while producing identical results).

Also, if a user sets the `algorithm` to Gotoh, but then changes the gap scores, the algorithm may flip back to Needleman-Wunsch if the gap scores are consistent with that algorithm.

I would suggest to make 'FOGSAA' (or 'fast global' or whatever you want to call it) one of the options for the `mode` attribute. If that option is chosen, the `algorithm` attribute automatically switches to `FOGSAA`.

michaelfm1211:
Note to self (and others following this PR, I guess): Tests are failing on macOS and Windows because of a bug in NumPy (see https://github.com/biopython/biopython/pull/4797). Passes on Linux and is working locally.

michaelfm1211:
I haven't had much time to work on this over the past month, but I should have more time soon. Right now the only feature yet to be implemented is support for different gap scores on the ends of strands. The bigger problem seems to be performance. From my rudimentary benchmarking with a few random sequences from GenBank, it looks like FOGSAA is running at least 10x slower than Needleman-Wunsch. Right now I'm not sure where this is coming from (my current guesses are inefficiencies in the priority queue, incorrect bounds, or some other sort of bug); I'll need to do some more testing.

michaelfm1211:
Odd. Tests are failing on macOS with Python 3.12.6. My dev environment is macOS with Python 3.12.5 the tests pass fine.

The exception being raised should only be raised if the lower bound is still less than the upper bound when all feasible paths have been exhausted, but a debug message I added says the lower bound is `0.200000` and the upper bound is also `0.200000`, so no exception should be raised. I don't really know what's going on but I suspect some kind of floating point comparison bug.

michaelfm1211:
I think this PR should be out of draft stage now. As long as I haven't missed anything, this FOGSAA implementation should give the same results as the Chakraborty code and implement all the features all the other PairwiseAligners have.

@mdehoon Do you want to take a look over this again? I know it's a lot of code, so no rush.

Bio/Align/__init__.py
```diff
@@ -3770,8 +3770,15 @@ class PairwiseAligner(_pairwisealigner.PairwiseAligner):
     and gap scores are negative or zero.  By default, the match score is 1,
     and the mismatch and gap scores are zero.  Based on the values of the gap
     scores, a PairwiseAligner object automatically chooses the appropriate
-    alignment algorithm (the Needleman-Wunsch, Smith-Waterman, Gotoh, or
-    Waterman-Smith-Beyer global or local alignment algorithm).
+    alignment algorithm (the Needleman-Wunsch, Smith-Waterman, Gotoh,
+    Waterman-Smith-Beyer, or Fast Optimal Global Sequence Alignment Algorithm
+    global or local alignment algorithm).
```
mdehoon, line 3775: I guess this should be 

(the Needleman-Wunsch, Smith-Waterman, Gotoh, or Waterman-Smith-Beyer global or local alignment algorithm, or the Fast Optimal Global Sequence Alignment Algorithm).

Just for my curiosity, is there a version of FOGSAA for local alignments?

Bio/Align/_pairwisealigner.c
```diff
@@ -1857,7 +1940,9 @@ Aligner_dealloc(Aligner* self)
 static PyObject*
 Aligner_repr(Aligner* self)
 {
-  const char text[] = "Pairwise aligner, implementing the Needleman-Wunsch, Smith-Waterman, Gotoh, and Waterman-Smith-Beyer global and local alignment algorithms";
+  const char text[] = "Pairwise aligner, implementing the Needleman-Wunsch, "
+      "Smith-Waterman, Gotoh, Waterman-Smith-Beyer, and Fast Optimal Global "
+      "Sequence Alignment Algorithm global and local alignment algorithms";
```
mdehoon, line 1945: Same comment here

mdehoon (commented):
Can you add a line or two to the Biopython tutorial documentation (chapter_align.rst ) to make users aware that FOGSAA is now available? Search engines will also be able to find it then.
Otherwise, it looks fine. Great job!

Bio/Align/__init__.py
michaelfm1211: I think one could theoretically adapt FOGSAA for local alignments, but I think it would almost always be slower than Smith-Waterman because FOGSAA would almost never be able to prune a branch. Also, the name is Fast Optimal *Global* Alignment Algorithm, so it really was never designed for local alignments.

Doc/Tutorial/chapter_pairwise.rst
```diff
@@ -320,6 +321,30 @@ alignments if segments with a score 0 can be added to the alignment. We
 follow the suggestion by Waterman & Eggert
 [Waterman1987]_ and disallow such extensions.
 
+If `aligner.mode` is set to `"fogsaa"`, then the Fast Optimal Global Alignment
+Algorithm [Chakraborty2013]_ with some modifications is used. This mode
+calculates a global alignment, but it is not like the regular `"global"` mode.
+It is best suited for long alignments between similar sequences. If the the
```
mdehoon, line 327: the the

Doc/Tutorial/chapter_pairwise.rst
```diff
@@ -320,6 +321,30 @@ alignments if segments with a score 0 can be added to the alignment. We
...
+match score is less than the mismatch score or any gap score, or if any gap
+score is greater than the mismatch score, then a warning is raised and the
+algorithm may return incorrect results. Unlike other modes that may return more
+than one alignment, FOGSAA always returns only one alignment.
```
mdehoon, line 331: Key point to include here is that FOGSAA relies on heuristic, and (unlike the other alignment algorithms) is not guaranteed to find the alignment with the highest score.

Doc/Tutorial/chapter_pairwise.rst
michaelfm1211: Oops, thanks for catching that!

Doc/Tutorial/chapter_pairwise.rst
michaelfm1211: I changed the description to explain how the heuristic is used. FOGSAA should be guaranteed to find the optimal alignment (one of the alignments returned by Needleman-Wunsch) if the assumptions of the heuristic are not violated. These are what the warnings are for: if a warning is raised then the heuristic *may* be incorrect.

Does the new explanation make this clear to users, or should I modify it?

michaelfm1211:
Odd, it looks like CircleCI failed to checkout the branch, but the rest of the tests work so it's probably not an issue with the code. I'm guessing this is just a CircleCI thing or just bad luck.

Doc/Tutorial/chapter_pairwise.rst
mdehoon: I think it's fine.  Users who want to know the details can look in the literature.

Bio/Align/_pairwisealigner.c
```diff
@@ -1585,6 +1609,24 @@ PathGenerator_next_waterman_smith_beyer_local(PathGenerator* self)
     }
 }
 
+static PyObject*
+PathGenerator_next_FOGSAA(PathGenerator* self)
+{
+    /* No need to create path because FOGSAA only finds one optimal alignment
+     * the .path fields should be populated by FOGSAA_EXIT_ALIGN. To indicate
+     * we've exausted the iterator, just set self->M[0][0].path to DONE */
```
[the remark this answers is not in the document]
michaelfm1211: Oops. I added another commit to fix that. Thanks!

mdehoon:
Thank you!

Issue #4788, opened 2024-08-08
Switch from MiniConda to MiniForge on AppVeyor


peterjc:
We currently use MiniConda which include's AnaConda's defaults channel, which is not 100% free https://docs.anaconda.com/miniconda/ - quote:

> Miniconda is free for anyone to use! However, access to Anaconda’s public repository of packages is only free to individuals and small organizations (<200 employees). A paid license is required for larger organizations and anyone embedding or mirroring Anaconda’s repository. See the [TOS](https://legal.anaconda.com/policies/en/) for details.

While I believe Biopython and OBF qualify, it would be prudent to avoid the defaults channel by switching to MiniForge and using just the community run conda-forge channel:

https://github.com/conda-forge/miniforge

For anyone already familiar with using conda at the command line on Windows, this should be a minor edit to https://github.com/biopython/biopython/blob/master/.appveyor.yml and perhaps some clarification in our installation instructions to steer users (we don't want Biopython users installing the non-free AnaConda content as many will be based at organizations with over 200 employees).

flaar94:
Sorry, I'm not very familiar with AppVeyer, but would that just require changing
```  
- ps: Start-FileDownload "https://repo.continuum.io/miniconda/Miniconda$env:PY_MAJOR_VER-latest-Windows-$env:PYTHON_ARCH.exe" C:\Miniconda.exe; echo "Finished downloading miniconda"
- cmd: C:\Miniconda.exe /S /D=C:\Py
```
with 
```
  - ps: Start-FileDownload "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Windows-$env:PYTHON_ARCH.exe" C:\Miniforge.exe; echo "Finished downloading miniforge"
  - cmd: C:\Miniforge.exe /S /D=C:\Py
```
?
I don't really have a good environment to test if it works well with AppVeyer's special Start-FileDownload command or how it handles bash commands (WSL doesn't want me running exe files). The pure ps approach would be 
`  - ps: start -Wait -FilePath C:\Miniforge.exe -ArgumentList "/S /D=C:\Py"`
I think.

peterjc:
Baring any surprises, yes, that could be all this needs. Do you want to try this on a personal fork (you many need to turn on AppVeyor integration), and if it works submit a pull request? Thank you.

If you can switch the command to powershell syntax, that is probably worth doing too.

[...]

PR #4795, opened 2024-08-20
Changed appveyor yaml to use miniforge instead of miniconda, and repl…

flaar94:
Testing

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,8 +31,8 @@ build_script:
         https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
         Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
         throw "There are newer queued builds for this pull request, failing early." }
-  - ps: Start-FileDownload "https://repo.continuum.io/miniconda/Miniconda$env:PY_MAJOR_VER-latest-Windows-$env:PYTHON_ARCH.exe" C:\Miniconda.exe; echo "Finished downloading miniconda"
-  - cmd: C:\Miniconda.exe /S /D=C:\Py
+  - ps: Start-FileDownload "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Windows-$env:PYTHON_ARCH.exe" C:\Miniforge.exe; echo "Finished downloading miniforge"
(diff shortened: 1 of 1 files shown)


peterjc:
Why drop 32-bit Windows? Does miniforge not support it?

flaar94:
> Why drop 32-bit Windows? Does miniforge not support it?

It doesn't look like they distribute 32 bit binaries, see https://github.com/biopython/biopython/issues/4788#issuecomment-2298054563 . Here I was just testing to see if the 64 bit version at least ran OK

flaar94:
Oops, did not intend to opened a pull request into the original repo. I was just testing. I'll try to find the automated process that did that

peterjc:
Looks like the failures are unrelated, see e.g. #4797

I suspect numpy 2.1.0, especially for the set_printoptions issues.

flaar94:
OK, that's good. I was wondering. So at least the 64 bit version seems to work. The 32 bit one failed to download because https://github.com/conda-forge/miniforge/releases/download/24.3.0-0/Miniforge3-Windows-x86.exe doesn't exist (but https://github.com/conda-forge/miniforge/releases/download/24.3.0-0/Miniforge3-Windows-x86_64.exe does)

peterjc:
Looking at https://github.com/conda-forge/miniforge there aren't any 32-bit Windows versions of the deprecated alternative mambaforge either.

Maybe you can try that workaround of installing a 64-bit miniforge, and then using it to create a 32-bit environment by setting ``CONDA_SUBDIR=win-32``?

If it doesn't work, we may have discuss dropping 32-bit Windows support.

flaar94:
Yeah, I'll give it a shot after work today. Regarding the numpy issue, it looks like as part of making `np.printoptions` function as a context manager, they changed `np.set_printoptions` to return a token 
```
@set_module('numpy')
def set_printoptions(precision=None, threshold=None, edgeitems=None,
                     linewidth=None, suppress=None, nanstr=None,
                     infstr=None, formatter=None, sign=None, floatmode=None,
                     *, legacy=None, override_repr=None):
    ...
    new_opt = _make_options_dict(precision, threshold, edgeitems, linewidth,
                                 suppress, nanstr, infstr, sign, formatter,
                                 floatmode, legacy)
    # formatter and override_repr are always reset
    new_opt['formatter'] = formatter
    new_opt['override_repr'] = override_repr

    updated_opt = format_options.get() | new_opt
    updated_opt.update(new_opt)

    if updated_opt['legacy'] == 113:
        updated_opt['sign'] = '-'

    token = format_options.set(updated_opt)
    return token
```

PR #4815, opened 2024-08-30
Standardizing on import numpy as np

### LINKED ISSUE #4238: Standardize on import numpy as np

john-science:

Closes #4238

This is a small change, where I converted every `import numpy` statement to `import numpy as np`.

This is, as the ticket suggest, just meant to help the repository converge on a standard. Assuming this change was made correctly, and there were not typos, such a change would be extremely safe and could not affect unit tests or runtime code.


john-science:
I see the linting failed.  Sorry, I didn't know you guys used `ruff`.  I will fix it.

mdehoon (approved):

peterjc (approved):
This looks great, happy to see this merged once the tests have finished.

peterjc:
That's weird - spelling errors from files you edited, fixed in 7ba26eb1a3ee83be125806ffd846272deca49869 but I would have expected those to have been flagged already based on how most pre-commit hooks behave.

Well, the AppVeyor and CircleCI tests passed so I _think_ we're good to merge anyway.

peterjc:
I understand the GitHub actions failure, which is due to recently added misspellings in the codespell tool.

When the pre-commit hooks are run on the pre-commit.ci service, this checks *all* the files. However, due to strict time limits we have to skip the slower hooks, including codespell. That means this didn't flag the two "misspellings" when the spell checker was updated.

On the other hand, we do run all the pre-commit hooks on GitHub Actions, but only on changed files. So again, this didn't flag the two "misspellings" immediately.

This was the first pull request to edit a file containing those words, and thus the first time we saw the failure.

Issue #4836, opened 2024-09-13
Type guards in the setter methods in the PDB module


kjczarne:
### Setup

I am reporting a problem with Biopython version, Python version, and operating
system as follows:

```
3.10.0 (default, Nov 10 2021, 13:28:23) [Clang 12.0.0 ]
CPython
macOS-14.0-arm64-arm-64bit
1.81
```

### Expected behaviour

With the following snippet of code:

```python
# For each of the CA atoms in the target structure, we can now update the coordinates:
for idx, atom in enumerate(target_atoms):
    if 'CA' in atom.get_id():
        atom.set_coord(superimposed[idx])
```

I would expect a type error. I just spent a good 30 min scratching my head why my downstream code fails and it turns out, in the above snippet I'm passing the entire `Atom` object as coords. The correct code would be:

```python
# For each of the CA atoms in the target structure, we can now update the coordinates:
for idx, atom in enumerate(target_atoms):
    if 'CA' in atom.get_id():
        atom.set_coord(superimposed[idx].coord)
```

### Actual behaviour

No error is thrown.

### Steps to reproduce

Using any arbitrary object in the `.set_coord()` method is permitted. There is not type checking performed. Only downstream, e.g. when the coordinates are getting unpacked as `x, y, z = atom.coords` does one realize that the wrong object has been set as a coordinate tuple.

**I am willing to work on this!** It is a UX aspect, not necessarily critical to the core capabilities of Biopython but it would save folks a lot of time to have some type bounds. I just need info whether the maintainers of Biopython would be interested in having this part improved and whether anybody is already working on this before I invest my time in this.

Thanks!

michaelfm1211:
Glancing over [Bio/PDB/Atom.py](https://github.com/biopython/biopython/blob/master/Bio/PDB/Atom.py), it looks like none of the setters (lines 307-361) perform type checking. We should probably add type checks on all of these. The types are all annotated in the constructor so I imagine this should be an easy change.

@peterjc @mdehoon Thoughts on this? Is there any reason these setters shouldn't have type checks?

flaar94:
I think adding type annotations to them would probably be valuable. There's some computation cost to runtime type guards, but as long as you aren't looping over millions of them it's probably fine. Also, you can just set the attributes directly if you want to do it quicker and avoid the type guard. I guess that code is ancient: from 2002, which might explain why they implemented so many getters/setters

peterjc:
[...]

PR #4841, opened 2024-09-17
Add type annotations to methods in Bio.PDB.Atom

michaelfm1211:

Related: https://github.com/biopython/biopython/issues/4836

The linked issue asks about runtime type checks in the Bio.PDB.Atom methods. @flaar94 suggested type checks might slow down code, but we should at least have type annotations so IDEs and mypy can catch mistakes.

### DIFF
--- a/Bio/PDB/Atom.py
+++ b/Bio/PDB/Atom.py
@@ -11,19 +11,15 @@ import copy
 import sys
 import warnings
 from typing import Optional
-from typing import TYPE_CHECKING
 from typing import TypeVar
 
 import numpy as np
 
-import Bio.PDB.Atom
 from Bio.Data import IUPACData
 from Bio.PDB.Entity import DisorderedEntityWrapper
 from Bio.PDB.PDBExceptions import PDBConstructionWarning
 from Bio.PDB.vectors import Vector
-
-if TYPE_CHECKING:
-    from Bio.PDB.Residue import Residue
+from Bio.PDB.Residue import Residue
 
 _AtomT = TypeVar("_AtomT", bound="Atom")
 
@@ -242,7 +238,7 @@ class Atom:
 
         return element
 
-    def _assign_atom_mass(self):
+    def _assign_atom_mass(self) -> float:
         """Return atom weight (PRIVATE)."""
         try:
             return IUPACData.atom_weights[self.element.capitalize()]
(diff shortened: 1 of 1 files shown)


Bio/PDB/Atom.py
```diff
@@ -308,23 +308,23 @@ def set_serial_number(self, n):
         """Set serial number."""
         self.serial_number = n
 
-    def set_bfactor(self, bfactor):
+    def set_bfactor(self, bfactor: Optional[float]):
```
peterjc, line 311: We still support Python 3.9 so can't use the https://peps.python.org/pep-0604/ syntax yet. On the bright side, pyupgrade and other tools can update this for us.

https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-604-typing-rewrites

Bio/PDB/Atom.py
```diff
@@ -387,31 +387,31 @@ def detach_parent(self):
...
         """Return standard deviations of anisotropic temperature factors."""
         return self.siguij_array
 
-    def get_anisou(self):
+    def get_anisou(self) -> Optional[np.ndarray]:
         """Return anisotropic B factor."""
         return self.anisou_array
 
-    def get_parent(self):
+    def get_parent(self) -> Optional[Residue]:
```
peterjc, line 402: This is breaking in the tests, ``NameError: name 'Residue' is not defined``

I guess either add the import, or try string based type annotation?

michaelfm1211:
@peterjc I just pushed a commit to fix that. The tests should now be passing.

Bio/PDB/Atom.py
```diff
@@ -11,19 +11,15 @@
 import sys
 import warnings
 from typing import Optional
-from typing import TYPE_CHECKING
 from typing import TypeVar
 
 import numpy as np
 
-import Bio.PDB.Atom
```
peterjc, line 19: Weird - this was importing itself?

peterjc (approved):
I'm happy to merge this after allowing a period for comments - remind me if there is no change after one week please.

Bio/PDB/Atom.py
flaar94: That's fascinating. I'm surprised it's not throwing a circular reference error, but I guess maybe it recognizes itself in the namespace and ignores the import? I think that TypeVar below is being created just to try to avoid circular references (I think that was more common when type checkers were less clever), so maybe someone accidentally broke it with a circular reference, and then the person who fixed it didn't clean up all the changes

peterjc: Could be - I didn't examine the history here.

michaelfm1211: Looking the `git blame Bio/PDB/Atom.py`, it looks like this line was added in #4593. My speculation is that the type annotations originally used `Atom` rather than the string `"Atom"` so it was imported, then when the `mypy` complained about using a type within itself the type annotation was changed to a string but removing the import was skipped over. I don't have any evidence for this though :)

michaelfm1211:
@peterjc Looks like there's been no further comments for the past two weeks (I meant to check in after one week, but I forgot; oops). The PR should be ready for merge.

PR #4874, opened 2024-10-25
Add SearchIO.InfernalIO module

drs:

already, or do not wish to be listed.

Add an Infernal search results parser and indexer. This parser handles plain text and tabular output. For tabular output, formats 1 (default), 2 and 3 are supported.


peterjc:
This will deserve a ``NEWS.rst`` entry.

Test failures with ``ModuleNotFoundError: No module named 'Bio.SearchIO.InfernalIO'``, you will need to manually add it to the package list in our ``setup.py`` (updating the packaging is something we ought to look at separately).

drs:
I fixed the `setup.py` and added the contribution in `NEWS.rst`.

NEWS.rst
```diff
@@ -38,6 +41,7 @@ e.g. motifs.write(motifs, "clusterbuster", precision=2)
 Many thanks to the Biopython developers and community for making this release
 possible, especially the following contributors:
 
+- Samuel Prince (first contribution)
```
peterjc, line 44: This should be alphabetical by first name, although it looks like that needs fixing anyway.

Bio/SearchIO/InfernalIO/infernal_tab.py
```diff
@@ -0,0 +1,363 @@
...
+    "seq_from": ("hit_start", int),
+    "seq_to": ("hit_end", int),
+    "strand": ("hit_strand", str),
+}
+
+
+class InfernalTabParser(_BaseInfernalParser):
+    """Parser for the Infernal tabular format."""
+
+    def __init__(self, handle, fmt=_DEFAULT_TAB_FORMAT):
```
peterjc, line 140: From a UI point of view, and without being familiar with Infernal, having a main SearchIO ``format`` argument (here ``infernal-tab`` I believe) and a secondary ``fmt`` argument seems confusing.

Does it make sense to think about ``format="infernal-tab1"``, ``format="infernal-tab2"`` etc?

peterjc:
Over to @bow to look at...

bow:
Hey @drs @peterjc Thanks for opening the PR! I'm just writing here to let you know that I am looking into it already ~ it's going to take me a while to go through the whole thing and I hope within the next 1-2 weeks I'll be able to reply more thoroughly.

Bio/SearchIO/InfernalIO/infernal_text.py
```diff
@@ -0,0 +1,543 @@
...
+# This file is part of the Biopython distribution and governed by your
+# choice of the "Biopython License Agreement" or the "BSD 3-Clause License".
+# Please see the LICENSE file that should have been included as part of this
+# package.
+
+
+"""Bio.SearchIO parser for Infernal plain text output format."""
+
+import re
+import operator
```
bow, line 12: Style suggestion (feel free to ignore):

```suggestion
import operator
import re
```

Bio/SearchIO/InfernalIO/infernal_text.py
```diff
@@ -0,0 +1,543 @@
...
+        """Iterate over query results."""
+        yield from self._parse_qresult()
+
+    def _read_until(self, bool_func):
+        """Read the file handle until the given function returns True (PRIVATE)."""
+        while True:
+            if not self.line or bool_func(self.line):
+                return
+            else:
+                self.line = read_forward(self.handle)
```
bow, line 78: Also a style suggestion, since the `else` is redundant:

```suggestion
            if not self.line or bool_func(self.line):
                return
            self.line = read_forward(self.handle)
```

bow (changes requested):
@drs Thank you again for opening this PR ~ impressive work! I am sorry it took me this long to get back to you.

I have added some comments / suggestions. There's only one minor typo that I'd like you to fix (the reason this is marked as 'request changes'). Most of the comments are suggestions and and questions about the code. So feel free to do what you think is best for those (including keeping current behavior).

peterjc:
If you are comfortable doing so, I would suggest rebasing this to the current master branch.

Doing a merge to resolve the conflicts will hinder our typical rebase-and-merge approach. But we could still do a squash-and-merge but that would lose your commit history.

drs:
I tried to rebase to the current master branch. I'm not very familiar with the process. Hopefully, I did that properly. @peterjc can you confirm if it's fine?

peterjc:
The rebase looks to have worked, although perhaps not perfectly as there are a few infernal tab tests failing.

Bio/SearchIO/InfernalIO/infernal_tab.py
```diff
@@ -0,0 +1,363 @@
...
+    "description": ("description", str),
+    "mdl_len": ("seq_len", int),
+}
+_COLUMN_HSP = {
+    "score": ("bitscore", float),
+    "evalue": ("evalue", float),
+    "bias": ("bias", float),
+    "gc": ("gc", float),
+    "mdl": ("model", str),
+    "trunc": ("truncated", str),
```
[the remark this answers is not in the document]
drs: `trunc` can be `no`, `5'`, `3'` or `5’&3'`. Changing the type to `bool` would remove some information. We could create new variables (ex. `5p_trunc` and `3p_trunc`) to keep all the information in a `bool`.

Bio/SearchIO/InfernalIO/infernal_tab.py
drs: That was my plan initially, but it's not straightforward to implement the inference that supports random access (i.e., `InfernalTabIndexer`). To initialize the `SearchIndexer`, we need the `fmt`, so we need to open the file and parse the header before we initialize `SearchIndexer`. My suggestion is to open the file for random access with `_open_for_random_access` from `Bio.File` (as in `SearchIndexer`;  https://github.com/biopython/biopython/blob/dda590d43f7b09eecab40982a4c1ac9fd49fc3b6/Bio/SearchIO/_index.py#L26), read the first two lines and use them to infer the header.

For `InfernalTabParser`, we can simply do: 
- If `fmt` is set, use the `fmt` and ignore the header
- Else read the first two lines of the handle, that should always be the header in a properly formatted file, and infer the header from these lines

I made a tentative update for this proposition in commit 373882b.

Bio/SearchIO/InfernalIO/infernal_tab.py
```diff
@@ -0,0 +1,363 @@
...
+                if sname in mapping:
+                    attr_name, caster = mapping[sname]
+                    if caster is not str:
+                        value = caster(value)
+                    parsed_dict[attr_name] = value
+
+        # adjust start and end coordinates according to strand
+        self._adjust_coords(frag)
+        # convert inclusion string to a bool
+        self._inclusion_str_to_bool(hsp)
```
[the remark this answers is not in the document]
drs: I agree with the suggestion and made this change in commit d0f8bf6.

Bio/SearchIO/InfernalIO/infernal_tab.py
```diff
@@ -0,0 +1,363 @@
...
+    "trunc": ("truncated", str),
+    "pass": ("pipeline_pass", int),
+    "inc": ("is_included", str),
+    "olp": ("olp", str),
+    "anyidx": ("anyidx", str),
+    "afrct1": ("afrct1", str),
+    "afrct2": ("afrct2", str),
+    "winidx": ("winidx", str),
+    "wfrct1": ("wfrct1", str),
+    "wfrct2": ("wfrct2", str),
```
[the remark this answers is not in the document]
drs: I agree. I changed the behavior in commit 45c8327 and updated documentation in commit 6625e99.

Bio/SearchIO/InfernalIO/infernal_tab.py
bow: The plan looks ok @drs. One comment about the inference logic itself: instead of doing this (paraphrased from code):

```
        if isinstance(fmt, int):
            if not 1 <= fmt <= 3:
                # <raise error>
            self.fmt = fmt
        else:
            self.fmt = <infer>
```

Let's do the check more explicitly (also to make the behavior less surprising):

```
        # The `isinstance` check below allows us to provide a consistent error
        # message, even when e.g. `fmt` happens to be unhashable.
        if isinstance(fmt, int) and fmt in {1, 2, 3}:
            self.fmt = fmt
        elif fmt is None:
            self.fmt = <infer>
        else:
            # raise exception
```

This would, e.g. raise an error when `fmt="1"` or `fmt="tabular-1"`.

Bio/SearchIO/InfernalIO/infernal_tab.py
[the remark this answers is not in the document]
bow: Aha :ok_hand: ~ no need to make a bool then, let's keep it as-is.

Bio/SearchIO/InfernalIO/infernal_tab.py
drs: I suggest that we drop the support for overriding the format and always infer from the header, as it can result in inconsistent behavior in some cases:

- There is no validation of the format when parsing the result rows. In most cases, when the format is incorrect, a `ValueError` (ex. `ValueError: invalid literal for int() with base 10: '-' `) is raised when casting the type in: 

https://github.com/drs/biopython-infernal/blob/f65381b7d54105b3aa9c3f440e1becb879c49966/Bio/SearchIO/InfernalIO/infernal_tab.py#L313-L318

But that's not always the case. If the user sets for format to 1, and the file format is 3, no error is raised, and the `description` will contain the three last columns `mdl len`, `seq len` and `description of target`. 

- Since the file is space separated, and the description can contain spaces, we cannot rely on the number of columns for validation.

drs: I removed the user-provided `fmt` argument. We still need to set the tabular format of the `InfernalTabParser` in the `InfernalTabIndexer` since the header cannot be passed to the parser. I keep the `fmt` argument as an internal value (now `_fmt`) for the `InfernalTabIndexer`.

drs:
@peterjc @bow I resolved most comments/suggestions (except contributions, which I will do at the end). I'm not sure that I understand why it's failing `codecov`. It there anything that I must do to address this? Is there anything that needs to be done on my end?

peterjc:
We're not so strict on the coverage reporting (it is a bit unreliable), but aspire to full coverage.

peterjc:
There are some git merge conflicts - if you are familiar with it, we would _prefer_ a `git rebase origin/master` to resolve them, rather than merges.

bow:
I agree with @peterjc ~ let's do a rebase to keep the history linear and clean. I can see the number of changed files in this commit is way too large. After a rebase, the coverage calculation might also be correct (i.e. it shouldn't fail because of files that are not part of the MR).

After the rebase, I don't have anything else from my end :slightly_smiling_face:.

drs:
@bow @peterjc I did the rebase. Everything looks good. I believe that the PR is ready to be merged.

I added this change to NEWS.rst under the upcoming v1.86.

bow:
Thank you @drs :tada: ~ I have just rebased this on top of `master` now :slightly_smiling_face:.

PR #4875, opened 2024-10-27
Melting Temp Efficiency Improvements

flaar94:

Built on top of PR #4866:
This draft pull request is intended to be eventually merged after PR #4866, and move discussion of further efficiency improvements into a dedicated PR

Current Changes:
- Modify string base set checks to use a faster subset-based approach


flaar94:
Regarding shifting from `string` to `bytes` methods, eg:
https://github.com/biopython/biopython/pull/4866#discussion_r1813895722

I think this is probably a good idea even if just for the simplicity of keeping things as one type. Unfortunately, I discovered that this will require a bit more care because some of the `seq` methods and functions that get called don't accept `bytes`, so we'll need to make some changes to allow that

PR #4891, opened 2024-11-25
Avoid loading of doctest: use Bio._utils instead

### LINKED ISSUE #4890: nuitka complains about loading doctest at Bio/Seq.py:3282
### Setup

Python: 3.12
Biopython: 1.84
Platform: Linux (GitHub CI)

### Expected behaviour

No loading doctest at Bio/Seq.py and use Bio._utils

### Actual behaviour

Nuitka-Plugins:WARNING: anti-bloat: Undesirable import of 'doctest' (intending to avoid 'unittest') in 'Bio.Seq' (at '/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/Bio/Seq.py:3282') encountered. It may slow down compilation.
Nuitka-Plugins:WARNING:     Complex topic! More information can be found at https://nuitka.net/info/unwanted-module.html

foxtran:

Closes #4890

In addition, fixes issues in Bio._utils since default_kwargs should be updated, not kwargs.

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -3277,6 +3277,7 @@ def complement_rna(sequence, inplace=False):
 
 
 if __name__ == "__main__":
+    from doctest import IGNORE_EXCEPTION_DETAIL
     from Bio._utils import run_doctest
 
-    run_doctest()
+    run_doctest(optionflags=IGNORE_EXCEPTION_DETAIL)
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -3276,14 +3276,7 @@ def complement_rna(sequence, inplace=False):
     return sequence.decode("ASCII")
 
 
-def _test():
-    """Run the Bio.Seq module's doctests (PRIVATE)."""
-    print("Running doctests...")
-    import doctest
-
-    doctest.testmod(optionflags=doctest.IGNORE_EXCEPTION_DETAIL)
-    print("Done")
-
-
 if __name__ == "__main__":
-    _test()
+    from Bio._utils import run_doctest
+
+    run_doctest()
--- a/Bio/_utils.py
+++ b/Bio/_utils.py
@@ -11,6 +11,7 @@ import os
 from typing import Any
 from typing import Callable
 from typing import cast
+from typing import Dict
 from typing import Optional
 from typing import Protocol
 from typing import TypeVar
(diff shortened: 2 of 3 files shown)


Bio/Seq.py
```diff
@@ -3276,14 +3276,8 @@ def complement_rna(sequence, inplace=False):
...
-    doctest.testmod(optionflags=doctest.IGNORE_EXCEPTION_DETAIL)
-    print("Done")
-
-
 if __name__ == "__main__":
-    _test()
+    from doctest import IGNORE_EXCEPTION_DETAIL
+    from Bio._utils import run_doctest
+
+    run_doctest(optionflags=IGNORE_EXCEPTION_DETAIL)
```
peterjc, line 3283: That looks like an oversight, did you see any more direct imports of doctest?

Bio/_utils.py
```diff
@@ -73,15 +74,15 @@ def run_doctest(target_dir: Optional[str] = None, *args: Any, **kwargs: Any) ->
...
+    doctest_attributes: Dict[str, Any] = {"optionflags": doctest.ELLIPSIS}
+    doctest_attributes.update(kwargs)
 
     cur_dir = os.path.abspath(os.curdir)
 
     print("Running doctests...")
     try:
         os.chdir(find_test_dir(target_dir))
-        doctest.testmod(*args, **kwargs)
+        doctest.testmod(*args, **doctest_attributes)
```
peterjc, line 85: So you're doing three things here:

* Type annotation, good
* Renaming a variable, because...
* Avoiding editing the passed in ``kwargs`` argument

Given they way we're using the ``run_tests`` function, I don't think we need to worry about side effects by editing the passed ``kwargs``, but this is cleaner.

peterjc (approved):

Bio/_utils.py
foxtran: > * Type annotation, good

It is because mypy said that it can not call doctest.testmod with Dict[str, int] without type annotation, so I did not have a change :(

> * Renaming a variable, because...

I needed to update default_kwargs, but it becomes not default, but some others...

> * Avoiding editing the passed in kwargs argument

Yep, it should not be written in this way, since before patch, it always set `optionflags` to ELLIPSIS even if in kwargs there was another `optionflags`

Bio/Seq.py
foxtran: I've rechecked with nuitka and it do not see :)
Compilation time was reduced significantly, may be it takes less than 5 minutes now :-) My binary size also lost half of size :D

peterjc:
Thank you!

PR #4899, opened 2024-12-16
Do not mention BAM when encountering an invalid bgzipped file

marcelm:

If a program reads both a BAM file and a bgzipped compressed FASTA or VCF file, the error message can confuse the user because it’s easy to overlook the "e.g.", and the user might think that the BAM file is the problematic one while it is the other.

We’ve had multiple reports of this for WhatsHap, which reads bgzipped references through pyfaidx. See these issues:

- https://github.com/whatshap/whatshap/issues/495
- https://github.com/whatshap/whatshap/issues/462
- https://github.com/whatshap/whatshap/issues/534

In WhatsHap, I will probably solve this by catching the exception and then printing an error message that also shows the name of the problematic file. However, this is a bit annoying because it’s merely a generic `ValueError` that bubbles up from pyfaidx and I will have to inspect its message in order to be able to distinguish it from other `ValueError`s. So I thought it would be nice to perhaps change this upstream as well.

### DIFF
--- a/Bio/bgzf.py
+++ b/Bio/bgzf.py
@@ -442,8 +442,7 @@ def _load_bgzf_block(handle, text_mode=False):
         raise StopIteration
     if magic != _bgzf_magic:
         raise ValueError(
-            r"A BGZF (e.g. a BAM file) block should start with "
-            r"%r, not %r; handle.tell() now says %r"
+            r"A BGZF block should start with %r, not %r; handle.tell() now says %r"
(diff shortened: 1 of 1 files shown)


peterjc:
This seems reasonable, assuming our test suite passes which it seems to.

[I am ignoring the numpy 2.2 related test failures, see e.g. #4898]

peterjc (approved):

PR #4903, opened 2025-01-02
Make appveyor use conda-forge while updating

flaar94:

This should fix the issue that was causing us to use conda's `defaults` channels again. They seem to have changed the `channels` list to include all of the URLs in the defaults channels instead of having just `defaults` which is what broke the previous code. It appears that the only place that we are actually using conda's defaults channels is in `conda update conda`. The package installations should be using pip and pypi.  Therefore the easiest fix seems to be to add arguments to force `update` to use conda-forge instead of the paid `defaults` channels. 

Note: Adding `defaults` to `denylist_channels` as shown in anaconda's docs actually seems to just break conda.

Context commit:
https://github.com/biopython/biopython/commit/cf7844c6e81b8d64b0e093b9cccdd2dbc42e04e0#diff-e1ab4c3cec24f841948eb867754d4ad0b9b6d11eaa6b5bfbe0357aceb0a49effR36

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,7 +35,7 @@ build_script:
   - ps: start -Wait -FilePath C:\Miniconda.exe -ArgumentList "/S /D=C:\Py"
   - SET PATH=C:\Py;C:\Py\Scripts;C:\Py\Library\bin;%PATH%
   - conda config --set always_yes yes
-  - conda update conda
+  - conda update conda --channel conda-forge --override-channels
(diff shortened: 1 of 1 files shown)


peterjc:
Do you want you name in the Biopython 1.895 release notes (``NEWS.rst``)?

flaar94:
> Do you want your name in the Biopython 1.895 release notes (`NEWS.rst`)?

No need. I'm already in the contrib.rst which is more than enough

PR #4939, opened 2025-02-15
Sphinx_rtd_theme update

### LINKED ISSUE #4923: Try updating from sphinx_rtd_theme v2 to v3
We are currently using ``sphinx_rtd_theme==2.0.0`` set in https://github.com/biopython/biopython/blob/master/.circleci/requirements-sphinx.txt when building the Tutorial and API docs. The setup is documented at https://github.com/biopython/biopython/tree/master/Doc

We ought to try upgrading to version 3, which would make the current version more obvious - see discussion on #4921

37Lime:

Closes #4923 

Updated `sphinx_rtd_theme==2.0.0` to `sphinx_rtd_theme==3.0.0`. Performed a local test and did not receive any conflicts with the current setup. Apologies if I missed something.

Best wishes.

### DIFF
--- a/.circleci/requirements-sphinx.txt
+++ b/.circleci/requirements-sphinx.txt
@@ -7,4 +7,4 @@ scipy
 numpydoc==1.7.0
 pygments
 sphinx==7.3.7
-sphinx_rtd_theme==2.0.0
+sphinx_rtd_theme==3.0.0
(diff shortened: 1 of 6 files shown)


peterjc:
Can you share a before/after screenshot? I'm curious if for example the version field is more obvious now (at the bottom of the left-hand column with v2).

37Lime:
Hello,

Thank you for the reply @peterjc.

I find the appearance equal concerning the version number (thus no pictures). Apologies if the err is on me, but I believe it should be correct (i.e., v3) since the update went through. 

However, I would say the version number is clear to the reader because it is also mentioned under the heading "**Biopython Documentation**" at the front page.

peterjc:
Merged, thank you.

PR #4942, opened 2025-02-18
include test outputs in gitignore

manulera:

These files are generated by `python setup.py test` and should not be included in the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,9 @@ ENV/
 
 # Testing
 **/.coverage
+
+# Files generated during testing
+other_trees.nex
+other_trees.xml
+tree1.nwk
+tree1.xml


manulera:
cc @peterjc

peterjc:
This should be fixed, but I disagree on the how. Instead of telling git to ignore these files, I think the tests should not leave these files behind - ideally by using a temprary location in the first place, or if these are part of  doctest where that would detract from the example, they should be cleanedup afterwards.

manulera:
Fair enough. I had a go at doing this properly, but it's really hard to follow how tests are run exactly when calling `python setup.py test` and since the whole test takes very long it's hard to isolate the problem and iterate.

My two cents:
* The below code worked when running the test as a python script (it deleted the files), but did not work when calling `python setup.py test`
* Calling `doctest.testmod()` does not run the `tearDown` function.

Screenshot 2025-02-19 at 18 52 46

peterjc:
See https://biopython.org/docs/latest/Tutorial/chapter_testing.html

Most of the test files can be run direclty, but to run the doctests alone is best done via `run_tests.py doctest`

PR #4977, opened 2025-03-21
Locale independent date serialization

### LINKED ISSUE #4972: Date ignored in Genbank writer with not-english locale
Hi,

I am writting genbank file, and sometime date are silently ignored and replace by 1980. My first suggestion is to add a warning somehow indicating that the date annotation is invalide.

When investigating [the code producing the date](https://github.com/biopython/biopython/blob/master/Bio/SeqIO/InsdcIO.py#L627-L659) if found that datetime instance are only considered if the abbreviated month is equivalent to the english abbreviation. With not english locale it thus sometime ignore the date  and silently replace it by `01-JAN-1980`.

If you confirm me that it is not the expected behavior to not work with non-english locale , I can work on a PR

Best regards,

Step to reproduce:
```python
"""
docker run -it python:3.12 bash
apt-get update && apt-get install -y locales
echo "fr_FR.UTF-8 UTF-8" >> /etc/locale.gen
echo "de_DE.UTF-8 UTF-8" >> /etc/locale.gen
echo "pt_BR.UTF-8 UTF-8" >> /etc/locale.gen
echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen
echo "en_GB.UTF-8 UTF-8" >> /etc/locale.gen
locale-gen
locale -a # check that all three lang are not there
pip install git+https://github.com/biopython/biopython.git
python # and run following script
"""

import locale
from datetime import datetime
from tempfile import NamedTemporaryFile

from Bio.Seq import Seq
import Bio.SeqIO

date = datetime.strptime('2025-10-01', '%Y-%m-%d')
record = SeqRecord(
    Seq('ATGC'),
    id="foo",
    annotations={
        "molecule_type": "dna",
        "topology": "circular",
        "date": date,
    },
)
this_year = str(date.year)
for l in [
    'en_GB.utf8',
    'de_DE.utf8',
    'fr_FR.utf8',
    'pt_BR.utf8',
    'en_US.utf8',
]:
    locale.setlocale(locale.LC_ALL, l)
    with NamedTemporaryFile(suffix='.gb') as file:
        with open(file.name, 'w') as gb_file:
            Bio.SeqIO.write(record, gb_file, "genbank")
        with open(file.name, 'r') as written_file:
            first_line = written_file.readlines()[0]
            print(f"In {l}: {first_line}")
            if this_year not in first_line:
                print(f"ERROR with {l} as date is {date.strftime('%d-%b-%Y').upper()}")

```

```log
'en_GB.utf8'
1
In en_GB.utf8: LOCUS       foo                        4 bp    dna     circular UNK 01-OCT-2025

'de_DE.utf8'
1
In de_DE.utf8: LOCUS       foo                        4 bp    dna     circular UNK 01-JAN-1980

ERROR with de_DE.utf8 as date is 01-OKT-2025
'fr_FR.utf8'
1
In fr_FR.utf8: LOCUS       foo                        4 bp    dna     circular UNK 01-JAN-1980

ERROR with fr_FR.utf8 as date is 01-OCT.-2025
'pt_BR.utf8'
1
In pt_BR.utf8: LOCUS       foo                        4 bp    dna     circular UNK 01-JAN-1980

ERROR with pt_BR.utf8 as date is 01-OUT-2025
'en_US.utf8'
1
In en_US.utf8: LOCUS       foo                        4 bp    dna     circular UNK 01-OCT-2025

```

bryan-brancotte:

Locale independent date serialization when writing GenBank file. Also added support for datetime.date.
Added test with various local known to make the former serialization fail (when available).

Show a warning when writing a file with an invalide date format

Closes #4972

### DIFF
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -647,7 +647,7 @@ class GenBankWriter(_InsdcWriter):
             "NOV",
             "DEC",
         ]
-        if isinstance(date, datetime) or isinstance(date, datetime_date):
+        if isinstance(date, datetime_date):
             date = f"{date.day:02d}-{months[date.month - 1]}-{date.year}"
         if not isinstance(date, str) or len(date) != 11:
             warnings.warn(
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -651,7 +651,7 @@ class GenBankWriter(_InsdcWriter):
             date = f"{date.day:02d}-{months[date.month - 1]}-{date.year}"
         if not isinstance(date, str) or len(date) != 11:
             warnings.warn(
-                f"Invalide date format provided {record.annotations['date']!r}, using default {default!r}",
+                f"Invalid date format provided {record.annotations['date']!r}, using default {default!r}",
                 BiopythonWarning,
             )
             return default
@@ -659,7 +659,7 @@ class GenBankWriter(_InsdcWriter):
             datetime(int(date[-4:]), months.index(date[3:6]) + 1, int(date[0:2]))
         except ValueError:
             warnings.warn(
-                f"Invalide date provided {record.annotations['date']!r}, using default {default!r}",
+                f"Invalid date provided {record.annotations['date']!r}, using default {default!r}",
                 BiopythonWarning,
             )
             date = default
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -650,10 +650,18 @@ class GenBankWriter(_InsdcWriter):
         if isinstance(date, datetime) or isinstance(date, datetime_date):
             date = f"{date.day:02d}-{months[date.month - 1]}-{date.year}"
         if not isinstance(date, str) or len(date) != 11:
+            warnings.warn(
+                f"Invalide date format provided {record.annotations['date']!r}, using default {default!r}",
+                BiopythonWarning,
+            )
             return default
         try:
             datetime(int(date[-4:]), months.index(date[3:6]) + 1, int(date[0:2]))
         except ValueError:
+            warnings.warn(
+                f"Invalide date provided {record.annotations['date']!r}, using default {default!r}",
+                BiopythonWarning,
+            )
             date = default
         return date
 
--- a/Bio/SeqIO/InsdcIO.py
+++ b/Bio/SeqIO/InsdcIO.py
@@ -31,7 +31,7 @@ http://www.ebi.ac.uk/imgt/hla/docs/manual.html
 """
 
 import warnings
-from datetime import datetime
+from datetime import datetime, date as datetime_date
 from string import ascii_letters
 from string import digits
 
@@ -633,9 +633,6 @@ class GenBankWriter(_InsdcWriter):
         # Cope with a list of one string:
         if isinstance(date, list) and len(date) == 1:
             date = date[0]
-        if isinstance(date, datetime):
-            date = date.strftime("%d-%b-%Y").upper()
-
         months = [
             "JAN",
             "FEB",
@@ -650,6 +647,8 @@ class GenBankWriter(_InsdcWriter):
             "NOV",
             "DEC",
         ]
+        if isinstance(date, datetime) or isinstance(date, datetime_date):
+            date = f"{date.day:02d}-{months[date.month - 1]}-{date.year}"
         if not isinstance(date, str) or len(date) != 11:
             return default
         try:
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -74,6 +74,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - Brandon Carter  <https://github.com/b-carter>
 - Brandon Invergo <https://github.com/brandoninvergo>
 - Brian Osborne <https://github.com/bosborne>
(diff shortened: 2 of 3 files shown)


bryan-brancotte:
The tests try using various locale, which are not by default in ubuntu-latest, and thus wont be tested in the CI, while they will be tested on a local computer if locale is installed manually. We can add it to the CI, but it add 30MB of download, some more time. I don't know what one would prefer

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 82f5f7626..84c9493ad 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,14 @@ jobs:
       run: |
         pre-commit run --files ${{ steps.changed-files.outputs.all_changed_files }}
 
+    - name: Install locale
+      run: |
+        sudo apt-get update
+        echo "fr_FR.UTF-8 UTF-8" | sudo tee -a /etc/locale.gen
+        echo "de_DE.UTF-8 UTF-8" | sudo tee -a /etc/locale.gen
+        echo "pt_BR.UTF-8 UTF-8" | sudo tee -a /etc/locale.gen
+        sudo apt-get -y install locales
+
   # Cache pip & python dependencies for Linux runner
   # since it is used in most actions.
   # Adapted from: https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d
```

P.S: diff not tested

Bio/SeqIO/InsdcIO.py
```diff
@@ -650,11 +647,21 @@ def _get_date(record):
             "NOV",
             "DEC",
         ]
+        if isinstance(date, datetime) or isinstance(date, datetime_date):
+            date = f"{date.day:02d}-{months[date.month - 1]}-{date.year}"
```
peterjc, line 651: You can check both classes at once with `isinstance(date, (datetime, datetime_date))` - see https://docs.python.org/3/library/functions.html#isinstance

However, we only need to check the `date` class as `datetime` is a subclass of `date` -- see https://docs.python.org/3/library/datetime.html#available-types or:

```pycon
>>> import datetime
>>> d = datetime.date(2025,6,30)
>>> d
datetime.date(2025, 6, 30)
>>> isinstance(d, datetime.date)
True
>>> isinstance(d, datetime.datetime)
False
>>> dt = datetime.datetime(2025, 6, 30)
>>> dt
datetime.datetime(2025, 6, 30, 0, 0)
>>> isinstance(dt, datetime.date)
True
>>> isinstance(dt, datetime.datetime)
True
```

peterjc (changes requested):
I wouldn't install the extra locales in the CI just for this test - the overhead seems excessive.

Bio/SeqIO/InsdcIO.py
```diff
@@ -650,11 +647,21 @@ def _get_date(record):
...
+            warnings.warn(
+                f"Invalide date format provided {record.annotations['date']!r}, using default {default!r}",
+                BiopythonWarning,
+            )
             return default
         try:
             datetime(int(date[-4:]), months.index(date[3:6]) + 1, int(date[0:2]))
         except ValueError:
+            warnings.warn(
+                f"Invalide date provided {record.annotations['date']!r}, using default {default!r}",
```
peterjc, line 662: typo `Invalide` -> `Invalid` (twice)

Bio/SeqIO/InsdcIO.py
```diff
@@ -650,11 +647,21 @@ def _get_date(record):
...
+                BiopythonWarning,
+            )
```
peterjc, line 664: Do these new warnings appear in the main test output? If so we should silence this in the test using the `warnings` module context manager and/or explicitly check the warning is triggered.

Bio/SeqIO/InsdcIO.py
bryan-brancotte: Hello, yes there are tested/triggerd, I used `with warnings.catch_warnings` [here](https://github.com/biopython/biopython/pull/4977/files#diff-2feda3dea27777471c9f0066eea2db0e68dbd4c4a9e5ca052043362656497f4eR1119) and [there](https://github.com/biopython/biopython/pull/4977/files#diff-2feda3dea27777471c9f0066eea2db0e68dbd4c4a9e5ca052043362656497f4eR1131) to check they are indeed raised. I don't see them in the output, so guessing that prevent them from being printed ?

Bio/SeqIO/InsdcIO.py
bryan-brancotte: Hello, yes there are tested/triggerd, I used `with warnings.catch_warnings` [here](https://github.com/biopython/biopython/pull/4977/files#diff-2feda3dea27777471c9f0066eea2db0e68dbd4c4a9e5ca052043362656497f4eR1119) and [there](https://github.com/biopython/biopython/pull/4977/files#diff-2feda3dea27777471c9f0066eea2db0e68dbd4c4a9e5ca052043362656497f4eR1131) to check they are indeed raised. I don't see them in the output, so guessing that prevent them from being printed ?

Bio/SeqIO/InsdcIO.py
bryan-brancotte: Thanks, I wan't aware of the isinstance with a tuple of class. Addressed

Bio/SeqIO/InsdcIO.py
peterjc: Ah - nice!

peterjc (approved):

PR #4984, opened 2025-03-27
Fix/seq feature fromstring

### LINKED ISSUE #4983: Decorator on `fromstring` method of `Location` class missing in Bio.SeqFeature
The method should be marked as classmethod or staticmethod. Considering that it's an ABC and `SimpleLocation` etc. are instantiated by name, staticmethod is probably the way to go.

https://github.com/biopython/biopython/blob/32b964861787b8e3e7123eb2ecf1ffe744c4298e/Bio/SeqFeature.py#L627-L630

FynnFreyer:

Closes #4983

The fromstring method of the abstract Location class is missing a method decorator, which causes the first argument to be interpreted as self. This doesn't lead to any problems when calling it from the class, but confuses tooling.

Note that the commit that added this (908ee298) speaks of a "fromstring class method", so taking this method in the instance context should be fine.

I agree to my contributions being dual licensed under the BSD 3-Clause and Biopython License
Signed-off-by: Fynn Freyer <fynn.freyer@googlemail.com>

### DIFF
--- a/Bio/SeqFeature.py
+++ b/Bio/SeqFeature.py
@@ -625,6 +625,7 @@ class Location(ABC):
         """Represent the Location object as a string for debugging."""
         return f"{self.__class__.__name__}(...)"
 
+    @staticmethod
     def fromstring(text, length=None, circular=False, stranded=True):
         """Create a Location object from a string.
 
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -138,6 +138,7 @@ please open an issue on GitHub or mention it on the mailing list.
 - Frederic Sohm <fsms at domain users.sourceforge.net>
 - Frederik Gwinner
 - Fredric Johansson <https://github.com/fredricj>
+- Fynn Freyer <https://github.com/FynnFreyer>
 - Fábio Madeira <https://github.com/biomadeira>
 - Gaetan Lehman <gaetan.lehmann at domain jouy.inra.fr>
 - Gavin E Crooks <https://github.com/gecrooks>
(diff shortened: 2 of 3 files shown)


peterjc:
This looks good, but could you add an explicit test for the broken behaviour it fixes? Thanks!

FynnFreyer:
The behaviour itself isn't broken, it's just that it wasn't decorated properly before.

But I guess I could check if `isinstance(Location.__dict__['fromstring'], staticmethod)` is true. That would distinguish between static and instance methods. Would that suffice?

FynnFreyer:
Really not sure what else to do. This is just an issue with tooling. I only noticed this, because my IDE told me that `Location.fromstring` expects a `Location` instead of a `str`. But, e.g., when running `Location.fromstring("1..207")` it just works, because there's no self that is passed.

FynnFreyer:
Okay, I came up with another way to test this

```python
# with old implementation
#   behaviour of CompoundLocation.fromstring would change
#   depending on whether we call from instance or class
f1 = SimpleLocation(10, 40)
f2 = SimpleLocation(50, 59)
instance = CompoundLocation([f1, f2])
spec = "10..40"
self.assertEqual(Location.fromstring(spec), instance.fromstring(spec))
```

This fails with the old implementation and succeeds with the new one. I'll add this to the test

FynnFreyer:
Caveat Emptor: This (second) kind of test would become nonsensical, as soon as `CompoundLocation` starts providing its own implementation for `fromstring`.

peterjc:
Ah - I hadn't sat down to try this but that fact that the code was usable despite the missing decorator annotation does explain how we missed it. Do you want to take a look too @mdehoon as the original author of #4128? Thanks!

FynnFreyer:
Should I remove the test then? While somewhat contrived, it does highlight a potential issue with the missing decorator, but it's also kind of a footgun, for something that basically amounts to a typo...

mdehoon (approved):

peterjc:
Thank you both!

PR #4991, opened 2025-04-29
fix model_n

jamesmkrieger:

Solves the issue reported in https://github.com/prody/ProDy/issues/2111, where biopython MMCIFIO _save_structure was writing model serial num 1 twice so there was one model too few and the number of atoms was wrong too.


peterjc:
Does this round trip? i.e. is there a `- 1` operation on parsing?

jamesmkrieger:
> Does this round trip? i.e. is there a `- 1` operation on parsing?

Actually, it doesn't so that might be a problem. Maybe it's better to not add 1 at all and leave model number `0` as 0 too.

PR #4998, opened 2025-05-19
improved error handling, should not show misleading message

### LINKED ISSUE #4996: retreive_pdb_file() in PDBlist.py has no error handling for HTTP Error or URL error
If the user is in an environment where the firewall or SSL cert errors prevent `urlretrieve(url, filename)` from downloading the requested PDB file, the function prints "Desired structure doesn't exist" but still returns a filename without processing the error. This can be resolved by adding some error catching:

```

    # Retrieve the file(s)
    if self._verbose:
        print(f"Downloading PDB structure '{pdb_code}'...")
    try:
        urlcleanup()
        urlretrieve(url, filename)
    except urllib.error.HTTPError as e:
        print(f"Failed to download PDB structure: '{pdb_code}': HTTP Error {e.code} - {e.reason}")
        return None
    except urllib.error.URLError as e:
        print(f"Failed to download PDB structure: '{pdb_code}': {e.reason}")
        return None
    except OSError as e:
        print(f"Desired structure not found or doesn't exist: " '{pdb_code}': {str(e)}")
        return None
    else:
        with gzip.open(filename, "rb") as gz:
            with open(final_file, "wb") as out:
                out.writelines(gz)
        os.remove(filename)
    return final_file
```

lee-t:
fixed PDBlist.retrieve_pdb_file(); improved error handling, should not show misleading message

Closes #4996

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -342,13 +342,16 @@ class PDBList:
             print(
                 f"Failed to download PDB structure: '{pdb_code}':"
                 f" HTTP Error {e.code} - {e.reason}")
+            return None
         except URLError as e:
             print(
                 f"Failed to download PDB structure: '{pdb_code}':"
                 f" URL Error {e.reason}")
+            return None
         except OSError as e:
             print(f"Desired structure not found or doesn't exist:"
                   f" '{pdb_code}': {str(e)}")
+            return None
         else:
             with gzip.open(filename, "rb") as gz:
                 with open(final_file, "wb") as out:
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -341,16 +341,20 @@ class PDBList:
         except HTTPError as e:
             print(
                 f"Failed to download PDB structure: '{pdb_code}':"
-                f" HTTP Error {e.code} - {e.reason}")
+                f" HTTP Error {e.code} - {e.reason}"
+            )
             return None
         except URLError as e:
             print(
                 f"Failed to download PDB structure: '{pdb_code}':"
-                f" URL Error {e.reason}")
+                f" URL Error {e.reason}"
+            )
             return None
         except OSError as e:
-            print(f"Desired structure not found or doesn't exist:"
-                  f" '{pdb_code}': {str(e)}")
+            print(
+                f"Desired structure not found or doesn't exist:"
+                f" '{pdb_code}': {str(e)}"
+            )
             return None
         else:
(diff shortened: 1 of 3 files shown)


Tests/test_PDB_PDBList.py
```diff
@@ -159,6 +159,20 @@ def test_invalid_file_format(self):
...
+        """Tests retrieving a non-existent molecule returns None."""
+        with self.make_temp_directory(os.getcwd()) as tmp:
+            pdblist = PDBList(pdb=tmp)
+            result = pdblist.retrieve_pdb_file("zzzz", file_format="pdb")
+            self.assertIsNone(result)
+
+    def test_retrieve_pdb_file_bad_url(self):
+        """Tests retrieving with bad server URL returns None."""
+        with self.make_temp_directory(os.getcwd()) as tmp:
+            pdblist = PDBList(server="http://invalid.server", pdb=tmp)
```
peterjc, line 172: As per https://www.rfc-editor.org/rfc/rfc2606.html please swap that round to eg http://something.invalid instead.

Bio/PDB/PDBList.py
```diff
@@ -336,8 +338,24 @@ def retrieve_pdb_file(
...
+                f"Failed to download PDB structure: '{pdb_code}':"
+                f" URL Error {e.reason}"
+            )
+            return None
+        except OSError as e:
+            print(
+                f"Desired structure not found or doesn't exist:"
+                f" '{pdb_code}': {str(e)}"
+            )
+            return None
```
peterjc, line 358: The explicit `return None` is the minimal bug fix, right?

peterjc:
As per my comment on #4996, I'm not sure we should catch the other two exceptions - but I don't use this code so am not the best judge here for the end-user perspective.

Bio/PDB/PDBList.py
lee-t: Essentially correct.  `except OSError  as e:`  would catch the raw exception for the print statement then `return None`  would solve the bug which the object is still returned after error.

lee-t:
I don't have a particularly strong reason for including both urllib.error subclasses other than I wanted to catch my particular error when I noticed this bug. I do agree that maybe a more minimal bugfix would suffice rather than import more urllib classes.

Tests/test_PDB_PDBList.py
lee-t: ba01c91ab4976839fc6c2516f599cd555bf01d4a

peterjc:
@JoaoRodrigues can you take a look please?

PR #5002, opened 2025-05-30
Added Documentation to CEAlign and QCPSuperimposer

### LINKED ISSUE #4768: Update the Structural Alignment section of the Tutorial to include CEAlign
We should update the (very sparse) section of the tutorial on structure alignment ([this one](https://biopython.org/DIST/docs/tutorial/Tutorial.html#sec295)) to include the CEAlign and QCPSuperimpose modules and more examples of how to use them.

dimi1729:

Closes #4768 

Since the "Analyzing structures" part of the documentation was getting crowded, I made a new section named "Superimposing two structures". I moved the previous documentation for using the Superimposer class and for "Mapping the residues of two related structures onto each other" into this new section, and added documentation for using CEAlign and QCPSuperimposer. I also added the citations in the docstrings of each of these classes to the bibliography.


Doc/Tutorial/chapter_pdb.rst
```diff
@@ -1846,6 +1803,112 @@ You can also get access to the molecular surface itself (via the
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, create an alignment file in FASTA format, then use the
+``StructureAlignment`` class. This class can also be used for alignments
+with more than two structures.
+
+Structural alignment with CEAligner
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want to align two structures that may not have identical sequences or lengths,
```
JoaoRodrigues, line 1886: I'd add that CEAlign is more suited for cases where the proteins have _low_ sequence identity, not necessarily not identical.

dimi1729: Can you actually elaborate on this a little (I do not want to add incorrect information due to my misunderstanding). I thought that the superimposed only worked with identical lists of atoms, so to impose two non-identical structures (either with low or high seq id) you would have to use cealign. Please correct me if I am wrong.

JoaoRodrigues: Structure alignment involves two steps: figuring out which atoms (or residues) correspond to which, and then finding the best rotation/translation moves that minimize the RMSD between those atoms. `Superimposer` (and `QCPSuperimpose`) does the second part only (it's basically a big wrapper on a SVD call). You need to provide two equal-sized lists of atoms to it, which in the docs we just happen to use something like `list(Structure.get_atoms())` between two equal-sized structures for simplicity. `CEAligner` does both parts, first figuring out the correspondence and then performing the superposition.

Scientifically, if the sequences of the two proteins are very similar (like >50%), you should use a regular sequence alignment and then pass it to `StructureAlignment`. We could change and break the API of `StructureAlignment` to run this alignment under the hood automatically (although there are quirks to it) to make it a one-step process, but that's a separate PR. If the sequences are more dissimilar, you should use `CEAligner`. Note that `CEAligner` will always work, but because it finds the correspondences based on structural similarity, you might sometimes end up with some small mismatches in the sequence, especially near termini/chain breaks or edges of secondary structure elements.

JoaoRodrigues: Adding to the above. You could have two structures of different sizes and still use Superimposer. You would run a sequence alignment, find the correspondences, and then call it. See an example here: https://gist.github.com/JoaoRodrigues/e3a4f2139d10888c679eb1657a4d7080

dimi1729: Thank you for the clarification, I will update the docs here and also for clarity I will add more to the Superimposer docs to show how to align similar proteins and then superimposing them.

However, I do think that that it would be more user friendly to have this be a one step process, so after this PR I plan on making another one to do this.

JoaoRodrigues: Thank you @dimi1729 , appreciate it! And yeah, go ahead and work on the other PR if you are willing to, that would be quite helpful. If you want to open an issue first, to discuss how the API should look etc, feel free.

JoaoRodrigues:
Merged, thank you again @dimi1729 !

PR #5018, opened 2025-06-20
Make `PDB.Chain.__getitem__` more lenient for indexing with `int`-like values 

### LINKED ISSUE #5017: `Bio.PDB.Chain` cannot be indexed with `np.int64`
I'm not sure if this is expected behaviour or not. 

I came across a confusing error (although this could be my fault) when attempting to index into a `Chain` instance from a protein structure parsed by `Bio.PDB`.  

I am using `np.random` in order to select a subset of residues from a `chain`.  
Selecting from the `residue` objects directly, results in a `ValueError`. 

```python
>>> import numpy as np
>>> from Bio.PDB import PDBParser
>>> parser = PDBParser() 
>>> s1 = parser.get_structure("7KN6", "7KN6.pdb")
>>> chain = s1[0]["A"]
>>> all_residues = list(chain.get_residues()) 
>>> np.random.choice(all_residues, size=10, replace=False)
# ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (119,) + inhomogeneous part.
```
I am therefore selecting the indexes (or sequence positions) of the residues:
```python
res_positions = [r.id[1] for r in all_residues]

random_positions = np.random.choice(res_positions, size=10, replace=False)
for rp in random_positions: 
    
    rp in chain # == False
    int(rp) in chain # == True 


```

However, when an `np.int64` is used to index into the `chain`, an `IndexError` is confusingly raised.  Is there room to make the `Chain.__getitem__` method more lenient when receiving `int`-like values?

kamurani:

Closes #5017

### DIFF
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -98,6 +98,8 @@ class Chain(Entity["Model", "Residue"]):
             id = int(id)
         except ValueError:
             pass
+        except TypeError:
+            pass
         if isinstance(id, int):
             id = (" ", id, " ")
         return id
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -94,7 +94,7 @@ class Chain(Entity["Model", "Residue"]):
          - id - int, residue resseq
 
         """
-        try: 
+        try:
             id = int(id)
         except ValueError:
             pass
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -6,7 +6,6 @@
 # package.
 
 """Chain class, used in Structure objects."""
-import numbers
 
 from typing import Optional
 from typing import TYPE_CHECKING
@@ -95,7 +94,7 @@ class Chain(Entity["Model", "Residue"]):
          - id - int, residue resseq
 
         """
-        if isinstance(id, numbers.Integral):
+        if isinstance(id, int):
             id = (" ", id, " ")
         return id
 
--- a/Bio/PDB/Chain.py
+++ b/Bio/PDB/Chain.py
@@ -94,6 +94,10 @@ class Chain(Entity["Model", "Residue"]):
          - id - int, residue resseq
 
         """
+        try: 
+            id = int(id)
+        except ValueError:
+            pass
         if isinstance(id, int):
             id = (" ", id, " ")
         return id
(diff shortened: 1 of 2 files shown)


kamurani:
As an alternative:


```python
import numpy as np 

n = np.int64(4) 

n == int(n)
# True 

isinstance(n, numbers.Integral) 
# True 

isinstance(n, int) 
# False 

```

JoaoRodrigues:
Thanks for the PR, what's the use case for this?

A quinta, 19/06/2025, 22:26, Kâmuran İmran ***@***.***>
escreveu:

JoaoRodrigues:
Oops, nevermind, didn't see the issue..

A sexta, 20/06/2025, 17:33, João Rodrigues ***@***.***>
escreveu:

Bio/PDB/Chain.py
```diff
@@ -119,6 +120,10 @@ def __contains__(self, id):
          - id - (string, int, string) or int
 
         """
+        try:
```
JoaoRodrigues: I'd include this attempted conversion in the `_translate_id` method and remove the `isinstance` check. This should solve your problem no?

JoaoRodrigues (changes requested):

JoaoRodrigues:
Looks good to me! Please have a look at the failed tests/CI, happy to merge once they clear.

kamurani:
@JoaoRodrigues sorry for the delay; should all be passing now.

Bio/PDB/Chain.py
```diff
@@ -94,6 +94,12 @@ def _translate_id(self, id):
          - id - int, residue resseq
 
         """
+        try:
+            id = int(id)
```
peterjc, line 98: This does not seem to accept octal or hex notation (without explicitly giving the base), which is good here.

peterjc:
Do we have a (small) test case this fixes?

JoaoRodrigues (approved):

kamurani:
@peterjc please let me know if this test I've added makes sense - and if there's anything else you want me to test.

kamurani:
@peterjc  just bumping this.

peterjc:
OK, I'll assume Joao is still happy and merge this. Thank you both!

PR #5022, opened 2025-06-27
Add option to ignore N-/C-terminus effects in isoelectric point calculation

FriedLabJHU:

This update introduces a new ignore_termini parameter to the IsoelectricPoint class, allowing users to exclude the N-terminus, C-terminus, or both from isoelectric point (pI) and charge calculations.

Key Changes:
Added ignore_termini argument to IsoelectricPoint.__init__, which accepts:

"Nterm" – excludes N-terminal charge and pK contribution

"Cterm" – excludes C-terminal charge and pK contribution

"both" – excludes both N- and C-terminal contributions

Purpose:
In many use cases (e.g., peptide segments derived from a larger protein), the termini of a fragment may not be chemically exposed and therefore do not contribute to the molecule’s charge in physiological contexts. This enhancement enables more biologically realistic pI calculations for internal peptide fragments by omitting artificial terminal effects.

Additionally, it allows for flexible modeling of peptides with only one exposed terminus, such as C-terminal peptides (ignore Nterm) or N-terminal peptides (ignore Cterm), which is useful in proteomics applications and fragment-based analyses.

### DIFF
--- a/Bio/SeqUtils/IsoelectricPoint.py
+++ b/Bio/SeqUtils/IsoelectricPoint.py
@@ -79,16 +79,32 @@ class IsoelectricPoint:
 
     """
 
-    def __init__(self, protein_sequence, aa_content=None):
+    def __init__(self, protein_sequence, aa_content=None, ignore_termini=None):
         """Initialize the class."""
         self.sequence = protein_sequence.upper()
+
         if not aa_content:
             from Bio.SeqUtils.ProtParam import ProteinAnalysis as _PA
-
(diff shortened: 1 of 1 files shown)


peterjc:
I thought there might be a similar option in one of the existing `Bio.SeqUtils` functions, but I can't immediately find it. If there is, consistent naming would be nice.

FriedLabJHU:
@peterjc as far as I am aware, there aren't any other functions that explicitly ignore the protein termini (other than by indexing). If there is an existing option meant for ignoring termini, then I will change this PR to match that naming.

PR #5045, opened 2025-08-26
Fix typo/bug in biopython.Bio.PDB.vectors vector_to_axis

Mikel-MG:

Currently, due to module shadowing, calling the `vector_to_axis()` function results in the following error

```
AttributeError: 'numpy.float64' object has no attribute 'cos'
```

### DIFF
--- a/Tests/test_PDB_vectors.py
+++ b/Tests/test_PDB_vectors.py
@@ -18,17 +18,15 @@ except ImportError:
         "Install NumPy if you want to use Bio.PDB."
     ) from None
 
-from Bio.PDB import calc_angle
-from Bio.PDB import calc_dihedral
-from Bio.PDB import m2rotaxis
-from Bio.PDB import refmat
-from Bio.PDB import rotaxis
-from Bio.PDB import rotmat
-from Bio.PDB.vectors import coord_space
-from Bio.PDB.vectors import get_spherical_coordinates
-from Bio.PDB.vectors import homog_trans_mtx
-from Bio.PDB.vectors import multi_coord_space
-from Bio.PDB.vectors import Vector
+from Bio.PDB import calc_angle, calc_dihedral, m2rotaxis, refmat, rotaxis, rotmat
+from Bio.PDB.vectors import (
+    Vector,
+    coord_space,
+    get_spherical_coordinates,
+    homog_trans_mtx,
+    multi_coord_space,
+    vector_to_axis,
+)
 
 
 class VectorTests(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


Mikel-MG:
Hello there, first PR to a open-source project here (apologies in advance for lack of know-how)

I have three questions

1. Should I have opened an issue to ask for feedback on the contribution? (I thought that, being such a small fix, there was no need for it, but I am having second thoughts now)
2. While running the required testing (`python3 setup.py test`) as well as the git commit hooks, I encountered a number of issues (failed tests) in parts of the code that I did not touch. Is this expected, and does it reflect legacy code / pending fixes, or could this be an issue with my setup?
3. For the sake of transparency, I want to disclose that I had to do the commit with the `--no-verify` flag, assuming that the CI workflow would work (it seems it did!). Is this discrepancy between "local" and "remote" testing worth opening an issue?

JoaoRodrigues:
Good catch @Mikel-MG, thank you!

Regarding your questions:
>Should I have opened an issue to ask for feedback on the contribution? (I thought that, being such a small fix, there was no need for it, but I am having second thoughts now)

For such a small and straightforward fix, no need. The PR is enough.

>While running the required testing (python3 setup.py test) as well as the git commit hooks, I encountered a number of issues (failed tests) in parts of the code that I did not touch. Is this expected, and does it reflect legacy code / pending fixes, or could this be an issue with my setup?

This should not be expected really, tests should pass on your machine as well. The tests are passing on all platforms in our testing infrastructure. What might be happening is that some dependencies are missing on your machine, and therefore some tests are skipped? I activated the tests to run here on your PR, so we'll see if your change does indeed affect any of them.

>For the sake of transparency, I want to disclose that I had to do the commit with the --no-verify flag, assuming that the CI workflow would work (it seems it did!). Is this discrepancy between "local" and "remote" testing worth opening an issue?

See my comment above. The tests didn't actually run automatically, I enabled them now. If they fail, have a look at the [Local Development](https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst#local-development) section of our contributors' guide to see how to install pre-commit and run the hooks. Since you committed already, you might have to force the hooks to run on the files you changed, e.g. `pre-commit run --files example.txt example2.txt`

Finally, would you mind adding a unit test to catch this? I find it quite surprising we don't have one already, but it shouldn't be hard to add. You can see examples of unit tests for the vector code [here](https://github.com/biopython/biopython/blob/master/Tests/test_PDB_vectors.py).

Mikel-MG:
Thanks for looking into this

Right, the check that failed (style) also failed on my local testing (which seems to imply that my setup was fine). The CI check throws:
```
Bio/PDB/vectors.py:551:5: F824 `global gtm` is unused: name is never assigned in scope
Bio/PDB/vectors.py:552:5: F824 `global gmry` is unused: name is never assigned in scope
Bio/PDB/vectors.py:553:5: F824 `global gmrz` is unused: name is never assigned in scope
Bio/PDB/vectors.py:553:5: F824 `global gmrz2` is unused: name is never assigned in scope
```

These F824 complaints refer to this part of the code (which, for clarity, is unrelated to my PR)
```python
global gtm
    global gmry
    global gmrz, gmrz2

    tm = gtm
    mry = gmry
    mrz = gmrz
    mrz2 = gmrz2

    # tx acs[1] to origin
    # tm = homog_trans_mtx(-a1[0][0], -a1[1][0], -a1[2][0])
    set_homog_trans_mtx(-a1[0], -a1[1], -a1[2], tm)
```

Where indeed, instead of doing something like `gmry=homog_trans_mtx(...)`, the `homog_trans_mtx()` function (and some similar ones) directly changes the values:

```python
def set_homog_trans_mtx(x: float, y: float, z: float, mtx: np.ndarray):
    """Update existing translation matrix to new values."""
    mtx[0][3] = x
    mtx[1][3] = y
    mtx[2][3] = z
```

Two follow-up points:
* If I understand this correctly, the checks fail due to a part of the code which was already in the repo, which makes me wonder, how did previous commits pass the tests? Does the CI only check for "affected" modules? Will it have to be fixed before my PR can be merged?
* I would be happy to add a unittest for the vector_to_axis function (will get back with a new commit in a few days)

peterjc:
We have CI running _most_ style checks on _all_ the code via https://results.pre-commit.ci/latest/github/biopython/biopython/master

We have CI running _all_ style checks on _changed_ code via GitHub Actions.

We opted not to run all the checks on all the code on each change as it is a bit slow 😞 

I think the F824 issue is due to updating past flake 7.2.0 in #4986, would you mind opening an issue on those failures please?

Bio/PDB/vectors.py
```diff
@@ -93,9 +93,9 @@ def vector_to_axis(line, point):
     :param point: vector defining the point
     """
     line = line.normalized()
-    np = point.norm()
+    norm_point = point.norm()
     angle = line.angle(point)
-    return point - line ** (np * np.cos(angle))
+    return point - line ** (norm_point * np.cos(angle))
```
peterjc, line 98: The fix looks good but suggests this method needs a test.

Mikel-MG:
Hello again,

I added some basic testing, let me know if you think I should add more.

I would like to point out that the pre-commit checks would not let me commit the new tests without reformatting the imported functions with flake8, I hope this is fine.

Bio/PDB/vectors.py
Mikel-MG: Agree, and done!

Mikel-MG:
Nice! I’m already listed in CONTRIB, and this doesn’t feel significant enough for NEWS, so feel free to squash and merge

peterjc:
Thank you!

PR #5047, opened 2025-08-28
fixes formatting for extremely high B-factor values

### LINKED ISSUE #5046: Edge case B-factor writing error
When writing a PDB file with B-factor values above 999, you will end up getting an invalid length B-factor value in the PDB. This is not a common occurence, however, with the rise of people using B-factor fields for other values - typically from computational tools, it is a real problem.

The issue arrises from the format string in `Bio.PDB.PDBIO`:

```python
_ATOM_FORMAT_STRING = (
    "%s%5i %-4s%c%3s %c%4i%c   %8.3f%8.3f%8.3f%s%6.2f      %4s%2s%2s\n"
)
```

Particularly, the `%6.2f` format code for the B-factor. This ensures that the **minimum** string length is 6 characters, not the maximum. For cases where the B-factor is above 999, you will have additional characters.

I have verified this in python3.10.12, and according to PEP3101:
```
[[fill]align][sign][#][0][minimumwidth][.precision][type]
```

Therefore, I propose that when a B-factor value is over 999, we set the precision to 1, and when over 9999 we set precision to 0. For normal cases this will have no effect as it is very rare to have B-factor values in such a range.

OWissett:

Closes #5046

### DIFF
--- a/Bio/PDB/PDBExceptions.py
+++ b/Bio/PDB/PDBExceptions.py
@@ -28,3 +28,7 @@ class PDBConstructionWarning(BiopythonWarning):
 # The SMCRA structure could not be written to file
 class PDBIOException(Exception):
     """Define class PDBIOException."""
+
+
+class PDBIOWarning(BiopythonWarning):
+    """Define class PDBIOWarning."""
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -35,17 +35,15 @@ def _format_b_factor(value: float) -> str:
     """
     _bfactor_str = ""
     if value < _MAX_B_FACTOR_2DP:
-        _rounded = round(value, 2)
-        if len(f"{_rounded:.2f}") > 6:
-            _bfactor_str = f"{_rounded:6.1f}"
+        if len(f"{value:.2f}") > 6:
+            _bfactor_str = f"{value:6.1f}"
         else:
-            _bfactor_str = f"{_rounded:6.2f}"
+            _bfactor_str = f"{value:6.2f}"
     elif value < _MAX_B_FACTOR_1DP:
-        _rounded = round(value, 1)
-        if len(f"{_rounded:.1f}") > 6:
-            _bfactor_str = f"{int(_rounded):6d}"
+        if len(f"{value:.1f}") > 6:
+            _bfactor_str = f"{value:6.0f}"
         else:
-            _bfactor_str = f"{_rounded:6.1f}"
+            _bfactor_str = f"{value:6.1f}"
     elif value < _MAX_B_FACTOR:
         _bfactor_str = f"{int(value):6d}"
     else:
--- a/Bio/PDB/PDBIO.py
+++ b/Bio/PDB/PDBIO.py
@@ -8,9 +8,8 @@
 import os
 import warnings
 
-from Bio import BiopythonWarning
 from Bio.Data.IUPACData import atom_weights
-from Bio.PDB.PDBExceptions import PDBIOException
+from Bio.PDB.PDBExceptions import PDBIOException, PDBIOWarning
 from Bio.PDB.StructureBuilder import StructureBuilder
 
 _ATOM_FORMAT_STRING = "%s%5i %-4s%c%3s %c%4i%c   %8.3f%8.3f%8.3f%s%s     %4s%2s%2s\n"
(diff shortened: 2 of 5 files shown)


Bio/PDB/PDBIO.py
```diff
@@ -239,7 +242,19 @@ def _get_atom_line(
...
                 charge,
             )
-            return _ATOM_FORMAT_STRING % args
+
+            _format_string = (
+                _ATOM_FORMAT_STRING
+                if bfactor < 999
+                else _ATOM_FORMAT_STRING_HIGH_B_FACTOR
+            )
+            if bfactor >= 9999:
```
JoaoRodrigues: Quite honestly, instead of throwing a warning and still writing a bad file, I'm more inclined to throwing a warning and truncating the value to a maximum. You can also, instead of having two different atom format strings, which can be annoying to maintain in the future, replace the b-factor `%6.2f` by `%6s` and then make the b-factor conversion when you write it.

Something like:

```python

MAX_B_FACTOR = 9999.9

def format_b_factor(value: float) -> str:
    """Return the b-factor value as a formatted string.
    
    Formats the b-factor value to fit the wwPDB spec of maximum
    6 characters, truncating the value if necessary.
    """
    if value < 1000:  # default
        return f"{value:6.2f}"
    elif value < 10_000:  # modify decimal places
        return f"{value:6.1f}"
    else:  # truncate
        warnings.warn(
            f"Truncated b-factor value {value!r} to {MAX_B_FACTOR} to fit the PDB spec. "
            "Consider writing mmCIF files instead.")
        return str(MAX_B_FACTOR)
```

And use this function around line 212.

OWissett:
Will do on Monday. I'm away from civilisation this weekend. I'll be up a mountain.

JoaoRodrigues:
> Will do on Monday. I'm away from civilisation this weekend. I'll be up a mountain.

Thanks and enjoy!

OWissett:
I have added unit tests, and restructured how the formatting works a little bit. Since technically the maximum value for a bfactor can be 999_999 (if we have no decimals) we can now handle those without emitting a warning.

I also have accounted for the case where rounding increases the number of digits.

Finally, I added a new warning class `PDBIOWarning`, as I feel it makes sense to have a dedicated warning for saving PDBs compared to other PDB warnings, in the same way we have a `PDBContructionWarning`

Bio/PDB/PDBExceptions.py
```diff
@@ -28,3 +28,7 @@ class PDBConstructionWarning(BiopythonWarning):
 # The SMCRA structure could not be written to file
 class PDBIOException(Exception):
     """Define class PDBIOException."""
+
+
+class PDBIOWarning(BiopythonWarning):
```
JoaoRodrigues, line 33: Nice, thanks!

Bio/PDB/PDBIO.py
```diff
@@ -24,6 +22,44 @@
...
+
+def _format_b_factor(value: float) -> str:
+    """Returns the b-factor value as a formatted string
+
+    Formats the b-factor value to fit the wwPDB specification of 6 characters
+    maximum, otherwise truncating to 6 characters if necessary.
+    """
+    _bfactor_str = ""
+    if value < _MAX_B_FACTOR_2DP:
+        _rounded = round(value, 2)
```
JoaoRodrigues: I'm not sure the rounding here is necessary since you're effectively doing it with the string formatting below.

Tests/test_PDB_PDBIO.py
```diff
@@ -26,6 +26,7 @@
 from Bio.PDB import Select
 from Bio.PDB.PDBExceptions import PDBConstructionWarning
 from Bio.PDB.PDBExceptions import PDBIOException
+from Bio.PDB.PDBIO import _MAX_B_FACTOR, PDBIOWarning
```
JoaoRodrigues: I'd import the warning on it's own, like above with `PDBIOException` (and from the Exceptions module)

Tests/test_PDB_PDBIO.py
```diff
@@ -381,6 +382,42 @@ def test_pdbio_write_unk_element(self):
...
+                    self.io.save(filename)
+            else:
+                self.io.save(filename)
+
+            try:
+                check_pdb(filename, expected_bfactor)
+            finally:
+                os.remove(filename)
+
+        test_b_factor(99.999999, 100)
```
JoaoRodrigues, line 412: If these fail, is it easy to figure out which sub-test failed if they're written like this? Would it be easier to write this as separate tests?

JoaoRodrigues (approved):
Minor comments, but looks good to me. If you can, add a small snippet to `NEWS.rst` explaining this change, as this can surprise some people.

Tests/test_PDB_PDBIO.py
OWissett: From when I was running the tests, if they fail it does show the failed assertion values, so you can see which case it is. I don't think it is too hard to follow, but I can make them different tests if you would prefer

Bio/PDB/PDBIO.py
OWissett: Done.

JoaoRodrigues:
Thanks a lot @OWissett, super nice contribution. I've enabled the full CI workflow to run on these changes and once the tests/checks pass, happy to merge.

JoaoRodrigues:
Merged, thanks again @OWissett !

PR #5070, opened 2025-09-19
Add support for terminal_columns in Alignment formatting, issue 5035

### LINKED ISSUE #5035: Bio.Align.PairwiseAligner missing format_alignment in Bio.pairwise2
The alignment object from Bio.Align.PairwiseAligner now only displays 60 residues per row, which is inconvenient for longer proteins. The 60-char limit came from old hardware. The format_alignment in Bio.pairwise2 can use the terminal space better, could we adopt that same logic, when we deprecate Bio.pairwise2.

Thank you for your consideration.

data2code:

Closes #5035 

Modify Bio.Align.__init__.py to support terminal_columns. If not specified, use all available terminal columns to show the alignment, otherwise, use user-specified value. Add Tests/test_Align_Alignment_format.py.

### DIFF
--- a/Tests/test_Align_Alignment.py
+++ b/Tests/test_Align_Alignment.py
@@ -26,6 +26,8 @@ from Bio.Seq import translate
 from Bio.SeqRecord import SeqRecord
 from Bio.SeqUtils import gc_fraction
 
+Align.Alignment.terminal_columns = 80
+
 
 class TestAlignment(unittest.TestCase):
     def test_empty_alignment(self):
@@ -3706,6 +3708,8 @@ class TestAlignmentFormat(unittest.TestCase):
         # query             0 DVQLQESGPGLVKP--SQSQSLTCTVTGYSITSDYAWNWIRQFP--GNKLEWMGYMS-------Y--------------SGSTRY---NPSLRSRISITRDTSKNQFFLQLKSVTTEDTATYFCARG-W 100
         #
         self.assertEqual(len(str(best_alignment).split("\n")), 4)
+
+        Align.Alignment.terminal_columns = 80
 
 
 if __name__ == "__main__":


mdehoon:
@data2code  Can you have a look at the failing checks?

data2code:
I don't know what's wrong. I ran "pytest Tests/test_Align_Alignment_format.py" and it was clean.

mdehoon:
@data2code  Have a look at the failing checks as shown on this web page.
All tests must pass, not just the new test in `test_Align_Alignment_format.py`.

data2code:
I believe I have fixed the style issue in the test script. The error currently is in the main code.
https://ci.appveyor.com/project/biopython/biopython/builds/52789718

I don't feel comfortable to add this new attribute to the Alignment class, as I am unfamiliar with the BioPython code base.
I am not sure I am the right person to create this PR.

peterjc:
The errors look like a problem not creating an attribute when the object is created? eg

```
File "C:\Py\Lib\site-packages\Bio\Align\__init__.py", line 2305, in _format_pretty
    if self.terminal_columns is None:
       ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Alignment' object has no attribute 'terminal_columns'
```

Black is still not happy - this could be the version of black:

```
black....................................................................Failed
- hook id: black
- exit code: 1

would reformat Tests/test_Align_Alignment_format.py
```

We still use black 24.10.0 for trailing whitespace in doctests.

mdehoon:
> The errors look like a problem not creating an attribute when the object is created? eg

> File "C:\Py\Lib\site-packages\Bio\Align\__init__.py", line 2305, in _format_pretty
    > if self.terminal_columns is None:
       > ^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'Alignment' object has no attribute 'terminal_columns'

The `Alignment` class needs a class variable `terminal_columns` that is `None` by default.

data2code:
Not sure what I can do here. The Pre-commit.ci gives failures without providing me the accurate feedback.

mdehoon:
@data2code  Check the failing tests as shown on this web page, where it says "4 failing checks"

data2code:
@mdehoon I only see failure flags without details. I am not a biopython developer; instead, I am a user who made a feature request. I am afraid I have done what I can with the issue. Could someone from the developer team help take over? Thanks.

mdehoon:
@data2code  
>Could someone from the developer team help take over?

There is no developer team. We are all just volunteers.

What happens if you click on the first failure (where it says "CI / style (pull_request)") ?

data2code:
This is all I see. An error without useful details.

peterjc:
Two files are not following the automated style checking using the tool `black`, which is part of the pre-commit checks we expect you te run locally. See https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

peterjc:
The main tests are looking better though, from AppVeyor which runs the tests on Windows:

```
ERROR: test_Align_Alignment_format
  File "C:\projects\biopython\Tests\run_tests.py", line 266, in runTest
    raise RuntimeError(f"No tests found in {name}")
RuntimeError: No tests found in test_Align_Alignment_format
Ran 501 tests in 576.243 seconds
```

Bio/Align/__init__.py
```diff
@@ -20,6 +20,7 @@
 import sys
 import types
 import warnings
+import shutil
```
peterjc: This would ideally be a few lines up (alphabetical sorting of standard library imports).

Bio/Align/__init__.py
```diff
@@ -2341,7 +2342,13 @@ def _format_pretty(self, matrix=None):
                 raise ValueError("Inconsistent coordinates")
         prefix_width = 10
         position_width = 10
-        line_width = 80
+
+        # support custom column size, see issue 5035
```
peterjc: I think the comment is redundant and could be removed.

Tests/test_Align_Alignment_format.py
```diff
@@ -0,0 +1,40 @@
+"""Tests for controling the number of terminal columns
+   Issue #5035
+"""
```
peterjc: This is not formatted in the expected style, but there are bigger issues with your test.

Tests/test_Align_Alignment_format.py
```diff
@@ -0,0 +1,40 @@
...
+# query            92 TYFCARG-W 100
+# 
+assert (len(str(best_alignment).split("\n")) == 12), "Expecting alignment to be output as two groups."
+
+best_alignment.terminal_columns = 200
+# target            0 -VQLQESDAELVKPGASVKIS--CKASGYTFT-DHVIHWVKQKPEQG--LEWIGYISPGNGDIKYNEKFKGKATLTADKSSSTAYMQLN-SL---------TS-------------EDSAVYLCKRGYY 100
+#                   0 -||||||...||||--|...|--|...||..|-|....|..|.|--|--|||.||.|-------|--------------|.||.|---|-||---------||-------------||.|.|.|.||-. 129
+# query             0 DVQLQESGPGLVKP--SQSQSLTCTVTGYSITSDYAWNWIRQFP--GNKLEWMGYMS-------Y--------------SGSTRY---NPSLRSRISITRDTSKNQFFLQLKSVTTEDTATYFCARG-W 100
+# 
+assert (len(str(best_alignment).split("\n")) == 4), "Expecting alignment to be output as one group."
```
peterjc: You have not defined any tests using the unittest framework, thus the error `No tests found in test_Align_Alignment_format`.

See https://biopython.org/docs/latest/Tutorial/chapter_testing.html

Also I think this need not be an entire new file.

peterjc (changes requested):

data2code:
@peterjc  Thank you, that's very helpful.

data2code:
@peterjc The test looks good on my local:

$ python Tests/test_Align_Alignment_format.py
test_terminal_columns (__main__.TestAlignmentFormat.test_terminal_columns) ... ok

Ran 1 test in 0.001s

OK

Any pointers for what to work on next? Thank you!

peterjc:
Fix the black style. Move the test into an appropriate existing `Bio.Align` test file. At that point we can ask GitHub Actions to run the full test suite (and see if the Circle CI apparent timeout is related to your changes or not).

[I have not considered the value of the change itself yet, as others seem to have.]

data2code:
I believe the test error was not due to the code I added to the test file. The error was triggered by other tests in the same file.

Bio/Align/__init__.py
```diff
@@ -2341,7 +2342,12 @@ def _format_pretty(self, matrix=None):
                 raise ValueError("Inconsistent coordinates")
         prefix_width = 10
         position_width = 10
-        line_width = 80
+
+        if not hasattr(self, "terminal_columns") or self.terminal_columns is None:
```
mdehoon: It is better to have a class variable `terminal_columns` on the `Alignment` class.

data2code:
@mdehoon Updated. Thanks for the suggestion.

mdehoon:
@data2code  Have a look at https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables

data2code:
Okay, move it to a class attribute. Thanks.

Tests/test_Align_Alignment.py
```diff
@@ -3667,6 +3667,47 @@ def test_mapall(self):
...
+        #
+        # target           54 GDIKYNEKFKGKATLTADKSSSTAYMQLN-SL---------TS-------------EDSA
+        #                  60 ----|--------------|.||.|---|-||---------||-------------||.|
+        # query            53 ----Y--------------SGSTRY---NPSLRSRISITRDTSKNQFFLQLKSVTTEDTA
+        #
+        # target           91 VYLCKRGYY 100
+        #                 120 .|.|.||-. 129
+        # query            92 TYFCARG-W 100
+        #
+        self.assertEqual(len(str(best_alignment).split("\n")), 12)
```
mdehoon, line 3703: Why not just test `str(best_alignment)` explicitly (using triple quotes)?

data2code: IMHO, this test should not be checking the exact alignment itself, but instead focus on the row count. The exact alginment might be altered in the future due small changes in alignment algorithm or the naming of the "target", "query" or other small formatting changes. The exact alignment should be checked by other test cases, while this test focuses on row counts. I hope it makes sense.

mdehoon: If at some point some bug is introduced that causes the printed alignment to be incorrect, then your test may not capture it, if it only checks the number of lines.

The algorithm by definition generates the optimal alignment. Even if there are any changes to the algorithm, the generated optimal alignment will still be the same.

data2code:
@mdehoon You have a valid point in believing the algorithm won't change, however, the formatting itself can change. E.g., someone may decided instead of using query and target, print Q and T instead. Then this test will break. Anyway, this test is to check the alignment layout, not the exact content of the alignment itself. There are other tests that change the validity of the content of the alignment. I think it's best for this test to focus on the formatting not the content. If it is okay with everyone, we should close this thread. I have learned a lot from this PR, thank you and Peter for your patience! Really appreciate all the guidance and feedback.

PR #5074, opened 2025-09-25
Handle possible non-existent attributes in clusterbuster.py

chenyenchung:

The commit bc0d58b4c4 added support for writing gap and weight attributes
to ClusterBuster format, but didn't account for motifs from other formats
(like JASPAR) that don't have these attributes. This caused AttributeError
when trying to convert JASPAR motifs to ClusterBuster format.

To trigger this bug, one can load a JASPAR-formatted motif (`motif.jaspar`):

```
>MA0030.2	FOXF2
A  [     3      7      0     27     27     27      0     27     16 ]
C  [    11      0      0      0      0      0     25      0      4 ]
G  [     8     20      0      0      0      0      0      0      2 ]
```

With version 1.85, the following script will result in `AttributeError: 'Motif' object has no attribute 'weight'`:

```
from Bio import motifs

with open('motif.jaspar', 'rt') as f:
  m = [motifs.read(f, 'jaspar')]

motifs.write(m, 'clusterbuster') // Fail
```

This fix adds hasattr() checks before accessing these optional attributes.

### DIFF
--- a/Bio/motifs/clusterbuster.py
+++ b/Bio/motifs/clusterbuster.py
@@ -93,13 +93,15 @@ def write(motifs, precision=0):
         except AttributeError:
             pass
         else:
-            lines.append(f"# WEIGHT: {weight}\n")
+            if weight:
+                lines.append(f"# WEIGHT: {weight}\n")
         try:
             gap = m.gap
         except AttributeError:
             pass
         else:
-            lines.append(f"# GAP: {gap}\n")
+            if gap:
+                lines.append(f"# GAP: {gap}\n")
         for ACGT_counts in zip(
             m.counts["A"], m.counts["C"], m.counts["G"], m.counts["T"]
         ):
--- a/Bio/motifs/clusterbuster.py
+++ b/Bio/motifs/clusterbuster.py
@@ -88,9 +88,9 @@ def write(motifs, precision=0):
     lines = []
     for m in motifs:
         lines.append(f">{m.name}\n")
-        if m.weight:
+        if hasattr(m, 'weight') and m.weight:
             lines.append(f"# WEIGHT: {m.weight}\n")
-        if m.gap:
+        if hasattr(m, 'gap') and m.gap:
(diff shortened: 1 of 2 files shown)


Bio/motifs/clusterbuster.py
```diff
@@ -88,9 +88,9 @@ def write(motifs, precision=0):
     lines = []
     for m in motifs:
         lines.append(f">{m.name}\n")
-        if m.weight:
+        if hasattr(m, 'weight') and m.weight:
             lines.append(f"# WEIGHT: {m.weight}\n")
```
mdehoon: Another way to do this is
```python
try:
    weight = m.weight
except AttributeError:
    pass
else:
    if weight:
        lines.append(f"# WEIGHT: {m.weight}\n")
```
which is consistent with Python's principle that It is Easier to Ask for Forgiveness than Permission.

I am not really familiar with clusterbuster, and I am wondering in which situation the `weight` attribute exists but `if m.weight` fails.

chenyenchung: Thank you for the timely review! I have revised the PR to use try-except to handle this. Regarding possible failure modes for the weight and gap attributes. Currently it is only set in the motifs module in `clusterbuster.py`, so any other format will miss them, while when a clusterbuster motif is loaded, missing weight and gap will result in the attribute to be exist as `None`, so I don't think there's currently a path for `if m.weight` to fail when it exists unless the user modify it.

mdehoon: OK, then please have a look at the failing tests. Once those are fixed, we can merge this.

chenyenchung: Got it! The error was caused by when I switched to try-except, I broke the handling when the attributes are `None`. In the current commit (f819dcc), I added `if m.weight:` and `if m.gap:` back to the export block, and now the offline tests passed locally. Thank you for your patience.

mdehoon: Thanks! Done.

PR #5079, opened 2025-10-15
Permit mixed case for genbank export masking

znruss:

Previously all genbank export was lowercase - it's not clear that the genbank spec actually requires this. This allows faithful case export.


peterjc:
The `ValueError` is good, but not the main change which I regard as breaking the defacto standard (even if not exlicit in the documentation), and a significant backward compatibility break which will break round-tripping (load and save).

PR #5057, opened 2025-09-05
added type annotations to most common oublic methods for Bio/Align, A…

dimi1729:
…lignIO, Graphics, HMM, KEGG, and PDB

- [X] I hereby agree to dual licence this and any previous contributions under both
the _Biopython License Agreement_ **AND** the _BSD 3-Clause License_.

- [X] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``
locally, and understand that continuous integration checks will be used to
confirm the Biopython unit tests and style checks pass with these changes.

- [X] I have added my name to the alphabetical contributors listings in the files
``NEWS.rst`` and ``CONTRIB.rst`` as part of this pull request, am listed
already, or do not wish to be listed. (*This acknowledgement is optional.*)

Addresses #2236 

Added type annotations to most getter functions in Align, AlignIO, Graphics, HMM, KEGG, and PDB modules. While this is not a completed effort, it will hopefully be useful to people who use these modules and like type hints. Since no added functionality was added and very little actual logic was changed, no additional tests were created, but all existing tests were ran.

PR #5087, opened 2025-10-24
Add type annotations

szabgab:

Addresses #2236   and does not touch files that are addressed in #5057

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -32,6 +32,7 @@ from typing import Union
 from Bio import BiopythonWarning
 from Bio.Data import CodonTable
 from Bio.Data import IUPACData
+from Bio.SeqRecord import SeqRecord
 
 
 def _maketrans(complement_mapping: dict[str, str]) -> bytes:
@@ -104,14 +105,14 @@ class SequenceDataAbstractBaseClass(ABC):
         assert self[:0] == b""
 
     @abstractmethod
-    def __len__(self):
+    def __len__(self) -> int:
         pass
 
     @abstractmethod
     def __getitem__(self, key):
         pass
 
-    def __bytes__(self):
+    def __bytes__(self) -> bytes:
         return self[:]
 
     def __hash__(self):
@@ -2734,7 +2735,7 @@ def back_transcribe(rna: Union[str, Seq, MutableSeq]) -> Union[str, Seq]:
 
 
 def _translate_str(
-    sequence: Seq,
+    sequence: str,
     table: Union[str, int, CodonTable.CodonTable],
     stop_symbol: str = "*",
     to_stop: bool = False,
@@ -2920,13 +2921,13 @@ def _translate_str(
 
 
 def translate(
-    sequence: Seq,
+    sequence: Union[str, Seq, MutableSeq],
     table: Union[str, int, CodonTable.CodonTable] = "Standard",
     stop_symbol: str = "*",
     to_stop: bool = False,
     cds: bool = False,
     gap: Optional[str] = None,
-) -> str:
+) -> Union[str, Seq]:
     """Translate a nucleotide sequence into amino acids.
 
     If given a string, returns a new string object. Given a Seq or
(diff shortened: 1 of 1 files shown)


Bio/Seq.py
```diff
@@ -2799,7 +2805,8 @@ def _translate_str(
     Bio.Data.CodonTable.TranslationError: Extra in frame stop codon 'TAG' found.
     """
     try:
-        table_id = int(table)
+        # TODO: only call int if table is of type int
```
peterjc, line 2809: This try/except is the check for it being an int, or int like.

Bio/Seq.py
```diff
@@ -3016,7 +3028,7 @@ def translate(
         return _translate_str(sequence, table, stop_symbol, to_stop, cds, gap=gap)
 
 
-def reverse_complement(sequence, inplace=False):
+def reverse_complement(sequence: Seq, inplace: bool = False) -> str:
```
peterjc: Surely `Union[str, Seq, MutableSeq, SeqRecord]` as per the docstring?

Bio/Seq.py
```diff
@@ -3081,7 +3093,7 @@ def reverse_complement(sequence, inplace=False):
     return sequence[::-1]
 
 
-def reverse_complement_rna(sequence, inplace=False):
+def reverse_complement_rna(sequence: Seq, inplace: bool = False) -> str:
```
peterjc: Surely `Union[str, Seq, MutableSeq, SeqRecord]` as per the docstring?

Bio/Seq.py
```diff
@@ -3146,7 +3158,7 @@ def reverse_complement_rna(sequence, inplace=False):
     return sequence[::-1]
 
 
-def complement(sequence, inplace=False):
+def complement(sequence: Seq, inplace: bool = False) -> str:
```
peterjc: Surely `Union[str, Seq, MutableSeq, SeqRecord]` as per the docstring?

Bio/Seq.py
```diff
@@ -3210,7 +3222,7 @@ def complement(sequence, inplace=False):
     return sequence.decode("ASCII")
 
 
-def complement_rna(sequence, inplace=False):
+def complement_rna(sequence: Seq, inplace: bool = False) -> str:
```
peterjc: Surely `Union[str, Seq, MutableSeq, SeqRecord]` as per the docstring?

Bio/Seq.py
```diff
@@ -2913,8 +2920,13 @@ def _translate_str(
 
 
 def translate(
-    sequence, table="Standard", stop_symbol="*", to_stop=False, cds=False, gap=None
-):
+    sequence: Seq,
```
peterjc: Surely `Union[str, Seq, MutableSeq]` as per the docstring?

Bio/Seq.py
```diff
@@ -2734,8 +2734,14 @@ def back_transcribe(rna):
 
 
 def _translate_str(
-    sequence, table, stop_symbol="*", to_stop=False, cds=False, pos_stop="X", gap=None
-):
+    sequence: Seq,
```
peterjc: No, as per the docstring, this should be a string.

peterjc (changes requested):
Lots to fix here - see comments.

Bio/Seq.py
szabgab: Wouldn't it be better to change this to `isinstance(table, int)`  and remove the try/except ?

szabgab:
Thanks for the feedback. I hope this one is going to be a step forward.

Bio/Seq.py
peterjc: It could be something else int-like but not a subclass. I would be cautious about changing this for fear of unintended side effects.

Bio/Seq.py
```diff
@@ -32,9 +32,10 @@
 from Bio import BiopythonWarning
 from Bio.Data import CodonTable
 from Bio.Data import IUPACData
+from Bio.SeqRecord import SeqRecord
```
peterjc, line 35: You'll have to remove that, or put it an if-typing conditional or something. It is making a circular import (see the CI failures).

PR #5099, opened 2025-11-07
Update ScanProsite to handle the new XML format from the ScanProsite API

### LINKED ISSUE #5073: ScanProsite.scan() just returns 2 matches while the web return 1010
```python
from Bio.ExPASy import ScanProsite

pattern = "[KLR]-[EKA]-M-x(2)-Q-[YF]-G-[YF]-[LV]-x-[ML]-F-[SGV]-x(2)-[WYF]-P-L-[AV]-[PA]-x(3)-[LV]-[ILV]-[VN]-N-x(2)-[EV]-x-[QRN]-x-D-x(2)-K"
url = "https://prosite.expasy.org/"

result=ScanProsite.scan(sig=pattern, mirror=url, output='xml', db='uniprot')

sequences=ScanProsite.read(result)
print(sequences) 
```

im running this code but i just get 2 matches. can someone point out the problem? 

Also, is uniprot the broadest db option available?

thank you

timjden:

Closes #5073 

The issue was resolved by the ScanProsite API team at the Swiss Institute of Bioinformatics (SIB), and I have included my correspondence with SIB in the issue comments.

While testing, it became clear that the XML response from the ScanProsite API has changed and breaks BioPython's parsing of the response. These changes are a quick fix for that.

Additional considerations for the future may include:

1) Better handling of the response from ScanProsite to inform the user if an error is in the API response, or in BioPython code itself
2) Use json as the default output parameter in `ScanProsite.scan()` because I suspect the structure of the response from the ScanProsite API is less likely to change in this format
3) And, if 3, then refactor the parsing of the ScanProsite API response to handle json
4) Warn the user if they have specified unsupported keyword argument values (e.g. in the case of #5073, the user specified "uniprot" as the db value which is not supported by the ScanProsite API and results in defaulting to the Swiss-Prot ("sp") database only.

@peterjc , perhaps you can let me know if you think I should take a stab at any/all of the above.

### Testing

Ran script:

```
from Bio.ExPASy import ScanProsite

pattern = "P-x(2)-G-E-S-G(2)-[AS]"
url = "https://prosite.expasy.org/"

sp_result = ScanProsite.scan(sig=pattern, mirror=url, output='xml', db='sp')
tr_result = ScanProsite.scan(sig=pattern, mirror=url, output='xml', db='tr')

sequences = ScanProsite.read(sp_result) + ScanProsite.read(tr_result)
print(len(sequences), "sequences found")
```

Output:

```
1089 sequences found
```

UI Output:

image

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -13,7 +13,12 @@ The latest news is at the top of this file.
 (In progress, not yet released): Biopython 1.87
 ===============================================
 
-Pending.
+Many thanks to the Biopython developers and community for making this release
+possible, especially the following contributors:
+
+- Michiel de Hoon
+- Peter Cock
+- Timothy Dennis (first contribution)
 
 28 October 2025: Biopython 1.86
 ===============================
@@ -65,14 +70,13 @@ trees using complex style options.
 Many thanks to the Biopython developers and community for making this release
 possible, especially the following contributors:
 
+- Cassie Bastress (first contribution)
 - Rachel Stern (first contribution)
 - Fabio Zanini
 - Michiel de Hoon
 - Oliver Wissett (first contribution)
 - Peter Cock
 - Samuel Prince (first contribution)
-- Cassie Bastress (first contribution)
-- Timothy Dennis (first contribution)
 
 15 January 2025: Biopython 1.85
 ===============================


Bio/ExPASy/ScanProsite.py
```diff
@@ -94,7 +94,7 @@ def feed(self, data, isFinal=0):
         # The error message is (hopefully) contained in the data that was just
         # fed to the parser.
         if self.firsttime:
-            if data[:5].decode("utf-8") != "<?xml":
+            if data[:22].decode("utf-8") != "<scanprosite_response>":
```
peterjc, line 101: Maybe if neither is true, treat as an error?

timjden: My thinking is that it's worth raising an error here if `<scanprosite_response>` isn't the first element because the code which parses the XML expects that and will break if the first element is `<?xml...>`.

timjden:
@peterjc thanks, I've added a few tests in `test_ExPASy.py`. They are passing but they do require internet so won't be triggered in the CI if I understand correctly.

image

timjden:
@peterjc I've added some offline tests and all checks in the CI pass now. Let me know if you're happy to merge when you have a moment. Thanks!

peterjc:
Looks good to me - no name in NEWS & CONTRIB?

Are you OK with a squash-and-merge to clean-up the history a little, or would you like to do a rebase yourself?

timjden:
Thanks @peterjc. I've added my name to NEWS and CONTRIB. Happy for you to squash and merge.

timjden:
@peterjc thanks for merging. Seems like the build got stuck though?

peterjc:
Looks like there was a left over `[ci skip]` in the squash-merge commit message? Forced it instead... will double check the main branch later.

PR #5106, opened 2025-11-21
[security] Make Entrez parser enforce TLS/SSL for downloads to avoid man-in-the-middle (MITM) attacks

hartwork:
- [x] I hereby agree to dual licence this under [both the _Biopython License Agreement_ **AND** the _BSD 3-Clause License_](https://github.com/biopython/biopython/blob/master/LICENSE.rst).

Addresses man-in-the-middle (MITM) issues with DTD and XSD downloads in class `Entrez.Parser.DataHandler` and its users.

CC @mdehoon

### DIFF
--- a/Tests/test_Entrez_parser.py
+++ b/Tests/test_Entrez_parser.py
@@ -9,6 +9,9 @@ import os
 import pickle
 import unittest
 from io import BytesIO
+from textwrap import dedent
+from unittest.mock import Mock, patch
+from urllib.parse import urlparse
 
 from Bio import Entrez
 from Bio import StreamModeError
@@ -12345,6 +12348,49 @@ We designed and generated pulmonary imaging biomarker pipelines to facilitate hi
         stream.seek(0)
         records = Entrez.parse(stream)
         self.assertRaises(CorruptedXMLError, next, records)
+
+
+class TlsEnforcementTest(unittest.TestCase):
+    """Test for DTD downloads enforcing TLS/SSL."""
+
+    def test_tls_enforced(self):
+        content = (
+            dedent(
+                """\
+                <?xml version="1.0"?>
+                <!DOCTYPE eInfoResult PUBLIC "-//NLM//DTD eInfoResult, 11 May 2002//EN"
+                    "https://www.ncbi.nlm.nih.gov/entrez/query/DTD/eInfo_020511.dtd">
+                <eInfoResult />
+                """
+            )
+            .replace("https", "http")
+            .encode("utf-8")
+        )
+
+        self.assertIn(b"http:", content)  # self-test, without TLS
+
+        handler = Entrez.Parser.DataHandler(
+            validate=True, escape=False, ignore_errors=False
+        )
+
+        handler.open_dtd_file = Mock(return_value=None)  # i.e. bypass cache read
+        handler.save_dtd_file = Mock()  # i.e. bypass cache write
+
+        mock_urlopen = Mock(
+            return_value=Mock(
+                read=Mock(return_value=b"<!ELEMENT eInfoResult (DbList|DbInfo|ERROR)>")
+            )
+        )
+
+        with patch("Bio.Entrez.Parser.urlopen", mock_urlopen):
+            handler.read(BytesIO(content))
+
+        expected_schemes = ["https"]  # i.e. now with TLS
+        actual_schemes = [
+            urlparse(call_.args[0]).scheme for call_ in mock_urlopen.call_args_list
+        ]
+
+        self.assertEqual(actual_schemes, expected_schemes)
 
 
 if __name__ == "__main__":


mdehoon:
This PR changes http->https on the fly.
URLs with http can arise for three reasons:
1) Because we hard-coded http in Bio/Entrez/Parser.py
2) Because the data returned by NCBI contain URLs with http
3) Because our tests contain URLs with http, if NCBI was using that in the past.

For case 1), we can just update Bio/Entrez/Parser.py.
For case 3), we can update the old tests using the data currently returned by NCBI, which hopefully uses https
For case 2), it is NCBI's responsibility to update their code.

hartwork:
> This PR changes http->https on the fly.

@mdehoon correct :smiley: 

> URLs with http can arise for three reasons:
> 
> 1. Because we hard-coded http in Bio/Entrez/Parser.py
> 
> 2. Because the data returned by NCBI contain URLs with http
> 
> 3. Because our tests contain URLs with http, if NCBI was using that in the past.
> 
> 
> For case 1), we can just update Bio/Entrez/Parser.py. For case 3), we can update the old tests using the data currently returned by NCBI, which hopefully uses https For case 2), it is NCBI's responsibility to update their code.

That picture is not complete, and BioPython user security is the responsibility of BioPython, whether NCBI adjusts their DTD files or not (even if that would be nice but as a bonus). 

If the pull request is not ready to merge, please let me know what you want me to adjust or how to move forward here precisely. Please give me something actionable. Thank you! :handshake:

mdehoon:
First step is to replace any hard-coded http in Bio/Entrez/Parser.py to https.
Next step is to raise an Exception in the `with_tls` if it finds that any URL starts with http (or ftp).
Then, run all the tests, including the online tests.
If all tests pass, then you can remove `with_tls`, and you are done.
If some tests don't pass, it is probably because the test XML data are old.
For the failing tests, use Bio.Entrez to perform the same query as was used to generate the test XML data. In most cases, the test query is written explicitly in a comment to each test.
Then run the tests again. Some may fail, because the test XML data in 2025 are different. Update those tests accordingly.
Probably, at that point, if you run the tests,`with_tls` will not detect any more URLs starting with http or ftp.
Then you can remove `with_tls`, and you are done.

hartwork:
> First step is to replace any hard-coded http in Bio/Entrez/Parser.py to https.

@mdehoon done. (Notable exceptions are use in XML namespaces where to my understanding strings need to match the original 1:1.)

> Next step is to raise an Exception in the `with_tls` if it finds that any URL starts with http (or ftp). Then, run all the tests, including the online tests. If all tests pass, then you can remove `with_tls`, and you are done. If some tests don't pass, it is probably because the test XML data are old. For the failing tests, use Bio.Entrez to perform the same query as was used to generate the test XML data. In most cases, the test query is written explicitly in a comment to each test. Then run the tests again. Some may fail, because the test XML data in 2025 are different. Update those tests accordingly.

Done.

> Probably, at that point, if you run the tests,`with_tls` will not detect any more URLs starting with http or ftp. Then you can remove `with_tls`, and you are done.

I don't think we can drop it because so far XML user input can contain arbitrary URLs that use non-TLS `http://` and are not confined to domain `www.ncbi.nlm.nih.gov`, and those would be vulnerable to MITM also and be allowed to pass. However arbitrary URL support is vulnerable to a form of [XXE](https://en.wikipedia.org/wiki/XML_external_entity_attack) (with all sorts of consequences…) and hence access to arbitrary URLs will need to be contained some way after this pull request anyway, in addition to protection against MITM. (That's issue (4) in my earlier e-mail to you.) So dropping does not seem like an option to me, at least not yet. Dropping the call to `urlopen` altogether maybe be an alternative to consider.

How would you like to continue with this pull request?

Tests/test_bgzf.py
```diff
@@ -518,5 +518,6 @@ def test_writer_passes_on_plain_file_handle(self):
 
 
 if __name__ == "__main__":
+    os.chdir(os.path.dirname(__file__))
```
peterjc: This is a documented expectation of the test framework - and if you 'fix' it here for `python Tests/test_bgzf.py` you risk make it harder to work out why the tests might fail via the the framework.

This is also irrevant to the HTTPS issue.

Bio/Blast/_parser.py
```diff
@@ -1152,7 +1152,7 @@ def _externalEntityRefHandler(self, context, base, systemId, publicId):
         assert base is None
         if systemId not in (
             "NCBI_BlastOutput.dtd",
-            "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd",
+            "https://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd",
```
mdehoon: What if systemID is "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd"?

Bio/Entrez/DTDs/INSD_INSDSeq.mod.dtd
```diff
@@ -35,7 +35,7 @@ $Revision: 192674 $
     URLs for the DDBJ, EMBL, and GenBank Feature Table Document:
     http://www.ddbj.nig.ac.jp/FT/full_index.html
     http://www.ebi.ac.uk/embl/Documentation/FT_definitions/feature_table.html
-    http://www.ncbi.nlm.nih.gov/projects/collab/FT/index.html
+    https://www.ncbi.nlm.nih.gov/projects/collab/FT/index.html
```
mdehoon: The DTDs should match exactly with the version from NCBI.

Bio/Phylo/_cdao_owl.py
```diff
@@ -1351,7 +1351,7 @@ def resolve_uri(s, namespaces=cdao_namespaces, cdao_to_obo=True, xml_style=False
             </owl:Restriction>
         </rdfs:subClassOf>
         <dc:description>A positional coordinate giving the source of a character state, used for molecular sequences.</dc:description>
-        <rdfs:comment>drawing from seqloc categories from NCBI at http://www.ncbi.nlm.nih.gov/IEB/ToolBox/SDKDOCS/SEQLOC.HTML#_Seq-loc:_Locations_on</rdfs:comment>
+        <rdfs:comment>drawing from seqloc categories from NCBI at https://www.ncbi.nlm.nih.gov/IEB/ToolBox/SDKDOCS/SEQLOC.HTML#_Seq-loc:_Locations_on</rdfs:comment>
```
mdehoon: See the comment at the bottom:

 

It's better to regenerate this using the current OWL API, instead of modifying the output to march what we think it should be.

Tests/Blast/broken5.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">
+<!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "https://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">
```
mdehoon: This DTD is not actually uses, so it's better to keep the output as is.

Tests/Blast/html_msgid_29_blastx_001.html
```diff
@@ -539,24 +539,24 @@
...
       <a href='http://www.nlm.nih.gov/accessibility.html'
       title='About using NCBI resources with assistive technology'>Accessibility</a> |
-      <a href='http://www.ncbi.nlm.nih.gov/About/glance/contact_info.html'
+      <a href='https://www.ncbi.nlm.nih.gov/About/glance/contact_info.html'
       title='How to get help, submit data, or provide feedback'>Contact</a> |
-      <a href='http://www.ncbi.nlm.nih.gov/sites/ehelp/feedback?Ncbi_App=blast'
+      <a href='https://www.ncbi.nlm.nih.gov/sites/ehelp/feedback?Ncbi_App=blast'
       title='How to get help, submit data, or provide feedback'>Send feedback</a>
    </p>
 </div>
```
mdehoon, line 562: Same here. These links are not actually used.

Tests/Blast/megablast_legacy.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "http://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">
+<!DOCTYPE BlastOutput PUBLIC "-//NCBI//NCBI BlastOutput/EN" "https://www.ncbi.nlm.nih.gov/dtd/NCBI_BlastOutput.dtd">
```
mdehoon: For this file and the other test files: Did you rerun Blast and capture the output, or did you change the existing output?

Tests/test_bgzf.py
hartwork: > This is also irrevant to the HTTPS issue.

@peterjc I see why their relation is arguable, true. I will drop that part now.

> This is a documented expectation of the test framework - and if you 'fix' it here for `python Tests/test_bgzf.py` you risk make it harder to work out why the tests might fail via the the framework.

A command that doesn't work from any folder — in particular from the repository root — is broken to me, and why document a constraint that can be dissolved as easily. I can make a dedicated pull request to fix it everywhere, in case you change your mind later.

hartwork: Dropped and pushed.

hartwork:
Hi @mdehoon,

we're now where it's clear that we are both wasting our time going forther with this pull request as is. I will go create an issue about the actual key issue — the XXE, not the MITM — now and then you or another team member can fix it anyway you like at the best use of your time and the best use of mine. I hope that works for you. That issue now is here: #5109

Thanks for your understanding, Sebastian

PR #5107, opened 2025-11-22
Make pre-commit CI more helpful

hartwork:
.. by showing the diff on failure and by also checking all files to uncover e.g. existing imports to now-removed files.

- [x] I hereby agree to dual licence this under both

Make pre-commit CI more helpful by (1) showing the diff on failure ~and by (2) also checking all files to e.g. uncover existing imports to about-to-be-removed files~.

PS: This is [in line with the official pre-commit GitHub Actions action](https://github.com/pre-commit/action/blob/1b06ec171f2f6faa71ed760c4042bd969e4f8b43/action.yml#L7-L19).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Run style checking via pre-commit
       run: |
-        pre-commit run --files ${{ steps.changed-files.outputs.all_changed_files }}
+        pre-commit run --show-diff-on-failure --files ${{ steps.changed-files.outputs.all_changed_files }}
 
   # Cache pip & python dependencies for Linux runner
   # since it is used in most actions.


peterjc:
We deliberately didn't run pre-commit on all the files due to the time needed - let's see how bad that is now as thing change...

hartwork:
> We deliberately didn't run pre-commit on all the files due to the time needed - let's see how bad that is now as thing change...

@peterjc thanks for giving it a try! I see `succeeded [..] in 2m 19s` at https://github.com/biopython/biopython/actions/runs/19598298380/job/56210621322?pr=5107 . What do you think?

peterjc:
Note the style check is deliberately first before everything else, and should be quick. This aviods trying the slow tests with a syntax error for example.

Compared to 30s for recent builds, over 2 minutes in not acceptable to me - that's adding over 1.5 mins to the all ready painful total run time.

The diff change might be worth taking though?

hartwork:
> Note the style check is deliberately first before everything else, and should be quick. This aviods trying the slow tests with a syntax error for example.

That's an interesting point.

I see two ways around it, but consider them as orthogonal, you will probably not like them:
- Duplicating the style check and making the copy run on all files. That adds coverage without increasing overall CI wall clock time.
- Making the slow tests run in any case, the damage of that may be less than anticipated.


> Compared to 30s for recent builds, over 2 minutes in not acceptable to me - that's adding over 1.5 mins to the all ready painful total run time.
> 
> The diff change might be worth taking though?

Cool! I will revert the `--all-files` part now.

hartwork:
Dropped.

peterjc:
There is something to be said for having GitHub Actions run a quick set of linting checks (maybe ruff but not flake8, and perhaps changed files only) that will abort the entire run if failed, and then in parallel with the main tests run the full stlye checks (on all files).

Is the diff-only change here good to squash-and-merge?

hartwork:
Hi Peter,

> Is the diff-only change here good to squash-and-merge?

yes, it's already a single commit only :+1: 

> There is something to be said for having GitHub Actions run a quick set of linting checks (maybe ruff but not flake8, and perhaps changed files only) that will abort the entire run if failed, and then in parallel with the main tests run the full stlye checks (on all files).

PS: Only running Ruff would be an easy fix to the pre-commit command line and significantly affect runtime:

#### All checks, all files: 91 seconds

```console
# time pre-commit run -a
check that executables have shebangs.....................................Passed
check json...............................................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
ruff check...............................................................Passed
black....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Passed
blacken-docs.............................................................Passed
rstcheck.................................................................Passed
doc8.....................................................................Passed
codespell................................................................Passed
Enforce https://doi.org/ style...........................................Passed
Contributor sorting......................................................Passed

real    1m31.636s
user    5m22.873s
sys     0m3.292s
```

#### Ruff only, all files: 1 second

```console
# time pre-commit run -a ruff-check
ruff check...............................................................Passed

real    0m0.425s
user    0m0.361s
sys     0m0.074s
```

By the way flake8 has near-100% overlap with Ruff in theory, maybe flake8 can be dropped altogether at some point.

I guess it all comes down to what the requirements are.

peterjc:
Beware of the caches in these timings - the CI will not have that.

Also we don't use flake8 for the base functionalty (ruff does that faster) but for the plugins (not available in ruff).

hartwork:
> Beware of the caches in these timings - the CI will not have that.

I'm aware that total CI runtime is more than running `pre-commit` on the command line, yes. But CI can also do some related caching (and already does) and the relative diff between the two runtimes above was my main point.

> Also we don't use flake8 for the base functionalty (ruff does that faster) but for the plugins (not available in ruff).

Ruff is coming for flake8 pulgins too, e.g. they already have at least parts of the plugins I found in `.flake8` and `.pre-commit-config.yaml`:

Found:
- https://docs.astral.sh/ruff/rules/#flake8-blind-except-ble
- https://docs.astral.sh/ruff/rules/#flake8-bugbear-b
- https://docs.astral.sh/ruff/rules/#flake8-commas-com
- https://docs.astral.sh/ruff/rules/#flake8-pie-pie
- https://docs.astral.sh/ruff/rules/#flake8-quotes-q
- https://docs.astral.sh/ruff/rules/#pycodestyle-e-w
- https://docs.astral.sh/ruff/rules/#pydocstyle-d
- https://docs.astral.sh/ruff/rules/#pyflakes-f

Missing?:
- flake8-rst-docstrings
- flake8-sfs

PR #5128, opened 2025-12-30
Add gc_content() function to Bio.SeqUtils

mjmedesr:
This PR adds a `gc_content()` function to `Bio.SeqUtils` that returns GC percentage (0-100). It is a convenience wrapper around the existing `gc_fraction()` function.

The function signature matches `gc_fraction` with the same `ambiguous` parameter options ("remove", "ignore", "weighted"). It multiplies the fraction by 100.

This addition is useful for users who prefer percentage values, e.g., in primer design or DNA nanostructure analysis.

Closes # (issue number to be added later)

### DIFF
--- a/Bio/SeqUtils/gc_content.py
+++ b/Bio/SeqUtils/gc_content.py
@@ -1,0 +1,24 @@
+"""GC content calculation."""
+
+from Bio.SeqUtils import gc_fraction
+
+
+def gc_content(seq, ambiguous="remove"):
+    """Calculate GC content percentage.
+
+    Parameters
+    ----------
+    seq : str, Seq, or SeqRecord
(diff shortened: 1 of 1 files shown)


peterjc:
No thank you, you are just recreating a previously removed function which was less useful than `gc_fraction`.

Quoting https://github.com/biopython/biopython/blob/master/DEPRECATED.rst

> Function 'GC' in Bio.SeqUtils was deprecated in Release 1.80, and removed in Release 1.82. Instead use function 'gc_fraction'.

As of Biopython 1.80 it was defined as follows: https://github.com/biopython/biopython/blob/biopython-180/Bio/SeqUtils/__init__.py#L139

PR #5142, opened 2026-01-16
Migrate to pyproject.toml

### LINKED ISSUE #5092: Warning: Package 'Bio.Align.substitution_matrices' is absent from the `packages` configuration.
```
❯ python setup.py build
running build
running build_py
creating build/lib.macosx-11.0-arm64-cpython-310/Bio
...
adding license file 'LICENSE.rst'
writing manifest file 'biopython.egg-info/SOURCES.txt'
/Users/peterjc/miniforge3/envs/biopy/lib/python3.10/site-packages/setuptools/command/build_py.py:212: _Warning: Package 'Bio.Align.substitution_matrices' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'Bio.Align.substitution_matrices' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'Bio.Align.substitution_matrices' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'Bio.Align.substitution_matrices' to be distributed and are
        already explicitly excluding 'Bio.Align.substitution_matrices' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************

!!
  check.warn(importable)
copying Bio/cpairwise2module.c -> build/lib.macosx-11.0-arm64-cpython-310/Bio
...
```

This seems odd as it **is** listed in `setup.py`:

```
❯ grep Bio.Align.substitution_matrices setup.py -C 4
PACKAGES = [
    "Bio",
    "Bio.Affy",
    "Bio.Align",
    "Bio.Align.substitution_matrices",
    "Bio.Align.substitution_matrices.data",
    "Bio.AlignIO",
    "Bio.Alphabet",
    "Bio.Blast",
    "Bio.CAPS",
--
    Extension("Bio.Align._alignmentcounts", ["Bio/Align/_alignmentcounts.c"]),
    Extension("Bio.Align._codonaligner", ["Bio/Align/_codonaligner.c"]),
    Extension("Bio.Align._pairwisealigner", ["Bio/Align/_pairwisealigner.c"]),
    Extension(
        "Bio.Align.substitution_matrices._arraycore",
        ["Bio/Align//substitution_matrices/_arraycore.c"],
    ),
    Extension("Bio.cpairwise2", ["Bio/cpairwise2module.c"]),
    Extension("Bio.Nexus.cnexus", ["Bio/Nexus/cnexus.c"]),
```

Used as follows:

```
❯ grep PACKAGES setup.py -C 4
REQUIRES = ["numpy"]

# --- set up the packages we are going to install
# standard biopython packages
PACKAGES = [
    "Bio",
    "Bio.Affy",
    "Bio.Align",
    "Bio.Align.substitution_matrices",
--
        "Topic :: Scientific/Engineering :: Bio-Informatics",
        "Topic :: Software Development :: Libraries :: Python Modules",
    ],
    cmdclass={"test": test_biopython},
    packages=PACKAGES,
    ext_modules=EXTENSIONS,
    include_package_data=True,  # done via MANIFEST.in under setuptools
    install_requires=REQUIRES,
    python_requires=">=%i.%i" % MIN_PY_VER,
```

Same warning using the more modern `python -m build` command.

[Moving from `setup.py` to `pyproject.toml` may resolve this?]

xelandernt:

 # Contents for the PR

Replaces `setup.py` packaging configuration with `pyproject.toml`, including everything associated with this change: ci, documentation, etc. (hopefully I didn't miss anything)

# Potential points of discussion

## Running tests

Currently `python setup.py test --offline` produces the following files in the root directory which are not gitignored (am I doing something wrong?):

```
other_trees.nex
other_trees.xml
tree1.nwk
tree1.xml
```

Since I opted for completely deleting `setup.py` the new equivalent way to write tests would be:
```
cd Tests
python run_tests.py
```
However, In my testing I had less errors when running the code bellow compared to above (8 vs 11)?
```
python Tests/run_tests.py
```
Is there a specific reason we need to change directory to the `Tests` folder?

## Keywords

`pyproject.toml` supports project `keywords` in the `project` section. I didn't add any, but I can add them if you propose a list.

## License

As per the [docs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files), since the Biopython license is not a valid [spdx](https://spdx.org/licenses/) license, I can write
```toml
[project]
license = "LicenseRef-Biopython OR BSD-3-Clause"
``` 
I hope this is correct.

## Setuptools

New minimum version is `74.1` since it supports `[[tool.setuptools.ext-modules]]`

Since `setuptools` is required to run tests (because `setuptoosl.find_packages()` is used), I added `setuptools` to the 
`dev` depedency group so they can be installed via `pip install -e . --group dev`. This prevents the `ModuleNotFoundError` for `setuptools` and is more friendly with other package managers such as `uv` 

## RPM Builds

Are now more involved, since its no longer a 1-liner, but as per the [docs](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead)
the command `python setup.py bdist_rpm` should not be used anymore. 

Potentially closes #5092


peterjc:
Could you move the strict zip changes to a new RP and revert the B905 ignores added in #5132 please?

peterjc:
The license metadata is a pain, see eg c809938fda0ba0a90696779539173ea8d881ad8f, but the license should be listed as "Biopython License Agreement" *only* as we have not finished the dual licensing, see #898.

peterjc:
You asked:

> Is there a specific reason we need to change directory to the Tests folder?

Many of our tests parse files (including doctests) and assume this. Changing that would be a lot of churn for little if any benefit.

xelandernt:
@peterjc 

- Change license to: `LicenseRef-Biopython-License-Agreement`
- Merged `master` into this branch to adress the linting issues.

CONTRIBUTING.rst
```diff
@@ -79,8 +79,9 @@ Local Testing
 Please always run the full test suite locally before submitting a pull
 request, e.g.::
 
-    $ pip install -e .
-    $ python setup.py test
+    $ pip install -e . --group dev
```
peterjc, line 82: What's the `--group dev` for?

peterjc:
Does this still follow our manifest? Is that related to the `include-package-data = true` line?

Update: Or maybe the `build-backend = "setuptools.build_meta"` bit?

pyproject.toml
```diff
@@ -0,0 +1,170 @@
+[build-system]
+requires = ["setuptools>=74.1", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "biopython"
+dynamic = ["version"]
+description = "Freely available tools for computational molecular biology."
+readme = "README.rst"
+requires-python = ">=3.10"
```
peterjc, line 10: This is much more elegant than the code we had in `setup.py` for the minimum version of Python 😄

pyproject.toml
```diff
@@ -0,0 +1,170 @@
+[build-system]
+requires = ["setuptools>=74.1", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "biopython"
+dynamic = ["version"]
```
peterjc, line 7: Is this enough to pull out `Bio.__version__` for us? That was always an ugly but widely used workaround in the `setup.py` days.

CONTRIBUTING.rst
xelandernt: >I added setuptools to the
dev depedency group so they can be installed via `pip install -e . --group dev`. This prevents the ModuleNotFoundError for setuptools and is more friendly with other package managers such as uv

So it basically installs the package and `setuptools`. Since with `pyproject.toml` the build backend is installed in a temporary environment during build time. If you do not add the `--group dev` flag and you don't have `setuptools` installed you cannot run tests after installing the package.

xelandernt:
>Does this still follow our manifest? Is that related to the include-package-data = true line?
>Update: Or maybe the build-backend = "setuptools.build_meta" bit?

Yes the `include-package-data = true` ensures that `MANIFEST.in` is respected.

pyproject.toml
xelandernt: Yes, in combination with:
```toml
[tool.setuptools.dynamic]
version = {attr = "Bio.__version__"}
```

xelandernt: An alternative approach to manually updating the `version` (either in your code or in `pyproject.toml`) is to extract the package version from the git tag. This is provided by tools such as `setuptools-scm`. However, since you are tagging releases with `biopython-{187}` this does not transfer easily to version`1.87`, for instance I don't think you would want `biopython-200` to be version `2.0`. Thus you would need to change your tagging strategy (which I guess at this point you do not want to do).

peterjc: We'd need a strong reason to change the tag stlye which dates back to CVS days. So this seems fine, thanks!

peterjc:
This is looking good - I need to test it locally still, and (given the importance and potential impact of these changes) a second review would be wise. Ideally someone can try a local compile under Windows too?

peterjc:
Referenced in https://lists.open-bio.org/pipermail/biopython/2026-January/017103.html - I would like to see this as part of our next release, Biopython 1.87

xelandernt:
I agree that it might be a good idea to keep `setup.py` for backwards compatibility in case someone if using it to install the project.

## Changes

- reintroduced `setup.py`
- Added deprecation message that will both raised as a warning and written to `stderr` when anything `python setup.py *` is invoked
 

@peterjc I would suggest to make an issue (or similar) that tracks the removal of `setup.py` in a future release.

setup.py
```diff
@@ -1,274 +1,19 @@
...
-            return ast.literal_eval(line.split("=")[1].strip())
-    return "Undefined"
-
-
-__version__ = get_version()
+_DEPRECATION_MESSAGE = (
+    "Invoking setup.py is deprecated and will be removed in a future release of Biopython.\n"
+    "Please use `pip install` or `python -m build` instead of `python setup.py` commands.\n"
+    "For further information see https://packaging.python.org/en/latest/discussions/setup-py-deprecated/\n"
+)
```
peterjc, line 13: Nice - just needs a note in `DEPRECATED.rst` (which I should have suggested earlier), and I would suggest `NEWS.rst` too (althought that file is more prone to merge conflicts if we delay merging this PR).

Doc/Tutorial/chapter_testing.rst
```diff
@@ -30,7 +30,8 @@ directory using the following:
 
 .. code:: console
 
-   $ python setup.py test
+   $ cd Tests
+   $ python run_tests.py
```
peterjc, line 34: The next lines and example can be dropped now (which also change to the tests sub-directory and run the command).

peterjc:
Nice. Almost there now.

I can see the value in using the issue tracker for to be removed deprecations (especially with milestones/labels), but we normally just search the DEPRECATED file for recent deprecations to see if any are ready to be removed.

Doc/biopdb_faq.tex
```diff
@@ -114,7 +114,7 @@ \section{Bio.PDB's installation}
 installed by default.  This module relies on a third party tool called
 flex (fast lexical analyzer generator).
 At the time of writing, in order to parse mmCIF files you'll have to
-install flex, then tweak your \verb|setup.py| file and (re)install
+install flex, then tweak your \verb|pyproject.toml| file and (re)install
```
mdehoon: Is this still true? I didn't find any flex-dependent code in Bio.PDB.

xelandernt: I looked through `NEWS.rst` and found the following:

```rst
15 July 2013: Biopython 1.62 beta released.
================================

...

The code for parsing 3D structures in mmCIF files was updated to use the
Python standard library's shlex module instead of C code using flex.
```


Therefore I removed that paragraph.

Issue #5135, opened 2026-01-07
Incompatible with NumPy 2.4


piecole:
At least one function/test is incompatible with NumPy updated to 2.4 (released 2025-12-20).

**NumPy 2.4:**
```
====================================================== FAILURES =======================================================
____________________________________________ VectorTests.test_coord_space _____________________________________________
TypeError: only 0-dimensional arrays can be converted to Python scalars

The above exception was the direct cause of the following exception:

self = <test_PDB_vectors.VectorTests testMethod=test_coord_space>

    def test_coord_space(self):
        """Confirm can generate coordinate space transform for 3 points."""
        # start with 3 points already aligned to axes
        point_set = (
            np.array([[2.0], [0.0], [2.0], [1.0]]),
            np.array([[0.0], [0.0], [0.0], [1.0]]),
            np.array([[0.0], [0.0], [2.0], [1.0]]),
        )
        # confirm get id matrix to transform to/from coord space
        homog_id = np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]])
>       mtxs = coord_space(point_set[0], point_set[1], point_set[2], True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Tests\test_PDB_vectors.py:208:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
Bio\PDB\vectors.py:559: in coord_space
    set_homog_trans_mtx(-a1[0], -a1[1], -a1[2], tm)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = array([-0.]), y = array([-0.]), z = array([-0.])
mtx = array([[1., 0., 0., 0.],
       [0., 1., 0., 0.],
       [0., 0., 1., 0.],
       [0., 0., 0., 1.]])

    def set_homog_trans_mtx(x: float, y: float, z: float, mtx: np.ndarray):
[...]

PR #5148, opened 2026-01-25
Fix `Entrez.Parser.DataHandler` URL security check continuity (related to #5109 and #5114)

### LINKED ISSUE #5109: [CVE-2025-68463] `Bio.Entrez.DataHandler`, `Bio.Entrez.read`, `Bio.Entrez.parse` are vulnerable to XXE and SSRF
Hi!

The code in class `Bio.Entrez.DataHandler` — that is also powering API functions `Bio.Entrez.read` and `Bio.Entrez.parse` — parses arbitrary XML content in a way where contained DTD and XSD URLs are requested via a HTTP `GET` request through stdlib function `urllib.request.urlopen` unless a file with the same basename — the last path component of the URL — matches a local file in the filesystem based DTD cache (or XSD cache respectively):

https://github.com/biopython/biopython/blob/d07dde90e0eacbc562d5ed011165890e69a7bb71/Bio/Entrez/Parser.py#L581-L584

https://github.com/biopython/biopython/blob/d07dde90e0eacbc562d5ed011165890e69a7bb71/Bio/Entrez/Parser.py#L1126-L1131

That `urlopen` allows an attacker to craft XML that when parsed makes the parser do arbitrary(!) HTTP `GET` requests which can be used to e.g. access internal network resources or cause denial of service: the code is vulnerable to [server-side request forgery](https://en.wikipedia.org/wiki/Server-side_request_forgery) via a form of "doctype XXE" ([**X**ML e**x**ternal **e**ntity attack](https://en.wikipedia.org/wiki/XML_external_entity_attack)).

**Demo**: [`biopython_doctype_xxe_demo.py`](https://github.com/user-attachments/files/24154447/biopython_doctype_xxe_demo.py)

The very same code in Biopython is also vulnerable to MITM (due to not enforcing TLS) and cache poisoning (..), but that is secondary and likely goes away for free once the SSRF issue is plugged properly.

Official Python >=3.15 docs are now warning of problems like these at https://docs.python.org/3.15/library/pyexpat.html#xml.parsers.expat.xmlparser.ExternalEntityRefHandler .

I hope that you will find some complete fix to this problem, so that users are safe from attacks through BioPython's XML parser.

Thanks and best, Sebastian

CC @mdehoon @peterjc

hartwork:
- [x] I hereby agree to dual licence this ~and any previous contributions~ under both

Closes #5109, follow-up to #5114

CC @peterjc

### DIFF
--- a/Bio/Entrez/Parser.py
+++ b/Bio/Entrez/Parser.py
@@ -1082,7 +1082,11 @@ class DataHandler(metaclass=DataHandlerMeta):
             handle.close()
 
     def verify_security(self, url, verify_hostname=True):
-        """Check if the given URL is from a trustable source."""
+        """Check if the given URL is from a trustable source.
+
+        When ``self.bypass_url_security`` evaluates to ``True``,
+        all URL security checks will be skipped.
+        """
         if not self.bypass_url_security:
             parts = urlparse(url)
             scheme = parts.scheme


Bio/Entrez/Parser.py
```diff
@@ -1082,20 +1082,19 @@ def save_xsd_file(self, filename, text):
...
+        """Check if the given URL is from a trustable source."""
+        if not self.bypass_url_security:
             parts = urlparse(url)
             scheme = parts.scheme
             hostname = parts.hostname
             hostnames = (
                 "www.ncbi.nlm.nih.gov",
                 "dtd.nlm.nih.gov",
                 "eutils.ncbi.nlm.nih.gov",
+                "jats.nlm.nih.gov",
```
mdehoon: My concern here is that NCBI may some day decide to add some new host name, and then our parser would immediately break and become useless. Can we set up a policy that any DTD that (a) came from these hosts, or (b) were included indirectly through any DTD in (a), is acceptable?

hartwork: @mdehoon hi!

> My concern here is that NCBI may some day decide to add some new host name, and then our parser would immediately break and become useless.

I believe that concern only applies to new DTD files that are not contained in Biopython's shipped file system cache that takes precedence.

> Can we set up a policy that any DTD that (a) came from these hosts, or (b) were included indirectly through any DTD in (a), is acceptable?

I have implemented that general idea now in new commit 391fdba0537be5a25d43b91f7294ab9ab2274735. The commit is easiest to review using `git show -U1 -w 391fdba0537be5a25d43b91f7294ab9ab2274735` which shows this output for me:

git_show_391fdba0537be5a25d43b91f7294ab9ab2274735

What do you think?

peterjc: I can image the NCBI moving to a higher level domain name or being renamed, but would accepting `*.nlm.nih.gov` suffice for the medium term?

hartwork: @peterjc interesting idea! I went and implemented that in commit 646c7931e15c3c29259a69da4356d78193e01b2a now instead, pushed.

Bio/Entrez/Parser.py
mdehoon: If you switch these two lines and call `verify_security` only if `len(self.dtd_urls) == 0`, then only the toplevel DTD and schemas will be checked.
This is still safe with your previous example (https://github.com/biopython/biopython/issues/5109#issuecomment-3792173161).

hartwork: @peterjc @mdehoon please agree on whether you want me to include or exclude non-parent DTD URLs from checking. With a vote for include, the PR already implements that and it would need buy-in from @mdehoon . With a vote for exclude, I would offer a hybrid between previous 391fdba0537be5a25d43b91f7294ab9ab2274735 and previous 646c7931e15c3c29259a69da4356d78193e01b2a for the next round of review. Please let me know.

hartwork: @peterjc @mdehoon any thoughts?

mdehoon: I think we can exclude non-parent DTD URLs. If we include non-parent DTD in the check, the parser becomes fragile, and will break the moment NCBI decides to include a DTD from a source we had not foreseen.

hartwork: @mdehoon so you have concerns about pattern `*.nlm.nih.gov` being broad enough to cover the future and would like to loosen non-parent checks to further reduce the chance of future breakage while reducing security in the sense that a broader pattern means blindly trusting additional parties, effectively relying on transitive trust a la "we trust anyone that NCBI will decide to trust with DTD hosting in the future". @peterjc do you have a minute to voice support or concerns about the direction that @mdehoon is in favor of?

mdehoon: > @mdehoon so you have concerns about pattern *.nlm.nih.gov being broad enough to cover the future and would like to loosen non-parent checks to further reduce the chance of future breakage while reducing security in the sense that a broader pattern means blindly trusting additional parties, effectively relying on transitive trust a la "we trust anyone that NCBI will decide to trust with DTD hosting in the future".

Yes.

peterjc: I'd lean to the more cautious only trust the NCBI and add exceptions later if needed, but will yield to @mdehoon as the maintainer of this code to take the more pragmatic choice.

mdehoon: > ... add exceptions later if needed ...

The problem is that the parser would break immediately, and becomes unusable until the next Biopython release comes out with the new websites added.

Note that the security issue only arises for people who are running the `Bio.Entrez` parser as a webserver, taking XML as input from the user. I doubt that this is a common occurrence.

peterjc: This really is about where to draw the line on security vs functionality/convience. Ultimately there could be security risks parsing any user-supplied data, but the risk here is higher than most of our plain-text parsers.

hartwork: @peterjc @mdehoon I have put a combination of both your ideas in now. Please let me know what you think :pray:

Bio/Entrez/Parser.py
```diff
@@ -1081,21 +1081,16 @@ def save_xsd_file(self, filename, text):
             handle.write(text)
             handle.close()
 
-    def verify_security(self, url):
-        """Check if the url is from a trustable sournce."""
-        if not self.secure:
+    def verify_security(self, url, verify_hostname=True):
+        """Check if the given URL is from a trustable source."""
```
peterjc: I'd be inclined to extend the docstring so say the method is a no-op if `.bypass_url_security` is True. But I won't block the PR over this polish.

hartwork: @peterjc addressed in commit b24914bcc79dcd4366e524c8140a5080a0fbf4ff now

NEWS.rst
```diff
@@ -23,6 +23,7 @@ possible, especially the following contributors:
 - Timothy Dennis (first contribution)
 - Ziyan Rao (first contribution)
 - Manuel Lera-Ramirez
+- Sebastian Pipping
```
peterjc, line 26: FYI, we sort this alphabetically, but leave it now unless there's anything else to change.

peterjc:
The tests and style checks look good (apart from the annoying #5135).

OK to merge @mdehoon?

peterjc:
Oh, and assuming we do merge this - are you OK with a squash-and-merge into a single commit as the PR history is getting hard to follow?

mdehoon:
Looks good to me, thanks!

hartwork:
> Oh, and assuming we do merge this - are you OK with a squash-and-merge into a single commit as the PR history is getting hard to follow?

@peterjc I see value in either approach here, please proceed as you see fits :+1:

mdehoon:
Thanks!

PR #5149, opened 2026-01-25
Refactor PSEA to use tempfile for secure, isolated execution

dipankatanu:
# Description
This PR updates Bio.PDB.PSEA.run_psea to use tempfile.TemporaryDirectory for executing the PSEA binary.

## Why this is needed
Currently, PSEA writes output files directly to the current working directory. This can lead to file name collisions in multi-threaded environments or permission errors in read-only file systems (common in HPC environments).

## Changes
Integrated tempfile to create a sandboxed execution environment.

Updated path handling using pathlib for better cross-platform compatibility.

Ensured automatic cleanup of temporary files after execution.

## Verification
Verified the logic with a standalone test script to ensure correct directory creation, file writing, and cleanup, as local C-extensions (cealign) were unavailable for full suite testing on my Windows environment.

PR #5150, opened 2026-01-25
Fix residue depth subprocess : Modernize process execution in ResidueDepth using subprocess.run

dipankatanu:
Description: > Replaced subprocess.call with subprocess.run in Bio.PDB.ResidueDepth.get_surface.
Removed shell=True to improve security and cross-platform reliability.
Switched to list-based argument passing to avoid shell injection risks.
Implemented check=True for automatic error handling of the MSMS executable.


peterjc:
You need to remove the PDB changes, which look to be from [#5149](https://github.com/biopython/biopython/pull/5149) (a duplicate of work on another PR just merged giving the conflict here now).

Which operating systems have you actually tried this on? Windows command line calls are rarely straight forward.

PR #5151, opened 2026-01-25
Refactor: Modernize string formatting in Atom and Chain using f-strings

dipankatanu:
Description:

This PR modernizes the string formatting in Bio.PDB.Atom and Bio.PDB.Chain by replacing legacy % formatting with f-strings.

Benefits:

Readability: Variables are embedded directly in the string, making the log messages and representations easier to parse.

Performance: f-strings are faster than the old % operator at runtime.

Consistency: Aligns these core modules with modern Python (3.6+) standards being adopted across Biopython.


peterjc:
See #5153, same applies. You have a bunch of unrelated changes here which do not match the description.

PR #5152, opened 2026-01-25
Add timeout support to NCBIWWW.qblast

dipankatanu:
Description: 
Added an optional timeout parameter to the qblast function to prevent infinite hangs during network requests. This allows users to specify a timeout (in seconds) that is passed directly to urllib.request.urlopen.

### DIFF
--- a/Bio/Blast/NCBIWWW.py
+++ b/Bio/Blast/NCBIWWW.py
@@ -93,6 +93,7 @@ def qblast(
     template_length=None,
     username="blast",
     password=None,
+    timeout=None,  # <--- New line added
 ):
     """BLAST search using NCBI's QBLAST server.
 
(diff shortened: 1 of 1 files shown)


Bio/Blast/NCBIWWW.py
```diff
@@ -278,7 +281,9 @@ def qblast(
             )
 
         request = Request(url_base, message, {"User-Agent": "BiopythonClient"})
-        handle = urlopen(request)
+        # Adding a timeout
+        handle = urlopen(request, timeout=timeout)
+        # handle = urlopen(request)
```
mdehoon, line 286: Please remove the comment lines.
Can the timeout parameter also be added to the qblast function in `Bio/Blast/__init__.py`?

peterjc:
You need to remove the PDB changes too, which look to be from #5149 (a duplicate of work on another PR just merged giving the conflict here now).

PR #5154, opened 2026-01-25
Fix PDB: Resolve TypeError in set_homog_trans_mtx for NumPy scalars

dipankatanu:
Description:

This PR fixes a TypeError and ValueError in Bio.PDB.vectors occurring with Python 3.12 and modern NumPy versions.

The Problem: > The set_homog_trans_mtx function was attempting to assign values directly to a NumPy array. In recent versions, if these values are passed as single-element arrays (common in PDB calculations), the assignment fails with only 0-dimensional arrays can be converted to Python scalars.

The Fix: Explicitly cast x, y, and z coordinates to float() within the function body to ensure they are treated as scalars before matrix assignment.


peterjc:
Next time please search for existing issues and PRs and add any insights to them before making a PR. i.e. #5135 and #5138. Your suggestion is at least shorter (if we ignore the unwanted changes on this and your other recent PRs).

PR #5155, opened 2026-01-25
Feat: Add NetworkX adapter for PDB NeighborSearch results

dipankatanu:
# Description:

This PR introduces a new utility module Bio.PDB.networks to facilitate the conversion of structural neighbor search results into networkx.Graph objects.

## Key Features:

Spatial-to-Graph Mapping: Directly converts atom/residue pairs into nodes and edges.

Weighted Edges: Automatically uses Euclidean distance (via the Entity - operator) as the edge weight.

Attribute Injection: Maps entity levels and IDs to node attributes for downstream network analysis.

## Use Case: 
This bridge allows researchers to instantly apply graph-theory metrics (centrality, clustering, community detection) to protein structures to identify key interaction hubs.

## Verification: 
 Logic verified via Tests/test_PDB_networkx.py.

### DIFF
--- a/Bio/PDB/networks.py
+++ b/Bio/PDB/networks.py
@@ -1,0 +1,33 @@
+import numpy as np
+from Bio.PDB.Entity import Entity
+
+
+def to_networkx(neighbor_pairs):
+    """Convert Biopython neighbor search results to a NetworkX Graph.
+
+    Arguments:
+    - neighbor_pairs: List of tuples (Entity, Entity) from NeighborSearch.search_all
+    """
+    try:
+        import networkx as nx
+    except ImportError:
+        raise ImportError(
+            "NetworkX is required for this function. Install it with 'pip install networkx'."
+        )
+
+    G = nx.Graph()
+
+    for entity1, entity2 in neighbor_pairs:
+        # Biopython '-' operator calculates Euclidean distance between Entities
+        distance = entity1 - entity2
+
+        # Add edge with distance as weight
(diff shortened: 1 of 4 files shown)


peterjc:
Please remove the unrelated changes from this PR. Ideally with `git rebase -i` to give a clearer history, but alternatively if this does get accepted we can squash-and-merge into one commit.

Bio/PDB/networks.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+from Bio.PDB.Entity import Entity
+
+
+def to_networkx(neighbor_pairs):
```
JoaoRodrigues, line 5: Would you be open to adding this method directly to the NeighborSearch module? I think that is cleaner than adding a new module.

You can also add type hints to the function so that it's easier to figure out the required input.

Tests/test_PDB_networkx.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+from Bio.PDB.Atom import Atom
+from Bio.PDB.networks import to_networkx
+
+class NetworkXExportTest(unittest.TestCase):
```
JoaoRodrigues, line 5: If you can, add a test that runs NeighborSearch and passes the output to your method. That prevents us from having errors in the future when/if we make changes to that output.

JoaoRodrigues (changes requested):
As @peterjc suggested, please rebase the PR so that the changes are limited to your contribution. Thanks!

PR #5157, opened 2026-02-04
Fix PDBList URL construction for non-RCSB Mirrors

### LINKED ISSUE #3987: Fix PDBList access to servers other than RCSB

kirandharmar867-maker:
Summary:
This PR fixes URL Construction in PDBList so it works correctly  With non-RCSB PDB mirrors.

What was changed :
> Updated the default PDB_Server To include the /pub/pdb prefix.
> removed Hardcoded /pub/pdb Segments From all constructed URLs, Making them relative to PDB_server.

Why This needed :
Some PDB mirrors (ex: EBl) use a Different Directory layout and do not work With hardcoded /pub/pdb paths.
by moving the prefix into the server root, PDBList now works consistently across Mrirors.

issue reference:
Fixes #3987

Scope:
this PR is only addresses URL Path construction as described in the issue.

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -88,7 +88,7 @@ class PDBList:
 
     def __init__(
         self,
-        server="https://files.wwpdb.org",
+        server="https://files.wwpdb.org/pub/pdb",
         pdb=None,
         obsolete_pdb=None,
         verbose=True,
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
Thanks @kirandharmar867-maker, great catch! The failing tests are due to another issue, so I'd be happy to merge this once that's fixed.

JoaoRodrigues (approved):

kirandharmar867-maker:
> Thanks @kirandharmar867-maker, great catch! The failing tests are due to another issue, so I'd be happy to merge this once that's fixed.

Oh! Thank You @JoaoRodrigues, 

defiantly I Will Fix It.

peterjc:
It's just that niggling #5135, so let's merge this now. Thanks both!

peterjc:
Correction - you need to agree to the licensing first, see https://github.com/biopython/biopython/blob/master/.github/PULL_REQUEST_TEMPLATE.md

peterjc:
Will you agree to the licensing explicitly please (more than just a thumbs up)?

PR #5158, opened 2026-02-08
Pdbparser thread safety

### LINKED ISSUE #5131: PDBParser is not thread-safe when used in ThreadPoolExecutor
**Summary**
Bio.PDB.PDBParser appears to have thread-safety issues when the same instance is shared across multiple threads in a ThreadPoolExecutor. This causes sporadic parsing failures with errors like "0 defined twice" and "'Atom' object has no attribute 'selected_child'".

**Environment**
Python: 3.11
BioPython: 1.86
OS: Linux (CentOS/Ubuntu)

```python
from concurrent.futures import ThreadPoolExecutor
from Bio.PDB import PDBParser
from pathlib import Path

def process_pdb(pdb_path, parser):
    structure = parser.get_structure(pdb_path.stem, str(pdb_path))
    return len(list(structure.get_residues()))

# Shared parser instance
parser = PDBParser(QUIET=True, PERMISSIVE=1)
pdb_files = list(Path("pdb_directory").glob("*.pdb"))

# This fails randomly with "0 defined twice" errors
with ThreadPoolExecutor(max_workers=64) as executor:
    results = list(executor.map(lambda p: process_pdb(p, parser), pdb_files))
```

**Expected Behavior**
Parsing should succeed for all valid PDB files.

**Actual Behavior**
Random failures with errors:

- "0 defined twice" (residue ID collision)
- "'Atom' object has no attribute 'selected_child'"

~70% of files fail when using 64 threads, while 100% succeed when processed sequentially.

**Workaround**
Create a new PDBParser instance for each thread/process:
```python
def process_pdb(pdb_path):
    parser = PDBParser(QUIET=True)  # New instance per call
    structure = parser.get_structure(pdb_path.stem, str(pdb_path))
    return len(list(structure.get_residues()))
```
Or use ProcessPoolExecutor instead of ThreadPoolExecutor.

**Suggestion**
Consider documenting that PDBParser is not thread-safe, or making internal state thread-local.

kirandharmar867-maker:
Summary: 
This PR adds Sort documentation note Clarifying that PDBParser instances are not thread-safe and
Should not be shared across threads.

Changes:
~ Added a new-line note to the new Parser Class docstring documenting thread-safety limitations.
~ no Functional or behavioral changes.

Motivation:
When using PDBParser with ThreadPoolExecutor, Sharing a single Parser instance across Threads can
lead to race conditions and sporadic Parsing errors.
This behavior was discussed in #5131 and is excepted due to internal mutable state.

documenting this explicitly helps user avoid subtle multithreading issues.

Related issue:
closes #5131

Checklist: 
documentation only change.
no API or behavior change.
Follows Biopython style.

### DIFF
--- a/Bio/PDB/PDBParser.py
+++ b/Bio/PDB/PDBParser.py
@@ -19,7 +19,11 @@ from Bio.PDB.StructureBuilder import StructureBuilder
 
 
 class PDBParser:
-    """Parse a PDB file and return a Structure object."""
+    """Parse a PDB file and return a Structure object.
+
+    Note: PDBParser instances are not thread-safe and should not be shared
+    across threads; create one parser per thread or use ProcessPoolExecutor.
+    """
 
     def __init__(
         self,
--- a/Bio/PDB/PSEA.py
+++ b/Bio/PDB/PSEA.py
@@ -53,7 +53,6 @@ def run_psea(fname, verbose=False):
             raise RuntimeError(f"Error running p-sea: {p.stderr}")
 
 
-
 def psea(pname):
     """Parse PSEA output file."""
     fname = run_psea(pname)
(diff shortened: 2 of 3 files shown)


peterjc:
Can you remove the unrelated PSEA work from this branch please?

PR #5168, opened 2026-02-24
fix: Deprecate query_letters and database_letters attributes

### LINKED ISSUE #1000: Deprecate record.database_letters, record.query_letters
From redmine [XML Blast parser: miscellaneous bug fixes and cleanup](https://redmine.open-bio.org/issues/2176)
Most of the cleanup, bug fixes were done. The last comment was:

> We could perhaps deprecate record.database_letters immediately, and at a later point, record.query_letters

There is a [TODO](https://github.com/biopython/biopython/blob/589bc075ace24cb29bcd69c9789963df7febeb4d/Bio/Blast/NCBIXML.py#L188) in the code regarding this.

@peterjc last comment

> I've updated CVS as per comment 12 to also use record.query_length, and comment 13 to also use record.database_length.
> 
> Before:
> 
> from Bio.Blast import NCBIXML
> for record in NCBIXML.parse(open("xbt007.xml")) :
> 
> ... print record.query_id
> ... print record.query_letters, record.query_length
> ... print record.num_letters_in_database, record.database_letters, record.database_length
> ... 
> gi|585505|sp|Q08386|MOPB_RHOCA
> 270 None
> 13958303 None None
> gi|129628|sp|P07175.1|PARA_AGRTU
> 222 None
> 13958303 None None
> 
> Now, with Bio/Blast/NCBIXML.py CVS revision 1.20 or 1.21,
> 
> from Bio.Blast import NCBIXML
> for record in NCBIXML.parse(open("xbt007.xml")) :
> 
> ... print record.query_id
> ... print record.query_letters, record.query_length
> ... print record.num_letters_in_database, record.database_letters, record.database_length
> ... 
> gi|585505|sp|Q08386|MOPB_RHOCA
> 270 270
> 13958303 None 13958303
> gi|129628|sp|P07175.1|PARA_AGRTU
> 222 222
> 13958303 None 13958303
> 
> We could perhaps deprecate record.database_letters immediately, and at a later point, record.query_letters

A404coder:

Closes #1000

## Changes
- Convert `query_letters` and `database_letters` to properties in `Header` class with deprecation warnings via `BiopythonParserWarning`
- Use `_query_letters` directly in parser internals to avoid triggering warnings during XML parsing
- Update all tests in `test_NCBIXML.py` and `test_NCBI_qblast.py` to use `query_length` instead of `query_letters`
- Add tests confirming deprecation warnings are raised for both `query_letters` and `database_letters`
- Add deprecation entry to `DEPRECATED.rst`
- Add contributor name to `NEWS.rst` and `CONTRIB.rst`

## Context
The issue has been open since 2016. The XML parser sets both `query_letters` and `query_length`, as well as `database_letters` and `database_length`. This change helps users migrate to the preferred attributes while maintaining backward compatibility.

## Testing
- All 24 existing NCBIXML tests pass with `query_length` replacing `query_letters`
- 2 new tests verify `BiopythonParserWarning` is raised when accessing deprecated attributes
- No warnings triggered during normal XML parsing (parser uses private attributes internally)

### DIFF
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -818,7 +818,6 @@ class BlastParser(_XMLparser):
         # Hack to record the claimed database size as database_length
         # (as well as in num_letters_in_database, see Bug 2176 comment 13):
         self._blast.database_length = self._blast.num_letters_in_database
-        # TODO? Deprecate database_letters next?
 
         # Hack to record the claimed database sequence count as database_sequences
         self._blast.database_sequences = self._blast.num_sequences_in_database
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -808,13 +808,13 @@ class BlastParser(_XMLparser):
             self._blast.query = self._header.query
         if not hasattr(self._blast, "query_id") or not self._blast.query_id:
             self._blast.query_id = self._header.query_id
-        if not hasattr(self._blast, "query_letters") or not self._blast.query_letters:
-            self._blast.query_letters = self._header.query_letters
+        if not self._blast._query_letters:
+            self._blast._query_letters = self._header._query_letters
 
         # Hack to record the query length as both the query_letters and
         # query_length properties (as in the plain text parser, see
         # Bug 2176 comment 12):
-        self._blast.query_length = self._blast.query_letters
+        self._blast.query_length = self._blast._query_letters
         # Perhaps in the long term we should deprecate one, but I would
         # prefer to drop query_letters - so we need a transition period
         # with both.
@@ -911,7 +911,7 @@ class BlastParser(_XMLparser):
         Important in old pre 2.2.14 BLAST, for recent versions
         <Iteration_query-len> is enough
         """
-        self._header.query_letters = int(self._value)
+        self._header._query_letters = int(self._value)
 
     def _set_record_query_id(self):
         """Record the identifier of the query (PRIVATE)."""
@@ -923,7 +923,7 @@ class BlastParser(_XMLparser):
 
     def _set_record_query_letters(self):
         """Record the length of the query (PRIVATE)."""
-        self._blast.query_letters = int(self._value)
+        self._blast._query_letters = int(self._value)
 
     # def _end_BlastOutput_query_seq(self):
     #     """The query sequence (PRIVATE)."""
--- a/Bio/Blast/NCBIXML.py
+++ b/Bio/Blast/NCBIXML.py
@@ -59,10 +59,12 @@ class Header:
 
     query               Name of query sequence.
     query_letters       Number of letters in the query sequence.  (int)
+                        DEPRECATED: Use query_length instead.
 
     database            Name of the database.
     database_sequences  Number of sequences in the database.  (int)
     database_letters    Number of letters in the database.  (int)
+                        DEPRECATED: Use database_length instead.
 
     """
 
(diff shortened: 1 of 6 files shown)


Bio/Blast/NCBIXML.py
```diff
@@ -74,11 +76,43 @@ def __init__(self):
...
+        self._database_letters = None
+
+    @property
+    def query_letters(self):
+        """DEPRECATED: Use query_length instead."""
+        warnings.warn(
+            "The query_letters attribute is deprecated. "
+            "Please use query_length instead.",
+            BiopythonParserWarning,
+            stacklevel=2,
```
peterjc, line 92: Why stack level 2?

peterjc:
This is what I expected on the code side, which is good.

You've ignored the pull request template - please address 
that.

There should be a test accessing the `.query_letters` and `.database_letters` properties and confirming the warning is present.

Otherwise, please update the tests to silence the new `BiopythonParserWarning: The query_letters attribute is deprecated. Please use query_length instead.` etc messages by using the preferred `.query_length`.

Also please add an entry to the `DEPRECATED` file.

Bio/Blast/NCBIXML.py
A404coder: Thanks for the question\! `stacklevel=2` is correct here for property getters — frame 1 is the property getter itself (where `warnings.warn` is called), and frame 2 is the caller's code that accesses `record.query_letters`. This way the warning points to the user's actual code line, not the property internals.

I also noticed the parser was internally accessing `query_letters` through the public property, which would trigger the warning during XML parsing. Fixed that by using `_query_letters` directly in the parser internals.

A404coder:
Thanks for the detailed review\! I've addressed the feedback in the latest commit:

- **Deprecation warning tests**: Added `test_deprecated_query_letters` and `test_deprecated_database_letters` that verify `BiopythonParserWarning` is raised when accessing these attributes.
- **Updated existing tests**: Replaced all `record.query_letters` usages with `record.query_length` in `test_NCBIXML.py` (19 occurrences) and `test_NCBI_qblast.py` (2 occurrences) to silence the deprecation warnings.
- **DEPRECATED.rst**: Added an entry for the deprecation of `query_letters` and `database_letters`.
- **Parser internals**: Changed the parser to use `_query_letters` directly instead of going through the public property, so no warnings are triggered during XML parsing.

I'll also fill in the PR template shortly. All tests pass locally.

Bio/Blast/NCBIXML.py
```diff
@@ -774,13 +808,13 @@ def _end_blast_record(self):
...
-        if not hasattr(self._blast, "query_letters") or not self._blast.query_letters:
-            self._blast.query_letters = self._header.query_letters
+        if not self._blast._query_letters:
+            self._blast._query_letters = self._header._query_letters
 
         # Hack to record the query length as both the query_letters and
         # query_length properties (as in the plain text parser, see
         # Bug 2176 comment 12):
-        self._blast.query_length = self._blast.query_letters
+        self._blast.query_length = self._blast._query_letters
```
peterjc, line 816: You might update the comment just below this?

peterjc:
By eye this looks good - something off in the formatting, black reports:

```
would reformat Tests/test_NCBIXML.py
would reformat Tests/test_NCBI_qblast.py
```

Bio/Blast/NCBIXML.py
A404coder: Good catch\! Updated the comment in the latest commit — removed the old "hack" / "transition period" wording and replaced it with a note that query_letters is now deprecated in favour of query_length.

A404coder:
Thanks for catching the formatting issue\! I've run black on both test files and also updated the outdated comment about the query_letters/query_length transition (as you suggested in the inline review). All should be clean now.

Bio/Blast/NCBIXML.py
peterjc: This old comment can go too.

peterjc:
Ah: "Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>"

We (Biopython) don't yet have a formal policy about this yet, but that is becoming more pressing. See also my blog post from late last year: https://blastedbio.blogspot.com/2025/11/thoughts-on-generative-ai-contributions.html

I was just involved in a conversation earlier today that "Good first issue" tagged issues (or their equivalents) ought to be explicitly reserved for humans, and not AI agents. Most projects (Biopython too) have deliberately left these "low hanging fruit" which the maintainers could easily do themselves, as real-world tasks to train beginners (even though this will often cost the maintainers more time). This is as an investment for the future (those people are then more likely to make further more valuable contribution here or elsewhere).

At least in this case the scope and nature of the changes is such that we don't need to worry about copyright issues.

peterjc:
Thank you for your understanding.

PR #5175, opened 2026-03-14
Replace assert with explicit if/raise in production code (#3600)

### LINKED ISSUE #3600: Do not use assert outside testing

Abdeltoto:

Closes #3600.

Replaces `assert` statements in production code with explicit `if/raise ValueError`
blocks, so input validation isn't silently skipped when Python runs with `-O`.

Files touched:

- `Bio/SeqIO/TabIO.py` — six asserts on tab/newline/CR in IDs and sequences,
  grouped into two `if/raise` blocks. Following @peterjc's review, the
  sequence-error message now references the record title rather than the
  sequence itself, so the exception text stays bounded for very large records.
- `Bio/ExPASy/Enzyme.py` — one assert on the PR line PROSITE prefix.
- `Bio/ExPASy/ScanProsite.py` — one assert on XML element-stack consistency.
- `Bio/motifs/alignace.py` — one assert on the `Motif` line format.

Code formatted with `black`. Manually checked that valid inputs still parse
and that the new `ValueError` paths fire on malformed inputs.

### DIFF
--- a/Bio/SeqIO/TabIO.py
+++ b/Bio/SeqIO/TabIO.py
@@ -119,12 +119,14 @@ class TabWriter(SequenceWriter):
         """Return record as tab separated (id(tab)seq) string."""
         title = _clean(record.id)
         seq = _get_seq_string(record)  # Catches sequence being None
-        assert "\t" not in title
-        assert "\n" not in title
-        assert "\r" not in title
-        assert "\t" not in seq
-        assert "\n" not in seq
-        assert "\r" not in seq
+        if "\t" in title or "\n" in title or "\r" in title:
+            raise ValueError(
+                f"Record id contains a tab, newline or carriage return: {title!r}"
+            )
+        if "\t" in seq or "\n" in seq or "\r" in seq:
+            raise ValueError(
+                f"Record sequence contains a tab, newline or carriage return: {seq!r}"
+            )
         return f"{title}\t{seq}\n"
 
     def write_record(self, record):
--- a/Bio/SeqIO/TabIO.py
+++ b/Bio/SeqIO/TabIO.py
@@ -125,7 +125,7 @@ class TabWriter(SequenceWriter):
             )
         if "\t" in seq or "\n" in seq or "\r" in seq:
             raise ValueError(
-                f"Record sequence contains a tab, newline or carriage return: {seq!r}"
+                f"Record sequence contains a tab, newline or carriage return: {title!r}"
             )
         return f"{title}\t{seq}\n"
 
--- a/Tests/test_SeqIO_TabIO.py
+++ b/Tests/test_SeqIO_TabIO.py
@@ -15,9 +15,9 @@ class TestTabIOInvalidInput(unittest.TestCase):
     """Tab format rejects records whose id or sequence contain reserved chars."""
 
     def test_tab_id_with_tab_raises(self):
-        """Reject tab-format write when record.id contains a tab character.
+        r"""Reject tab-format write when record.id contains a tab character.
 
-        Note: ``_clean`` already replaces ``\\n`` and ``\\r`` in the id with
+        Note: ``_clean`` already replaces ``\n`` and ``\r`` in the id with
         spaces before this check runs, so only the tab branch of the id check
         is reachable from a normal write call.
         """
--- a/Tests/test_SeqIO_TabIO.py
+++ b/Tests/test_SeqIO_TabIO.py
@@ -1,0 +1,43 @@
+# Copyright 2026 by the Biopython contributors. All rights reserved.
+# This code is part of the Biopython distribution and governed by its
+# license. Please see the LICENSE file that should have been included
+# as part of this package.
+"""Tests for SeqIO TabIO module — error paths."""
+
+import unittest
+
+from Bio.Seq import Seq
+from Bio.SeqIO.TabIO import TabWriter
+from Bio.SeqRecord import SeqRecord
+
+
+class TestTabIOInvalidInput(unittest.TestCase):
+    """Tab format rejects records whose id or sequence contain reserved chars."""
+
+    def test_tab_id_with_tab_raises(self):
+        """Reject tab-format write when record.id contains a tab character.
+
+        Note: ``_clean`` already replaces ``\\n`` and ``\\r`` in the id with
+        spaces before this check runs, so only the tab branch of the id check
+        is reachable from a normal write call.
+        """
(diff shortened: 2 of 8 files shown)


peterjc:
Please read https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst

mdehoon (approved):

Bio/SeqIO/TabIO.py
```diff
@@ -119,12 +119,14 @@ def to_string(cls, record):
...
-        assert "\n" not in seq
-        assert "\r" not in seq
+        if "\t" in title or "\n" in title or "\r" in title:
+            raise ValueError(
+                f"Record id contains a tab, newline or carriage return: {title!r}"
+            )
+        if "\t" in seq or "\n" in seq or "\r" in seq:
+            raise ValueError(
+                f"Record sequence contains a tab, newline or carriage return: {seq!r}"
+            )
```
peterjc, line 129: There is a risk here that the sequence is extremely long, so best not used in the exception. I would consider using the record title instead.

peterjc (changes requested):
The changes are generally good, thank you.

I have a code suggestion, but you need to run the code formatting tool black - and more critically you haven't followed the pull request template (see #5171).

Abdeltoto:
Thanks for the review, @peterjc — all three points addressed in the latest push:

- `TabIO.py`: switched the sequence-error message to reference `{title!r}` instead of `{seq!r}`, matching the pattern already used a few lines above (L124). The exception text now stays bounded even for very large records.
- Ran `black` on the four touched files — only `Enzyme.py` needed a wrap.
- PR description rewritten to follow the template (license + contributing checkboxes + `Closes #3600` and a short summary).

Ready for another look whenever you have a minute.

peterjc:
The changes look good, thank you.

How easy does it look to add test triggering the new exceptions using unittest's `.assertRaises(...)`? I think they can all be done with a malformed input.

Abdeltoto:
Thanks @peterjc — added in `3cfd0df`:

- `Tests/test_Enzyme.py` → `test_pr_line_invalid_prefix` (Enzyme record with a PR line that doesn't start with `"PROSITE; "`).
- `Tests/test_motifs.py` → `test_alignace_malformed_motif_line` (line starts with `"Motif"` — passes the prefix check — but the first whitespace-delimited token differs).
- `Tests/test_SeqIO_TabIO.py` (new file) → three tests covering the tab branch of the id check and both branches of the sequence check.
- `Tests/test_ExPASy_offline.py` → `test_scanprosite_mismatched_end_element`.

Two small notes on the design choices:

1. **ScanProsite**: `endElement` cannot be tripped via the SAX parser with well-formed XML (Expat enforces matching), and malformed XML raises `ExpatError` before reaching our handler. Since the check is genuinely a defensive invariant on `self.element`, I drive `ContentHandler` directly with mismatched start/end names — happy to switch to a parser-driven approach if you prefer.

2. **TabIO id check**: `_clean` already replaces `\n` and `\r` with spaces before the `if "\t" in title or "\n" in title or "\r" in title` runs, so only the tab branch of the id check is reachable from a normal `SeqIO.write` call. I left the existing `\n`/`\r` clauses in place (defensive in case `_clean` ever changes) and noted it in the test docstring rather than narrowing the production check. Let me know if you'd rather I drop the unreachable clauses from `TabIO.py` itself.

All 15 tests in the four touched files pass locally, `black --target-version py312` reports no diffs.

Abdeltoto:
Hi @peterjc — following up on your Apr 27 question: commit \3cfd0df\ adds \ssertRaises\ coverage for all four new \ValueError\ paths (Enzyme PR line, ScanProsite handler mismatch via direct handler drive, TabIO via \TabWriter.to_string\, AlignACE malformed motif line). Details are in the PR reply thread.

Could you please give this another look when you have a moment? If the earlier \CHANGES_REQUESTED\ is stale relative to the new commits, a fresh review (or dismiss + re-review) would unblock the merge queue. Thank you again for the thorough feedback.

Tests/test_SeqIO_TabIO.py
```diff
@@ -0,0 +1,43 @@
...
+class TestTabIOInvalidInput(unittest.TestCase):
+    """Tab format rejects records whose id or sequence contain reserved chars."""
+
+    def test_tab_id_with_tab_raises(self):
+        """Reject tab-format write when record.id contains a tab character.
+
+        Note: ``_clean`` already replaces ``\\n`` and ``\\r`` in the id with
+        spaces before this check runs, so only the tab branch of the id check
+        is reachable from a normal write call.
+        """
```
peterjc, line 23: See the pre-commit warning from the slashes:

```
D301 Use `r"""` if any backslashes in a docstring
  --> Tests/test_SeqIO_TabIO.py:18:9
   |
17 |       def test_tab_id_with_tab_raises(self):
18 | /         """Reject tab-format write when record.id contains a tab character.
19 | |
20 | |         Note: ``_clean`` already replaces ``\\n`` and ``\\r`` in the id with
21 | |         spaces before this check runs, so only the tab branch of the id check
22 | |         is reachable from a normal write call.
23 | |         """
   | |___________^
24 |           record = SeqRecord(Seq("ACGT"), id="bad\tid", description="")
25 |           with self.assertRaisesRegex(ValueError, "Record id contains"):
   |
help: Add `r` prefix
```

peterjc:
Thank you. This is a guess, but there's something funny about your recent comments - are you using a coding agent LLM AI tool? If so, please see the ongoing policy discussion on our mailing list: https://mailman.open-bio.org/pipermail/biopython/2026-April/017113.html

Abdeltoto:
Hi Peter, sorry about the noise. I pushed a fix for the D301 docstring warning (r""" on that TabIO test). Thanks again for the review / the mailing-list link — I’ve read it and I’ll keep updates here tighter and clearly mine going forward.

peterjc:
@Abdeltoto we've not merged #5241 yet, but we do have agreement that any AI usage should at least be declared. So once again, was this generated with an AI tools?

PR #5176, opened 2026-03-14
Add AFDBList class for PDBList-like AlphaFold downloads (#3990)

### LINKED ISSUE #3990: Access to Alphafold database of 3D modules in a PDBList-like interface
This has been somewhat raised in #3853, but it would be potentially interesting to add a small interface to bulk (or not) download Alphafold models and build a structured database that users could keep updated, This module could be based off `PDBList.py` and could initially be limited to downloading structures by species, accession(s), but later on potentially allow for more complex queries. I could see this being a nice pipeline piece combined with our BLAST and PDB modules.

Abdeltoto:
Hi,

I worked on issue #3990. The idea was to add a PDBList-like interface for the AlphaFold database, so users can download predicted structures by UniProt ID in a familiar way.

I added an `AFDBList` class that mirrors the main `PDBList` workflow:

- `__init__(server, pdb, verbose)` – same pattern as PDBList
- `retrieve_pdb_file(uniprot_id, pdir, file_format, overwrite)` – same idea as PDBList’s `retrieve_pdb_file`, but for UniProt accessions

Supported formats are mmCIF (default) and PDB. When there are multiple models for a single accession (e.g. isoforms), the canonical one is preferred by default, with an optional `model_index` parameter to choose another.

I also switched the API base URL from alphafold.com to alphafold.ebi.ac.uk (the official EBI server) and added the missing `AFDBList` export to `Bio.PDB`.

Example usage::

    from Bio.PDB import AFDBList

    afdb = AFDBList()
    path = afdb.retrieve_pdb_file("P00520")
    # or as PDB format
    path = afdb.retrieve_pdb_file("P00520", file_format="pdb")

Closes #3990

### DIFF
--- a/Bio/PDB/__init__.py
+++ b/Bio/PDB/__init__.py
@@ -67,6 +67,9 @@ from .PDBIO import Select
 
 # Download from the PDB
 from .PDBList import PDBList
+
+# Download from the AlphaFold database
+from .alphafold_db import AFDBList
 from .PDBMLParser import PDBMLParser
 from .PDBParser import PDBParser
 from .Polypeptide import CaPPBuilder
--- a/Bio/PDB/alphafold_db.py
+++ b/Bio/PDB/alphafold_db.py
@@ -140,7 +140,7 @@ class AFDBList:
     def __init__(
         self,
         server: str = "https://alphafold.ebi.ac.uk",
-        pdb: Optional[str] = None,
+        pdb: str | None = None,
         verbose: bool = True,
     ):
         """Set up the download interface.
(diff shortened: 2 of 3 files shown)


Abdeltoto (commented):
perfect

gabufle:
Tested on Windows 11 with Python 3.13.2. 
Retrieve_pdb_file('P00520') successfully downloaded AF-P00520-F1-model_v6.cif. The basic functionality works.
I did have one note, though: should an invalid accession raise an exception rather than returning None, so pipelines can catch it explicitly?

peterjc:
You have not followed the PR template which has some vital questions.

Also, was this was AI generated? If so we would close it as per the policy since #5229. See also the proposed even stricter no AI policy #5241.

PR #5183, opened 2026-03-30
Declare Python 3.10 support deprecated.

peterjc:
No obvious way to add a warning with `pyproject.toml` like we used to do with `setup.py`? CC @xelandernt

### DIFF
--- a/DEPRECATED.rst
+++ b/DEPRECATED.rst
@@ -7,6 +7,10 @@ followed by code (modules, methods, functions).
 Python
 ======
 
+Python 3.10
+-----------
+First supported in release 1.80. Support deprecated as of Release 1.88.
+
 Python 3.9
 ----------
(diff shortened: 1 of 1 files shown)


xelandernt:
@peterjc I don't think there is a way to raise a DeprecationWarning on install with `pyproject.toml`.

You could but it into `__init__.py` but that will raise the warning every time the package is imported.

However, I would argue since python 3.10 reaches EOL this year, users should expect deprecation. Furthermore, if users continue to use python 3.10 and have no upper bound on their `biopython` version, they will simply get the latest version that did support python 3.10. Their application can never break until they upgrade their python version, and when they do that they should test everything anyway.

peterjc:
Thanks, I agree while it would be a nice to have a warning for our users, it is hardly going to be a surprise given this is driven by the Python 3.10 EOL.

mdehoon (approved):

PR #5194, opened 2026-04-14
Support minimal MEME files without `Background letter frequencies` line.

ghuls:
Support minimal MEME files without `Background letter frequencies` line as this line is not required by the minimal MEME format (and background frequencies default to 0.25 for each nucleotide).

Minimal MEME files available on https://www.factorbook.org/downloads do not contain a `Background letter frequencies` line.

Some `for line in handle` invocations are rewritten to `for line in iter(handle.readline, "")` so handle.tell() can be used and the file position can be rewinded to the position before the `_read_background` call in case no `Background letter frequencies` line was found.
See: https://github.com/python/cpython/issues/81217

Closes #...


Bio/motifs/minimal.py
```diff
@@ -117,15 +119,15 @@ def _read_background(record, handle):
                 )
             break
     else:
-        raise ValueError(
-            "Improper input file. File should contain a line starting background frequencies."
-        )
+        # If no background letter frequencies line was found,
+        # rewind the input file to the old position.
+        handle.seek(handle_pos)
```
peterjc: There is no reason to assume you can use `handle.seek()` is there? This could be `sys.stdin` couldn't it?

ghuls: True, it wouldn't work for stdin. Force pushed a version that should work with sys.stdin too.

peterjc:
@mdehoon this looks better now - your thoughts?

PR #5203, opened 2026-04-24
Warn instead of error for invalid strand type in LOCUS line (issue #4…

### LINKED ISSUE #4846: Biopython too sensitive to spacing differences.
### Setup

system as follows:

```python
```

```
3.9.19 | packaged by conda-forge | (main, Mar 20 2024, 12:55:20) 
[Clang 16.0.6 ]
CPython
macOS-14.1.2-arm64-arm-64bit
1.83
```

### Expected behaviour

SeqIO.read works on both files.

### Actual behaviour

```
>>> from Bio import SeqIO
>>> SeqIO.read("./pAV-CAG-GFP-correct.gb", "gb")
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/File.py", line 72, in as_handle
    with open(handleish, mode, **kwargs) as fp:
TypeError: expected str, bytes or os.PathLike object, not TextIOWrapper

During handling of the above exception, another exception occurred:

  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/SeqIO/__init__.py", line 662, in read
    record = next(iterator)
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/SeqIO/Interfaces.py", line 85, in __next__
    return next(self.records)
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/GenBank/Scanner.py", line 515, in parse_records
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/GenBank/Scanner.py", line 498, in parse
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/GenBank/Scanner.py", line 464, in feed
    self._feed_first_line(consumer, self.line)
  File "/Users/pnguyen/miniconda3/envs/biopython/lib/python3.9/site-packages/Bio/GenBank/Scanner.py", line 1404, in _feed_first_line
    raise ValueError(
ValueError: LOCUS line does not have valid strand type (Single stranded, ...):
LOCUS       pAV-CAG-GFP-corr        2394 bp DNA                     01-JAN-1980

>>> SeqIO.read("./pAV-CAG-GFP-correct -- PN.gb", "gb")
SeqRecord(seq=Seq('CCTGCAGGCAGCTGCGCGCTCGCTCGCTCACTGAGGCCGCCCGGGCGTCGGGCG...AGG'), id='.', name='pAV-CAG-GFP-corr', description='synthetic circular DNA', dbxrefs=[])
```

### Steps to reproduce

```
(biopython) Peters-MacBook-Pro:Downloads pnguyen$ diff pAV-CAG-GFP-correct.gb "pAV-CAG-GFP-correct -- PN.gb"
1c1
< LOCUS       pAV-CAG-GFP-corr        2394 bp DNA                     01-JAN-1980
---
> LOCUS       pAV-CAG-GFP-corr        2394 bp    DNA                  01-JAN-1980
```

CameronCat:
…846)

When the LOCUS line does not have a valid strand type at the expected position (e.g. files exported by SnapGene), fall back to whitespace-based parsing and issue a BiopythonParserWarning instead of raising a ValueError. The fallback recovers locus name, size, molecule type, topology, division, and date when present.

Fixes #4846

Closes #4846

Adds a whitespace-based fallback to the GenBank LOCUS line parser for the strand-type column check. Previously this raised ValueError on SnapGene-exported files; now it warns and continues, recovering name, size, molecule type, topology, division, and date. Tests added in Tests/test_GenBank_locus_spacing.py.


mdehoon:
I am not a big fan of warnings, as they are easy to miss, and get generated only once in a session.

Maybe `SeqIO.parse` should have an optional argument specifying whether minor problems should raise an exception (the default), issue a warning, or just be ignored. For example, the `parse` function in `Bio.motifs` has such an argument.

CameronCat:
Thanks for the review!
That's a fair concern about warnings being easy to miss.
I just wanted to ask a few questions:
Do you want the strictness control at the SeqIO level, or threaded down into the scanner? Following the Bio.motifs.parse pattern would put it on parse() directly. I am happy to do that if it's the preferred approach.

For backward compatibility, should the default remain raising ValueError, or shift to "warn"? I'd would keep "raise" as the default so we don't silently change behavior for current users, but I would be happy to hear what you have to say.

mdehoon:
> Do you want the strictness control at the SeqIO level, or threaded down into the scanner? Following the Bio.motifs.parse pattern would put it on parse() directly. I am happy to do that if it's the preferred approach.

Yes I agree.

> For backward compatibility, should the default remain raising ValueError, or shift to "warn"? I'd would keep "raise" as the default so we don't silently change behavior for current users, but I would be happy to hear what you have to say.

Yes I agree with this also. But let's hear what @peterjc  thinks, as he is the main maintainer of `Bio.SeqIO`.

peterjc:
I think @mdehoon and I have rather different views of Python's built in warnings system, and how to use it. I am a big fan of warnings and want them on by default - you can silence them as needed: https://docs.python.org/3/library/warnings.html

Also relevant is the example of our PBD code with a permissive toggle:

```
         - PERMISSIVE - Evaluated as a Boolean. If false, exceptions in
           constructing the SMCRA data structure are fatal. If true (DEFAULT),
           the exceptions are caught, but some residues or atoms will be missing.
           THESE EXCEPTIONS ARE DUE TO PROBLEMS IN THE PDB FILE!.
```

That's permissive by default (sometimes using `PDBConstructionWarning` but for other issues silently passed), but permissive=false meaning strict raises exceptions.

But we don't need to fight that battle here. The NCBI themselves have signalled they are moving away from a strict column based LOCUS line, and we already deal with a lot of border line cases in the format.

I have no objection in principle to solving #4846 as suggested (downgrade the error to a warning). My concern here is adding even more special casing to one of the messiest bits of the GenBank parser to cope with malformed output from SnapGene (again).

I would rather not make guesses (Zen of Python), and lean to a minimal fall back - eg https://github.com/biopython/biopython/blob/7a9c76cce8c6a58db791be2b12a135af210cedf2/Bio/GenBank/Scanner.py#L1559

CameronCat:
Thanks, that makes sense. I'll simplify the fallback to just extract locus name and size and drop the heuristic guessing for molecule type, topology, division, and date. I'll update the tests to match.

Bio/GenBank/Scanner.py
```diff
@@ -1404,9 +1405,20 @@ def _feed_first_line(self, consumer, line):
                     "expected position:\n" + line
                 )
             if line[44:47] not in ["   ", "ss-", "ds-", "ms-"]:
+                warnings.warn(
```
peterjc: I'd move the warning into the if-clause otherwise you can get warning and error.

Tests/test_GenBank_locus_spacing.py
```diff
@@ -0,0 +1,60 @@
+"""Tests for GenBank LOCUS line whitespace fallback parsing (issue #4846)."""
```
peterjc: I think I'd just add this to `test_GenBank.py` (and since that is so large now, it might be worth then thinking about breaking it up - for instance all the LOCUS line testing might deserve it's own `test_GenBank_locus.py` file?).

CameronCat: I moved the test case into LineOneTests.test_topology_genbank in test_GenBank.py and removed the standalone file. I left the larger split (LOCUS tests into their own test_GenBank_locus.py) out of this PR to keep it focused, but I am happy to open a separate PR for that reorganization if it would be useful.

peterjc:
Merged, thank you!

PR #5204, opened 2026-04-27
Fix FastaBlastIterator typo that mis-pairs IDs with sequences

RitataLU:
In FastaBlastIterator.__next__, the line that should advance self._line to the next record's title was misspelled as self_line, creating a local variable instead of updating instance state. As a result, every record after the first was yielded with the first record's id, name, and description while the inner loop kept consuming subsequent '>' lines from the stream — silent data corruption with no error raised.

Also make zero-record cases loud rather than silent: SimpleFastaParser now emits a BiopythonParserWarning, and FastaBlastIterator and FastaPearsonIterator raise ValueError, when input is non-empty but contains no '>' line. Empty streams continue to yield zero records.

Add Tests/Fasta/multi_blast.pro and four new tests covering the multi-record fasta-blast path and the warn/raise behavior. The existing tests only used single-record fixtures, which is why the typo had not been caught.

Closes #...

### DIFF
--- a/Bio/SeqIO/FastaIO.py
+++ b/Bio/SeqIO/FastaIO.py
@@ -16,6 +16,7 @@ You are expected to use this module via the Bio.SeqIO functions.
 from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
 from Bio import BiopythonDeprecationWarning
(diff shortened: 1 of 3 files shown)


mdehoon:
This seems like a good bug to fix..

PR #5207, opened 2026-05-06
Document how to modify feature locations

### LINKED ISSUE #901: order of compound locations written as EMBL is reversed
When reading an EMBL file with annotations and exporting a record from there back to EMBL format, the order of the elements of reverse strand compound locations is reversed, e.g.:

```
FT   CDS             complement(join(92561..93210,84225..92550))
```
instead of 
```
FT   CDS             complement(join(84225..92550,92561..93210))
```
as it was in the input file. This confuses downstream processing tools such as ENA's pre-submission validator, which now flags this up as having a too short intron (<10bp) -- this is wrong here but goes away once I manually switch the order back to ascending coordinates. It might be helpful here to have the exporter preserve the original order, or allow for different sorting.

omniscoder:
Fixes #901.

This adds a short tutorial section explaining that `SimpleLocation`/`FeatureLocation` and `CompoundLocation` expose read-only `start` and `end` properties, and shows how to rebuild locations when coordinates need to change.

It also documents that compound parts should be kept in biological order, which matters in particular for reverse-strand features.

Validation:
- `git diff --check`
- `chapter_seq_annot` doctests passed via the `Tests/test_Tutorial.py` harness (13 doctests)
- full `test_Tutorial.py` in this environment still has an unrelated pre-existing failure in `chapter_uniprot`

### DIFF
--- a/Doc/Tutorial/chapter_seq_annot.rst
+++ b/Doc/Tutorial/chapter_seq_annot.rst
@@ -611,6 +611,55 @@ Biopython. It has been designed so that dealing with fuzziness is not
 that much more complicated than dealing with exact positions, and
 hopefully you find that true!
 
+Modifying locations
+^^^^^^^^^^^^^^^^^^^
+
+The ``start`` and ``end`` attributes of ``SimpleLocation`` and
+``CompoundLocation`` are read-only. Older code may still call
+``SimpleLocation`` by its alias ``FeatureLocation``. For compound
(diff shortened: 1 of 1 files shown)


peterjc:
Please read the CONTRIBUTING file, we cannot merge this as is.

Also it doesn't seem to address the actual issue. Is this just some AI generated text? See this the on going mailing list discussions from https://mailman.open-bio.org/pipermail/biopython/2026-April/017113.html

PR #5209, opened 2026-05-06
Add missing SeqIO iterator context manager docstrings

omniscoder:
Refs #2116.

This is a one-file D105 cleanup for `Bio/SeqIO/Interfaces.py`. It adds the missing docstrings on `SequenceIterator.__enter__` and `SequenceIterator.__exit__` and makes no behavioral changes.

Validation:
- `git diff --check`
- `./.venv/bin/python -m flake8 --isolated --select D1 Bio/SeqIO/Interfaces.py`

### DIFF
--- a/Bio/SeqIO/Interfaces.py
+++ b/Bio/SeqIO/Interfaces.py
@@ -122,9 +122,11 @@ class SequenceIterator(ABC, Generic[AnyStr]):
         return self
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


peterjc:
Please read the CONTRIBUTING file, we cannot merge this as is. Also for this issue it would be better to do a batch of files at once rather than a PR for a one line fix.

PR #5211, opened 2026-05-11
Tests: add tests for six_frame_translations and nt_search in Bio.SeqUtils

amitsjc:
Both functions had no test coverage. This PR adds test_six_frame_translations and test_nt_search to Tests/test_SeqUtils.py.

test_six_frame_translations covers: return type, header fields (GC_Frame, nucleotide counts, length, GC%), start/stop codon translation, abbreviated vs full sequence display, the docstring RNA example, and alternate genetic codes.

test_nt_search covers: single and multiple exact matches, no-match case, ambiguous IUPAC codes N/R/Y, and position accuracy.

All 13 tests in test_SeqUtils.py pass.

### DIFF
--- a/Tests/test_SeqUtils.py
+++ b/Tests/test_SeqUtils.py
@@ -15,8 +15,10 @@ from Bio.SeqRecord import SeqRecord
 from Bio.SeqUtils import CodonAdaptationIndex
 from Bio.SeqUtils import gc_fraction
 from Bio.SeqUtils import GC_skew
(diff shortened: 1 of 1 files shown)


peterjc:
You haven't used the PR template. We can't accept any PRs without the relevant information - see the contributing file. Note on the mailing list we're also actively discussing rejecting LLM/AI contributions.

Also there is a little test coverage of `six_frame_translations` as a doctest.

PR #5213, opened 2026-05-15
Fix degenerate consensus for fully ambiguous columns

charliesheh:
Summary
Fix `motifs.degenerate_consensus` handling for fully ambiguous columns.
Previously, a motif column derived from ambiguous bases such as `N` could produce a partially ambiguous IUPAC code such as `V` instead of preserving full ambiguity as `N`.

Changes
- Updated `degenerate_consensus` logic to avoid selecting a three-base degenerate code when the top three nucleotide counts sum to zero
- Added regression coverage for fully ambiguous motif columns

Testing
- `python -m pytest test_motifs.py -k degenerate_consensus`
- `python -m pytest test_motifs.py`

### DIFF
--- a/Bio/motifs/matrix.py
+++ b/Bio/motifs/matrix.py
@@ -207,7 +207,7 @@ class GenericPositionMatrix(dict):
                 key = nucleotides[0]
             elif 4 * sum(counts[:2]) > 3 * sum(counts):
                 key = "".join(sorted(nucleotides[:2]))
-            elif counts[3] == 0:
+            elif counts[3] == 0 and sum(counts[:3]) > 0:
                 key = "".join(sorted(nucleotides[:3]))
             else:
                 key = "ACGT"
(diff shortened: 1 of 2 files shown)


peterjc:
You haven't used the PR template. We can't accept any PRs without the relevant information - see the contributing file. Note on the mailing list we're also actively discussing rejecting LLM/AI contributions.

charliesheh:
Thank you for the feedback. I apologize for missing the PR template. I’ve updated the PR description with the requested information from the contribution guidelines.

I reproduced the reported behavior locally, traced the issue to the degenerate consensus handling for fully ambiguous columns, implemented the fix, and added a regression test for the edge case.

peterjc:
This looks very much like AI and you missed the point. This PR cannot be accepted as is (even without us explicitly banning AI).

PR #5231, opened 2026-06-02
fix(PDB): extend obsolete file handling to mmCIF and XML formats

### LINKED ISSUE #3988: Extend PDBList obsolete handling to mmCIF and mmtf formats as well as assemblies

mayank-dev-15:
## Description

Previously, update_pdb() only moved obsolete PDB-formatted files to the obsolete folder. Now also moves mmCIF (.cif) and XML (.xml) formatted obsolete files, matching the multi-format support already present in retrieve_pdb_file().

Fixes biopython/biopython#3988

## Changes
- Updated update_pdb() to loop over (pdb, mmCif, xml) formats when moving obsolete files
- Added verbose logging for failed moves
- Changed flat_tree and non-flat_tree paths to handle all formats

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -385,32 +385,34 @@ class PDBList:
                 # something has gone wrong.
 
         # Move the obsolete files to a special folder
-        # NOTE: This should be updated to handle multiple file types and
-        # assemblies. As of now, it only looks for PDB-formatted files.
-        # Using pathlib will be probably the best approach here, to build
-        # and index of which files we have efficiently (or glob them).
+        # NOTE: This handles multiple file types (pdb, mmCif, xml, mmtf, bundle).
         for pdb_code in obsolete:
-            if self.flat_tree:
-                old_file = os.path.join(self.local_pdb, f"pdb{pdb_code}.{file_format}")
-                new_dir = self.obsolete_pdb
-            else:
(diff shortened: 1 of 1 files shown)


GeneratedScript-RL:
@mayank-dev-15 You didnt use the PR format so this will likely be rejected. Additionally you did not add yourself to `CONTRIB.rst`.

Additionally, did you read `CONTRIBUTING.rst`? I see you have pre-commit.ci errors which you wouldn't have missed in the first place if you did have pre commit installed.

peterjc:
Is this AI generated? See #5229

mayank-dev-15:
No

mayank-dev-15:
Why do i use AI but there are some tools on my VS codium
And i have to build my base For GSOC 2027

PR #5244, opened 2026-06-23
Calculate TLEN when writing SAM alignments

### LINKED ISSUE #5232: TLEN calculation when writing sam file
When writing sam files, the TLEN on each alignment is set to 0, this is hard coded in [sam.py](https://github.com/biopython/biopython/blob/master/Bio/Align/sam.py#L245)

Could this be calculated from the coordinates or CIGAR string and written into the sam file? It would be very helpful for downstream processing.

GeneratedScript-RL:

Resolves the incorrect Template Length (TLEN) calculation when writing SAM files.

Closes #5232

### DIFF
--- a/Bio/Align/sam.py
+++ b/Bio/Align/sam.py
@@ -114,7 +114,7 @@ class AlignmentWriter(interfaces.AlignmentWriter):
                 line = "\t".join(fields) + "\n"
                 stream.write(line)
 
-    def format_alignment(self, alignment, md=None):
+    def format_alignment(self, alignment, md=None, tlen=None):
         """Return a string with a single alignment formatted as one SAM line."""
         if not isinstance(alignment, Alignment):
             raise TypeError("Expected an Alignment object")
@@ -242,7 +242,11 @@ class AlignmentWriter(interfaces.AlignmentWriter):
             pnext = 0
         else:
             pnext += 1  # 1-based coordinates
-        tLen = 0
+        if tlen is None:
+            try:
+                tlen = alignment.tlen
+            except AttributeError:
+                tlen = 0
         fields = [
             qName,
             str(flag),
@@ -252,7 +256,7 @@ class AlignmentWriter(interfaces.AlignmentWriter):
             cigar,
             rnext,
             str(pnext),
-            str(tLen),
+            str(tlen),
             query,
             qual,
         ]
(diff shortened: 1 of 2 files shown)


Bio/Align/sam.py
```diff
@@ -368,6 +372,83 @@ def format_alignment(self, alignment, md=None):
         line = "\t".join(fields) + "\n"
         return line
 
+    def write_alignments(self, stream, alignments):
+        """Write alignments to the output file, and return the number of alignments."""
+        if not isinstance(alignments, (list, tuple)):
+            return super().write_alignments(stream, alignments)
+        tlens = self._calculate_tlens(alignments)
```
mdehoon, line 379: Why calculate the `tlen` values in advance, instead of as part of the `for` loop starting on the next line?

mdehoon:
Did you use AI for this PR?

Bio/Align/sam.py
GeneratedScript-RL: From my understanding/research TLEN depends on the mate’s coordinates so it cannot always be computed from the current alignment alone. I calculate the values up front for list or tuple inputs to build mate pairs once and keep the write loop simple while preserving streaming behavior for iterators

GeneratedScript-RL:
No

Bio/Align/sam.py
peterjc: TLEN is inferred Template LENgth (insert size), so yes for paired end reads you need to look beyond the current read in isolation.

GeneratedScript-RL:
Happy to make any changes if needed.

Bio/Align/sam.py
mdehoon: > I calculate the values up front for list or tuple inputs to build mate pairs once and keep the write loop simple while preserving streaming behavior for iterators

There is also the `.rewind` method, if you need to loop over the alignments more than once, and the alignments are not a list or tuple.

Still, for large files, the `records` in `_calculate_tlens` could become huge, and may not fit in memory. Another approach would be to sort the alignments first by name. Then you just need to keep one record in memory. I haven't checked, but I guess this is how samtools would do it. In any case, if you require the alignments to be sorted (or at least grouped) by name, then your code can be applied to very large files also.

GeneratedScript-RL:
Hello! Any updates?

mdehoon:
@GeneratedScript-RL  Did you see my comments about sorting the alignments?

GeneratedScript-RL:
image

PR #5246, opened 2026-06-25
Guard Tree.collapse() against children without branch lengths

### LINKED ISSUE #5245: Tree.collapse() raises type error on trees without branch lengths
Reproducible on the newest version in master

## Expected

Tree.collapse() and collapse_all() should remove the target node and relink its children without raising an error, even when the tree has no branch lengths (like a cladogram)

The sibling method prune() already handles a missing branch length so collapse() should in theory behave similarly

## Actual

```
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'int'
```
In BaseTree.py collapse() adds the removed node's branch length to each of its children using child.branch_length += extra_length

When child has its branch length set to None (the normal case for cladograms and any tree parsed without branch lengths), the expression fails because None cannot be added to a number. The sibling method prune() guards the same operation with a check for if child.branch_length is not None, but collapse() does not

## How to reproduce

A cladogram with no branch lengths:
```python
from io import StringIO
from Bio import Phylo
tree = Phylo.read(StringIO("(((A,B),C),D);"), "newick")
tree.collapse_all()
```
The single-node form, where only some children have a branch length:
```python
tree = Phylo.read(StringIO("((A,B):1.0,C:2.0);"), "newick")
tree.collapse(tree.clade[0])
```
Both raise TypeError

## Versions
```
3.13.0 (v3.13.0:60403a5409f, Oct  7 2024, 00:37:40) [Clang 15.0.0 (clang-1500.3.9.4)]
CPython
macOS-26.5.1-arm64-arm-64bit-Mach-O
1.88.dev0
```

## Note on related work
I looked at some other PRs that touch this file (#4634 #5243) and neither fixes this crash so I am reporting it separately

I have a fix ready that mirrors the guard already used in prune() and I am happy to open a pull request :)

lizapoz:
- [ x] I hereby agree to dual licence this and any previous contributions under both

- [ x] I have read the ``CONTRIBUTING.rst`` file, have run ``pre-commit``

- [x ] I have added my name to the alphabetical contributors listings in the files

Closes #5245

Tree.collapse() added the removed node's branch length to each child with child.branch_length += extra_length, which raised TypeError when a child had branch_length None (cladograms / trees parsed without branch lengths). Mirror the guard already used in the sibling prune() method.

### DIFF
--- a/Bio/Phylo/BaseTree.py
+++ b/Bio/Phylo/BaseTree.py
@@ -593,7 +593,8 @@ class TreeMixin:
         popped = parent.clades.pop(parent.clades.index(path[-1]))
         extra_length = popped.branch_length or 0
         for child in popped:
-            child.branch_length += extra_length
+            if child.branch_length is not None:
+                child.branch_length += extra_length
         parent.clades.extend(popped.clades)
         return parent
 
--- a/Tests/test_Phylo.py
+++ b/Tests/test_Phylo.py
@@ -481,6 +481,12 @@ class MixinTests(unittest.TestCase):
         for clade in d2:
             self.assertAlmostEqual(d1[clade], d2[clade])
 
+    def test_collapse_without_branch_lengths(self):
+        tree = Phylo.read(StringIO("(((A,B),C),D);"), "newick")
+        tree.collapse_all()
+        self.assertEqual(len(tree.get_terminals()), 4)
+        self.assertEqual(len(tree.get_nonterminals()), 1)
+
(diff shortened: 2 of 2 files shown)


etal (approved):
LGTM, thanks for the fix!

peterjc:
I'm ignoring the CodeCov issue on CircleCI and merging this anyway. We should open an issue for that (I was initially hoping it was a temporary issue).

Thank you both.

PR #5247, opened 2026-06-25
Fix PDBList URL construction for non-RCSB mirrors

### LINKED ISSUE #3987: Fix PDBList access to servers other than RCSB

anshjaiswal12:
## Summary

- Move the `/pub/pdb` path prefix into the configurable `server` root so mirror-specific paths work without duplicating `/pub/pdb` in every URL builder.
- Default server is now `https://files.wwpdb.org/pub/pdb`; EBI users can pass `https://ftp.ebi.ac.uk/pub/databases/pdb`.
- Add offline tests that mock downloads and verify RCSB, EBI, and PDBj URLs.

Fixes #3987

## Test plan

- [x] `python -m unittest test_PDB_PDBList.TestPDBListURLConstruction`
- [x] `python -m unittest test_PDB_PDBList.TestPBDListGetList.test_default_server_includes_pdb_path_prefix`
- [x] `python -m unittest test_PDB_PDBList.TestPBDListGetList.test_mirror_server_urls`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Bio/PDB/PDBList.py
+++ b/Bio/PDB/PDBList.py
@@ -86,9 +86,12 @@ class PDBList:
     http://www.pdb.org/.
     """
 
+    # Default wwPDB mirror; EBI uses https://ftp.ebi.ac.uk/pub/databases/pdb instead.
+    DEFAULT_SERVER = "https://files.wwpdb.org/pub/pdb"
+
     def __init__(
         self,
-        server="https://files.wwpdb.org",
+        server=None,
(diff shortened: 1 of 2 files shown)


peterjc:
Thank you for declaring you used AI.

See policy change in #5229 explicitly rejecting AI contributions for the 'good first issue' PRs. See also #5241 for a potential blanket no-AI policy.

PR #5251, opened 2026-06-25
Feat: Add optional title parameter to plotting function

### LINKED ISSUE #5250: Add title parameter to Phylo plot
I think it would be beneficial (and easy) to add an optional `title` parameter to `Phylo.draw`.
Current version:
```py
Phylo.draw(tree, do_show=False)
plt.title("My Plot Title")
plt.show()
```

Updated version:
```py
Phylo.draw(tree, title="My Plot Title")
```

dawnandrew100:
* Feat: Add optional title parameter to plotting function

* Add name to NEWS.rst contributor list

* Add name to contrib.rst contributor list

Closes #5250

### DIFF
--- a/Bio/Phylo/_utils.py
+++ b/Bio/Phylo/_utils.py
@@ -286,6 +286,7 @@ def draw(
     show_confidence=True,
     # For power users
     axes=None,
+    title=None,
     branch_labels=None,
     label_colors=None,
     *args,
@@ -331,6 +332,8 @@ def draw(
         axes : matplotlib/pylab axes
             If a valid matplotlib.axes.Axes instance, the phylogram is plotted
             in that Axes. By default (None), a new figure is created.
+        title : str
+            A string specifying an optional title for the plot.
         branch_labels : dict or callable
             A mapping of each clade to the label that will be shown along the
             branch leading to it. By default this is the confidence value(s) of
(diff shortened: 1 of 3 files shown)


mdehoon:
Wouldn't it be possible to just call matplotlib's title function after generating the figure?

dawnandrew100:
The main purpose of this addition would just be for succinctness for the end user. But yes, it is also possible to call the title function after generating the figure.

peterjc:
Then maybe just extend any suitable example in our docs to do that, rather than expanding the function?

dawnandrew100:
Expanding an example instead of the function definitely makes sense and I can do that if I happen to find one.

A follow up question for you (and/or the team?) is would there any circumstance where making this minor, optional change to the function would be okay? Like adding this new argument to the end of the parameters in case there are people who don't use keyword arguments and use positional arguments instead? This way this addition wouldn't affect any existing workflows?

Basically, I'm just trying to gather if there is a scenario where this minor change would be okay, or if making a (admittedly insignificant) change to this function is a no go in general.

Bio/Phylo/_utils.py
```diff
@@ -286,6 +286,7 @@ def draw(
     show_confidence=True,
     # For power users
     axes=None,
+    title=None,
```
peterjc: This should be moved down to be the final optional argument for backward compatibility (as the user may be giving all the old arguments positionally).

peterjc:
We take backward compatibility seriously (as does numpy for example), so in practise when changing existing methods/functions usually all you can do is add new optional arguments, or add new meanings to existing arguments.

dawnandrew100:
Makes sense
I realised that probably was important after I had already opened the PR, apologies!

PR #5252, opened 2026-06-27
Fix 5235 - Endian issue with binary CIF 

### LINKED ISSUE #5235: Endian issue with binary CIF on ppc64 and sparc64 arch
On ppc64, Py_buffer.format is reported as "<H" in _bcif_helper.integer_unpack(). The code assumes the first character is the type code (H, h, B, b) and does not handle valid PEP-3118 endian prefixes. This causes test_PDB_binary_cif to fail. Additionally, the function returns Py_None while an exception is set, leading to a subsequent SystemError.
https://buildd.debian.org/status/fetch.php?pkg=python-biopython&arch=ppc64&ver=1.87%2Bdfsg-1&stamp=1778686569&raw=0

Will-Tyler:
## Acknowledgements

## Description

This pull request attempts to fix #5235.

### DIFF
--- a/Bio/PDB/bcifhelpermodule.c
+++ b/Bio/PDB/bcifhelpermodule.c
@@ -2,6 +2,7 @@
 #include <Python.h>
 #include <math.h>
 #include <stdint.h>
+#include <string.h>
 
 static void
 integer_unpack_u8(Py_buffer *in_view, Py_buffer *out_view)
@@ -153,7 +154,8 @@ integer_unpack(PyObject *self, PyObject *args)
         goto exit;
     }
 
-    const char format = in_view.format[0];
+    const size_t l = strlen(in_view.format);
+    const char format = in_view.format[l - 1];
 
     if (format == 'B') {
         integer_unpack_u8(&in_view, &out_view);
--- a/Bio/PDB/binary_cif.py
+++ b/Bio/PDB/binary_cif.py
@@ -117,6 +117,9 @@ def _integer_packing_decoder(column):
         dtype = np.dtype("<i4")
 
     data = column["data"]["data"]
+    # Convert to native endianness
+    data = data.astype(data.dtype.newbyteorder("="))
+
     assert byte_count == data.dtype.itemsize
     assert np.issubdtype(data.dtype, np.unsignedinteger) == is_unsigned
     decoded_data = np.empty((src_size,), dtype)


peterjc:
@schoekek are you able to test this please?

schoekek:
The patch works at this point, but there is a new problem:

```text
FAIL: test_get_structure (test_PDB_binary_cif.TestBinaryCIFParser.test_get_structure)

  ...
  File ".../Bio/PDB/binary_cif.py", line 95, in _run_length_decoder
    assert len(decoded_data) == encoding["srcSize"]
AssertionError
```

A tutorial doctest also fails as a consequence:

```
test_Tutorial.TutorialTestCase.test_doctests
```

because the BinaryCIF parser is unable to read the structure correctly.

### Observations

The assertion itself appears to be a consequence rather than the root cause. The failure occurs during the decoding chain:

```
ByteArray
 -> IntegerPacking
 -> RunLength
```

`_run_length_decoder()` itself only performs

```python
decoded_data = np.repeat(data[::2].astype(dtype), data[1::2])
```

which looks endian-independent.

The more likely source seems to be the C extension used by the IntegerPacking decoder:

```
Bio/PDB/bcifhelpermodule.c
```

In particular, `integer_unpack()` and the helper functions directly cast the input buffer to native integer types, e.g.

```c
uint16_t *in_data = in_view->buf;
int16_t *in_data = in_view->buf;
```

while the Python code explicitly creates NumPy arrays with little-endian dtypes:

```python
np.dtype("<i2")
np.dtype("<u2")
np.dtype("<i4")
np.dtype("<u4")
```

This may cause incorrect interpretation of the buffer contents on big-endian systems.

I have not yet verified that this is the exact root cause, but it seems consistent with the observed behaviour:

* works on little-endian architectures,
* fails on big-endian architectures,
* failure occurs after IntegerPacking during RunLength decoding.

Here are the two buildd runs in the Debian environment:

ppc64: https://buildd.debian.org/status/fetch.php?pkg=python-biopython&arch=ppc64&ver=1.87%2Bdfsg-2&stamp=1782747833&raw=0
sparc64: https://buildd.debian.org/status/fetch.php?pkg=python-biopython&arch=sparc64&ver=1.87%2Bdfsg-2&stamp=1782748735&raw=0

peterjc:
That sounds like a question for Will - I would have expected `Bio/PDB/bcifhelpermodule.c` to decode from the on-disk binary layout (which I assumed is mandated by the file format) the appropriate type for use within Python (Endian converted as needed) - but have not looked at the details here at all.

Will-Tyler:
Thank you @schoekek for the pointers, and I agree that endianness could be the issue.

The Binary CIF specification encodes all the data as little endian. (See [here](https://github.com/molstar/BinaryCIF/blob/master/encoding.md#encoding-process).) But the C code uses native types like you mentioned.

I think the simplest fix will be to [change the byte ordering](https://numpy.org/doc/stable/user/byteswapping.html#changing-byte-ordering) from little endian to native endianness on the Python side (`Bio/PDB/binary_cif.py`).

schoekek:
The failure has moved further into the decoder:

AssertionError
assert len(decoded_data) == encoding["srcSize"]

The test passes on little-endian architectures, so there may still be an endianness-related issue after integer_unpack() on big-endian systems.

https://buildd.debian.org/status/fetch.php?pkg=python-biopython&arch=ppc64&ver=1.87%2Bdfsg-3&stamp=1782905907&raw=0

Here is the patch I used: https://salsa.debian.org/med-team/python-biopython/-/blob/master/debian/patches/Fix-Endian-issue-with-binary-CIF.patch?ref_type=heads

dbolser:
I came at this from a different angle (an out-of-bounds write in the same
file, reported to Peter by email) and ended up reading this patch closely.
Five things, most of them small. I have no big-endian machine, so everything
below is from reading rather than testing — @schoekek is the only one here who
can confirm any of it.

**The output array is still little-endian, which may be why the failure moved.**

`_integer_packing_decoder` converts the input with the new
`astype(...newbyteorder("="))`, but the destination is still allocated from an
explicitly little-endian dtype:

```python
    if is_unsigned:
        dtype = np.dtype("<u4")
    else:
        dtype = np.dtype("<i4")
    ...
    decoded_data = np.empty((src_size,), dtype)
    _bcif_helper.integer_unpack(data, decoded_data)
```

`integer_unpack` writes through `uint32_t *`/`int32_t *`, i.e. in native order,
so on a big-endian machine it fills a little-endian-typed array with big-endian
values. Reading `decoded_data` back through NumPy then byte-swaps them. That
would leave this decoder returning wrong values even with the input conversion
in place, and its output feeds whichever decoder pops next — which fits
@schoekek's report that the failure moved on to
`assert len(decoded_data) == encoding["srcSize"]` at line 87 rather than
disappearing. Allocating with `np.dtype(np.uint32)` / `np.dtype(np.int32)`
would match what the C actually writes.

**`format[l - 1]` reads out of bounds if the format string is empty.**

```c
    const size_t l = strlen(in_view.format);
    const char format = in_view.format[l - 1];
```

If `strlen` returns 0 then `l - 1` is `SIZE_MAX`. Today's `format[0]` on an
empty string reads the terminator and falls through to the existing error
branch harmlessly, so this is a new failure mode on the line being changed.
Neither version guards `in_view.format` against NULL either; `PyBUF_FORMAT`
should mean it is always set, but `strlen(NULL)` is a worse way to discover
otherwise than `format[0]` was.

**Taking the last character is looser than it looks.** It gets `"<B"` right,
but it also silently accepts `"ii"` as `i`. Skipping a leading character from
`<>=!@` and then requiring exactly one code left is the same length and
rejects what it should.

**`astype` without `copy=False`.** `_delta_decoder`, two functions up, already
does `data.astype(dtype, copy=False)`. As written the new line copies the whole
array on every little-endian machine, where the conversion is a no-op.

**Unrelated but adjacent:** the assertion that caught this,
`assert len(decoded_data) == encoding["srcSize"]`, disappears under `python -O`.
Given it is checking a length derived from file content against a field of the
same file, it is doing real validation work rather than asserting an internal
invariant, and would be better as a raised exception.

For what it is worth, I implemented a variant of this in a fork: converting the
input to native order as you do here, allocating the output natively, and
having the C helper parse an optional byte-order character and *refuse*
anything that is not native rather than misread it. The refusal turned out to
be the useful part — it means a big-endian machine gets a clear error naming
the format string instead of quietly wrong coordinates, which would have made
#5235 much faster to diagnose. Happy to share it if that is useful, though I
should say plainly that it is AI-written, which I gather is a complication for
contributions here.

---

*Reviewed by Claude (Claude Fable 5, Anthropic), working with @dbolser, who is
posting this on my behalf. Nothing above was executed; there is no big-endian
machine at this end.*

peterjc:
Contributing AI-written code is a problem (I want to outright reject that in #5241 but that has not been agreed), but the ideas are valuable - and indeed I'm expecting bug reports from issues found with AI tools. Also note we *do* seem to have general agreement that all written text on the discussions (GitHub and mailing list) should be human (with an exception of AI usage for translation into English).

dbolser:
> Contributing AI-written code is a problem (I want to outright reject that in #5241 but that has not been agreed), but the ideas are valuable - and indeed I'm expecting bug reports from issues found with AI tools. Also note we _do_ seem to have general agreement that all written text on the discussions (GitHub and mailing list) should be human (with an exception of AI usage for translation into English).

Noted. 

Meanwhile the patch is merged in the fork if you want to test.

PR #5260, opened 2026-07-15
Type-annotate Bio.PDB.PDBMLParser and fix header resolution parsing

AlsoTheZv3n:
## Summary
Adds type annotations to `Bio.PDB.PDBMLParser` and removes its `ignore_errors`
entry from `.mypy.ini`, so the module is now type-checked like the rest of the
package. Along the way this fixes a latent bug in the header parsing.

Main changes:
- Annotate the parsed-tree parameters as `ElementTree.ElementTree` (they were
  annotated as the `ElementTree` *module*, which is not a type).
- Make XML element access explicit about `None`: `Element.find(...)` returns
  `Element | None` and `.text` is `str | None`. Two small helpers (`_find`,
  `_text`) raise `PDBConstructionException` when a required element/text is
  missing (matching the sibling `MMCIFParser`), instead of an opaque
  `AttributeError`. Optional header text fields stay `None`-tolerant, as before.
- Fix an ElementTree truthiness bug (`if element:` → `if element is not None:`).

## Behaviour
For valid PDBML, atom/coordinate data is unchanged (the existing test still
passes). Two intentional differences:
- `resolution` was previously always `None`, because `if element:` is falsy for
  a childless leaf element. It is now read correctly and matches the mmCIF
  parser (1GBT/6WG6/3JQH: 2.0 / 3.54 / 2.201 Å).
- A missing *required* element now raises a descriptive `PDBConstructionException`
  instead of an opaque `AttributeError`.

## Tests
- `test_get_structure` previously compared only the last of its three structures
  (the `assertEqual` sat outside the loop); it is moved inside and now also
  asserts the parsed `resolution` matches the mmCIF parser for each entry.
- Added `test_missing_required_element_raises` and
  `test_optional_header_text_may_be_absent`.

## Verification
`mypy` (v1.16.1) is green on `Bio`/`BioSQL` with the exclusion removed
("no issues found in 298 source files"), `pre-commit run --all-files` is clean,
and the `test_PDB_PDBMLParser` suite passes.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -34,9 +34,6 @@ ignore_errors = True
 [mypy-Bio.PDB.kdtrees]
 ignore_missing_imports = True
 
-[mypy-Bio.PDB.PDBMLParser]
-ignore_errors = True
-
 [mypy-Bio.PDB.PICIO]
 ignore_errors = True
 
--- a/Bio/PDB/PDBMLParser.py
+++ b/Bio/PDB/PDBMLParser.py
@@ -10,6 +10,7 @@ See https://pdbml.wwpdb.org/.
 """
 
 from os import PathLike
+from typing import cast
 from typing import TextIO
 from typing import Union
 from xml.etree import ElementTree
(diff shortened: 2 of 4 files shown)


peterjc:
Type annotation improvements are welcome in principle, thank you.

You've not used the PR template...

I suspect this from your profile, but was this AI generated? See proposed no-AI policy on #5241.

AlsoTheZv3n:
Closing this — it was AI-assisted, which #5241 proposes to disallow, so it's not a fit for the project's direction. Thanks.

PR #5264, opened 2026-07-26
Add api_key parameter to qblast and deprecate username/password

LarytheLord:
Following the discussion, this adds `api_key` support to `Bio.Blast.qblast` (the current function) rather than patching the old `Bio.Blast.NCBIWWW.qblast`, and deprecates the old function as @peterjc suggested.

Changes:
- `Bio.Blast.qblast` gains an `api_key` argument. The NCBI now uses API keys for higher request-rate limits.
- Its `username`/`password` arguments are deprecated, since the NCBI no longer supports HTTP Basic authentication for QBLAST.
- `Bio.Blast.NCBIWWW.qblast` is deprecated in favour of `Bio.Blast.qblast`. The warning notes that the newer function returns a bytes stream (parsed with `Bio.Blast.parse`) rather than a text handle, so calling code will need updating.
- Added `NEWS.rst` and `DEPRECATED.rst` entries, and a test that the deprecation warning is raised.

Closes the api_key request in #5059.

### DIFF
--- a/Bio/Blast/NCBIWWW.py
+++ b/Bio/Blast/NCBIWWW.py
@@ -31,6 +31,7 @@ from urllib.request import install_opener
 from urllib.request import Request
 from urllib.request import urlopen
 
+from Bio import BiopythonDeprecationWarning
 from Bio import BiopythonWarning
 from Bio._utils import function_with_previous
 
@@ -130,6 +131,15 @@ def qblast(
     https://blast.ncbi.nlm.nih.gov/doc/blast-help/urlapi.html
 
     """
+    warnings.warn(
+        "Bio.Blast.NCBIWWW.qblast is deprecated; use Bio.Blast.qblast instead. "
+        "Note that Bio.Blast.qblast returns a bytes stream to be parsed with "
+        "Bio.Blast.parse, rather than the text handle returned here, so calling "
+        "code will need updating. The username and password parameters are also "
+        "no longer supported by the NCBI. This function will be removed in a "
+        "future release.",
+        BiopythonDeprecationWarning,
+    )
     programs = ["blastn", "blastp", "blastx", "tblastn", "tblastx"]
     if program not in programs:
         raise ValueError(
--- a/Bio/Blast/__init__.py
+++ b/Bio/Blast/__init__.py
@@ -38,6 +38,7 @@ from xml.parsers import expat
 
 import numpy as np
 
+from Bio import BiopythonDeprecationWarning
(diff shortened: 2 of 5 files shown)


mdehoon:
@LarytheLord  Did you see the newer qblast function in Bio.Blast?

peterjc:
@mdehoon Why not deprecate the whole of `Bio.Blast.NCBIWWW.qblast` at this point in favour of `Bio.Blast.qblast` (which was added two years ago in Biopython 1.84)? The message will need a little guidance for the unicode string vs bytes change.

LarytheLord:
Ah, I'd missed `Bio.Blast.qblast` — thanks for the pointer. Looked at it now and it has the same gap: no `api_key`, and the same `username`/`password` that the NCBI no longer supports.

So the more useful move is probably to put the `api_key` (and the username/password deprecation) on `Bio.Blast.qblast`, since that's the one going forward, and deprecate the whole of `Bio.Blast.NCBIWWW.qblast` like @peterjc suggested, rather than patching the old one.

Happy to redo the PR that way — add `api_key` + deprecate `username`/`password` on `Bio.Blast.qblast`, and a `BiopythonDeprecationWarning` on `NCBIWWW.qblast` as a whole. Want me to go ahead?

peterjc:
Sounds good, @mdehoon OK with you? I've not scripted any online BLAST searches in a long time (I just run BLAST locally for big jobs or use the browser for small jobs), so am not in a position to judge.

mdehoon:
@LarytheLord  Yes that sounds good, thanks!

peterjc:
You're not (yet) following the PR template which is required.

Are you using AI tools here? Please be aware that we're actively discussing policy there - see #5241 and the mailing list - and disclosing this would be required even if we don't settle on an outright ban as I would prefer.

LarytheLord:
yes, I used AI assistance on this, and I should have said so upfront and filled in the PR template. I'm sorry for that. I've now read #5241 and the mailing list threads, and I understand and respect where Biopython stands on this.

I'm closing this PR. The small parts I put together with assistance I'll redo by hand. If I come back to the api_key / qblast work in #5059, it'll be a fresh PR written manually by me, following CONTRIBUTING.rst and the template properly, with no AI tools involved.

Sorry for the noise here, and thanks for being direct about the policy.

peterjc:
I thought this was clear from your earlier #5169 and #5170, but a non-AI PR would be welcome. Thank you.

PR #5276, opened 2026-08-04
Fix ArrayCore initialization reference handling

### LINKED ISSUE #5271: Incorrect reference handling in _arraycore initialization failure paths
This issue does not affect users with working NumPy installations, the affected code runs only when NumPy initialization fails. It is a defensive hardening opportunity I encountered while chasing a separate memory issue through my project's dependencies.

The affected file is `Bio/Align/substitution_matrices/_arraycore.c` on the current Biopython `master` branch.

#### Expected behaviour

If `numpy.ndarray` is missing, invalid, or unsuitable as a base type, `_arraycore` should abort initialization while releasing exactly the temporary references it owns.

#### Actual behaviour

The initializer decrements the NumPy module reference after retrieving `numpy.ndarray`, then decrements the same reference again when the attribute is missing or is not a type. Other early returns can retain the temporary `numpy.ndarray` reference.

This can corrupt or imbalance CPython reference counts during failed module initialization.

#### Steps to reproduce

Load `_arraycore` in a fresh CPython process with a stub `numpy` module where `ndarray` is either missing or not a type. Both cases exercise the incorrect cleanup paths.

I have a tested fix ready that balances the temporary references and uses `PyModule_AddType` for the successful ownership transfer. I can open a pull request if the maintainers would like this failure path hardened.

stanbot8:

## Summary of Changes

Fixed a few memory issues in `PyInit_arraycore`:
- `basemodule` was decremented twice while `baseclass` leaked. Both references are now released once, and `basetype` remains a non-owning alias. `Bio/Align/substitution_matrices/_arraycore.c`: `185-220, 227, 239, 247, 251`
- `mod` leaked after later initialization failures. Moved module creation after base validation. Failure paths release it. `Bio/Align/substitution_matrices/_arraycore.c`: `184, 236-241, 248`
- A failed `PyType_Ready()` could poison retries. Retries now fail with the exact initialization error. `Bio/Align/substitution_matrices/_arraycore.c`: `225-231, 243-250`
- `Array_Type`'s permanent static reference was incorrectly decremented. Registration now leaves it untouched. `Bio/Align/substitution_matrices/_arraycore.c`: `253`
- Test `test_arraycore_import_preserves_references` covers the success, invalid/missing, and retry cases. `Tests/test_align_substitution_matrices.py`

Closes #5271

### DIFF
--- a/Bio/Align/substitution_matrices/_arraycore.c
+++ b/Bio/Align/substitution_matrices/_arraycore.c
@@ -179,13 +179,10 @@ static struct PyModuleDef module = {
 PyMODINIT_FUNC
 PyInit__arraycore(void)
 {
-    int result;
     PyObject *basemodule;
     PyObject *baseclass;
-    PyTypeObject* basetype;
-
-    PyObject *mod = PyModule_Create(&module);
-    if (!mod) return NULL;
+    PyObject *mod;
+    PyTypeObject *basetype;
 
     // Import the module containing the base class
     basemodule = PyImport_ImportModule("numpy");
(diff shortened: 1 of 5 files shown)


Bio/Align/substitution_matrices/_arraycore.c
```diff
@@ -195,39 +192,65 @@ PyInit__arraycore(void)
...
+        Py_DECREF(baseclass);
+        Py_DECREF(mod);
+        return NULL;
+    }
+    Py_DECREF(baseclass);
 
-    result = PyModule_AddObjectRef(mod, "Array", (PyObject*)&Array_Type);
-    Py_DECREF(&Array_Type);
-    if (result == -1) {
+    if (PyModule_AddObjectRef(mod, "Array", (PyObject *)&Array_Type) < 0) {
```
mdehoon, line 253: Don't we need to decref Array_Type here?

Tests/test_align_substitution_matrices.py
```diff
@@ -4,30 +4,159 @@
...
+
+if len(sys.argv) == 4 and sys.argv[1] == "_arraycore_import_probe":
+    mode = sys.argv[2]
+    if mode == "success":
+        _check_successful_arraycore_import(sys.argv[3])
+    elif mode == "retry":
+        _check_retried_arraycore_import(sys.argv[3])
+    else:
+        _check_failed_arraycore_import(mode, sys.argv[3])
+    raise SystemExit
```
mdehoon: Can these be converted to regular unittests?

Bio/Align/substitution_matrices/_arraycore.c
stanbot8: `Array_Type` is static, so there's no reference to release here. `PyModule_AddObjectRef()` gives the module a reference that is released automatically during cleanup.

Tests/test_align_substitution_matrices.py
stanbot8: I converted the four probe cases to `unittest.TestCase` but kept the subprocess isolation so each test starts clean. I also rebased onto the latest `master`. Does this look better?

stanbot8: I reconsidered the test arrangement and moved the import cases to their own module. The main test file no longer has to moonlight as a subprocess target.

Tests/test_align_substitution_matrices.py
```diff
@@ -13,21 +20,39 @@
...
 from Bio.Align import substitution_matrices
 from Bio.Data import IUPACData
 
+
 nucleotide_alphabet = IUPACData.unambiguous_dna_letters
 protein_alphabet = IUPACData.protein_letters
 
 
 class TestBasics(unittest.TestCase):
+    @unittest.skipUnless(sys.implementation.name == "cpython", "requires CPython")
```
mdehoon, line 33: Why is Cython needed?

stanbot8: Cython isn't needed. The test uses CPython's `sys.getrefcount()` to verify ownership.

PR #5283, opened 2026-08-10
Handle empty QUAL title line without raising IndexError

eeshsaxena:

---

### Problem

The QUAL parsers in `Bio/SeqIO/QualityIO.py` extract the record id with `descr.split()[0]`. A record whose title line is empty (a bare `>`) makes `descr.split()` return an empty list, so parsing raises `IndexError` instead of reading the record:

```python
import io
from Bio import SeqIO
list(SeqIO.parse(io.StringIO('>\n10 20 30\n'), 'qual'))
# IndexError: list index out of range
```

`FastaIO` already handles this case (it falls back to an empty id via `try: title.split(None, 1)[0] except IndexError: ''`), so FASTA parses a bare `>` fine while QUAL crashes.

### Fix

Fall back to an empty id when the title line is empty, in both QUAL iterator paths, matching `FastaIO`'s behaviour.

### Test

Added `test_qual_empty_title`: a bare `>` record parses with `id == ''` and its qualities intact. It raises `IndexError` before this change and passes after. The full `test_SeqIO_QualityIO.py` suite passes (53 tests).

### DIFF
--- a/Bio/SeqIO/QualityIO.py
+++ b/Bio/SeqIO/QualityIO.py
@@ -1086,7 +1086,11 @@ class FastqIteratorAbstractBaseClass(SequenceIterator[str]):
             self.line = None
 
         descr = title_line
-        id = descr.split()[0]
+        try:
+            id = descr.split(None, 1)[0]
+        except IndexError:
+            # Empty title line (a bare ">"), matching FastaIO behaviour.
+            id = ""
         name = id
 
         if not quality_string.isascii():
@@ -1538,7 +1542,11 @@ class QualPhredIterator(SequenceIterator):
             raise StopIteration
         while True:
             descr = line[1:].rstrip()
-            id = descr.split()[0]
+            try:
+                id = descr.split(None, 1)[0]
+            except IndexError:
+                # Empty title line (a bare ">"), matching FastaIO behaviour.
+                id = ""
             name = id
 
             qualities: list[int] = []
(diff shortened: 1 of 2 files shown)


peterjc:
The change is reasonable, but I wonder where you found such Qual files and indeed why you are using them in this day-and-age?

Also, please use the PR template, and for the purposes of our on-going AI policy discussion (see eg #5241), was any of this AI generated?

eeshsaxena:
Fair questions.

On the files: I don't have real .qual files lying around, and you're right that they're a relic. I hit this while poking at the QualityIO parsers for robustness, feeding them deliberately malformed input. A blank title line on the QUAL record is what trips it, the parser indexes into the split result and raises IndexError instead of a clean parse error. So it's a "don't crash on malformed input" fix rather than anything to do with me relying on QUAL day to day.

On the template: my bad, I'll redo the description against it.

On the AI question, since you're asking straight: I used AI to help test the solution, generating the malformed cases and checking the behaviour. Wanted to be upfront about that given the policy thread.

peterjc:
You need to check some of the boxes in the PR template...

If you wrote the fix yourself, that's fine. The test being AI generated is annoying as I don't want to merge that in violation of my preferred no-AI stance on #5241, but is much like something I'd have written for this. PR edited with a shorter human-written test case.

peterjc:
You need to agree to the dual licensing for this PR to be merged. Thank you.

eeshsaxena:
Happy to. I agree to the dual licensing terms for this contribution. Let me know if you need it noted anywhere specific.

peterjc:
Your comment here is enough. Thank you.

Assuming the full test suite passes (and it should), I will do a squash-and-merge to keep the AI test out of the main branch history.

PR #5290, opened 2026-08-19
Pairwise aligner convenience functions

mdehoon:
This PR adds an `align` and `score` function to `Bio.Align` as convenience functions.
These functions take `seqA`, `seqB`, and further arguments `*args` and `**kwargs` as input.
Using these arguments, these functions construct a `PairwiseAligner`, and runs the aligner on the `seqA` and `seqB` sequences.
The function obtains the mode and strand from the `**kwargs` if provided explicitly.
Otherwise, it will look for `"global"` or `"local"`, and `"+"` or `"-"`, in the `*args`, and set the mode and strand accordingly, defaulting to `"global"` and `"+"` if not provided.

While these functions are inefficient compared to creating a `PairwiseAligner` first and applying it to many sequences, they provide an interface similar to `Bio.pairwise2`, and may help users to transition from `Bio.pairwise2`.

Let's hear from users if there are any comments or suggestions, before merging this.

### DIFF
--- a/Bio/Align/__init__.py
+++ b/Bio/Align/__init__.py
@@ -48,7 +48,6 @@ from Bio.Seq import SequenceDataAbstractBaseClass
 from Bio.SeqRecord import _RestrictedDict
 from Bio.SeqRecord import SeqRecord
 
-
 # Import errors may occur here if a compiled _pairwisealigner.c file or
 # compiled _codonaligner.c file (_pairwisealigner.pyd or _pairwisealigner.so,
 # or _codonaligner.pyd or _codonaligner.so) is missing or if the user is
@@ -4446,7 +4445,6 @@ AlignmentCounts object returned by the .counts method of an Alignment object."""
 
     def align(self, seqA, seqB, strand="+"):
         """Return the alignments of two sequences using PairwiseAligner."""
-        # self.warn_defaults_changed()  # FIXME remove this after 1.87 is out
         if isinstance(seqA, (bytes, Seq, MutableSeq, SeqRecord)):
             sA = bytes(seqA)
             sA = np.frombuffer(sA, dtype=np.uint8).astype(np.int32)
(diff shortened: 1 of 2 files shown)


peterjc:
Worth adding to the DEPRECATION file entry for pairwise2? Cross reference #5273.

mdehoon:
@peterjc 
> Worth adding to the DEPRECATION file entry for pairwise2?

Yes I think so, but let's wait to see if we get any comments on these convenience functions.

In particular, I am wondering if we should use `lalign`, `galign` or `localalign`, `globalalign` for local and global alignments, respectively, instead of the current `align(..., "local", ...)`.
Also, maybe we can drop the `score` convenience function; I doubt it would be used much.

peterjc:
Given the historical separation, I lean slightly to separate helper functions for local vs global. And I'd go for the more explicit names ` localalign` and `globalalign` (or even `local_align` and `global_align`) over `lalign` and `galign`.

CC @MarkusPiotrowski

peterjc:
Looks like the tests and code are out of sync now - did you forget to commit the updated tests perhaps?

peterjc:
Try this for the RST bullet point list:

```
    * m: (match_score, mismatch_score).
    * g: gap_score.
    * i: insertion_score (if one value);
      (open_insertion_score, extend_insertion_score) if two values.
    * d: deletion_score (if one value);
      (open_deletion_score, extend_deletion_score) if two values.
    * o: open_gap_score (if one value);
      (open_insertion_score, open_deletion_score) if two values.
    * x: extend_gap_score (if one value);
      (extend_insertion_score, extend_deletion_score) if two values.
    * e: end_gap_score (if one value);
      (end_insertion_score, end_deletion_score) if two values.
    * s: substitution_matrix.
```

mdehoon:
@peterjc  Thanks

PR #5293, opened 2026-08-22
Return NotImplemented for unsupported Seq comparison types

### LINKED ISSUE #5292: Bio.Seq.Seq ordering operators raise opaque TypeError instead of NotImplemented
## Bug report

`Bio.Seq.Seq`'s ordering operators fall back to `self._data < other` (a `bytes` comparison) for unsupported operand types, which raises an opaque `TypeError` rather than returning `NotImplemented`:

```python
def __lt__(self, other):
    if isinstance(other, _SeqAbstractBaseClass):
        return self._data < other._data
    elif isinstance(other, str):
        return self._data < other.encode("ASCII")
    else:
        return self._data < other
```

### Problem

For an unsupported `other` (e.g. an `int`, `None`, or a `bytearray`), `bytes < other` raises `TypeError` immediately. Returning `NotImplemented` instead would let Python fall back to the reflected operation or raise a standard, clearer comparison error — the conventional behavior for rich-comparison operators.

Example: `Seq('ATG') < 5` currently raises `TypeError: '<' not supported between instances of 'bytes' and 'int'`, exposing the internal `bytes` representation.

### Expected

Return `NotImplemented` (or `False` for `__eq__`/comparisons with clearly incompatible types) instead of relying on `bytes`'s `TypeError`.

### Location

`Bio/Seq.py` — `Seq.__lt__`, `__le__`, `__gt__`, `__ge__` (and `__eq__`).

AdvaitS:
Fixes #5292

`Seq`'s rich-comparison operators (`__lt__`, `__le__`, `__gt__`, `__ge__`) fell back to `self._data < other` for unsupported operand types, which raised an opaque `TypeError` exposing the internal `bytes` representation. This changes them to return `NotImplemented` for unsupported types (so Python raises a standard `Seq`-level error), while preserving `str`/`bytes`/`bytearray` comparison.

Before: `Seq('ACG') < 5` → `TypeError: '<' not supported between instances of 'bytes' and 'int'`
After:  `Seq('ACG') < 5` → `TypeError: '<' not supported between instances of 'Seq' and 'int'`

## Test log

```
$ python -m unittest test_seq.TestSeqStringMethods.<comparison tests>
Ran 8 tests ... OK
```

### DIFF
--- a/Bio/Seq.py
+++ b/Bio/Seq.py
@@ -454,8 +454,9 @@ class _SeqAbstractBaseClass(ABC):
             return self._data < other._data
         elif isinstance(other, str):
             return self._data < other.encode("ASCII")
-        else:
+        elif isinstance(other, (bytes, bytearray)):
             return self._data < other
+        return NotImplemented
 
     def __le__(self, other):
         """Implement the less-than or equal operand."""
@@ -463,8 +464,9 @@ class _SeqAbstractBaseClass(ABC):
             return self._data <= other._data
         elif isinstance(other, str):
             return self._data <= other.encode("ASCII")
-        else:
+        elif isinstance(other, (bytes, bytearray)):
             return self._data <= other
+        return NotImplemented
 
     def __gt__(self, other):
         """Implement the greater-than operand."""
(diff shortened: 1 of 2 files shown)


peterjc:
Is this AI generated? Please see #5241 for a potential explicit policy on that.

Also, please reopen using our PR template.

mdehoon:
> This changes them to return NotImplemented for unsupported

contradicts

> Before: Seq('ACG') < 5 → TypeError: '<' not supported between instances of 'bytes' and 'int'
> After: Seq('ACG') < 5 → TypeError: '<' not supported between instances of 'Seq' and 'int'

@AdvaitS  Was this generated by AI?

PR #5296, opened 2026-08-27
Add Windows ARM64 build

ndabas:

As the title says.

I have added Windows ARM64 to the build and test matrices. Note that setup-python only has WoA builds starting from 3.11 so the matrices reflect that.

Additionally, there was a test that was failing on Windows (all architectures) because it was hard-coded to LF line endings, and the `.gitattributes` in this repo configured .txt files to have CRLF endings on checkout. I have made a minor change there so that the test does not depend on the line ending at all, and it now passes on all platforms. Hopefully adding that to the same PR is fine; let me know if not.


peterjc:
I have no objection to testing on Windows ARM64 - that's a good idea. But with this the test matrix which was already overly large gets considerably bigger. That's a _lot_ of compute for relatively little gain.

If you're comfy with the syntax, how could you feel about say just doing the oldest and the newest Python on every architecture, with a full range of Python restricted to one of the faster builds (Linux maybe?).

ndabas:
> If you're comfy with the syntax, how could you feel about say just doing the oldest and the newest Python on every architecture, with a full range of Python restricted to one of the faster builds (Linux maybe?).

Thanks, but correct me if I am wrong -- it looks like this is how it is setup already (for all OSs + architectures) -- there is one test on the oldest supported Python version and another one on the newest. The changes in this PR are only adding two test jobs for Windows ARM64, one on Python 3.11 and another one on 3.14.

What I could do is to add tests on every supported version to the Linux job, and possibly clean up the other matrices a bit to reduce duplication of the Python version lists.

Let me know your thoughts (or if I misunderstood what you were saying!)

peterjc:
Ah - you're right. I'd not looked at this recently enough.

The _full_ matrix (all those Python versions for all those architectures) is being run for the build stage - but that is super fast.

A much reduced list is being used for the actual (rather slow) testing - and your PR should only expand that slightly.

Let's run the tests now...

peterjc:
Looks good - although a little slower than the Windows x86 tests.

Do you think we need to add a mention of this to the NEWS file? If so please add that to this PR.

(I'm not sure if we should try to do ARM64 wheels in our next release via https://github.com/biopython/biopython-wheels yet given https://github.com/pypa/cibuildwheel says that is still experimental - but one day!)

If you can see any sensible cleanup of other matrices that could be a good follow up PR?

peterjc (approved):

ndabas:
> Looks good - although a little slower than the Windows x86 tests.

Yes, that's because the Windows 11 ARM runner does not currently have Python 3.11 pre-installed, so installing that adds a minute or two. The tests on 3.14 take about the same time as the other tests on Windows.

> Do you think we need to add a mention of this to the NEWS file? If so please add that to this PR.

Sure, I think that would be nice, I'll go ahead and do that.
 
> (I'm not sure if we should try to do ARM64 wheels in our next release via https://github.com/biopython/biopython-wheels yet given https://github.com/pypa/cibuildwheel says that is still experimental - but one day!)

I think you can go ahead -- they've been saying that for years haha, there are plenty of other projects who have released Windows ARM64 wheels built using cibuildwheel. To be clear they say that the CI/build runner bits are experimental, not the built artifacts themselves -- as those are built using standard build mechanisms like PEP 517/518.

I think I'll go ahead and create a PR in the biopython-wheels repo as well to add support for Windows ARM64.
 
> If you can see any sensible cleanup of other matrices that could be a good follow up PR?

Sure, I'll do some experiments to see if we can reduce a bit of repetition, and will create a PR if it looks significant enough.

ndabas:
Added to NEWS -- let me know if that looks okay!

NEWS.rst
```diff
@@ -17,6 +17,8 @@ This release of Biopython supports Python 3.10, 3.11, 3.12, 3.13, 3.14 and the
 Python 3.15 release candidate. It has also been tested on PyPy3.10 v7.3.19.
 Python 3.10 is approaching end of life, our support for it is now deprecated.
 
+This release adds support for Python 3.11+ on Windows for ARM64.
+
```
peterjc, line 20: If we've sure we can do the wheels, then yes. I was thinking a weaker statement like the one about PyPy, that we're testing on Windows ARM64 now.

ndabas: I'm sure we can, if you're sure as well. However this being the first release for WoA, I can weaken the language a bit.

ndabas:
I've added a PR for this in the wheels repo: https://github.com/biopython/biopython-wheels/pull/22

NEWS.rst
ndabas: Done.

peterjc (approved):

ndabas:
@peterjc It seems the checks got stuck because I force-pushed my branch with a `[skip ci]` commit, is there a way for you to trigger those manually, or I can push it again without the skip? (In which case you would need to approve it again I suppose.)

peterjc:
It did seem to be in an odd state - rebased anyway as the previous commit had passed all the tests.

Thank you.

ndabas:
Great, thank you!

PR #5297, opened 2026-08-29
Check offset and row length in PrintedAlignmentParser.feed()

### LINKED ISSUE #5272: PrintedAlignmentParser.feed() does not bound offset, and its length check cannot fire
`PrintedAlignmentParser.feed()` adds a caller-supplied `offset` to the line
buffer without validating it, and the guard meant to reject lines of differing
length cannot fire. All of the below reproduces on a clean 1.87 install
(CPython 3.12.3, Linux, `pip install biopython==1.87`).

The first one is reachable from the public API and is the one I would look at
first, because it returns process memory to the caller rather than raising.

### 1. Ragged blocks are accepted, and the coordinates contain heap contents

`Alignment.parse_printed_alignment()` feeds each line straight to the parser.
Given rows of differing printed width it does not complain:

```python
from Bio.Align import Alignment

lines = [b"ACGT--ACGTACGT--AC", b"ACGTTTACGT"]   # 18 and 10 wide
sequences, coordinates = Alignment.parse_printed_alignment(lines)
```

No error. `coordinates` then holds values like

```
[0, 4, 4, 8, -2, 12, 12, 12, 12, 12, 14, 5273202989963956549,
 7235349650388502096, 4049075123956623165, ...]
```

Those large integers are not coordinates. Unpacking a few of them as little
endian `int64` gives `b"PF'), id"` and `b"', dbxre"` — text out of the process
heap, arriving in a NumPy array the caller is expected to use.

### 2. An out-of-range `offset` reads out of bounds, and segfaults when large

```python
from Bio.Align._aligncore import PrintedAlignmentParser

PrintedAlignmentParser().feed(b"ACGT", 100000)   # returns (1, b'\x02')
PrintedAlignmentParser().feed(b"ACGT", 1 << 40)  # exit 139, segmentation fault
```

`feed(self, line, offset=0)` is documented in the module as skipping the first
`offset` bytes, so a caller has no reason to expect either result.

### Where it comes from

Reading `Bio/Align/_aligncore.c` on current master:

**The offset is never bounded.** Line 141 accepts any `Py_ssize_t`, line 143
adds it, and line 148 dereferences the result:

```c
    if (!PyArg_ParseTuple(args, "S|n:feed", &line, &offset)) return NULL;

    buffer = PyBytes_AS_STRING(line) + offset;

    s = buffer;
    ...
    if (*s == '-') row[i++] = 0;
```

`PyBytes_GET_SIZE(line)` is never consulted.

**The length check is a tautology.** `m` is defined as `s - buffer` on the line
above the comparison, so `buffer + m` is `s` by construction and the `else if`
is never true:

```c
    m = s - buffer;
    if (n == 0) self->m = m;
    else if (buffer + m != s) {
        PyErr_Format(PyExc_ValueError,
                     "line has length %zd (expected %zd)", m, self->m);
```

The intended comparison looks like `m != self->m`. As written, the error it
formats — which already prints both lengths — can never be raised, so a ragged
block is parsed as though every row were the width of the first, and the row
that is too short reads past its own buffer.

### Suggested shape of a fix

Bound `offset` against `PyBytes_GET_SIZE(line)` and raise `ValueError` when it
is outside, and change the comparison to `m != self->m`. I would treat the
second as the more urgent of the two: the offset argument is at least
explicitly passed by a caller, whereas a ragged block can come from a malformed
file and produces no error at all.

### What I checked, and what I did not

Run rather than inferred: all three reproductions above, on a clean 1.87
install, including unpacking the returned integers back to bytes. Read from
current master rather than run: the quoted source and its line numbers. Not
checked: any platform other than Linux / CPython 3.12.3, whether older releases
behave the same, and how far past the buffer the ragged case actually reads —
I did not run it under a sanitizer.

I have a fix for both in a fork and am happy to open a pull request if that
would help, though I am conscious `CONTRIBUTING.rst` restricts AI-assisted
contributions, so I am reporting rather than assuming.

---

*Investigated by Claude (Claude Fable 5, Anthropic), working with @dbolser,
who is posting this on my behalf.*

trinik15:

Closes #5272.

In feed() in Bio/Align/_aligncore.c, the offset was added to the line buffer
pointer without checking it against the line length, so out-of-range offsets
read out of bounds and large offsets segfault. feed() now raises ValueError
for offsets outside 0 <= offset < len(line).

The row length check compared buffer + m against s, where m = s - buffer, so
it was always false and ragged rows were accepted; parsing a short row then
read past its own buffer, which is where the garbage coordinates in the
issue's first reproduction come from. The check now compares m against
self->m and raises ValueError for rows of differing width. The feed()
docstring documents the new ValueError.

No internal caller is affected: parse_printed_alignment calls feed() without
an offset, and the bigmaf reader passes an offset for a position it just read
from the buffer.

The tests in Tests/test_Align_printed_alignment.py fail on unpatched master;
with this patch they all pass, and the alignment parser test files that use
parse_printed_alignment still pass.

Assisted-by: Z.ai Code


peterjc:
You've not used the PR template, but I think you've covered the key points like the licensing. Was this fix AI generated which would fall foul of the current policy proposal on #5241 - or human generated? Thanks!

(Skipping the PR template seems to be common with AI tooling.)

trinik15:
hi yes this was ai assisted i reviewed and understood every change before pushing, i've read the #5241 draft and i see this wouldnt be accepted, ive added an assisted by line. If you'd rather close this i will rework the patch myself

peterjc:
That policy isn't agreed yet (obviously I hope it will be before too long), which is putting any AI PRs on hold. Some submitters are waiting, others have closed their PRs - it depends in part on their motivations I suppose.

If you're happy to redo the fix and write a test by hand - having understood the problem from this AI assisted investigation - that'd be good. We could merge that without waiting on the AI policy being settled. Thank you.

trinik15:
Superseded by #5298: the fix and tests were rewritten by hand after understanding the problem from this AI-assisted investigation, as discussed above. Closing in favour of the new pull request.

PR #5300, opened 2026-09-01
Improve Seq comparison TypeError messages

### LINKED ISSUE #5292: Bio.Seq.Seq ordering operators raise opaque TypeError instead of NotImplemented

ktirumal2006:

Closes #5292

Keep the existing `TypeError` behavior for incompatible sequence ordering
comparisons, following the maintainer discussion, but report the public
`Seq` or `MutableSeq` type instead of leaking the internal `bytes` or
`bytearray` representation.

Add regression coverage for `<`, `<=`, `>`, and `>=` with incompatible
operands on both sequence types.

Tests:
- `python test_Seq_objs.py` — 69 passed
- `pre-commit run --files Bio/Seq.py Tests/test_Seq_objs.py` — passed
- `python run_tests.py --offline` — the changed `test_Seq_objs` group passed;
  two unrelated failing modules reproduced identically on the unmodified base
  revision (`test_Align_aligncore` and `test_SearchIO_infernal_text_index`)


peterjc:
I take it this was based on the discussion on #5293?

Did you use any AI tools for this please? See #5241

PR #5301, opened 2026-09-02
Issue #3600 - Replace assert statements with if statements in Bio.Align.hhr

ssarmin:
Replaced assert statements with if-statements and explicit error raising in Bio/Align/hhr.py.

Addresses part of #3600.


Bio/Align/hhr.py
```diff
@@ -66,21 +66,30 @@ def _read_header(self, stream):
...
             if line.strip() == "":
                 break
             counter += 1
             word, _ = line.split(None, 1)
-            assert int(word) == counter
+            if int(word) != counter:
+                raise ValueError(
+                    "Unexpected alignment number %s in summary table (expected %d)"
+                    % (word, counter)
+                )
```
peterjc, line 81: I'd use f-string formatting here (but % like C code is fine especially if the rest of the file does that).

peterjc (approved):
I'm happy. For bonus points do you think you can trigger any of these new exceptions (eg with malformed data) and turn that into a test?

ssarmin:
> I'm happy. For bonus points do you think you can trigger any of these new exceptions (eg with malformed data) and turn that into a test?

Thanks for the review!

For the bonus testing suggestion, my plan is to add malformed `.hhr` files under `Tests/HHsuite/`, paired with test cases in `test_Align_hhr.py` asserting the ValueError.

For now, I am going to start with just one test case and get your feedback before writing the rest. Does this approach sound good?

ssarmin:
I just pushed the tests: ended up adding all 4 for the total-length parsing checks (Q, Q Consensus, T, T Consensus) instead of just one, since they followed the same pattern.

Each of the tests uses a malformed copy of the existing allx.hhr file with the parens stripped from one field, paired with a test asserting the ValueError message.

Please let me know if this approach looks good.

peterjc:
@mdehoon are you happy to merge this?

mdehoon (approved):

mdehoon:
Looks good to me, thanks!

Issue #5309, opened 2026-09-10
QCPSuperimposer returns a wrong rotation when the reference point set is collinear


ChenziqiAdam:
When the reference set is collinear and the moving set is the same points after a non-trivial rigidrotation, `QCPSuperimposer` returns `get_rms()` ≈ 0, but returns the identity rotation matrix, so `get_transformed()` is not aligned at all and the true RMSD of the transformed coordinates is large. 

I think this may be the same symptom class as #5004, on an input that #5005 did not cover.

Code for reproduction:

```python
import numpy as np
from Bio.PDB.qcprot import QCPSuperimposer
from Bio.SVDSuperimposer import SVDSuperimposer

# reference: three collinear points
ref = np.array([[0., 0., 0.], [1., 0., 0.], [2., 0., 0.]])

# moving set: the same points after a rigid rotation + translation
theta = 0.9
ax = np.array([1., 2., 3.]); ax /= np.linalg.norm(ax)
K = np.array([[0, -ax[2], ax[1]], [ax[2], 0, -ax[0]], [-ax[1], ax[0], 0]])
R = np.eye(3) + np.sin(theta) * K + (1 - np.cos(theta)) * K @ K
mob = ref @ R.T + np.array([5., -2., 7.])

for name, S in [("QCP", QCPSuperimposer), ("SVD", SVDSuperimposer)]:
    s = S(); s.set(ref, mob); s.run()
    t = s.get_transformed()
    rmsd_true = np.sqrt(((ref - t) ** 2).sum() / 3)
    print(f"{name}: get_rms()={s.get_rms():.3e}   RMSD(transformed)={rmsd_true:.3e}")
```

Output:

```
QCP: get_rms()=0.000e+00   RMSD(transformed)=6.845e-01
SVD: get_rms()=1.125e-15   RMSD(transformed)=1.125e-15
```

peterjc:
A co-linear reference set sounds like a pathological input set where you'd expect this to fail, but if you think this is a threshold issue the changes in the issues you point to sound like a good starting point to investigate.

Over to @JoaoRodrigues who is better placed than me to comment.

PR #5321, opened 2026-09-18
Fix #5309: QCPSuperimposer wrong rotation for collinear reference

rollroyces:
## Summary

MOSS-TTS-Nano installs cleanly and `is_available()` returns ok, but every generation fails because the engine's venv ships torchaudio 2.7 with no I/O backend (no soundfile, no torchcodec, no ffmpeg). The engine's own bundled reference clip is the first thing it tries to load, so the very first generation — not a user audio file — fails with `Couldn't find appropriate backend to handle uri .../engines/moss-tts-nano/MOSS-TTS-Nano/assets/audio/zh_1.wav`.

The fix is two-line in `SPECS["moss-tts-nano"]`: add `soundfile` to `install_args` so the editable install pulls a torchaudio backend into this engine's own venv, and tighten `probe_code` with `assert torchaudio.list_audio_backends()` so a future upstream change that drops the backend will fail the verify step loudly instead of crashing hours later at first audio read.

Fixes #2100.

## Changes

- **`backend/services/sidecar_install.py`** (`SPECS["moss-tts-nano"]`):
  - `install_args = ("-e", "{checkout}", "soundfile")` — pulls `soundfile` alongside the editable install. Same venv, no parent-app surface change.
  - `probe_code = "import moss_tts_nano_runtime, torchaudio; assert torchaudio.list_audio_backends(), 'torchaudio has no I/O backend (install soundfile or torchcodec)'"` — runs in the verify step after install. The assert message names the two viable backends so the failure is actionable.
- **`tests/test_sidecar_install.py`**:
  - Extend `test_new_specs_install_recipe` parametrisation to include `moss-tts-nano` with the new `install_args` (`["-e", "{c}", "soundfile"]`), so the install recipe for this engine is regression-tested just like moss-tts-v15 / confucius4-tts / dots-tts.
  - New `test_moss_tts_nano_probe_asserts_an_audio_backend_is_present` — asserts the probe_code is non-empty, references `torchaudio.list_audio_backends()` (the assertion), and compiles cleanly (it runs through `python -c` on Windows where backslashes in source paths could otherwise break the literal).

## Why this PR

The reporter verified the fix manually: `uv pip install --python <moss-nano venv>/bin/python soundfile` was sufficient. The two-part install_args + probe_code split means the engine ships a working audio path on the next clean install, and a future upstream refactor that removes soundfile (e.g. switches to torchcodec) fails fast with a clear message rather than hiding behind a first-generation crash.

## Type

- [x] 🐛 Bug fix

## Testing

- `uv run --frozen pytest tests/test_sidecar_install.py -q` → **138 passed** (was 137, +1 new dedicated test + 1 new parametrised case)
- `uv run --frozen pytest tests/test_sidecar_install.py -k "moss_tts_nano or new_specs" -v` → all targeted tests green
- `HF_HUB_OFFLINE=1` per `AGENTS.md` honest-CI guidance

## Checklist

- [x] Forked the repo and branched from `main`
- [x] One feature/fix per PR
- [x] PR title uses Conventional Commit format with scope, referencing the issue
- [x] All template sections filled
- [x] Tests added that fail without the source change and pass with it (revert `install_args` → existing test fails on `pip[5:] == [...]`)
- [x] CHANGELOG Unreleased entry — deferred to maintainer per their active reorganisation of the section (#2180 follow-up applies the same)


peterjc:
The PR desciption has nothing to do with the code changes or to Biopython. At best this is a copy-paste error. If so, please read and follow the CONTRIBUTING file before trying again.

